PR #20, opened 2018-03-22
Add PreferredPaymentMethodDelegate to CheckoutViewController

ochococo:
Should partially solve the Issue https://github.com/Adyen/adyen-ios/issues/21

# 🤔 Rationale

- Server can send payment setup with only one available payment method, this way we can save the user one useless tap
- Use case: preferred payment method is preselected before Adyen payment was initiated but we want to use Quick Integration
- We can accept only one payment type ex. Apple Pay

# 💻 Getting there

While this PR is a first step to have single payment method flow using Quick Integration I think we need something much better.

I need more time to propose changes to the codebase giving us the ultimate solution to this problem but in the mean time this addition is an improvement and would not break compatibility. In the final soultion the public API won't change so the faster we introduce it, the better.

# ⚙️ How to use it?

Implement the delegate:

```swift
extension PaymentInteractionMediator: PreferredPaymentMethodDelegate {
    func preferredMethod(_ controller: CheckoutViewController, available availableMethods: [Adyen.PaymentMethod]) -> Adyen.PaymentMethod? {
        return availableMethods.first // Why not? ;)
    }
}
```

Set a delegate:

```swift
checkoutViewController.paymentMethodDelegate = paymentInteractionMediator
```

That's it.

### DIFF
--- a/Adyen/UI/Checkout/CheckoutViewController.swift
+++ b/Adyen/UI/Checkout/CheckoutViewController.swift
@@ -42,7 +42,10 @@ public final class CheckoutViewController: UIViewController, PaymentRequestDeleg
     
     /// The delegate for payment processing.
     internal(set) public weak var delegate: CheckoutViewControllerDelegate?
-    
+
+    /// The delegate for payment method.
+    public weak var paymentMethodDelegate: PreferredPaymentMethodDelegate?
+
     /// The delegate for card scanning functionality for card payments.
     public weak var cardScanDelegate: CheckoutViewControllerCardScanDelegate?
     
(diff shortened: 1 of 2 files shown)


ochococo:
@joostvandijk 👀

joostvandijk:
Hi @ochococo,

Thanks a lot for your contribution! I have a couple of questions:
- How would this work when the payment method requires a view controller to be displayed modally, such as Apple Pay. First, the CheckoutViewController is presented, and then, the Apple Pay dialogue is presented? I'm worried that this would look a bit odd. For payment methods where a view controller is pushed, such as cards, it would look fine, I assume.
- Why would this require an additional protocol method? I think it would be nicer for the SDK to automatically do this behavior if only one payment method is returned from `/setup`.

Please note that we are working on a solution that would enable this, but it will take some time for it to arrive. If we're able to implement this now, without any changes to the public interface, then we'd gladly add it.

ochococo:
Hi @joostvandijk 

Ad. 1 - It just works as you've described. I agree it looks a bit weird but if we only support one method (ex. ApplePay) we should not present list with one item, I would say it just as odd to have to tap on the only method on a list.

Ad. 2 - Yeah I agree - we could do this inside the SDK without any external call but I wanted to left the default flow unchanged for other consumers of this code.

I will update this PR to just start the payment if only one is returned without changes to public interfaces, will you accept that change?

ochococo:
@joostvandijk I would like to help you with the ultimate solution to this problem, it would be good for us to discuss the strategy and maybe I could implement it for you?

jkutyn:
Hi @ochococo. Sorry for the delayed update. That's a great idea. We wanted to put into place a more robust solution that works out of the box, so we are working on building this behaviour into our core functionality. We will keep you updated on the progress.

Issue #21, opened 2018-03-22
Preferred payment method on Quick Integration flow


ochococo:
- Server can send payment setup with only one available payment method, this way we can save the user one useless tap
- Use case: preferred payment method is preselected before Adyen payment was initiated but we want to use Quick Integration
- We can accept only one payment type ex. Apple Pay

joostvandijk:
This has been fixed in 2.4.0.

ochococo:
Thanks @joostvandijk :)

PR #27, opened 2018-04-29
Feature/preferred payment in quick integration improved ux

ochococo:
Should solve the Issue https://github.com/Adyen/adyen-ios/issues/21

# 🤔 Rationale

- Server can send payment setup with only one available payment method, this way we can save the user one useless tap
- Use case: preferred payment method is preselected before Adyen payment was initiated but we want to use Quick Integration
- We can accept only one payment type ex. Apple Pay

# ⚙️ How to use it?

Implement the delegate:

```swift
extension PaymentInteractionMediator: PreferredPaymentMethodDelegate {
    func preferredMethod(_ controller: CheckoutViewController, available availableMethods: [Adyen.PaymentMethod]) -> Adyen.PaymentMethod? {
        return availableMethods.first // Why not? ;)
    }
}
```

Set a delegate:

```swift
checkoutViewController.paymentMethodDelegate = paymentInteractionMediator
```

That's it.

# 🤔 Caveats?

Some payment plugins can present instead of push, internal logic can suggest otherwise but for example ApplePay plugin will ignore the navigation mode.

### DIFF
--- a/Adyen.xcodeproj/project.pbxproj
+++ b/Adyen.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		1A64D0B32094C71600448864 /* LoadingViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1A64D0B12094C71600448864 /* LoadingViewController.swift */; };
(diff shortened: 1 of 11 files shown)


joostvandijk:
Hi @ochococo,

Thank you for your contribution! We just released 2.4.0, which includes this functionality. Now, when only one payment method is available, the UI will go straight to that payment method.

Kind regards,

Joost

PR #36, opened 2018-07-20
Swift 4.2 support

rvandijke:
Add support for Swift 4.2 syntax.

All modifications are protected by #if swift(>=4.2) to also support for the current versions of Swift.


rvandijke:
Closing PR as this conflicts with V2

PR #40, opened 2018-08-07
Added a shouldDismissOnDidFinish property to CheckoutController 

jasonhuge:
**Rationale**
Currently the CheckoutController is dismissed before the didFinish delegate method is called.  
However, In the didFinish delegate method we still have more work to do. Such as making a call to our backend so that we can communicate with Adyen and update our records.  
Although it makes sense that CheckoutController is dismissed once it is finished.  Visually as a user it looks awkward dismissing the Adyen activity indicator before all backend communication is finished.  

**Solution**
Adding a shouldDismissOnDidFinish property that defaults to true I think should fix this issue.  
By defaulting shouldDismissOnDidFinish to true the workflow kept the same as it is now, but it gives the us the option to set to false which would allow us to dismiss the CheckoutController when see fit.

### DIFF
--- a/Adyen/UI/Checkout/CheckoutController.swift
+++ b/Adyen/UI/Checkout/CheckoutController.swift
@@ -46,6 +46,9 @@ public final class CheckoutController {
     /// Determines whether the preselected payment method should be shown when available. Default value is `true`.
     public var showsPreselectedPaymentMethod = true
     
+    /// Determines whether the presenter should automatically dismiss when the didFinish method is called
+    public var shouldDismissOnDidFinish = true
+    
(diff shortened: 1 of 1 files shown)


diegomarcon:
Hi Jason,

Thanks for submitting a pull request and pointing to this issue.

On latest version ([2.1.1](https://github.com/Adyen/adyen-ios/releases/tag/2.1.1)) a new callback (`willFinish`) was added to `CheckoutControllerDelegate`. 

You can do any remaining task when this method is invoked, as the checkout will not be dismissed until `completionHandler` is called.

Cheers,

PR #50, opened 2018-11-08
Change Card Validators and CardType enum access level to public

EugeneKutia:
**Intention**
Before v2.x.x update there was a public `CardValidator` class.
It's useful in case somebody needs to customise card input screen but still intends to use Adyen validators.
Also `CardType` have been public before v2.x.x allowing to have custom logic for a specific card type.

**Solution**
Would like to bring back a public access level for `Card Validators` and `CardType`  with this PR.

### DIFF
--- a/AdyenCard/Utilities/CardExpiryValidator.swift
+++ b/AdyenCard/Utilities/CardExpiryValidator.swift
@@ -4,13 +4,15 @@
 // This file is open source and available under the MIT license. See the LICENSE file for more info.
 //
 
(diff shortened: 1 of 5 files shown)


joostvandijk:
Hi @EugeneKutia,

Thanks for your contribution! We're working on making a public version of the card validator, and we plan to release this in the near future. It will be similar in functionality to the card validator in 1.0. The card validator in 2.0 isn't really designed for public use.

For now, I'll close this PR, as we should have a solution for this in the next couple releases.

Kind regards,

Joost

PR #96, opened 2019-07-31
Expose public initialiser for CardExpiryDateValidator, CardNumberValidator

Xanndarr:
In v3 of the framework release, you exposed the three validators in `AdyenCard/Validators/`, this adds public initialisers to these validators so they can be used in custom components


joostvandijk:
Hi @Xanndarr,

Thanks for the contribution! This is a good change and clearly an oversight on our part. We'll merge this internally so it can be part of the next release.

PR #101, opened 2019-08-20
Fail challenge if there is no existing transaction

moia-hari:
Added calling `ActionComponentDelegate.didFail(_)` incase there is no valid transaction when trying to call the handle Challenge method.

### Motivation and Context
We had this situation 
 - Tried to authorize a payment 
 - First, got the request to perform IdentifyShopper (good)
 - Performed and send back the result 
 - Second, requested to perform `ChallengeShopper` (expected)
 - Performed and send back the result and now things got weird
 - Third, requested again to perform `ChallengeShopper` ( 😕)
 - Calling SDK's `handle(_ action: ThreeDS2ChallengeAction)` on ThreeDS2Component just returned
 - The app keeps waiting for a callback to happen, but nothing ( 😞)

Wouldn't it be good if the call fails? ( 🤔)

### Where to start reviewing
- I am not sure if the `ComponentError.noActiveTransaction` is the best name for this scenario

### DIFF
--- a/Adyen/Components/Base/ComponentError.swift
+++ b/Adyen/Components/Base/ComponentError.swift
@@ -12,7 +12,7 @@ public enum ComponentError: Error {
     /// Indicates the component was cancelled by the user.
     case cancelled
 
-    /// Indicates the component failed to perform the intended action.
-    case failed
+    /// Indicates the component failed since there was not active transaction.
+    case noActiveTransaction
     
 }
--- a/Adyen/Components/Base/ComponentError.swift
+++ b/Adyen/Components/Base/ComponentError.swift
@@ -11,5 +11,8 @@ public enum ComponentError: Error {
     
     /// Indicates the component was cancelled by the user.
     case cancelled
+
+    /// Indicates the component failed to perform the intended action.
+    case failed
     
 }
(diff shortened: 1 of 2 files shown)


joostvandijk:
Hi @moia-hari,

Thanks for your contribution! A couple comments:
- This scenario shouldn't happen. `ChallengeShopper` should only be given once. Are you sure your requests/responses are correct? Do you use the right value for `paymentData` when submitting to `/payments/details`?
- Even if it shouldn't happen, we should handle this scenario. I do think it's a `ThreeDS2Component`-specific error, so we should probably add a `ThreeDS2ComponentError` enum instead of using the generic `ComponentError`.

I'll continue work on this in our internal repository, so we can add it in the next release.

moia-hari:
@joostvandijk 

Thank you for your comments. 

- That was my understanding too, currently, we are in talks with Adyen support to figure out what's going on here.
- Perfect, `ThreeDS2ComponentError` sounds better. I just wanted this scenario handled.

I will close this PR for now and wait for the next release from you.

moia-zsolt:
For the record: Android SDK has the same problem, I offered the same fix as a PR: https://github.com/Adyen/adyen-android/pull/97

joostvandijk:
Hi @moia-hari,

This issue has been addressed in yesterday's update (3.1.2).

Best,

Joost

PR #103, opened 2019-08-20
Fix component view controller child view layout

demalex:
Hi @adyen-git-manager! We are working with CC payments and faced one issue during view controller presentation. The issue appears if any payment view controller is presented modally (which uses `ComponentViewController` inside).

We can work around it by installing payment view controller as child view controller, but fix for the issue is very simple.

Please, have a look.

### DIFF
--- a/Adyen/UI/Controller/ComponentViewController.swift
+++ b/Adyen/UI/Controller/ComponentViewController.swift
@@ -47,13 +47,6 @@ public final class ComponentViewController: UIViewController {
     }
     
     /// :nodoc:
-    public override func viewWillLayoutSubviews() {
-        super.viewWillLayoutSubviews()
-        
-        children.first?.view.frame = view.bounds
-    }
-    
-    /// :nodoc:
     public override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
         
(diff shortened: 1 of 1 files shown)


joostvandijk:
Hi @demalex,

Thanks for your contribution! I'm not exactly sure what the use case is, are you embedding the view controller's view directly into your view hierarchy?

The change is quite minor and seems to just have the same behaviour, except for using auto layout instead of frame calculation. I'll make some minor changes (we still need the `loadView` to avoid a bug where an app's xib with the same name is loaded), but I'll merge this internally. I'm interested in hearing more about your usecase though.

Best,

Joost

demalex:
Hi @joostvandijk,

Thank you for feedback.

I see the point with `loadView`, I will bring it back.

The bug is present in `master` branch. I was checking `AdyenUIHost` example,  iOS 12.2, iPhone Xr Simulator. When the `Card` option is selected, `CardComponent`s view controller does not appear because child's view layout is not correct.

Our use case is different. We can't just use `CardComponent`s view controller, because this approach does not fit our architecture. So, we created container view controller and added component view controller as child view controller. This allows us to use our architecture and at the same time fix child layout problem. 

But, if you try to use (present) plain `CardComponent`s view controller modally but it is own, the bug appears.

Let me know if you need any additional information.

Thanks!

joostvandijk:
Hi @demalex,

I shipped your changes with the latest release, 3.1.2. Can you please check if this fixes your issue?

Best,

Joost

demalex:
Hi @joostvandijk,

Thank you very much! Issue is fixed! 🎉

PR #107, opened 2019-09-10
made init public

sonic555gr:
There are a few structures that are public but their initialisers are internal. 

We needed to create a RedirectComponent based on a WKWebView as we want control on how it looks like on screen. 

To parse the data from the returnURL we need to initialise a RedirectDetails Object but the init was internal instead of public. 

This is to fix the issue.

### DIFF
--- a/Adyen/Components/Redirect/RedirectDetails.swift
+++ b/Adyen/Components/Redirect/RedirectDetails.swift
@@ -16,7 +16,7 @@ public struct RedirectDetails: AdditionalDetails {
     ///
     /// - Parameter:
     ///   - returnURL: The URL through which the user returned to the app after a redirect.
-    internal init(returnURL: URL) {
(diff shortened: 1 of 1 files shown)


joostvandijk:
Hi @sonic555gr,

Thanks for your contribution, and my apologies for not getting back to you earlier. As with the other pull request, I'd like to know a little bit more about your usecase. Why do you still want to use `RedirectDetails` without the `RedirectComponent`? Is it just for parsing the query parameters?

Best,

Joost

sonic555gr:
@joostvandijk Yes. I recreated the component but a custom WebView. it just seems safer to use your SDK for as much as possible.

joostvandijk:
Hi @sonic555gr ,

Alright! I'll merge these changes internally and ship them with the next release.

Best,

Joost

PR #108, opened 2019-09-17
Fix: Call completion in `IssuerListComponent.stopLoading`

d4rkd3v1l:
Completion is called on `stopLoading` of any `PresentableComponent` but `IssuerListComponent`. 
Therefore it's just not possible to rely  on completion being called and will lead to a potentially broken app flow, if nonetheless doing so.


joostvandijk:
Thanks for the fix! We'll merge this in our internal repository and ship it with the next release.

PR #109, opened 2019-09-17
Card details public initializer

d4rkd3v1l:
In order to use our own custom `CardComponent` / ViewController we need to access the `CardDetail`'s initializer from "outside". And honestly I don't see, why it must be internal. But I'm open getting convinced of the contrary :-)

### DIFF
--- a/AdyenCard/CardDetails.swift
+++ b/AdyenCard/CardDetails.swift
@@ -36,7 +36,7 @@ public struct CardDetails: PaymentMethodDetails {
     ///   - paymentMethod: The used card payment method.
     ///   - encryptedCard: The encrypted card to read the details from.
     ///   - holderName: The holder name if available.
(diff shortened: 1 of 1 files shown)


joostvandijk:
Hi @d4rkd3v1l,

The whole concept of our components is that you use it with the provided UI. If you don't use our component's view controller, what use do you have for the `CardDetails`? It'd be good if I could know a little bit more about your usecase.

Best,

Joost

d4rkd3v1l:
We have our own CreditCard form, which is preferred by our designer and just fits better to our CI.
But for sure, we want to implement as least as possible, so therefore I think it's just awesome to be able to provide own components and more customized UIs, without having to implement everything by ourselves. 😀

PS: I removed the the other commit/PR, that shouldn't have been part of this. 😬

joostvandijk:
Hi @d4rkd3v1l,

I understand. I'll merge these changes internally and ship them with the next release.

Best,

Joost

PR #114, opened 2019-09-26
Check IBAN minimum length to avoid a crash on rearrange

anas10:
Hello,

The rearrange function is expecting 4 characters, however it does not check if a string with at least 4 characters is passed to it which causes a crash.

```
/// Rearranges the string to prepare for IBAN validation.
/// The country code and check numbers (the four initial characters) are moved to the end of the string.
/// The string is expected to be at least 4 characters.
private func rearrange(_ string: String) -> String {
```

I am creating this PR because we would like to be able to rely on the validators without writing extra checks on our end to prevents crashes inside an Adyen class.

Thanks you 😄


joostvandijk:
Thanks for this! I'll merge this with our internal repository and ship it with the next release.

PR #132, opened 2019-12-11
Added Swedish translations.

mattiosdev:


joostvandijk:
Hi @mattiosdev,

Thanks for your contribution! As this relates to an older version of our product, we'll not do an immediate update, but we can merge the files internally.

In the meantime, you can still use custom localization by having the strings present in your main app bundle. They will be picked up by the SDK (see [here](https://github.com/Adyen/adyen-ios/blob/v2/AdyenInternal/Assets/Localization.swift#L21)).

Best,

Joost

mattiosdev:
Hi,

I’ve added Estonian translations inside Localizable.strings file. Everything works perfectly but “Cancel” button is not translated, it’s still in English.

 “cancelButton” = “Tühista”;

Looks like it’s related to code inside Appearance.swift line 251:

func cancelButtonItem(target: Any, selector: Selector) -> UIBarButtonItem {
var cancelButtonItem: UIBarButtonItem
if let cancelButtonImage = navigationBarAttributes.cancelButtonImage {
cancelButtonItem = UIBarButtonItem(image: cancelButtonImage, style: .plain, target: target, action: selector)
cancelButtonItem.accessibilityLabel = ADYLocalizedString(“cancelButton")
} else {
cancelButtonItem = UIBarButtonItem(barButtonSystemItem: .cancel, target: target, action: selector)
}
return cancelButtonItem
}


I also tried to add image. Image works good but text is empty. When I go to Credit card view then back button is translated correctly. It looks like the problem is only with cancel button.

Best regards,
Mateusz Fraczek

Issue #135, opened 2020-01-14
AmountFormatter, maximumFractionDigits for COP should be 2


fiatdv:
Thanks for all your support. We are using your pod to format currency in our app, among other things. For Colombia, country code COP, iOS is returning 0 but it should be 2 according to https://en.wikipedia.org/wiki/ISO_4217. I believe that we need to add an entry for COP on your maximumFractionDigits as follows to fix the iOS COP error. Please advise. Thanks again.

maximumFractionDigits -

        case "COP":
            // iOS returns 0 instead.
            return 2

mohammedDehairy:
@fiatdv this issue should be solved with current version, will close the issue if you don't mind.

PR #136, opened 2020-01-14
COP ISO-4217 Support as iOS returns 0

fiatdv:
Solution for: https://github.com/Adyen/adyen-ios/issues/135


joostvandijk:
Hey @fiatdv,

Thanks for your contribution! This seems to be an issue indeed. We'll merge this internally and release it with the next update.

Best,

Joost

PR #148, opened 2020-03-20
Added support for Apple Pay billing and shipping address information.

airspeed:
**Motivation**

When one user already has configured a shipping/billing address in his wallet, we could avoid asking his shipping/billing information again after the payment is authorized with Adyen. This would save him time and effort.

**Solution**

Apple Pay already supports (with `PKPaymentAuthorizationViewController` over `PKPaymentRequest`) user-defined billing and shipping addresses, so we just need to expose those data in the `ApplePayDetails` data type.

**Alternatives considered**

The only alternative is like now to ask the user again after the payment authorization over Adyen for his billing/shipping contact information.


joostvandijk:
Hi @airspeed,

Thanks for your contribution! I think it looks very good, and it's definitely something we should have. Some comments:
- Could you add the `requiredBillingContactFields`/`requiredShippingContactFields` to the `DropInConfiguration.PaymentMethodsConfiguration.ApplePayConfiguration` as well, and pass those where the Drop-in instantiates the Apple Pay component?
- Can you add default values for `requiredBillingContactFields`/`requiredShippingContactFields` in the initializer to avoid a breaking API change?

Best,

Joost

airspeed:
> Hi @airspeed,
> 
> Thanks for your contribution! I think it looks very good, and it's definitely something we should have. Some comments:
> 
> * Could you add the `requiredBillingContactFields`/`requiredShippingContactFields` to the `DropInConfiguration.PaymentMethodsConfiguration.ApplePayConfiguration` as well, and pass those where the Drop-in instantiates the Apple Pay component?
> * Can you add default values for `requiredBillingContactFields`/`requiredShippingContactFields` in the initializer to avoid a breaking API change?
> 
> Best,
> 
> Joost

@joostvandijk 
Will definitely do, thank you! Hope to get it done through the weekend. 👍

airspeed:
@joostvandijk 
I have implemented the two requested points and also updated the test definitions.
Unfortunately the `PKContactField` API requires iOS 11, so I have also updated the  parameter docs to reflect this.

With some flexible naming and some additional duplication it would be probably possible to use `@available` compile time checks to make `requiredBillingContactFields/requiredShippingContactFields` only available to apps targeting iOS 11 and above, but I thought as an API vendor it would probably be more interesting for you to give this feature also to apps targeting iOS 10, although for their iOS >11 customers only.

joostvandijk:
Thanks @airspeed, looks good! We'll merge this internally and include this with the next release.

PR #205, opened 2020-08-12
Dev coios 250 mb way component

mohammedDehairy:
## Summary
1.  Renamed MBWayAwaitComponent to PollingAwaitComponent and MBWayActionDetails to AwaitActionDetails because they are not necessarily for MBWay.
2. Made small refactporing in PollingAwaitComponent to fix some bugs.
3. Unit tested PollingAwaitComponent.
4. Added headers and query parameters to the APIEnvronment protocol that can be injected in all requests performed by the APIClient's.
5. Fixed adyenPrint global function and made it simpler

## Tested scenarios
1. Fixed Unit tests.
2. Unit tested PollingAwaitComponent.

**Fixed issue**:  
1. Fixed an issue that happens if in the `PollingAwaitComponent` all the retries has been exhausted, and no conclusion reached, then nothing happens, I made so that either the `delegate.didFail` or `delegate.didProvide` is called.
2. Fixed an issue with AdyenPrint, where JSON is printed in an ugly way, also made the logic much simpler.
3. Fixed small issue in `SimpleScheduler` and `BackoffScheduler` where the maximum number of retries can be exceeded by one.


AdyenTests/Adyen Tests/Components/APIClientMock.swift
```diff
@@ -0,0 +1,33 @@
...
+
+enum DummyError: Error {
+    case dummy
+}
+
+final class APIClientMock: APIClientProtocol {
+
+    var mockedResponse: Response?
+
+    var mockedError: Error?
```
descorp, line 20: Maybe we can use a mocked sequence of network results, so we can also cover cases with multiple re-tries and "pending-pending-error" scenarios ?

Something like:

```
internal enum NetworkResult {
    case .responce(Responce)
    case .error(Error)
}

final class APIClientMock: APIClientProtocol {
    var mockedResults: [NetworkResult] = []
...
```

Adyen/Utilities/Environment.swift
```diff
@@ -12,6 +12,12 @@ public struct Environment: APIEnvironment {
     /// :nodoc:
     public var baseURL: URL
     
+    /// :nodoc:
+    public var headers: [String: String] = ["Origin": "https://test.org", "Content-Type": "application/json"]
+    
+    /// :nodoc:
+    public var queryParameters: [URLQueryItem] = [URLQueryItem(name: "token", value: "test_YP5RJK3TL5EELNPAS7RHIRVRAIMYZXOH")]
```
descorp, line 19: The `token` value is generated for a specific `origin in the backoffice, so it is different for different Accounts (and different `origin`).
It should be a parameter, I believe..

Adyen/Components/Await/AwaitComponent.swift
```diff
@@ -81,7 +81,7 @@ public final class AwaitComponent: AnyAwaitComponent {
             let scheduler = BackoffScheduler(queue: .main)
             let baseAPIClient = APIClient(environment: environment)
```
descorp, line 82: What about injecting `apiClient` into AwaitComponent? Why does it have to know how to generate one?
Maybe we can use a Builder pattern here, so AwaitComponent does not have to know about dependencies of any underling component?

Adyen/Utilities/Environment.swift
mohammedDehairy: the client key injection will be the subject my next pull request :-) , I don't want to make huge PR again :-D .

Adyen/Components/Await/AwaitComponent.swift
mohammedDehairy: I don't want to enable client code to inject the `apiClient`, for the `AwaitComponent` to work as expected the apiClient HAS to be a `RetryAPICliet` with a `BackoffScheduler`, `RetryAPICliet` with a `BackoffScheduler` is an integral part of how `AwaitComponent` work, and I enabled injecting the `apiClient` in an internal initializer for `Unit tests` only :-) .

AdyenTests/Adyen Tests/Components/APIClientMock.swift
mohammedDehairy: Thats a good idea actually, Thanks!

Adyen/Components/Await/AwaitComponent.swift
mohammedDehairy: please check the last commit, I Renamed `AnyAwaitComponent` into `AnyAwaitActionHandler`, moved the payment method specific `AwaitActionHandler` building into `AwaitActionHandlerProvider` builder and modified unit tests AwaitComponentTests

AdyenTests/Adyen Tests/Components/APIClientMock.swift
mohammedDehairy: done

PR #206, opened 2020-08-13
Implemented client key injection into PaymentMethodsConfiguration, an…

mohammedDehairy:
## Summary
Implemented client key injection into PaymentMethodsConfiguration, and propagating the client key to DropInComponent and all the paymentMethod components, also added unit tests.

## Tested scenarios
1. Tested `ComponentManager` injecting the client key into its created components.
2. Tested MBWay payment method decoding from `paymentMethods/` endpoint response.
3. Tested `ComponentManager` scrapping `MBWayComponent` in case client key is not configured.

### DIFF
--- a/Adyen/Components/Base/Component.swift
+++ b/Adyen/Components/Base/Component.swift
@@ -12,6 +12,10 @@ public protocol Component: AnyObject {
     /// Defines the environment used to make networking requests.
     var environment: Environment { get set }
     
+    /// The client key that corresponds to the webservice user you will use for initiating the payment.
+    /// See https://docs.adyen.com/user-management/client-side-authentication for more information.
+    var clientKey: String? { get set }
+    
 }
 
 public extension Component {
@@ -26,6 +30,17 @@ public extension Component {
         }
         set {
             objc_setAssociatedObject(self, &AssociatedKeys.environment, newValue, objc_AssociationPolicy.OBJC_ASSOCIATION_RETAIN_NONATOMIC)
+        }
+    }
+    
+    /// :nodoc:
+    var clientKey: String? {
+        get {
+            environment.clientKey
+        }
+        
+        set {
+            environment.clientKey = newValue
         }
     }
     
--- a/AdyenDropIn/DropInActionComponent.swift
+++ b/AdyenDropIn/DropInActionComponent.swift
@@ -102,7 +102,7 @@ public final class DropInActionComponent: ActionComponent {
     private func perform(_ action: AwaitAction) {
         guard environment.clientKey != nil else {
             // swiftlint:disable:next line_length
-            assertionFailure("Failed to instantiate AwaitComponent because client key is not configured, Please supply the client key in the PaymentMethodsConfiguration if using DropInComponent, or DropInActionComponent.environment.clientKey if using DropInActionComponent separately.")
+            assertionFailure("Failed to instantiate AwaitComponent because client key is not configured. Please supply the client key in the PaymentMethodsConfiguration if using DropInComponent, or DropInActionComponent.clientKey if using DropInActionComponent separately.")
             return
         }
         let component = AwaitComponent(style: awaitComponentStyle)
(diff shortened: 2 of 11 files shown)


AdyenDropIn/DropInActionComponent.swift
```diff
@@ -100,6 +100,11 @@ public final class DropInActionComponent: ActionComponent {
     }
     
     private func perform(_ action: AwaitAction) {
+        guard environment.clientKey != nil else {
+            // swiftlint:disable:next line_length
+            assertionFailure("Failed to instantiate AwaitComponent because client key is not configured, Please supply the client key in the PaymentMethodsConfiguration if using DropInComponent, or DropInActionComponent.environment.clientKey if using DropInActionComponent separately.")
```
descorp, line 105: `configured, Please supply ` -> `configured. Please supply `

descorp, line 105: `DropInActionComponent.environment.clientKey if using DropInActionComponent separately`

Maybe we can move `clientKey` as calculated property of DropInActionComponent? For convenience.
`environment.clientKey` is a bit more "internal" implementation detail..

AdyenDropIn/Utilities/ComponentManager.swift
```diff
@@ -130,7 +132,12 @@ internal final class ComponentManager {
         return component
     }
     
-    private func createMBWayComponent(_ paymentMethod: MBWayPaymentMethod) -> MBWayComponent {
+    private func createMBWayComponent(_ paymentMethod: MBWayPaymentMethod) -> MBWayComponent? {
+        guard configuration.clientKey != nil else {
+            // swiftlint:disable:next line_length
+            adyenPrint("Failed to instantiate MBWayComponent because client key is not configured, Please supply the client key in the PaymentMethodsConfiguration.")
```
descorp, line 138: `configured, Please` -> `configured. Please`

descorp, line 138: Should we mention DrpoInACtionComponent here, too?

mohammedDehairy: here its means that they are using dropInComponent, and `DropInComponent` is instantiating the `DrpoInACtionComponent` and making sure the client key is supplied to it.

mohammedDehairy: done

AdyenDropIn/DropInActionComponent.swift
mohammedDehairy: 👍 good idea

mohammedDehairy: 👍

Adyen/Components/Base/Component.swift
```diff
@@ -29,6 +33,17 @@ public extension Component {
         }
     }
     
+    /// :nodoc:
```
descorp, line 36: We can actually add here some user-friendly description. This is something that component implementers could possibly use..

Adyen/Components/Base/Component.swift
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2019 Adyen N.V.
+// Copyright (c) 2020 Adyen N.V.
```
descorp, line 2: This could be reverted 😅

Adyen/Components/Base/Component.swift
mohammedDehairy: there is documentation in the `Component` protocol for the `clientKey` var, right? we have kind of a convention that if a class or struct inherits from/implements a class/protocol, the overriding declaration/implementation should show :nodoc:, as the parent class/protocol has the documentation, so this falls under this rule/convention I guess.

Adyen/Components/Base/Component.swift
mohammedDehairy: reverted it.

PR #207, opened 2020-08-14
Stop Polling when the await component cancel button is clicked by the…

mohammedDehairy:
## Summary
Implemented stoping polling once the user cancels the `AwaitComponent`.

## Tested scenarios
Testing calling `didCancel()` on `PollingAwaitComponent` in `PollingAwaitComponentTests`

**Fixed issue**:  
Fixed the case when the user clicks the cancel button of the `AwaitComponent`, and stop polling

### DIFF
--- a/Adyen/Components/Base/PresentableComponent.swift
+++ b/Adyen/Components/Base/PresentableComponent.swift
@@ -13,8 +13,17 @@ public protocol Localizable {
     var localizationParameters: LocalizationParameters? { get set }
 }
 
+/// :nodoc:
+/// Represents any object than can handle a cancel event.
+public protocol Cancellable {
+    
+    /// :nodoc:
+    /// Called when the user cancels the component.
+    func didCancel()
+}
+
 /// A component that provides a view controller for the shopper to fill payment details.
-public protocol PresentableComponent: Component {
+public protocol PresentableComponent: Component, Cancellable {
     
     /// The payment information.
     var payment: Payment? { get set }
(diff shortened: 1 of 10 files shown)


AdyenDropIn/DropInComponent.swift
```diff
@@ -159,8 +159,9 @@ public final class DropInComponent: NSObject, PresentableComponent {
         }
     }
     
-    private func didSelectCancelButton(isRoot: Bool, component: PaymentComponent?) {
+    private func didSelectCancelButton(isRoot: Bool, component: PresentableComponent) {
         guard !paymentInProgress else { return }
+        component.didCancel()
```
descorp, line 164: This will do nothing, righ?
Default implementation is blank...

Adyen/Components/Base/PresentableComponent.swift
```diff
@@ -13,8 +13,17 @@ public protocol Localizable {
...
+public protocol Cancellable {
+    
+    /// :nodoc:
+    /// Called when the user cancels the component.
+    func didCancel()
+}
+
 /// A component that provides a view controller for the shopper to fill payment details.
-public protocol PresentableComponent: Component {
+public protocol PresentableComponent: Component, Cancellable {
```
descorp, line 26: Could you explain?
I like an idea, but I don't really see that it is used anywhere...

AdyenDropIn/DropInComponent.swift
mohammedDehairy: `PollingAwaitComponent` use it to stop polling.

Adyen/Components/Base/PresentableComponent.swift
mohammedDehairy: `PollingAwaitComponent` use it to stop polling.

PR #221, opened 2020-09-25
Fix binlookup

descorp:
### Changes

* fix payload encryption on BinLookup
* add throttling to BinLookup request

### DIFF
--- a/Adyen/Utilities/Throttler.swift
+++ b/Adyen/Utilities/Throttler.swift
@@ -1,0 +1,42 @@
+//
+// Copyright (c) 2020 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+// Throttles requests.
+public final class Throttler {
+    
+    private var workItem: DispatchWorkItem = DispatchWorkItem(block: {})
+    private var previousRun: Date = Date.distantPast
+    private let queue: DispatchQueue
+    private let minimumDelay: TimeInterval
+    
+    public init(minimumDelay: TimeInterval, queue: DispatchQueue = DispatchQueue.main) {
+        self.minimumDelay = minimumDelay
+        self.queue = queue
+    }
+    
+    /// Throttle a block of code after `minimumDelay`.
+    /// - Parameter block: Block of code to be throttled.
+    public func throttle(_ block: @escaping () -> Void) {
+        // Cancel any existing work item if it has not yet executed
+        workItem.cancel()
+        
+        // Re-assign workItem with the new block task, resetting the previousRun time when it executes
+        workItem = DispatchWorkItem {
+            [weak self] in
+            self?.previousRun = Date()
+            block()
+        }
+        
+        // If the time since the previous run is more than the required minimum delay
+        // => execute the workItem immediately
+        // else
+        // => delay the workItem execution by the minimum delay time
+        let delay = previousRun.timeIntervalSinceNow > minimumDelay ? 0 : minimumDelay
+        queue.asyncAfter(deadline: .now() + Double(delay), execute: workItem)
+    }
+}
--- a/AdyenCard/Components/Card/CardComponent.swift
+++ b/AdyenCard/Components/Card/CardComponent.swift
@@ -4,7 +4,7 @@
 // This file is open source and available under the MIT license. See the LICENSE file for more info.
 //
 
-import Adyen
+import Foundation
 
 /// Delegate for observing user's activity on `CardComponent`.
 public protocol CardComponentDelegate: class {
@@ -221,7 +221,7 @@ public final class CardComponent: PaymentComponent, PresentableComponent, Locali
                                       localizationParameters: localizationParameters)
         
         observe(item.$binValue) { [weak self] bin in
-            self?.binDidReceived(bin)
+            self?.didReceived(bin: bin)
         }
         
         item.identifier = ViewIdentifierBuilder.build(scopeInstance: self, postfix: "numberItem")
@@ -280,7 +280,7 @@ public final class CardComponent: PaymentComponent, PresentableComponent, Locali
         return footerItem
     }()
     
-    private func binDidReceived(_ bin: String) {
+    private func didReceived(bin: String) {
         let quickDetection = supportedCardTypes.adyen.types(forCardNumber: bin)
         self.securityCodeItem.selectedCard = quickDetection.first
         self.numberItem.numberFormatDidChange(detectedCards: quickDetection)
--- a/AdyenCard/Components/Card/CardComponent.swift
+++ b/AdyenCard/Components/Card/CardComponent.swift
@@ -4,7 +4,7 @@
 // This file is open source and available under the MIT license. See the LICENSE file for more info.
 //
 
-import Foundation
+import Adyen
 
 /// Delegate for observing user's activity on `CardComponent`.
 public protocol CardComponentDelegate: class {
@@ -28,6 +28,7 @@ public final class CardComponent: PaymentComponent, PresentableComponent, Locali
     
     private static let publicBinLenght = 6
     private let cardTypeProvider: CardTypeProvider
+    private let throttler = Throttler(minimumDelay: 0.5)
     
     /// Card Component errors.
     public enum Error: Swift.Error {
@@ -220,10 +221,7 @@ public final class CardComponent: PaymentComponent, PresentableComponent, Locali
                                       localizationParameters: localizationParameters)
         
         observe(item.$binValue) { [weak self] bin in
-            guard let self = self else { return }
-            
-            self.requestCardType(for: bin, update: item)
-            self.cardComponentDelegate?.didChangeBIN(String(bin.prefix(CardComponent.publicBinLenght)), component: self)
+            self?.binDidReceived(bin)
         }
         
         item.identifier = ViewIdentifierBuilder.build(scopeInstance: self, postfix: "numberItem")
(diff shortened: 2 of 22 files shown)


AdyenCard/Components/Card/CardComponent.swift
```diff
@@ -4,7 +4,7 @@
 // This file is open source and available under the MIT license. See the LICENSE file for more info.
 //
 
-import Foundation
+import Adyen
```
mohammedDehairy, line 7: why is this changed?

AdyenCard/Components/Card/CardComponent.swift
```diff
@@ -220,10 +221,7 @@ public final class CardComponent: PaymentComponent, PresentableComponent, Locali
                                       localizationParameters: localizationParameters)
         
         observe(item.$binValue) { [weak self] bin in
-            guard let self = self else { return }
-            
-            self.requestCardType(for: bin, update: item)
-            self.cardComponentDelegate?.didChangeBIN(String(bin.prefix(CardComponent.publicBinLenght)), component: self)
+            self?.binDidReceived(bin)
```
mohammedDehairy, line 224: `self?.didReceive(bin)` sounds much better IMO :-)

AdyenCard/Components/Card/CardComponent.swift
```diff
@@ -282,12 +280,22 @@ public final class CardComponent: PaymentComponent, PresentableComponent, Locali
...
+    
+    private func requestCardTypes(for bin: String) {
         cardTypeProvider.requestCardType(for: bin, supported: self.supportedCardTypes) { [weak self] cardTypes in
             guard let self = self else { return }
             
-            self.securityCodeItem.selectedCard = cardTypes.first
-            item.detectedCardsDidChange(detectedCards: cardTypes)
+            self.numberItem.detectedCardsDidChange(detectedCards: cardTypes)
             self.cardComponentDelegate?.didChangeCardType(cardTypes, component: self)
         }
```
mohammedDehairy, line 300: what happens if there is no internet connection, or if the call failed for any reason, should we report the locally detected card types?

Adyen/Utilities/Throttler.swift
```diff
@@ -0,0 +1,42 @@
+//
+// Copyright (c) 2020 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+// Throttles requests.
+public final class Throttler {
```
mohammedDehairy, line 10: why do we need to throttle bin lookup calls?

AdyenCard/Components/Card/CardComponent.swift
descorp: it will do so. there are two fallback inside `cardTypeProvider ` : on "no publicKey" and "unsuccessful binLookup request"

Adyen/Utilities/Throttler.swift
descorp: you type 16 digits + make mistakes and corrections. Every change is a network call.
Throttling will save you some network data.

descorp: Alternatively, we can move it into CardNumberItem 🤔

AdyenCard/Components/Card/CardComponent.swift
descorp: good question :D

AdyenCard/Components/Card/CardComponent.swift
descorp: 👌

Adyen/Utilities/Throttler.swift
mohammedDehairy: hmm 🤔 not sure if its worth the complexity TBH, it will save the shopper maximum 100KB of data? is this worth a delay 0.5 between calls and the complexity? this is not a rhetorical question 😊.

mohammedDehairy: ok, I guess it make sense :-)

AdyenCard/Components/Card/CardComponent.swift
```diff
@@ -282,12 +280,22 @@ public final class CardComponent: PaymentComponent, PresentableComponent, Locali
...
+    }
+    
+    private func requestCardTypes(for bin: String) {
         cardTypeProvider.requestCardType(for: bin, supported: self.supportedCardTypes) { [weak self] cardTypes in
             guard let self = self else { return }
             
-            self.securityCodeItem.selectedCard = cardTypes.first
-            item.detectedCardsDidChange(detectedCards: cardTypes)
+            self.numberItem.detectedCardsDidChange(detectedCards: cardTypes)
             self.cardComponentDelegate?.didChangeCardType(cardTypes, component: self)
```
mohammedDehairy, line 299:
```
self.numberItem.didChange(detectedCards: cardTypes)
```
would be more swifty IMO.

also 

```
self.cardComponentDelegate?.didChange(cardTypes, component: self)
```

but I guess its too late for this one, right?

Adyen/Utilities/Throttler.swift
mohammedDehairy, line 10: does this need to be `public`?

descorp: nope, it only used in AdyenCard

AdyenCard/Form/FormCardNumberItem.swift
```diff
@@ -49,21 +49,20 @@ internal final class FormCardNumberItem: FormTextItem {
...
         }
     }
     
-    internal func detectedCardsDidChange(detectedCards: [CardType]) {
+    internal func didChange(detectedCards: [CardType]) {
         let newDetectedCardTypes = Set<CardType>(detectedCards)
         guard self.previouslyDetectedCardTypes != newDetectedCardTypes else { return }
         
         previouslyDetectedCardTypes = newDetectedCardTypes
-        cardNumberFormatter.cardType = newDetectedCardTypes.first
```
mohammedDehairy, line 66: why remove this line? the formatter should adjust to the true card types detected by the backend, right?

AdyenCard/Utilities/CardType/CardTypeProvider.swift
```diff
@@ -32,45 +38,50 @@ internal final class CardTypeProvider: Component {
     /// - Parameters:
     ///   - bin: Card's BIN number. If longer than `minBinLength` - calls API, otherwise check local Regex,
     ///   - caller:  Callback to notify about results.
```
mohammedDehairy, line 40: maybe change the docs as well: `caller -> completion`

descorp: It moved to line 52. `didChange (detectedCards:)` is not looking nicely under throttling, so I've moved it out.

AdyenCard/Components/Card/CardComponent.swift
descorp: renamed

AdyenTests/Card Tests/Mocks/PublicKeyProviderMock.swift
```diff
@@ -1,9 +1,7 @@
...
-//  PublicKeyProviderMock.swift
-//  AdyenTests
+// Copyright (c) 2020 Adyen N.V.
 //
-//  Created by Vladimir Abramichev on 24/08/2020.
-//  Copyright © 2020 Adyen. All rights reserved.
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
 //
 
 @testable import Adyen
```
mohammedDehairy, line 7: do we still need this :-) ?

AdyenTests/Card Tests/Utilities/CardEncryptorCardTests.swift
```diff
@@ -145,4 +145,36 @@ class CardEncryptorCardTests: XCTestCase {
 
         XCTAssertNotNil( try? CardEncryptor.encryptedToken(for: card, holderName: nil, publicKey: key))
     }
+
+    // MARK: - Test encrypting BIN
+
+    func testEncryptExpiryBINShouldReturnNilWithNilBIN() {
```
mohammedDehairy, line 151: I think the name should be `testEncryptBINShouldReturnNilWithEmptyBIN`, right?

AdyenCard/Form/FormCardNumberItem.swift
descorp: It moved to line 52. `cardNumberFormatter` is mostly detecting Amex / no-Amex, so local regex-based recognition is more appropriate.

AdyenTests/Card Tests/Utilities/CardTypeProviderTests.swift
```diff
@@ -1,9 +1,7 @@
...
-//  CardTypeProviderTests.swift
-//  AdyenTests
+// Copyright (c) 2020 Adyen N.V.
 //
-//  Created by Vladimir Abramichev on 24/08/2020.
-//  Copyright © 2020 Adyen. All rights reserved.
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
 //
 
 @testable import AdyenCard
```
mohammedDehairy, line 7: empty file :-)

AdyenCard/Utilities/CardType/CardTypeProvider.swift
descorp: ups :D

AdyenTests/Card Tests/Mocks/PublicKeyProviderMock.swift
descorp: copy paste..

AdyenTests/Card Tests/Utilities/CardTypeProviderTests.swift
descorp: what do you mean?

AdyenTests/Card Tests/Utilities/CardEncryptorCardTests.swift
descorp: why not

AdyenCard/Form/FormCardNumberItem.swift
mohammedDehairy: 👍 ok

AdyenTests/Card Tests/Utilities/ThrottlerTests.swift
```diff
@@ -0,0 +1,33 @@
...
+            }
+        }
+
+        DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(1)) {
+            expectation.fulfill()
+        }
+
+        wait(for: [expectation], timeout: 2)
+        XCTAssertEqual(counter, 1)
+    }
```
mohammedDehairy, line 30: nice test 👍

AdyenTests/Card Tests/Utilities/CardTypeProviderTests.swift
mohammedDehairy: sorry, I mean, why change the file header comments?

descorp: it is nicer :D

mohammedDehairy (approved):

PR #222, opened 2020-09-29
Fixed mb way component

mohammedDehairy:
## Summary
 1. Remove MBWay email field.
 2. Fixe an issue with DropInNavigationController, where the preferred content size of content view controllers can be wrong.
 3. Made the Phone number field in MBWayComponent accept any phone number from anywhere in the world.
 4. Fixed Unit tests to reflect that any phone number is accepted.
 5. Created a generic RegularExpressionValidator and made EmailValidator and PhoneNumberValidator inherit from it.

## Tested scenarios
Covered the new logic with unit tests and modified the old unit tests.

### DIFF
--- a/Adyen/Formatters/PhoneNumberFormatter.swift
+++ b/Adyen/Formatters/PhoneNumberFormatter.swift
@@ -1,0 +1,34 @@
+//
+// Copyright (c) 2020 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+/// :nodoc:
+open class PhoneNumberFormatter: Adyen.Formatter {
+    
+    /// :nodoc:
+    public init() {}
+    
+    /// :nodoc:
+    open func formattedValue(for value: String) -> String {
+        return sanitizedValue(for: value)
+    }
+    
+    /// :nodoc:
+    open func sanitizedValue(for value: String) -> String {
+        let sanitizedString = value.filter { isValid(character: $0) }
+        
+        // Converts all non-latin digits to latin digits
+        let latinString = sanitizedString.applyingTransform(StringTransform.toLatin, reverse: false) ?? sanitizedString
+        
+        // Filters out non-ASCII digits
+        return latinString.filter { $0.isASCII && isValid(character: $0) }
+    }
+    
+    private func isValid(character: Character) -> Bool {
+        character.isNumber || character == "+"
+    }
+}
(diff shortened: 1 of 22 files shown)


Adyen/Validators/RegularExpressionValidator.swift
```diff
@@ -0,0 +1,32 @@
+//
+// Copyright (c) 2020 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+/// Validates a phone number.
```
descorp, line 9: Comment is outdated :D

AdyenDropIn/Presentation/DropInNavigationController.swift
```diff
@@ -49,13 +49,15 @@ internal final class DropInNavigationController: UINavigationController {
     // MARK: - Keyboard
     
     @objc private func keyboardWillChangeFrame(_ notification: NSNotification) {
-        guard let topViewController = topViewController as? WrapperViewController,
-            topViewController.requiresInput,
-            let bounds = notification.userInfo?[UIResponder.keyboardFrameEndUserInfoKey] as? CGRect
-        else { return }
+        if let bounds = notification.userInfo?[UIResponder.keyboardFrameEndUserInfoKey] as? CGRect {
```
descorp, line 52: I believe there was cases when updating `keyboardRect ` is not needed like stored card payment.

Adyen/Formatters/PhoneNumberFormatter.swift
```diff
@@ -0,0 +1,34 @@
+//
+// Copyright (c) 2020 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+/// :nodoc:
+open class PhoneNumberFormatter: Adyen.Formatter {
```
descorp, line 10: Are we going to use this in Qiwi payment as well?

AdyenDropIn/Presentation/DropInNavigationController.swift
mohammedDehairy: stored payment methods works fine, also i had a problem with AwaitComponent where the preferred height takes into account the keyboard height even though the await view has no fields, and when investigated, found out the thekeyboardRect wasn't reset when leaving the MBWay form, hence the change.

Adyen/Formatters/PhoneNumberFormatter.swift
mohammedDehairy: don't think so, no need to change Qiwi

Adyen/Validators/RegularExpressionValidator.swift
mohammedDehairy: 👍  fixed

descorp (approved):

PR #226, opened 2020-10-01
Swift encryption

descorp:
### Changes

* reduce Obj-c code
* rewrite RSA from Obj-c into Swift
* code cleanup


AdyenTests/Adyen Tests/Components/APIClientMock.swift
```diff
@@ -15,7 +15,7 @@ enum Dummy: Error {
     case dummyError
 
     /// This is not a real public key, this is just a random string with the right pattern.
-    internal static var dummyPublicKey = "59554|59YWNVYSILHQWVXSIYY8XVK5HMLEAFT2JPJBDVHUD2798K12GKE652PYLJYYNBR0HVN0AYLC38VIU0TSBC9JTQZ4AHOHPPIGVH985H6EI5HAFZXZAM0QIXBAYEP180X0MM6HRHZONIM62TI9US8NXHXNKYSRE8ASJLY3KED6KDD6SY4I29CUY5FYTN8XEQ8NS8M0ECUAG0GV08XAX19HEX8IQ35SNRY8P9G0YOTTEFYC8QGM7N4PYRUWTSOEJV8W9AKJ8ZLR851OA0P7NZOJXZ2EOYNWSORS9RL4HGXVXGANDYXOWCD7XYPHJD6EPYGRUDV87EOT5FHR574DJW5881Y88Y2QR6R9W1WG5N0CV3WJGELJ971OR0S0PTKHOFW7PXRRDVQU1TT4Q8KJJLZ2VHS1BYP0VFQY1FOADWZ2YPGXDT6KPSN6OJ81G9B9BO7LMGYIONUDWQZQM41O27RROX44I89WRLHZHNYP5NEF2ACTF1AJHA4SNTUN9Z93HYQ2"
+    internal static var dummyPublicKey = "1B2F|B22CD14EF451C993318F5BDE9DA1AAEE32269D1D13590DC23761113F0E01D7D0A19910BA1DDE9348C29DCE4ED32AB121ABF6C9AF771DAB547BD4261A50FCE3686747D581455A31C166CCD84E0B3FAB2B23F9E8CB69864C520C60247A5E64C669D0719A31132AF3CEE94DB56AB2840B967FCFC65DFAD092121AE8C6435EAB95D1"
```
mohammedDehairy, line 18: why did you change this?
unit tests has to pass with the same old example, changing the example to fit the new code defies the purpose of unit tests, right?.

not to mention the new key doesn't match the old key length, why is that?

AdyenTests/Card Tests/Utilities/CardEncryptorCardTests.swift
```diff
@@ -21,14 +25,13 @@ class CardEncryptorCardTests: XCTestCase {
         let card = CardEncryptor.Card(number: "test_number")
         let key = "test_invalid_key"
         XCTAssertThrowsError(try card.encryptedNumber(publicKey: key, date: Date())) { error in
-            XCTAssertTrue(error is CardEncryptor.Error, "Thrown Error is not CardEncryptor.Error")
-            XCTAssertEqual(error as! CardEncryptor.Error, CardEncryptor.Error.encryptionFailed, "Thrown Error is not CardEncryptor.Error.encryptionFailed")
-            XCTAssertEqual(error.localizedDescription, CardEncryptor.Error.encryptionFailed.errorDescription)
+            XCTAssertTrue(error is Cryptor.Error, "Thrown Error is not CardEncryptor.Error")
+            XCTAssertEqual(error as! Cryptor.Error, Cryptor.Error.rsaEncryptionError, "Thrown Error is not CardEncryptor.Error.encryptionFailed")
+            XCTAssertEqual(error.localizedDescription, Cryptor.Error.rsaEncryptionError.errorDescription)
```
mohammedDehairy, line 30: Error types are part of the public interface, will client code be broken by this change?

AdyenTests/Adyen Tests/Components/APIClientMock.swift
descorp: Basically this is fully functional RSA modulus and exponent. The old `dummyPublicKey` was a random set of data.
I have changed it so we can actually test the encryption code in unit tests.

Does it have to match? I can do the same length, no issues with that.
In theory, if you know a structure of public key modulus and exponent, you can try to perform an attack on RSA.
So I just did a `dummyPublicKey` that looks like nothing but our actual card encryption key :D

AdyenTests/Card Tests/Utilities/CardEncryptorCardTests.swift
descorp: It is possible if client code expect to catch a specific type of the error.
I don't think it is a common thing to do, but to be on a safe side I can wrap `Cryptor.Error` into `CardEncryptor.Error. encryptionFailed(error: Cryptor.Error)`

mohammedDehairy: why would you think its uncommon, many clients build their own Card forms and use this code to encrypt card details before sending to us, I don't see a reason why they would be uncommon.

AdyenTests/Adyen Tests/Components/APIClientMock.swift
mohammedDehairy: don't think card public key length is variable, as far as I know the card public key has to match this regex "([0-9]){5}\|([A-Z]|[0-9]){512}", exactly 5 digits + "|" + 512 alphanumeric characters, Android and iOS use this regex to validate the card public key provided by the client.

and knowing the structure of the public key is not a problem as at all, RSA cryptography is not that fragile :-D , otherwise no point of encryption, tbh I have no idea what you're talking about :-D .

AdyenTests/Card Tests/Utilities/CardEncryptorCardTests.swift
descorp: Done. `CardEncryptor.Error.encryptionFailed` is deprecated :D

AdyenTests/Adyen Tests/Components/APIClientMock.swift
descorp: Fixed

AdyenCard/Utilities/Encryption/CardEncryptor.swift
```diff
@@ -73,6 +25,8 @@ public enum CardEncryptor {
         
     }
     
+    // MARK: - Encrypts a card
```
mohammedDehairy, line 28: `Card Encryption` I think is better

AdyenCard/Utilities/Encryption/CardEncryptor.swift
```diff
@@ -283,3 +229,99 @@ public enum CardEncryptor {
         
     }
 }
+
+extension CardEncryptor {
+
+    public struct Bin: Encodable {
```
mohammedDehairy, line 235: this should be `internal` I think

AdyenTests/Card Tests/Utilities/CardEncryptorCardTests.swift
mohammedDehairy: still the merchants expect to receive `CardEncryptor.Error`, if its deprecated but is not used at all, it is still breaking change.

if an encryption error happens and merchants expecting `CardEncryptor.Error`, then their code is broken.

descorp: Ok. I'll hide `Cryptor.Error` behind `CardEncryptor.Error.encryptionFailed`

AdyenCard/Utilities/Encryption/CardEncryptor.swift
descorp: 👍

AdyenCard/Utilities/Encryption/CardEncryptor.swift
descorp: 👍 
even `private` is fine

mohammedDehairy (approved):

PR #231, opened 2020-10-05
SwiftUI adaptation

mohammedDehairy:
## Summary
1. Created a custom view modifier to present any UIViewController modally from SwiftUI view.
2. Created a sample application for SwiftUI.
3. Moved the common example code between UIKit app and swiftUI app into PaymentsController.


descorp (approved):

MattCheetham:
Thanks for this. Just having a go at working with it now. I'm manually compiling and importing the Adyen frameworks. Do I need to re-compile just the top level Adyen.framework or are there changes throughout? It looks like there might be no changes at all, just example code. I'm just trying to present the drop-in SDK so will have a look now to see if I can work it out from the example code.

mohammedDehairy:
Hi @MattCheetham ,

 The changes related to SwiftUi adaptation is only in `Adyen.framework`.

Mohamed

MattCheetham:
This seems to be working well for us. Definitely looks and behaves much better and I was able to get a payment through and processed. We've got a slight hiccup our side with 3DS2 so I can't test out that part yet but will hopefully be able to by the end of the week. Thanks for your work on this, it's kept us on track with our release plan and made our lives much easier.

MattCheetham:
Can confirm we got this all up and running for us and we're planning to launch shortly. The only issue we ran into was that due to the way the view is presented with SwiftUI, we were updating and object in the background after submitting card payment details and it was causing SwiftUI to re-render the view which caused the drop-in the view controller to immediately be dismissed.

We've worked around it by avoiding updating any objects that our checkout view relies on until payment is 100% complete (post 3DS challenge!) but just wanted to flag in case you see other users running into the same issue. Not entirely sure what you could do about that really.

mohammedDehairy:
Thanks @MattCheetham  for sharing the issue, if you can describe how to reproduce it, we'll try to mitigate it.

MattCheetham:
In SwiftUI, any modifications to @State or @Published objects cause a full recreation of the SwiftUI view stack.

For us, we added the `.present` method to our view and when the user hits pay, we set the view controller and the user begins payment. After the user enters their card details and we get the `didSubmit` call, we make multiple requests to our back end to process the payment. One of these calls updates the `checkout` object that powers some of the UI in the view where were are presenting the adyen view controller from. In updating this object, SwiftUI calls a redraw on the view which unsets the viewcontroller property, causing the view to be dismissed.

I'm fairly new to SwiftUI so it's entirely possible this is something that can be avoided and we found a way to avoid updating our checkout object until later anyway so it's not an issue for us right now but worth me writing down in case someone else has a similar problem in the future.

PR #239, opened 2020-10-14
Fix: improve card type detection visualisation

descorp:
### Changes

* CardNumberItem is no longer responsible for card type logo presentation logic - simply shows what is in there. All logic for limiting of max card type logos shown, invalid and empty states are moved to CardComponent. 
* rename some internal API
* refactor some internal logic

### DIFF
--- a/AdyenCard/Components/Card/CardComponent.swift
+++ b/AdyenCard/Components/Card/CardComponent.swift
@@ -29,6 +29,7 @@ public final class CardComponent: PaymentComponent, PresentableComponent, Locali
     /// :nodoc:
     internal var cardTypeProvider: AnyCardTypeProvider
     
+    private var topElements: [CardType]
     private static let publicBinLenght = 6
     private let throttler = Throttler(minimumDelay: 0.5)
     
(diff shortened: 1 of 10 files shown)


AdyenTests/Card Tests/Form/FormCardNumberItemTests.swift
```diff
@@ -90,19 +90,19 @@ class FormCardNumberItemTests: XCTestCase {
         
         // Clearing the field should bring back both logos.
         item.value = ""
-        cardTypeProvider.requestCardType(for: item.value, supported: supportedCardTypes) { response in
-            item.didChange(detectedCards: response, for: item.value)
```
descorp, line 94: `FormCardNumberItem ` no longer responsible for handling visual state. If no cards recognised - no card shown.

AdyenCard/Components/Card/CardComponent.swift
```diff
@@ -29,6 +29,7 @@ public final class CardComponent: PaymentComponent, PresentableComponent, Locali
     /// :nodoc:
     internal var cardTypeProvider: AnyCardTypeProvider
     
+    private var topElements: [CardType]
```
mohammedDehairy, line 32: can we have a more descriptive name? `topElements` can be any thing.

descorp: updated

mohammedDehairy (approved):

mohammedDehairy (approved):

PR #240, opened 2020-10-15
Support Swift Package manager

mohammedDehairy:
## Summary
1. Created the Package.swift file.
2. Added the different products and targets.
3. Modified many swift files to explicitly import other targets they depend on.

## Tested scenarios
1. Tested 3DS2 SDK integration for debug builds and release builds.
2. Tested WeChat pay.
3. Tested building release builds for the SDK.


.github/workflows/pr_scan.yml
```diff
@@ -5,27 +5,30 @@ on:
...
   build:
     env:
       sonarToken: ${{ secrets.SONAR_TOKEN }}
     runs-on: macos-latest
     steps:
     - uses: actions/checkout@v2
-    
+    - uses: maxim-lobanov/setup-xcode@v1
+      with:
+        xcode-version: latest-stable
```
descorp, line 17: let's not use any 3rd party code.
Just set `echo '::set-env name=XCODE_VERSION::/Applications/Xcode_12.1.app/Contents/Developer'`

AdyenCard/Components/Card/CardComponentExtensions.swift
```diff
@@ -107,26 +108,9 @@ extension CardComponent {
...
-        }
-    }
-    
-    /// Initializes the card component for stored cards.
-    ///
-    /// - Parameters:
-    ///   -  paymentMethod: The stored card payment method.
-    ///   -  publicKey: The key used for encrypting card data.
-    ///   -  style: The Component's UI style.
-    @available(*, deprecated, message: "Use init(paymentMethod:clientKey:style:) instead.")
```
descorp, line 126: Why (only) this deprecated method is removed?
I don't see "deprecated: ***" commit message :P

AdyenCard/Utilities/Encryption/Extensions/StringExtension.swift
```diff
@@ -5,29 +5,18 @@
 //
 
 import CommonCrypto
+import Foundation
 
 extension String {
 
-    internal func sha1() -> String {
```
descorp, line 11: Why this is removed?

Common Example code/Domains/Components/ComponentsViewController.swift
```diff
@@ -6,7 +6,7 @@
 
 import UIKit
 
-internal final class ComponentsViewController: UIViewController, Presenter {
+internal class ComponentsViewController: UIViewController, Presenter {
```
descorp, line 9: why `final` removed ?

AdyenCard/Components/Card/CardComponentExtensions.swift
mohammedDehairy: The removed function is a duplicate of another function:

one has this signature:

```
(paymentMethod: AnyCardPaymentMethod, publicKey: String, style: FormComponentStyle = FormComponentStyle())
```

and the other has this signature:

```
(paymentMethod: StoredCardPaymentMethod, publicKey: String, style: FormComponentStyle = FormComponentStyle())
```

and `StoredCardPaymentMethod ` is a `AnyCardPaymentMethod `, so code beat flagged this as code duplication.

AdyenCard/Utilities/Encryption/Extensions/StringExtension.swift
mohammedDehairy: its not used :-) , right?

Common Example code/Domains/Components/ComponentsViewController.swift
mohammedDehairy: removed the final qualifier to reuse the code in the deleted AdyenSPMUIHost target, but since the target is deleted, we can put back `final`.

AdyenCard/Components/Card/CardComponentExtensions.swift
descorp: ok, sounds fair :D

AdyenCard/Utilities/Encryption/Extensions/StringExtension.swift
descorp: ok, good point! Indeed it is no longer used!

descorp (approved):

PR #254, opened 2020-11-03
Quick fix

descorp:
### Changes

* quick fix for AppelPay dissmis

### DIFF
--- a/Adyen/Components/Apple Pay/ApplePayComponentExtensions.swift
+++ b/Adyen/Components/Apple Pay/ApplePayComponentExtensions.swift
@@ -14,9 +14,11 @@ extension ApplePayComponent: PKPaymentAuthorizationViewControllerDelegate {
     
     /// :nodoc:
     public func paymentAuthorizationViewControllerDidFinish(_ controller: PKPaymentAuthorizationViewController) {
-        controller.dismiss(animated: true) { [dismissCompletion] in
+        controller.dismiss(animated: true) { [weak self] in
+            guard let self = self else { return }
+            
             self.delegate?.didFail(with: ComponentError.cancelled, from: self)
-            dismissCompletion?()
+            self.dismissCompletion?()
             self.dismissCompletion = nil
         }
         self.paymentAuthorizationViewController = nil


mohammedDehairy (approved):

MattCheetham:
Hey @descorp this change has re-broken our Apple Pay closure from being called. Are you able to give any context to the reason for switching it back? Just trying to understand why this works for you and not me.

descorp:
hm... 

@MattCheetham  what Xcode / Swift version are you using?

We've faced an issue on DropIn with only local retain '[dismissCompletion]' on Xcode 12.0 (12A7209)

MattCheetham:
I'm running Xcode 12.1 (12A7403) with Swift 5.

I tried moving it to deinit and it had no effect, in fact, it seems that deinit was never even called on ApplePayComponent. I'll keep looking and see if I can find another solution.

MattCheetham:
I found another solution. As long as `self.dismissCompletion?()` is called before `self.delegate?.didFail(with: ComponentError.cancelled, from: self)` I have no issues.

The whole code block would look like:
```    
        controller.dismiss(animated: true) { [weak self] in

            self.dismissCompletion?()
            self.dismissCompletion = nil
        }
        self.paymentAuthorizationViewController = nil
    }
```

This is because calling the `didFail` delegate calls `didFail` on `DropInComponent` which itself calls `stopLoading(withSuccess: false)`. Essentially calling stop with a nil completion handler.

This in turn calls `rootComponent.stopLoading(withSuccess: success, completion: completion)` which will be the method on ApplePayComponent. 

The implementation of `stopLoading` on `ApplePayComponent` sets `dismissCompletion = completion`, which is nil. Thereby nilling out the closure before I can then call it. By moving the closure call above the didFail call, it should fix the issue.

It's worth noting that I'm getting this on a failed transaction (Just finishing getting set up with a test device now to try a successful one) so it may only be affecting a failure. I of course may also be completely wrong but my trailing through the SDK seems to point to this as the cause.

descorp:
@MattCheetham 

Indeed, there is a race condition: the whole flow relays on a single fact that **PKPaymentAuthorizationViewController** calls `func paymentAuthorizationViewControllerDidFinish(_ controller:)` multiple times.
The complication occurs because ApplePay are the only *PaymentComponent* that needs to dismiss itself...

`rootComponent.stopLoading(withSuccess:, completion:)` sets a `dismissCompletion`; however `rootComponent.stopLoading(withSuccess:, completion:)` is called after the dismissal 🤯

Could you check if [this PR](https://github.com/Adyen/adyen-ios/pull/256) solves it for you?

PR #272, opened 2020-11-17
feature: zero payment amount is treated as a 0 authorization transaction and the button title shows a proper title as web components

mohammedDehairy:
feature: zero payment amount is treated as a 0 authorization transaction and the button title shows a proper title as web components.

### DIFF
--- a/Adyen/Utilities/Localization.swift
+++ b/Adyen/Utilities/Localization.swift
@@ -89,16 +89,40 @@ private func attempt(_ input: LocalizationInput) -> String? {
     return nil
 }
 
+/// :nodoc:
+public enum PaymentStyle {
+    case needsRedirectToThirdParty(String)
+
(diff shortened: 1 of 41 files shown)


Adyen/Utilities/Localization.swift
```diff
@@ -80,16 +80,42 @@ private func attempt(_ input: LocalizationInput) -> String? {
     return nil
 }
 
+/// :nodoc:
+public enum PaymentStyle {
+    /// Indicates that the payment method needs a redirect to a third party app or website like MBWay.
+    case needsRedirectToThirdParty(String)
```
descorp, line 86: `needsRedirectToThirdParty ` is never used ?

descorp (dismissed):

descorp (approved):

PR #276, opened 2020-11-20
chore: Small fixes to unit tests and fixed many warnings

mohammedDehairy:
chore: Small fixes to unit tests and fixed many warnings

### DIFF
--- a/AdyenCard/Utilities/Encryption/Cryptor/AES.swift
+++ b/AdyenCard/Utilities/Encryption/Cryptor/AES.swift
@@ -155,13 +155,14 @@ extension Cryptor.AES {
                        blockX: blockX)
 
             // encrypt
-            AESCCM.encryptXor(key: key,
-                              encodeBytes: encodeBytesLength,
-                              counter: counter,
-                              pointer: pointer,
-                              length: count,
-                              blockA: blockA,
-                              blockS: blockS)
+            let context = AESCCM.EncryptXorContext(key: key,
+                                                   encodeBytes: encodeBytesLength,
+                                                   counter: counter,
+                                                   pointer: pointer,
+                                                   length: count,
+                                                   blockA: blockA,
+                                                   blockS: blockS)
+            AESCCM.encryptXor(context: context)
         }
 
         private enum AESCCM {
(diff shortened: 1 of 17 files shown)


AdyenTests/Card Tests/CardComponentTests.swift
```diff
@@ -247,7 +247,7 @@ class CardComponentTests: XCTestCase {
         let method = CardPaymentMethod(type: "bcmc", name: "Test name", fundingSource: .credit, brands: ["any_test_brand_name"])
         let sut = CardComponent(paymentMethod: method,
                                 clientKey: "test_client_key")
-        sut.showsLargeTitle = false
+        sut._showsLargeTitle = false
```
descorp, line 250: I think we can delete those :)

AdyenTests/Adyen Tests/Components/Apple Pay/ApplePayComponentTest.swift
```diff
@@ -69,6 +69,23 @@ class ApplePayComponentTest: XCTestCase {
         waitForExpectations(timeout: 2)
     }
     
+    func testApplePayViewControllerIsResetAfterAuthorization() {
```
descorp, line 72: As I mentioned yesterday - this test case is excessive:

we already have tests for `paymentAuthorizationViewController(didAuthorizePayment:)` and `stopLoading`.

AdyenDropIn/Utilities/ComponentManager.swift
```diff
@@ -80,7 +80,6 @@ internal final class ComponentManager {
             return nil
         }
         
-        cardComponent?.showsLargeTitle = false
```
descorp, line 83: default value is `true`
I believe you have fixed and tested it everywhere?

AdyenCard/Utilities/Encryption/Cryptor/AES.swift
```diff
@@ -232,16 +233,20 @@ extension Cryptor.AES {
...
-                                            encodeBytes: Int,
-                                            counter: UInt,
-                                            pointer: Pointer,
-                                            length: Int,
-                                            blockA: MutableBytes,
-                                            blockS: MutableBytes) {
-                setCounter(blockA, encodeBytes, counter)
-                aesEncrypt(block: blockA, with: key, saveTo: blockS)
-                memxor(pointer, blockS, length)
+            internal struct EncryptXorContext {
```
descorp, line 236: I wonder if there is a better way to do this 🤔
It is better than 7 params, tho

AdyenDropIn/Utilities/ComponentManager.swift
mohammedDehairy: in those components, they check if `showsLargeTitle` is true AND `_isDropIn` is false, so this should cover this scenario.

AdyenTests/Adyen Tests/Components/Apple Pay/ApplePayComponentTest.swift
mohammedDehairy: its necessary to test that the `viewController` is nilled after calling stopLoading, right?

AdyenTests/Card Tests/CardComponentTests.swift
mohammedDehairy: don't think so, they are still needed, to setup the test preconditions.

AdyenTests/Adyen Tests/Components/Apple Pay/ApplePayComponentTest.swift
descorp: it is already [tested here](https://github.com/Adyen/adyen-ios/blob/d478372a95e4d798120e42e3385f3e20350f68cc/AdyenTests/Adyen%20Tests/Components/Apple%20Pay/ApplePayComponentTest.swift#L37-L43)

AdyenDropIn/Utilities/ComponentManager.swift
descorp: 👍

descorp (approved):

AdyenTests/Adyen Tests/Components/Apple Pay/ApplePayComponentTest.swift
mohammedDehairy: ok, I will remove it, but in another PR, is that okay?

PR #277, opened 2020-11-20
Callback on dismiss SFSafariViewController 

descorp:
### Changes

* add callback on SFSafariViewController dismiss manually


Adyen/Components/Redirect/WebRedirectComponent.swift
```diff
@@ -15,6 +15,8 @@ internal final class WebRedirectComponent: NSObject, PresentableComponent, Actio
     
     /// Indicates the component's UI style.
     private let style: RedirectComponentStyle?
+
+    private var alreadyCanceled = false
```
mohammedDehairy, line 19: this is not used, right?

mohammedDehairy (approved):

PR #296, opened 2020-12-02
feature: made a separate configuration for showing cvc field for normal card component and stored card component, to be configured separatly

mohammedDehairy:
feature: made a separate configuration for showing cvc field for normal card component and stored card component, to be configured separatly.

### DIFF
--- a/AdyenTests/Card Tests/CardComponentTests.swift
+++ b/AdyenTests/Card Tests/CardComponentTests.swift
@@ -517,7 +517,7 @@ class CardComponentTests: XCTestCase {
                                              holderName: "holderName")
         let sut = CardComponent(paymentMethod: method,
                                 clientKey: "test_client_key")
-        sut.showsSecurityCodeField = false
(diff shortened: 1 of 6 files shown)


AdyenTests/Card Tests/CardComponentTests.swift
```diff
@@ -517,7 +517,7 @@ class CardComponentTests: XCTestCase {
                                              holderName: "holderName")
         let sut = CardComponent(paymentMethod: method,
                                 clientKey: "test_client_key")
-        sut.showsSecurityCodeField = false
```
descorp, line 520: so the field `sut.showsSecurityCodeField ` was actually useless in this test?

mohammedDehairy: no, it wasn't, it was affecting both form and stored card pop at the same time, I made them separate.

descorp (approved):

PR #298, opened 2020-12-04
chore: get rid of the dependency on the details object

mohammedDehairy:
chore: get rid of the dependency on the details object.

### DIFF
--- a/Common Example code/Networking/Environment.swift
+++ b/Common Example code/Networking/Environment.swift
@@ -14,9 +14,9 @@ internal enum DemoServerEnvironment: APIEnvironment {
     internal var baseURL: URL {
         switch self {
         case .beta:
-            return URL(string: "https://checkout-beta.adyen.com/v64")!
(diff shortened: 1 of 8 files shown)


Common Example code/Networking/Environment.swift
```diff
@@ -14,9 +14,9 @@ internal enum DemoServerEnvironment: APIEnvironment {
     internal var baseURL: URL {
         switch self {
         case .beta:
-            return URL(string: "https://checkout-beta.adyen.com/v64")!
+            return URL(string: "https://checkout-beta.adyen.com/v100")!
         case .test:
-            return URL(string: "https://checkout-test.adyen.com/v64")!
```
descorp, line 19: ApplePay is not working on version higher than v53. Probably not a good idea to pump version for now...

mohammedDehairy: Sorry, that's a [WIP] PR, but I will take this into consideration :-) .

descorp (approved):

PR #300, opened 2020-12-07
chore: Modified URLSessionHelpers extension to avoid returning invalid JSON api response body when the response status code is not 200

mohammedDehairy:
chore: Modified URLSessionHelpers extension to avoid returning invalid JSON api response body when the response status code is not 200.

### DIFF
--- a/Common Example code/PaymentsController.swift
+++ b/Common Example code/PaymentsController.swift
@@ -185,6 +185,7 @@ internal final class PaymentsController {
             }
         case let .failure(error):
             currentComponent?.stopLoading(withSuccess: false) { [weak self] in
+                self?.presenter?.dismiss(completion: nil)
                 self?.presentAlert(with: error)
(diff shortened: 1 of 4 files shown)


Common Example code/PaymentsController.swift
```diff
@@ -185,6 +185,7 @@ internal final class PaymentsController {
             }
         case let .failure(error):
             currentComponent?.stopLoading(withSuccess: false) { [weak self] in
+                self?.presenter?.dismiss(completion: nil)
```
descorp, line 188: Does it show the alert if you try ApplePay on v66? 🤔

descorp (approved):

Common Example code/PaymentsController.swift
mohammedDehairy: yes, it does.

PR #311, opened 2020-12-18
Add exclusive Apple Pay icon

descorp:
### Changes

* create a list of brand-protected icons
* add custom icon for Apple Pay


Adyen/UI/Views/NetworkImageView.swift
```diff
@@ -37,7 +37,18 @@ open class NetworkImageView: UIImageView {
...
+        DispatchQueue.main.async {
+            self.image = image
+            self.dataTask = nil
+        }
+    }
+
     private func loadImage(from url: URL) {
+        if url.lastPathComponent.hasPrefix("applepay") {
+            return setImage(UIImage(named: "apple_pay", in: Bundle.coreInternalResources, compatibleWith: nil)!)
+        }
```
mohammedDehairy, line 50: why do we use a local version? that means if Apple changes its icon specs, we would have to update the SDK, using the remote one is better because its easier to change, right?

mohammedDehairy:
This PR should be to `release-3.8.2` branch not develop

mohammedDehairy:
This contains changes not intended for 3.8.2 release, your feature branch should be from `release-3.8.2` branch, not develop.

mohammedDehairy (changes requested):
This contains changes not intended for 3.8.2 release, your feature branch should be from release-3.8.2 branch, not develop.

PR #312, opened 2020-12-21
Brand-protected icons for payment methods

descorp:
### Changes

* add custom icon for ApplePay
* allow brand protected icons remain untampered on DropIn

### DIFF
--- a/Adyen/Helpers/UIViewHelpers.swift
+++ b/Adyen/Helpers/UIViewHelpers.swift
@@ -61,7 +61,7 @@ public extension AdyenScope where Base: UIView {
             let radius = value * min(base.bounds.height, base.bounds.width)
             base.layer.cornerRadius = radius
         case .none:
-            break
+            base.layer.cornerRadius = 0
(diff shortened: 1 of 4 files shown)


Adyen/UI/Views/NetworkImageView.swift
```diff
@@ -34,10 +34,26 @@ open class NetworkImageView: UIImageView {
...
+        }
+    }
+
     private func loadImage(from url: URL) {
+        let parts = url.lastPathComponent.split(separator: "@")
+        if parts.count > 1,
+           let internalName = brandProtectedImages[String(parts[0])],
+           let image = UIImage(named: internalName, in: Bundle.coreInternalResources, compatibleWith: nil) {
+            return setImage(image)
+        }
```
mohammedDehairy, line 55: why use a local image? this way we need to update the SDK whenever we want to update apple icon, why not update the image on the backend?

mohammedDehairy (approved):

PR #320, opened 2020-12-24
feature: Added the Actions module and moved all action components and model objects to it

mohammedDehairy:
feature: Added the Actions module and moved all action components and model objects to it.


Adyen/UI/UI Style/CoreColors.swift
```diff
@@ -10,65 +10,78 @@ import UIKit
 extension UIColor {
     
     /// :nodoc:
-    internal enum AdyenCore {
```
descorp, line 13: I believe we called this `AdyenCore` because there was some collisions on Pods?

AdyenActions/UI/View Controllers/Await View Controller/AwaitView.swift
```diff
@@ -13,7 +14,7 @@ internal final class AwaitView: UIView {
     internal lazy var icon: UIImageView = {
         var image = UIImage(named: viewModel.icon)
         if image == nil {
-            image = UIImage(named: viewModel.icon, in: Bundle.coreInternalResources, compatibleWith: nil)
```
descorp, line 16: This is interesting one.. 
If we have  BLIK component in **Components** but `blik` icon in **Actions** ?

Adyen/UI/UI Style/CoreColors.swift
mohammedDehairy: no I recall we called it `AdyenCore` because in DropIn there was another extension to UIColor which has some duplicates and we called it `AdyenDropIn`, now there is only one extension in `Adyen` target with public modifiers to avoid code duplication and removed the extension in DropIn.

AdyenActions/UI/View Controllers/Await View Controller/AwaitView.swift
mohammedDehairy: Yeah, because the `blik` icon is used in the await action component when the payment method is `blik`, same for MBWay.

descorp (approved):

PR #328, opened 2021-01-05
Make ModalViewControllers public

descorp:
### Changes

* make ModalViewControllers public
* refactor WrapperViewController.swift


Adyen/UI/Presentation/ModalViewController.swift
```diff
@@ -44,10 +43,7 @@ internal final class ModalViewController: UIViewController {
     }
     
     /// :nodoc:
-    internal var isRoot: Bool = false
-    
-    /// :nodoc:
-    internal var cancelButtonHandler: ((Bool) -> Void)?
+    public var isRoot: Bool = false
```
mohammedDehairy, line 46: why is this public? this is internal logic

descorp: true..
could be even private

mohammedDehairy:
i don't understand, all merchants need is a wrapper UINavigationController with a nav bar title and a cancel button, but ModalViewController does a lot more than this, making it public doesn't seem right to me

i am afraid this will make merchants even more confused, because ModalViewController does a lot more complicated logic than what they need

PR #335, opened 2021-01-12
chore: Implemented binlookup v2 and update the cvc field when the detected card brand is one for which cvc is optional

mohammedDehairy:
chore: Implemented binlookup v2 and update the cvc field when the detected card brand is one for which cvc is optional.

### DIFF
--- a/AdyenCard/Components/Card/CardComponent.swift
+++ b/AdyenCard/Components/Card/CardComponent.swift
@@ -19,7 +19,7 @@ public protocol CardComponentDelegate: AnyObject {
     /// Called when `CardComponent` detected card type(s) in entered PAN.
     /// - Parameter value: Array of card types matching entered value. Null - if no data entered.
     /// - Parameter component: The `CardComponent` instance.
-    func didChangeCardType(_ value: [CardType]?, component: CardComponent)
+    func didChangeCardBrand(_ value: [CardBrand]?, component: CardComponent)
 }
 
 /// Stored card configuration.
@@ -39,7 +39,7 @@ public class CardComponent: PaymentComponent, PresentableComponent, Localizable,
     internal var cardPublicKeyProvider: AnyCardPublicKeyProvider
     
     /// :nodoc:
-    internal var cardTypeProvider: AnyCardTypeProvider
+    internal var cardTypeProvider: AnyCardBrandProvider
 
     private static let maxCardsVisible = 4
     private static let publicBinLenght = 6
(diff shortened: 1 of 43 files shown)


AdyenCard/Components/Card/CardComponent.swift
```diff
@@ -314,11 +314,13 @@ public class CardComponent: PaymentComponent, PresentableComponent, Localizable,
     }
     
     private func requestCardTypes(for bin: String) {
-        cardTypeProvider.requestCardTypes(for: bin, supported: self.supportedCardTypes) { [weak self] cardTypes in
+        let parameters = CardBrandProviderParameters(bin: bin, supportedTypes: supportedCardTypes)
```
descorp, line 317: there are only 2 parameters, what is the point of wrapping it in standalone struct?

descorp, line 317: `supportedCardBrans` ?

AdyenCard/Utilities/CardType/CardBrand.swift
```diff
@@ -0,0 +1,66 @@
...
+                  showsExpiryDate: Bool = true) {
+        self.type = type
+        self.isSupported = isSupported
+        self.cvcPolicy = cvcPolicy
+        self.isLuhnCheckEnabled = isLuhnCheckEnabled
+        self.showsExpiryDate = showsExpiryDate
+    }
+
+    private enum CodingKeys: String, CodingKey {
+        case type = "brand"
```
descorp, line 58: this whole card / brand naming could be confusing..

AdyenTests/Card Tests/Utilities/CardEncryptorCardTests.swift
```diff
@@ -196,20 +196,14 @@ class CardEncryptorCardTests: XCTestCase {
 
         var payload = Data()
         let expected: Data
-        if #available(iOS 13.0, *) {
```
descorp, line 199: Nope, this will fail on iOS 12 pipeline that we run on master

AdyenCard/Components/Card/CardComponent.swift
mohammedDehairy: Yeah, tell that to CodeBeat :-D , it was complaining about AnyCardBrandProvider main function having too many parameters :-).

AdyenTests/Card Tests/Utilities/CardEncryptorCardTests.swift
mohammedDehairy: the whole function has `@available(iOS 13.0, *)` iOS 12 won't execute this test case, right?.

Xcode was complaining with a warning.

AdyenCard/Utilities/CardType/CardBrand.swift
mohammedDehairy: what do you suggest?

AdyenTests/Card Tests/Utilities/CardEncryptorCardTests.swift
descorp: nope, It executes no matter what :(

AdyenCard/Utilities/CardType/CardBrand.swift
descorp: Nothing I am afraid - we have to draw this line somewhere..
I put the comment when I saw it, but now I guess what you did here is nicest alternative :D

AdyenTests/Card Tests/Utilities/CardEncryptorCardTests.swift
mohammedDehairy: reverted the change.

descorp (approved):

PR #349, opened 2021-01-25
chore: created the generic BaseFormComponent to handle all components with any combinations of first name, last name, phone, and email forms

mohammedDehairy:
chore: created the generic BaseFormComponent to handle all components with any combinations of first name, last name, phone, and email forms


Adyen/Components/BaseFormComponent/FormElements.swift
```diff
@@ -0,0 +1,119 @@
...
+
+    /// :nodoc:
+    var style: StyleType { get }
+
+    /// :nodoc:
+    func build(_ itemBuilder: BaseFormBuilder) -> ItemType
+}
+
+/// :nodoc:
+internal struct FirstNameElement: BaseFormElement {
```
descorp, line 29: Do we really need specifically `FirstNameElement` and `LastNameElement ` ?
What if we next have 'MiddleName', `AddressStreet`, `AddressCity` e.t.c ?

How about jut TextElement with params, so we can extend list easily?


`

AdyenComponents/Qiwi Wallet/QiwiWalletComponent.swift
```diff
@@ -19,16 +19,26 @@ public final class QiwiWalletComponent: PhoneBasedPaymentComponent {
...
+    override public func submitButtonTitle() -> String {
+        ADYLocalizedString("adyen.continueTo", localizationParameters, paymentMethod.name)
     }
 
-    override internal func createPaymentDetails() -> PaymentMethodDetails {
-        QiwiWalletDetails(paymentMethod: paymentMethod,
-                          phonePrefix: phoneNumberItem.prefix,
-                          phoneNumber: phoneNumberItem.value)
+    override public func getPhoneExtensions() -> [PhoneExtension] { qiwiWalletPaymentMethod.phoneExtensions
+    }
```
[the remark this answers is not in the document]
mohammedDehairy: I think this a GitHub issue, because in Xcode is perfectly formatted.

Adyen/Components/BaseFormComponent/FormElements.swift
mohammedDehairy: the idea, is that `FormTextInputItem` is very generic, but the `FirstNameElement` and `LastNameElement` are specific to have the corresponding item  configuration specifically using the correct localization, if I make it generic then how is it different than `FormTextInputItem `.

the idea is that `Elements` are a less abstract layer on top of `FormItems`

descorp (approved):

PR #364, opened 2021-02-15
Limit CVC for stored card

descorp:
### Changes

* limit CVC for stored card

### DIFF
--- a/AdyenCard/Components/Stored Card/StoredCardAlertManager.swift
+++ b/AdyenCard/Components/Stored Card/StoredCardAlertManager.swift
@@ -12,10 +12,11 @@ import UIKit
 
 internal final class StoredCardAlertManager: NSObject, UITextFieldDelegate, Localizable {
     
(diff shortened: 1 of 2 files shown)


AdyenCard/Components/Stored Card/StoredCardAlertManager.swift
```diff
@@ -32,11 +33,28 @@ internal final class StoredCardAlertManager: NSObject, UITextFieldDelegate, Loca
...
+        }
+    }
+
+    private var maxCharactersCount: Int {
+        guard let brand = CardType(rawValue: paymentMethod.brand) else { return 4 }
+        switch brand {
+        case .americanExpress:
+            return 4
+        default:
+            return 3
```
mohammedDehairy, line 54: max here should be 4 in any case, since its 4-4 for Amex, and 3-4 for any other brand, right?

mohammedDehairy (approved):

PR #382, opened 2021-03-03
Refactor payment cancelation

descorp:
### Changes

* simplify ApplePay flow
* simplify DropIn state management
* add FinalizableComponent

### Code cleanup

* remove `withSuccess success: Bool` from `LoadingComponent.stopLoad(completion: (() -> Void)?)`
* `disableCloseButton: Bool` from  `presentationDelegate.present(component: presentableComponent)`

### DIFF
--- a/AdyenComponents/Apple Pay/ApplePayComponent.swift
+++ b/AdyenComponents/Apple Pay/ApplePayComponent.swift
@@ -9,7 +9,7 @@ import Foundation
 import PassKit
 
 /// A component that handles Apple Pay payments.
-public class ApplePayComponent: NSObject, PaymentComponent, PresentableComponent, Localizable {
+public class ApplePayComponent: NSObject, PresentableComponent, PaymentComponent, Localizable, FinalizableComponent {
 
     /// The Apple Pay payment method.
     public var paymentMethod: PaymentMethod {
@@ -19,6 +19,7 @@ public class ApplePayComponent: NSObject, PaymentComponent, PresentableComponent
     /// Apple Pay component configuration.
     internal let configuration: Configuration
     internal var paymentAuthorizationViewController: PKPaymentAuthorizationViewController?
+    internal var paymentAuthorizationCompletion: ((PKPaymentAuthorizationStatus) -> Void)?
     
     /// The delegate of the component.
     public weak var delegate: PaymentComponentDelegate?
@@ -79,6 +80,12 @@ public class ApplePayComponent: NSObject, PaymentComponent, PresentableComponent
     
     /// :nodoc:
     public var localizationParameters: LocalizationParameters?
+
+    /// Finalizes ApplePay payment after being proccessed by payment provider.
+    /// - Parameter success: The status of the payment.
+    public func didFinalize(with success: Bool) {
+        paymentAuthorizationCompletion?(success ? .success : .failure)
+    }
     
     // MARK: - Private
     
@@ -87,6 +94,7 @@ public class ApplePayComponent: NSObject, PaymentComponent, PresentableComponent
             let request = configuration.createPaymentRequest()
             paymentAuthorizationViewController = ApplePayComponent.createPaymentAuthorizationViewController(from: request)
             paymentAuthorizationViewController?.delegate = self
+            paymentAuthorizationCompletion = nil
         }
         return paymentAuthorizationViewController!
     }
@@ -104,7 +112,6 @@ public class ApplePayComponent: NSObject, PaymentComponent, PresentableComponent
         paymentAuthorizationViewController?.dismiss(animated: true) { [weak self] in
             guard let self = self else { return }
             self.paymentAuthorizationViewController = nil
-
             completion?()
         }
     }
--- a/AdyenComponents/Apple Pay/ApplePayComponent.swift
+++ b/AdyenComponents/Apple Pay/ApplePayComponent.swift
@@ -9,48 +9,22 @@ import Foundation
 import PassKit
 
 /// A component that handles Apple Pay payments.
-public class ApplePayComponent: NSObject, PaymentComponent, PresentableComponent, Localizable, LoadingComponent {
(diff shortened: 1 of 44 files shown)


AdyenComponents/Apple Pay/ApplePayComponent.swift
```diff
@@ -9,48 +9,23 @@ import Foundation
...
-    
-    /// A list of fields that you need for a shipping contact in order to process the transaction.
-    /// Ignored on iOS 10.*.
-    public var requiredShippingContactFields: Set<PKContactField> {
-        configuration.requiredShippingContactFields
-    }
-    
     /// Initializes the component.
     ///
-    /// - Warning: `stopLoading()` must be called before dismissing this component.
```
mohammedDehairy, line 51: maybe change it to `/// - Warning: `didFInalize()` must be called before dismissing this component.`

AdyenComponents/Apple Pay/ApplePayComponent.swift
```diff
@@ -83,94 +57,62 @@ public class ApplePayComponent: NSObject, PaymentComponent, PresentableComponent
...
-        
+        let request: PKPaymentRequest = configuration.createPaymentRequest()
+        guard let paymentAuthorizationViewController = ApplePayComponent.createPaymentAuthorizationViewController(from: request) else {
+            throw UnknownError(
+                errorDescription: "Failed to instantiate PKPaymentAuthorizationViewController because of unknown error"
+            )
+        }
+
         self.configuration = configuration
+        self.paymentAuthorizationViewController = paymentAuthorizationViewController
```
mohammedDehairy, line 68: `self.paymentAuthorizationViewController` doesn't have to be optional, right?

AdyenComponents/Apple Pay/ApplePayComponentExtensions.swift
```diff
@@ -15,7 +15,7 @@ extension ApplePayComponent: PKPaymentAuthorizationViewControllerDelegate {
     
     /// :nodoc:
     public func paymentAuthorizationViewControllerDidFinish(_ controller: PKPaymentAuthorizationViewController) {
-        if isUserCancel {
```
mohammedDehairy, line 18: weren't this necessary to differentiate between when user cancels actively and when the payment is finished?

AdyenDropIn/DropInComponent.swift
```diff
@@ -212,9 +214,7 @@ extension DropInComponent: PaymentComponentDelegate {
     
     /// :nodoc:
     public func didFail(with error: Error, from component: PaymentComponent) {
-        paymentInProgress = false
```
mohammedDehairy, line 215: we don't need to set it to `false` here?

AdyenDropIn/DropInComponent.swift
```diff
@@ -239,8 +239,6 @@ extension DropInComponent: ActionComponentDelegate {
     /// :nodoc:
     public func didFail(with error: Error, from component: ActionComponent) {
         if case ComponentError.cancelled = error {
-            paymentInProgress = false
```
mohammedDehairy, line 242: we don't need to set it to false here?

Demo/Common/IntegrationExample.swift
```diff
@@ -74,30 +60,23 @@ internal final class IntegrationExample {
     }
 
     internal func finish(with resultCode: PaymentsResponse.ResultCode) {
-        let success = resultCode == .authorised || resultCode == .received || resultCode == .pending
+        if let finalizableComponent = currentComponent as? FinalizableComponent {
+            let success = resultCode == .authorised || resultCode == .received || resultCode == .pending
+            finalizableComponent.didFinalize(with: success)
```
mohammedDehairy, line 65: Maybe `DropInComponent.didFinalize()` should also call `stopLoading`?

AdyenComponents/Apple Pay/ApplePayComponent.swift
descorp: 👍

AdyenComponents/Apple Pay/ApplePayComponent.swift
descorp: It has to be because PKPaymentAuthorizationViewController can't be presented twice. We have to kill and recreate it every time

AdyenComponents/Apple Pay/ApplePayComponentExtensions.swift
descorp: It was.. hm..
I'll do one more regression, to check if we break something... :D

AdyenDropIn/DropInComponent.swift
descorp: It is set in `userCancel` now

AdyenDropIn/DropInComponent.swift
descorp: It is set in `userCancel` now

Demo/Common/IntegrationExample.swift
descorp: Why not , no harm in it

AdyenComponents/Apple Pay/ApplePayComponent.swift
mohammedDehairy: 👍 yeah, forgot about that

AdyenComponents/Apple Pay/ApplePayComponentExtensions.swift
mohammedDehairy: now `delegate?.didFail(with: ComponentError.cancelled, from: self)` will be called when the payment is complete in addition to when user cancel, right? which merchants were complaining about.

descorp: I have checked. Seems all good

mohammedDehairy (approved):

Issue #369, opened 2021-02-19
SPM Warning


iKorich:
**Describe the bug**
`
Showing All Messages
resource 'Localizable.strings' in target 'Adyen' is missing the default localization 'en-US'; the default localization is used as a fallback when no other localization matches

`

**To Reproduce**
1. Add Adyen via SPM
2. Build
3. See warning

**Expected behavior**
no warnings

**Smartphone (please complete the following information):**
 - Device: any
 - OS: 14.4
 - SDK Version 3.8.3

mohammedDehairy:
Hi @iKorich ,

Thanks for reaching out.

Please refer to [this comment](https://github.com/Adyen/adyen-ios/issues/146#issuecomment-733570453).

```
we have tested the fallback whenever the device language is one which is not supported by the SDK, so localization and fallback to en-US works fine, but this warning is till persisting for some reason, but don't think its harmful in any way.
```

MattCheetham:
Just wanted to voice that this one bothers me too. Currently the only warning in our project. Do we know if this is an Xcode bug we can file with Apple or is it something the framework is doing wrong?

descorp:
Hey @MattCheetham 

Unfortunately, we have no solid conclusion if this is an Xcode bug or just our SDK doing something wrong.
We know that localization is working as expected; this is just a warning, not a blocker.

We will investigate further on how to bypass the warning...
Any information that can shed light on it is welcome :)

descorp:
Thanks to @MattCheetham for solving this!

I am going to close this ticket for now. It is available on develop and will be publicly released in v4 beta this week.
Feel free to reopen if you still have the issue.

PR #394, opened 2021-03-11
Fixes localization warning when using SPM

MattCheetham:
## Summary
This PR tackles the issue described in #369 where compiling the Adyen SDK using Swift Package Manager results in a warning about a missing localization file.

## Background
At this time, I do believe this to be a bug with the way Apple has implemented the `defaultLocalization` parameter in Swift Package Manager. Apple's own documentation states that `defaultLocalization` should be a [LanguageTag](https://developer.apple.com/documentation/swift_packages/languagetag) which is an IETF language tag, in conformance with [RFC5654](https://tools.ietf.org/html/rfc5646).

The documentation of RFC5646 states that language tags should be case insensitive but does recommend capitalising the country code. Unfortunately it seems that Swift Package Manager does differentiate between upper and lower case characters and has some sort of issue identifying a language if you specify an upper case country code. Thankfully, simply lowercasing the entire string removes the warning.

I have run the unit tests on the project and everything passed so as far as I can tell, the change is harmless and simply fixes the issue. I suspect that Apple simply never tested it with a language-country code and only expected a fallback to be a top level language code.

## Tested scenarios
All project unit tests run and passed successfully. I suspect you may want to run some manual tests your side though before merging this.

**Fixed issue**:  #369

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -5,7 +5,7 @@ import PackageDescription
 
 let package = Package(
     name: "Adyen",
-    defaultLocalization: "en-US",
+    defaultLocalization: "en-us",
     platforms: [
         .iOS(.v11)
     ],


descorp (approved):

descorp:
Thanks a lot @MattCheetham ! 
You are the hero of the day 🦸‍♂️ 

Who had knew it should be lowercase...

mohammedDehairy:
Wow, amazing!.

Thanks @MattCheetham  for diving deep into the issue, it was indeed very annoying warning.

mohammedDehairy (approved):

PR #411, opened 2021-03-15
Format project

mohammedDehairy:
Format project.


AdyenActions/Actions/Action.swift
```diff
@@ -64,7 +64,7 @@ public enum Action: Decodable {
         case sdk
         case qrCode
         case await
-        case voucher
+            case voucher
```
mohammedDehairy, line 67: 🤔 not sure what is this

PR #437, opened 2021-03-24
Fix CI

descorp:
### Changes

* resolve swiftlint warnings
* improve carthage integration tests
* revert local dependency to Carthage
* ci cleanup


Adyen.xcodeproj/project.pbxproj
```diff
@@ -5191,6 +5213,7 @@
 		F95B6DA725274551002C9062 /* Release */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
```
mohammedDehairy, line 5216: which target is this? why is it turned on?

mohammedDehairy (approved):

PR #442, opened 2021-03-31
[WIP] Dynamic View Controller

descorp:
### Changes

* add FoldableFormItem
* add DynamicViewController
* propagate DynamicViewController up to DropIn presentation level
* some minor refactoring

### Unrelated changes

* diners regex update
* AVS check for issuing country

### DIFF
--- a/Adyen/UI/Form/FormItemViewBuilder.swift
+++ b/Adyen/UI/Form/FormItemViewBuilder.swift
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2020 Adyen N.V.
+// Copyright (c) 2021 Adyen N.V.
 //
 // This file is open source and available under the MIT license. See the LICENSE file for more info.
 //
@@ -58,5 +58,17 @@ public struct FormItemViewBuilder {
     /// :nodoc:
     public func build(with item: FormSeparatorItem) -> FormItemView<FormSeparatorItem> {
         FormSeparatorItemView(item: item)
+    }
+
+    /// Builds `FormSeparatorItemView` from `FormSeparatorItem`.
+    /// :nodoc:
+    public func build(with item: FormFoldableItem) -> FormItemView<FormFoldableItem> {
+        FormFoldableItemView(item: item)
+    }
+
+    public static func renderItem(_ item: FormItem) -> AnyFormItemView {
+        let itemView = item.build(with: FormItemViewBuilder())
+        itemView.accessibilityIdentifier = item.identifier
+        return itemView
     }
 }
--- a/Adyen/UI/Form/FormViewController.swift
+++ b/Adyen/UI/Form/FormViewController.swift
@@ -10,8 +10,6 @@ import UIKit
 /// :nodoc:
 @objc(ADYFormViewController)
 open class FormViewController: UIViewController, Localizable, DynamicViewController {
-
-    private let notificationCenter = NotificationCenter.default
     
     private lazy var itemManager = FormViewItemManager(itemViewDelegate: self)
     
@@ -32,8 +30,6 @@ open class FormViewController: UIViewController, Localizable, DynamicViewControl
         self.style = style
         super.init(nibName: nil, bundle: nil)
     }
-
-    deinit { notificationCenter.removeObserver(self) }
     
     @available(*, unavailable)
     public required init?(coder: NSCoder) {
@@ -127,11 +123,6 @@ open class FormViewController: UIViewController, Localizable, DynamicViewControl
         
         view.backgroundColor = style.backgroundColor
         formView.backgroundColor = style.backgroundColor
-
-        notificationCenter.addObserver(self,
-                                       selector: #selector(keyboardWillChangeFrame(_:)),
-                                       name: UIResponder.keyboardWillChangeFrameNotification,
-                                       object: nil)
     }
     
     /// :nodoc:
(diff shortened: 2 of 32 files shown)


Adyen/UI/Form/FormItemViewBuilder.swift
```diff
@@ -59,4 +59,16 @@ public struct FormItemViewBuilder {
     public func build(with item: FormSeparatorItem) -> FormItemView<FormSeparatorItem> {
         FormSeparatorItemView(item: item)
     }
+
+    /// Builds `FormSeparatorItemView` from `FormSeparatorItem`.
+    /// :nodoc:
+    public func build(with item: FormFoldableItem) -> FormItemView<FormFoldableItem> {
+        FormFoldableItemView(item: item)
```
mohammedDehairy, line 66: I prefer `FoldableFormItemView`, the parent class name should have the subclass qualifier at the beginning:

```
parent class: FormItemView
subclass: FoldableFormItemView
```

```
parent class: ParentClass
subclass: ChildParentClass
```

Adyen/UI/Form/Items/Foldable/FormFoldableItem.swift
```diff
@@ -0,0 +1,41 @@
...
+// Copyright (c) 2021 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+/// A form item that presents single FormItem, that could be hide or showed dynamicaly.
+/// :nodoc:
+public final class FormFoldableItem: CompoundFormItem {
```
mohammedDehairy, line 11: same as [here](https://github.com/Adyen/adyen-ios/pull/442/files#r605736167)

Adyen/UI/Form/Items/Split/FormSplitItemView.swift
```diff
@@ -9,39 +9,24 @@ import UIKit
 
 /// A view representing a split item.
 internal final class FormSplitItemView: FormItemView<FormSplitItem> {
```
mohammedDehairy, line 11: `SplitFormItemView` sounds better IMO

descorp: Everything is called FormBlablaItem. 
I believe those should align 🤔

Adyen/UI/Form/FormItemViewBuilder.swift
descorp: We can rename, but this would be a breaking change.
Let's address it in another PR

Adyen/UI/Form/Items/Split/FormSplitItemView.swift
mohammedDehairy: yeah, true, but I prefer the other way tbh

Adyen/UI/Form/FormViewController.swift
```diff
@@ -198,17 +154,6 @@ open class FormViewController: UIViewController, Localizable {
...
-        guard let keyboardFrame = notification.userInfo?[UIResponder.keyboardFrameEndUserInfoKey] as? CGRect else { return }
-        var heightOffset = keyboardFrame.origin.y - UIScreen.main.bounds.height
-        
-        if #available(iOS 11.0, *) {
-            heightOffset += min(abs(heightOffset), view.safeAreaInsets.bottom)
-        }
-        
-        bottomConstraint?.constant = heightOffset
-    }
-
```
mohammedDehairy, line 211: why is this removed?

Adyen/UI/UI Style/CoreFonts.swift
```diff
@@ -17,11 +17,7 @@ extension UIFont {
         }
         
         internal static var largeTitle: UIFont {
-            if #available(iOS 11.0, *) {
-                return UIFont.preferredFont(forTextStyle: .largeTitle).adyen.font(with: .bold)
-            } else {
-                return .systemFont(ofSize: 32.0, weight: .bold)
-            }
+            .systemFont(ofSize: 32.0, weight: .bold)
```
mohammedDehairy, line 20: this should be:
```
 UIFont.preferredFont(forTextStyle: .largeTitle).adyen.font(with: .bold)
```

Adyen/UI/UI Style/CoreColors.swift
```diff
@@ -82,13 +82,7 @@ extension UIColor {
 
         /// :nodoc:
         public static var componentLoadingMessageColor: UIColor {
-            if #available(iOS 11.0, *) {
-                return UIColor(named: "awaitLoadingMessageColor",
-                               in: Bundle.coreInternalResources,
-                               compatibleWith: nil) ?? componentPlaceholderText
-            } else {
-                return componentPlaceholderText
-            }
```
mohammedDehairy, line 91: this should be:
```
UIColor(named: "awaitLoadingMessageColor",
                                in: Bundle.coreInternalResources,
                                compatibleWith: nil) ?? componentPlaceholderText
```

Adyen/UI/View Controllers/ViewControllerDelegate.swift
```diff
@@ -0,0 +1,61 @@
...
+            objc_getAssociatedObject(self, &AdyenViewControllerAssociatedKeys.dynamicContentDelegate) as? DynamicViewControllerDelegate
+        }
+        set {
+            objc_setAssociatedObject(self,
+                                     &AdyenViewControllerAssociatedKeys.dynamicContentDelegate,
+                                     newValue,
+                                     objc_AssociationPolicy.OBJC_ASSOCIATION_ASSIGN)
+        }
+    }
+
```
mohammedDehairy, line 56: we use this pattern a lot, it makes me uncomfortable, its not the swift way, what's wrong in implementing a delegate property in all conforming classes?

AdyenCard/Utilities/CardType/CardBrandProvider.swift
```diff
@@ -26,7 +26,7 @@ internal struct CardBrandProviderParameters {
 /// Provide cardType detection based on BinLookup API.
 internal final class CardBrandProvider: AnyCardBrandProvider {
     
-    private static let minBinLength = 6
+    private static let minBinLength = 10
```
mohammedDehairy, line 29: according to Nicholas its 11 actually

Adyen/UI/Form/FormViewController.swift
descorp: I am still hovering on this one as well..

IN one hand: 
We have too many sources of trust for keyboard events. 
Too many things happening in the same time in different parts of the app and it is not easy to control animation.

In the other:
This will lead to changes in Component integration. Also previously we had a white "background" under keyboard, I will have to do something extra.

Adyen/UI/UI Style/CoreFonts.swift
descorp: oh, good one!

Adyen/UI/UI Style/CoreColors.swift
descorp: 👍

Adyen/UI/View Controllers/ViewControllerDelegate.swift
descorp: good point!

AdyenCard/Utilities/CardType/CardBrandProvider.swift
descorp: Yep. We do `bin > minBinLength ` :D

mohammedDehairy: minimum means its an allowed value, so if the minimum is 10 then it should be `if bin >= minBinLength ` :-D

descorp: 👍

PR #443, opened 2021-04-06
fix: Prevent URLSession response from being cached

kaphacius:
### Changes

Prevents caching of URLSession response by implementing a method from `URLSessionDataDelegate`.
**Note:** Implementing this requires `APIClient` to inherit from `NSObject`, which might impact performance. Another possible solution would be a separate delegate class, that only implements this method and is a property of `APIClient`


mohammedDehairy (approved):

kaphacius:
Changed the way data is cached by only allowing in-memory caching. Instead of relying on delegate callbacks, we set an `.ephemereal` URLSession configuration

mohammedDehairy (approved):

descorp (approved):

PR #445, opened 2021-04-07
Apple Pay token as base 64

descorp:
### Changes

* send Apple Pay token as Base64

### DIFF
--- a/AdyenComponents/Apple Pay/ApplePayComponentExtensions.swift
+++ b/AdyenComponents/Apple Pay/ApplePayComponentExtensions.swift
@@ -26,7 +26,7 @@ extension ApplePayComponent: PKPaymentAuthorizationViewControllerDelegate {
                                                    didAuthorizePayment payment: PKPayment,
                                                    completion: @escaping (PKPaymentAuthorizationStatus) -> Void) {
         paymentAuthorizationCompletion = completion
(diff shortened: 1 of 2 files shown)


AdyenComponents/Apple Pay/ApplePayComponentExtensions.swift
```diff
@@ -26,7 +26,7 @@ extension ApplePayComponent: PKPaymentAuthorizationViewControllerDelegate {
                                                    didAuthorizePayment payment: PKPayment,
                                                    completion: @escaping (PKPaymentAuthorizationStatus) -> Void) {
         paymentAuthorizationCompletion = completion
-        let token = String(data: payment.token.paymentData, encoding: .utf8) ?? ""
+        let token = payment.token.paymentData.base64EncodedString()
```
mohammedDehairy, line 29: are we sure this won't break merchant integrations?.

They might be depending on the token structure, I wouldn't do this, unless we add it to the breaking changes and migration guide.

descorp: > add it to the breaking changes and migration guide.

we can do that. Good idea.

In general - it should be safe. Even our documentation is mentioning `token` as Base64 :D
All information inside is for Apple, and if someone really need it - they can decode it from Base64

mohammedDehairy (approved):

PR #447, opened 2021-04-07
Big refactoring

descorp:
### Changes

* make `minBinLength` 11
* move ViewControllerDelegate in standalone file
* improve UIView helpers
* move finalPresentationFrame to ModalViewController
* remove `#available(iOS 11.0, *)`
* remove environment from CardNumberItem. Use LogoProvider instead.

### DIFF
--- a/Adyen/Helpers/UIViewConstraintsHelper.swift
+++ b/Adyen/Helpers/UIViewConstraintsHelper.swift
@@ -13,7 +13,8 @@ extension AdyenScope where Base: UIView {
     /// Attach top, bottom, left and right anchores of this view to corespondent anchores inside specified view.
     /// IMPORTANT: both view should be in the same hierarcy.
     /// - Parameter view: Container view
-    public func anchore(inside view: UIView, with padding: UIEdgeInsets = .zero) {
+    @discardableResult
+    public func anchore(inside view: UIView, with padding: UIEdgeInsets = .zero) -> [NSLayoutConstraint] {
         let constraints = [
             base.topAnchor.constraint(equalTo: view.topAnchor, constant: padding.top),
             base.bottomAnchor.constraint(equalTo: view.bottomAnchor, constant: padding.bottom),
@@ -21,12 +22,14 @@ extension AdyenScope where Base: UIView {
             base.rightAnchor.constraint(equalTo: view.rightAnchor, constant: padding.right)
         ]
         NSLayoutConstraint.activate(constraints)
+        return constraints
     }
 
     /// Attach top, bottom, left and right anchores of this view to corespondent anchores inside specified view.
     /// IMPORTANT: both view should be in the same hierarcy.
     /// - Parameter view: Container view
-    public func anchore(inside margines: UILayoutGuide, with padding: UIEdgeInsets = .zero) {
+    @discardableResult
+    public func anchore(inside margines: UILayoutGuide, with padding: UIEdgeInsets = .zero) -> [NSLayoutConstraint] {
         let constraints = [
             base.topAnchor.constraint(equalTo: margines.topAnchor, constant: padding.top),
             base.bottomAnchor.constraint(equalTo: margines.bottomAnchor, constant: padding.bottom),
@@ -34,6 +37,7 @@ extension AdyenScope where Base: UIView {
             base.rightAnchor.constraint(equalTo: margines.rightAnchor, constant: padding.right)
         ]
         NSLayoutConstraint.activate(constraints)
+        return constraints
     }
 
     /// Wrap the view inside a container view with certain edge insets
--- a/Adyen/UI/Form/FormItemViewBuilder.swift
+++ b/Adyen/UI/Form/FormItemViewBuilder.swift
@@ -59,4 +59,10 @@ public struct FormItemViewBuilder {
     public func build(with item: FormSeparatorItem) -> FormItemView<FormSeparatorItem> {
         FormSeparatorItemView(item: item)
     }
+
+    public static func renderItem(_ item: FormItem) -> AnyFormItemView {
+        let itemView = item.build(with: FormItemViewBuilder())
+        itemView.accessibilityIdentifier = item.identifier
+        return itemView
+    }
 }
--- a/Adyen/UI/Form/FormItemViewBuilder.swift
+++ b/Adyen/UI/Form/FormItemViewBuilder.swift
@@ -60,7 +60,7 @@ public struct FormItemViewBuilder {
         FormSeparatorItemView(item: item)
     }
 
-    public static func renderItem(_ item: FormItem) -> AnyFormItemView {
+    public static func build(_ item: FormItem) -> AnyFormItemView {
         let itemView = item.build(with: FormItemViewBuilder())
         itemView.accessibilityIdentifier = item.identifier
         return itemView
--- a/Adyen/UI/Form/Items/Split/FormSplitItemView.swift
+++ b/Adyen/UI/Form/Items/Split/FormSplitItemView.swift
@@ -16,7 +16,7 @@ internal final class FormSplitItemView: FormItemView<FormSplitItem> {
     ///
     /// - Parameter item: The item represented by the view.
     internal required init(item: FormSplitItem) {
-        views = item.subitems.map(FormSplitItemView.renderItem)
+        views = item.subitems.map(FormSplitItemView.build)
         super.init(item: item)
         
         addSubview(stackView)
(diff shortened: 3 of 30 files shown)


Adyen/UI/Form/FormItemViewBuilder.swift
```diff
@@ -59,4 +59,10 @@ public struct FormItemViewBuilder {
     public func build(with item: FormSeparatorItem) -> FormItemView<FormSeparatorItem> {
         FormSeparatorItemView(item: item)
     }
+
+    public static func renderItem(_ item: FormItem) -> AnyFormItemView {
```
mohammedDehairy, line 63: this name is misleading, its not actually rendering, its building, what's wrong with `build`?

Adyen/UI/Form/Items/Split/FormSplitItemView.swift
```diff
@@ -54,5 +39,11 @@ internal final class FormSplitItemView: FormItemView<FormSplitItem> {
         stackView.spacing = 16
         return stackView
     }()
+
+    private static func renderItem(_ item: FormItem) -> AnyFormItemView {
```
mohammedDehairy, line 43: also here its just `building` not `rendering`

Adyen/UI/Form/FormItemViewBuilder.swift
descorp: Good point 🤔

mohammedDehairy (approved):

Adyen/Helpers/UIViewConstraintsHelper.swift
```diff
@@ -13,27 +13,31 @@ extension AdyenScope where Base: UIView {
     /// Attach top, bottom, left and right anchores of this view to corespondent anchores inside specified view.
     /// IMPORTANT: both view should be in the same hierarcy.
     /// - Parameter view: Container view
-    public func anchore(inside view: UIView, with padding: UIEdgeInsets = .zero) {
+    @discardableResult
+    public func anchore(inside view: UIView, with padding: UIEdgeInsets = .zero) -> [NSLayoutConstraint] {
```
kaphacius, line 17: i think it is misspelled, should be `anchor` (both a noun and verb). probably multiple places in the project

mohammedDehairy (approved):

Adyen/UI/View Controllers/Secured View Controller/SecuredViewController.swift
```diff
@@ -105,54 +97,42 @@ public final class SecuredViewController: UIViewController {
...
-        
-        blurConstraints = constraints
-        
-        NSLayoutConstraint.activate(constraints)
+        blurEffectView.translatesAutoresizingMaskIntoConstraints = false
+        blurConstraints = blurEffectView.adyen.anchore(inside: view)
         
-        UIView.animate(withDuration: 0.2, animations: {
-            self.blurEffectView.effect = self.blurEffect
+        UIView.animate(withDuration: 0.2, animations: { [weak self] in
```
kaphacius, line 119: not sure we need `[weak self]` here, since `self` does not retain UIView. same in a few other places.

Adyen/UI/View Controllers/Secured View Controller/SecuredViewController.swift
```diff
@@ -105,54 +97,42 @@ public final class SecuredViewController: UIViewController {
...
-                NSLayoutConstraint.deactivate(blurConstraints)
-            }
-            self.blurEffectView.removeFromSuperview()
-            self.$blurEffectView.reset()
-            
-            self.view.backgroundColor = self.style.backgroundColor
-        })
+        UIView.animate(withDuration: 0.2,
+                       animations: { [weak self] in self?.blurEffectView.effect = nil },
+                       completion: { [weak self] _ in
```
kaphacius, line 127: in places where there is a lot of `[weak self]` checks sometimes it is better to check whether `self` is `.some` at the beginning of the closure, so we don't need to execute it at all.

AdyenCard/Components/Card/CardComponent.swift
```diff
@@ -25,6 +25,7 @@ public protocol CardComponentDelegate: AnyObject {
 /// A component that provides a form for card payments.
 public class CardComponent: PaymentComponent, PresentableComponent, Localizable, Observer, LoadingComponent {
 
+    private let publicBinLenght = 6
```
kaphacius, line 28: length*

AdyenDropIn/Presentation/DropInNavigationController.swift
```diff
@@ -114,18 +116,14 @@ extension DropInNavigationController: UINavigationControllerDelegate {
...
 extension DropInNavigationController: UIViewControllerTransitioningDelegate {
-    
-    /// :nodoc:
-    public func presentationController(forPresented presented: UIViewController,
-                                       presenting: UIViewController?,
-                                       source: UIViewController) -> UIPresentationController? {
+
+    internal func presentationController(forPresented presented: UIViewController,
+                                         presenting: UIViewController?,
+                                         source: UIViewController) -> UIPresentationController? {
```
kaphacius, line 122: can we make this method return a non-optional value? same at line #109

AdyenCard/Utilities/BinLookup/BinLookupResponse.swift
```diff
@@ -17,7 +17,7 @@ internal struct BinLookupResponse: Response {
     internal let requestId: String
 
     /// :nodoc:
-    internal let issuingCountryCode: String
+    internal let issuingCountryCode: String?
```
kaphacius, line 20: is there a case where this can be `nil`? when decoding a JSON?

Adyen/Helpers/UIViewConstraintsHelper.swift
descorp: 🙀

AdyenCard/Utilities/BinLookup/BinLookupResponse.swift
descorp: Yes, when decoding a JSON

In normal condition:

```
{
  "issuingCountryCode" : "US",
  "brands" : [
    {
      "cvcPolicy" : "required",
      "brand" : "visa",
      "enableLuhnCheck" : true,
      "showExpiryDate" : true,
      "showSocialSecurityNumber" : false,
      "supported" : true
    }
  ],
  "requestId" : "AF41AC8D-1F7E-48D1-98B3-4DABED429F2D"
}
```


If BIN is unknown the result is :

```
{
  "requestId" : "BA050BE8-3AE0-4D87-9FBA-6DC4E2C382B4"
}
```

AdyenDropIn/Presentation/DropInNavigationController.swift
descorp: How do you mean?
Those methods are in UIKit: `UIViewControllerTransitioningDelegate` and `UINavigationControllerDelegate`.

AdyenCard/Components/Card/CardComponent.swift
descorp: 👍

Adyen/Helpers/UIViewConstraintsHelper.swift
descorp: PR for it [here](https://github.com/Adyen/adyen-ios/pull/448)

AdyenCard/Components/Card/CardComponent.swift
descorp: PR for it [here](https://github.com/Adyen/adyen-ios/pull/448)

Adyen/UI/View Controllers/Secured View Controller/SecuredViewController.swift
descorp: For me this is more a force of a habit..
Technically it does retain `self`, but only for 0.2 sec :D
I agree - most of the time `[weak self]` is unnecessary; in the other hand it is fail-safe - no harm in having it and it protects you from potential mistakes.

What do you think @mohammedDehairy ?
Should we use `[weak self]`:

[ ] ONLY when needed 
[ ] ALWAYS
[ ] without restrictions

Adyen/UI/View Controllers/Secured View Controller/SecuredViewController.swift
descorp: Indeed.
We usually use `guard let self = self else { return }`

Adyen/UI/View Controllers/Secured View Controller/SecuredViewController.swift
mohammedDehairy: here its most probably not needed, but adding [weak self] doesn't hurt either, so no need to have a hard rule about it, I wouldn't mind adding it if its force of habit :-) .

AdyenDropIn/Presentation/DropInNavigationController.swift
kaphacius: was not aware of that 😅

PR #471, opened 2021-04-26
Add script for generating localization keys

kaphacius:
This is the first iteration of the script. The idea is to generate compile time constants for localization keys.
I only applied them in `QRCodeView` and `QRCodeComponent` to demonstrate how they would be used.

The actual file generated is `Adyen/Assets/Generated/ADYLocalizationKey.swift`

Question: is the `public func ADYLocalizedString(_ key: ADYLocalizationKey, _ parameters: LocalizationParameters?, _ arguments: CVarArg...) -> String` supposed to be called by the end-users? For now I made a copy of it that accepts the newly created `ADYLocalizationKey` instance as the `key` parameter for demo purposes. But ideally we would want to have only 1 version of the function, to eliminate the possibility of using a wrong localization key.


Adyen/Utilities/Localization.swift
```diff
@@ -40,6 +40,30 @@ public func ADYLocalizedString(_ key: String, _ parameters: LocalizationParamete
...
+///
+/// :nodoc:
+///
+/// - Parameters:
+///   - key: The key used to identify the localized string.
+///   - parameters: The localization parameters.
+///   - arguments: The arguments to substitute in the templated localized string.
+/// - Returns: The localized string for the given key, or the key itself if the localized string could not be found.
+public func ADYLocalizedString(_ key: ADYLocalizationKey, _ parameters: LocalizationParameters?, _ arguments: CVarArg...) -> String {
+    let possibleInputs = buildPossibleInputs(key.key, parameters)
```
descorp, line 56: You can just reuse `public func ADYLocalizedString(_ key: String, _ parameters: LocalizationParameters?, _ arguments: CVarArg...) -> String` right?

kaphacius: Yes, but I did not want to make changes in the whole project before we agree on the script :)

Adyen/Assets/Generated/ADYLocalizationKey.swift
```diff
@@ -0,0 +1,107 @@
+//
+// Copyright (c) 2021 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+// This file is autogenerated. Please do not modify it.
+
+/// :nodoc:
+public struct ADYLocalizationKey {
```
mohammedDehairy, line 8: why not an enum? why a struct?

mohammedDehairy, line 8: also what's up with the `ADY` prefix, its not necessary at all, its not objective c :-) .

kaphacius: No preference really. I thought you were against big enums :)

kaphacius: The function `ADYLocalizedString` has it, so i thought there was some magic to it. I can make it `enum LocalizationKey: String`

AdyenCard/Form/FormCardSecurityCodeItemView.swift
```diff
@@ -15,7 +15,7 @@ internal final class FormCardSecurityCodeItemView: FormTextItemView<FormCardSecu
         accessory = .customView(cardHintView)
         observe(item.$selectedCard) { [weak self] cardsType in
             let number = cardsType == CardType.americanExpress ? "4" : "3"
-            let localization = ADYLocalizedString("adyen.card.cvcItem.placeholder.digits", item.localizationParameters)
+            let localization = localizedString(.cardCvcItemPlaceholderDigits, item.localizationParameters)
```
descorp, line 18: We can use here `localizedString(. cardCvcItemPlaceholderDigits, item.localizationParameters, number)`

kaphacius: 👍

descorp (dismissed):

mohammedDehairy (dismissed):

mohammedDehairy (approved):

PR #473, opened 2021-04-28
ZipCodeOonly billing address field

descorp:
### Changes

* add BillingAddressInformation
* unblock `zipCode` mode
* add single field for `zipCode` mode

### DIFF
--- a/Adyen/Components/Base/AddressInfo.swift
+++ b/Adyen/Components/Base/AddressInfo.swift
@@ -9,8 +9,8 @@ import Foundation
 /// The model for address data.
 public struct AddressInfo: Equatable, Encodable {
 
-    internal let invalidCountry = "ZZ"
-    internal let invalidValue = "N.A."
+    internal static let invalidCountry = "ZZ"
+    internal static let invalidValue = "null"
 
     /// Create new instance of AddressInfo
     public init(city: String? = nil,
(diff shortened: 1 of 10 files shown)


Adyen/Components/Base/AddressInfo.swift
```diff
@@ -9,6 +9,26 @@ import Foundation
 /// The model for address data.
 public struct AddressInfo: Equatable, Encodable {
 
+    internal let invalidCountry = "ZZ"
+    internal let invalidValue = "N.A."
```
kaphacius, line 13: maybe these could be static constants, since they relate to the whole type.

Adyen/Helpers/StringHelpers.swift
```diff
@@ -24,6 +24,11 @@ public extension AdyenScope where Base == String? {
 
 /// :nodoc:
 public extension AdyenScope where Base == String {
+
+    /// Returns nil string is empty or actual value.
+    var nullIfEmpty: String? {
```
kaphacius, line 29: maybe `nilIfEmpty`?

Adyen/Components/Base/AddressInfo.swift
descorp: 👍

Adyen/Helpers/StringHelpers.swift
descorp: yep, `null` is so C# 😅

mohammedDehairy (approved):

kaphacius (approved):

PR #474, opened 2021-04-30
fix: COIOS-405 Add intermediate Apple Pay component with an Pay button

kaphacius:
This PR attempts to fix COIOS-405 by adding an intermediate component shown with an Pay, as per Apple's guidelines.


AdyenDropIn/Components/PreApplePayComponent.swift
```diff
@@ -0,0 +1,141 @@
...
+    private func createModel(with amount: Payment.Amount) -> PreApplePayView.Model {
+        PreApplePayView.Model(
+            hint: amount.formatted,
+            style: PreApplePayView.Model.Style(
+                hintLabel: TextStyle(font: .preferredFont(forTextStyle: .footnote), color: UIColor.Adyen.componentSecondaryLabel),
+                backgroundColor: UIColor.Adyen.componentBackground))
+    }
+    
+    /// :nodoc:
+    private func makeViewController() -> UIViewController {
```
descorp, line 72: You already have `lazy var viewController: UIViewController`

Demo/UIKit/ComponentsViewController.swift
```diff
@@ -32,11 +32,13 @@ internal final class ComponentsViewController: UIViewController, Presenter {
...
             [
                 ComponentsItem(title: "Card", selectionHandler: presentCardComponent),
-                ComponentsItem(title: "ApplePay", selectionHandler: presentApplePayComponent),
                 ComponentsItem(title: "iDEAL", selectionHandler: presentIdealComponent),
                 ComponentsItem(title: "SEPA Direct Debit", selectionHandler: presentSEPADirectDebitComponent),
                 ComponentsItem(title: "MB WAY", selectionHandler: presentMBWayComponent),
                 ComponentsItem(title: "Convenience Stores", selectionHandler: presentConvenienceStore)
+            ],
+            [
+                ComponentsItem(title: "Apple Pay", selectionHandler: presentApplePayComponent)
```
descorp, line 41: This should be a PKPaymentButton

AdyenDropIn/Presentation/WrapperViewController.swift
```diff
@@ -61,7 +61,7 @@ internal final class WrapperViewController: UIViewController {
 
 extension ModalViewController {
 
-    private var leastPresentableHeightScale: CGFloat { 0.3 }
+    private var leastPresentableHeightScale: CGFloat { 0.25 }
```
descorp, line 64: maybe we don't really need a bottom limit 🤔

kaphacius: maybe 🤔 i had to decrease it to match the design.

AdyenDropIn/Components/PreApplePayComponent.swift
kaphacius: 👍

AdyenDropIn/DropInComponent.swift
```diff
@@ -132,11 +132,13 @@ public final class DropInComponent: NSObject, PresentableComponent {
...
+        case let component as PreApplePayComponent:
+            component.presentationDelegate = self
+            navigationController.present(asModal: component)
         case let component as PresentableComponent where component.requiresModalPresentation:
             navigationController.present(asModal: component)
-        case let component as PresentableComponent:
-            let exceptions = (component.viewController is UIAlertController) || (component is ApplePayComponent)
-            navigationController.present(component.viewController, customPresentation: !exceptions)
+        case let component as PresentableComponent where component.viewController is UIAlertController:
+            navigationController.present(component.viewController, customPresentation: true)
```
mohammedDehairy, line 141: `customPresentation` should be false, right?

kaphacius: yes, you are correct 👍

AdyenDropIn/Presentation/WrapperViewController.swift
descorp, line 64: Is it look ok on mini device?

Adyen/UI/View Controllers/ADYViewController.swift
```diff
@@ -4,32 +4,39 @@
...
+        self.view = _view
     }
     
-    override internal var preferredContentSize: CGSize {
+    override public var preferredContentSize: CGSize {
         get {
             let targetSize = CGSize(width: UIScreen.main.bounds.width,
                                     height: UIView.layoutFittingCompressedSize.height)
-            return qrCodeView.systemLayoutSizeFitting(targetSize,
+            return view.systemLayoutSizeFitting(targetSize,
```
mohammedDehairy, line 39: something wrong with formatting I think

Adyen/UI/View Controllers/ADYViewController.swift
```diff
@@ -4,32 +4,39 @@
...
 import UIKit
 
-internal final class QRCodeViewController: UIViewController {
+/// :nodoc:
+/// `ADYViewController` serves as a height-aware `UIViewController`
+final public class ADYViewController: UIViewController {
     
-    private let qrCodeView: QRCodeView
+    /// :nodoc:
+    private let _view: UIView
```
mohammedDehairy, line 14: we have private vars with a `_` prefix, but only for `_isDropin` because it has to be public but yet used only internally so the `_` has a point, but here its not really the swift way I would say.

AdyenDropIn/Components/PreApplePayComponent.swift
```diff
@@ -0,0 +1,93 @@
...
+internal final class PreApplePayComponent: Localizable, PresentableComponent, FinalizableComponent, PaymentComponent {
+    
+    /// :nodoc:
+    internal let paymentMethod: PaymentMethod
+    
+    /// :nodoc:
+    private var payment: Payment? { _payment }
+    
+    /// :nodoc:
+    private let _payment: Payment
```
mohammedDehairy, line 23: `_` here too, not sure if we should start this trend.

Adyen/UI/View Controllers/ADYViewController.swift
descorp: `contentView` ?

mohammedDehairy (approved):

PR #475, opened 2021-04-30
chore: Use en-US localization as a fallback if current language does not have a translation

kaphacius:
This PR adds a fallback localization mechanism. In case current language does not have a translation for a specific key, an English translation will be used. We consider en-US `Localizable.strings` to be the most complete version.

### DIFF
--- a/Adyen/Utilities/Localization.swift
+++ b/Adyen/Utilities/Localization.swift
@@ -29,9 +29,10 @@ private struct LocalizationInput {
 ///   - arguments: The arguments to substitute in the templated localized string.
 /// - Returns: The localized string for the given key, or the key itself if the localized string could not be found.
 public func localizedString(_ key: LocalizationKey, _ parameters: LocalizationParameters?, _ arguments: CVarArg...) -> String {
-    let possibleInputs = buildPossibleInputs(key.key, parameters)
-
-    let result = attempt(possibleInputs) ?? fallbackLocalizedString(key: key.key)
+    
+    // Use fallback in case attempt result is nil or empty
+    let result = attempt(buildPossibleInputs(key.key, parameters))
+        .flatMap(\.adyen.nilIfEmpty) ?? fallbackLocalizedString(key: key.key)
     
     guard !arguments.isEmpty else {
         return result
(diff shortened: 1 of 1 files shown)


Demo/Configuration.swift
```diff
@@ -18,11 +18,11 @@ internal enum Configuration {
 
     static let appName = "Adyen Demo"
 
-    static let amount = Payment.Amount(value: 17408, currencyCode: "EUR")
+    static let amount = Payment.Amount(value: 17408, currencyCode: "BRL")
 
     static let reference = "Test Order Reference - iOS UIHost"
     
-    static let countryCode = "NL"
+    static let countryCode = "BR"
```
descorp, line 25: don't commit this. It also breaks Swift UI tests

Adyen/Utilities/Localization.swift
```diff
@@ -41,7 +42,16 @@ public func localizedString(_ key: LocalizationKey, _ parameters: LocalizationPa
 }
 
 private func fallbackLocalizedString(key: String) -> String {
-    NSLocalizedString(key, tableName: nil, bundle: Bundle.coreInternalResources, comment: "")
+    let localizedFallback = NSLocalizedString(key, tableName: nil, bundle: Bundle.coreInternalResources, comment: "")
+    
+    if localizedFallback != key && localizedFallback.isEmpty == false {
```
descorp, line 47: It can't be equals `key`, right? 🤔

descorp: Why can add `var nonEmpty: Bool ` extension for String 😆

kaphacius: if the key is missing from the `Localizable.strings`, `NSLocalizedString` will return the key. here we need to check for that

Demo/Configuration.swift
kaphacius: 👍

Adyen/Utilities/Localization.swift
```diff
@@ -41,7 +42,16 @@ public func localizedString(_ key: LocalizationKey, _ parameters: LocalizationPa
...
+        return localizedFallback
+    } else {
+    // Fallback to en-US
+        return Bundle.coreInternalResources.path(forResource: "en-US", ofType: "lproj")
+            .flatMap(Bundle.init(path: ))
+            .map { NSLocalizedString(key, tableName: nil, bundle: $0, comment: "") } ?? key
```
mohammedDehairy, line 53: why not put this in the `buildPossibleInputs`?

kaphacius: this is more of a fallback option, so i thought it should be in `fallbackLocalizedString`

Adyen/Utilities/Localization.swift
mohammedDehairy: yeah, actually you're right, here is better.

but can we unit test this in `ADYLocalizationTests`?

kaphacius: will do! 👍

Adyen/Utilities/Localization.swift
mohammedDehairy: yeah, actually you're right, here is better.

but can we unit test this in `ADYLocalizationTests`?

kaphacius: unfortunately i have not found a reliable way to test this

mohammedDehairy: 👍 no problem.

mohammedDehairy (approved):

PR #478, opened 2021-05-03
Full address with country dropdown

descorp:
## Open PR

* refactor FormItems
* make title and placeholder Observable
* add CompoundFormItem and SelfRenderingFormItemDelegate


Adyen/UI/Form/Items/Address/AddressViewModel.swift
```diff
@@ -0,0 +1,82 @@
+//
+// Copyright (c) 2021 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+internal enum AddressFields: String, CaseIterable {
```
mohammedDehairy, line 9: `AddressField` singular makes more sense to me than a plural :-)

Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
```diff
@@ -54,7 +52,7 @@ public final class FullFormAddressItem: FormValueItem<AddressInfo, AddressStyle>
 
     internal lazy var countrySelecItem: FormRegionPickerItem = {
         let locale = localizationParameters?.locale.map { Locale(identifier: $0) } ?? Locale.current
-        let countries = RegionRepository.localCountryFallback(for: locale as NSLocale)
+        let countries = RegionRepository.localCountryFallback(for: locale as NSLocale).sorted { $0.name < $1.name }
```
mohammedDehairy, line 55: just `sorted()` will work, right?

Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
```diff
@@ -64,76 +62,98 @@ public final class FullFormAddressItem: FormValueItem<AddressInfo, AddressStyle>
...
+    private func createTextItem(for field: AddressFields, from viewModel: AddressViewModel) -> FormItem {
+        let item = FormTextInputItem(style: style.textField)
+        item.validationFailureMessage = validationMessage
+        item.title = localizedString(viewModel.labels[field], localizationParameters)
+        item.placeholder = localizedString(viewModel.placeholder[field], localizationParameters)
 
-        return cityTextItem
-    }()
+        if viewModel.optionalFields.contains(field) {
+            item.title.map { item.title = $0 + " \(optionalMessage)" }
```
mohammedDehairy, line 116: that's weird, you're setting `item.title` inside the map closure? is that what you mean?
```
item.title = item.title.map { $0 + " \(optionalMessage)" }
```

or maybe
```
item.title.append(" \(optionalMessage)")
```
?

Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
```diff
@@ -64,76 +62,98 @@ public final class FullFormAddressItem: FormValueItem<AddressInfo, AddressStyle>
...
     /// :nodoc:
     override public func build(with builder: FormItemViewBuilder) -> AnyFormItemView {
         builder.build(with: self)
     }
 
 }
+
+private func localizedString(_ key: LocalizationKey?, _ parameters: LocalizationParameters?, _ arguments: CVarArg...) -> String? {
+    guard key != nil else { return nil }
+    return localizedString(key.unsafelyUnwrapped, parameters, arguments) as String
```
mohammedDehairy, line 158: why do we need `unsafelyUnwrapped`? why not like this?:

```
guard let key = key else { return nil }
     return localizedString(key, parameters, arguments) as String
```

Adyen/UI/Form/Items/Basic Items/FormVerticalStackItemView.swift
```diff
@@ -45,3 +50,15 @@ internal final class FormVerticalStackItemView<FormItemType: FormItem>: FormItem
...
 }
+
+extension FormVerticalStackItemView: SelfRenderingFormItemDelegate {
+
+    internal func didUpdatedItems(_ items: [FormItem]) {
+        stackView.arrangedSubviews.forEach { $0.removeFromSuperview() }
+        views = items.map(FormVerticalStackItemView.build)
+        views.forEach(stackView.addArrangedSubview)
+        stackView.setNeedsLayout()
+        views.first { $0.canBecomeFirstResponder }?.becomeFirstResponder()
```
mohammedDehairy, line 61: you need to call `self.updatePreferredContentSize()`in case the height changed , right?

Adyen/UI/Form/Items/FormItem.swift
```diff
@@ -53,3 +54,16 @@ public protocol ValidatableFormItem: FormItem {
 /// A form item that requires keyboard input or otherwise custom input view.
 /// :nodoc:
 public protocol InputViewRequiringFormItem: FormItem {}
+
+/// Delegate to the view all events that requireschange in corespondent FormView changes.
+/// :nodoc:
+public protocol SelfRenderingFormItemDelegate: AnyObject {
```
mohammedDehairy, line 60: does this need to be public?

Adyen/UI/Form/Items/Value/FormValueItem.swift
```diff
@@ -24,7 +24,7 @@ public protocol FormValueItemStyle: TintableStyle {
 open class FormValueItem<ValueType: Equatable, StyleType: FormValueItemStyle>: FormItem {
 
     /// :nodoc:
-    public var subitems: [FormItem] = []
+    public var subitems: [FormItem] { [] }
```
mohammedDehairy, line 27: why? :-D

Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
descorp: no, those are just `Region` structs. For `sorted` to work - need to adopt `Comparable` protocol.

Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
descorp: here `map` is like `if let ...`. I only want to append if `item.title` is not nil.

Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
descorp: because `key ` is enum and you can't do `guard let key = key` on enums. It have to be unwrapped to be checked, else enum is only checked as `guard case let key = .KEY`

Adyen/UI/Form/Items/Basic Items/FormVerticalStackItemView.swift
descorp: hm.. I believe because there are keyboard event - it will be called eventually.
I don't like it, but it works 🤔

Adyen/UI/Form/Items/Value/FormValueItem.swift
descorp: For override. `public var subitems: [FormItem] = []` is **{get set}**, but `public var subitems: [FormItem] { [] }` is **{get}**

Adyen/UI/Form/Items/FormItem.swift
descorp: Not for now 👍

Adyen/UI/Form/Items/Address/AddressViewModel.swift
descorp: Good one! 👍

Adyen/UI/Form/Items/Basic Items/FormVerticalStackItemView.swift
mohammedDehairy: its fine for now

Adyen/UI/Form/Items/Address/AddressViewModel.swift
```diff
@@ -0,0 +1,82 @@
...
+                                                  .appartment: .apartmentSuiteFieldTitle],
+                                         placeholder: [.city: .cityFieldPlaceholder,
+                                                       .houseNumberOrName: .houseNumberFieldPlaceholder,
+                                                       .street: .streetFieldPlaceholder,
+                                                       .stateOrProvince: .provinceOrTerritoryFieldPlaceholder,
+                                                       .postalCode: .postalCodeFieldPlaceholder,
+                                                       .appartment: .apartmentSuiteFieldPlaceholder],
+                                         optionalFields: [.appartment],
+                                         schema: AddressFields.allCases.map { .item($0) })
+        switch countryCode {
```
kaphacius, line 45: nice use of `switch` 👍

Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
```diff
@@ -54,7 +52,7 @@ public final class FullFormAddressItem: FormValueItem<AddressInfo, AddressStyle>
 
     internal lazy var countrySelecItem: FormRegionPickerItem = {
```
kaphacius, line 53: countySelec**t**Item*

Adyen/UI/Form/Items/Value/FormValueItem.swift
mohammedDehairy: then its can be let, right?:
```
public let subitems: [FormItem] = []
```

Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
kaphacius: i'm a big fan of using `map` everywhere, but here might be not the best case. the `transform` closure ideally should be w/o side-effects. i think in this case `if let` suits better.

Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
```diff
@@ -64,76 +62,98 @@ public final class FullFormAddressItem: FormValueItem<AddressInfo, AddressStyle>
...
-        cityTextItem.validationFailureMessage = validationMessage
-        cityTextItem.identifier = ViewIdentifierBuilder.build(scopeInstance: self, postfix: "provinceOrTerritory")
+        bind(item: item, to: field)
+        return item
+    }
 
-        return cityTextItem
-    }()
+    private func bind(item: FormRegionPickerItem, to field: AddressFields) {
+        observers[field].map { remove($0) }
```
kaphacius, line 126: `observers[field].map(remove(_:))` might be possible too, looks a bit cleaner

Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
kaphacius: i think that's a perfect place to use map instead of this function
```
item.title = viewModel.labels[.stateOrProvince]
            .map { localizedString($0, localizationParameters) }
```

Adyen/UI/Form/Items/Basic Items/FormVerticalStackItemView.swift
```diff
@@ -45,3 +50,15 @@ internal final class FormVerticalStackItemView<FormItemType: FormItem>: FormItem
     }
 
 }
+
+extension FormVerticalStackItemView: SelfRenderingFormItemDelegate {
+
+    internal func didUpdatedItems(_ items: [FormItem]) {
```
kaphacius, line 56: `didUpdateItems~*

Adyen/UI/Form/Items/Value Pickers/FormRegionPickerItem.swift
```diff
@@ -16,15 +16,26 @@ public typealias RegionPickerItem = BasePickerElement<Region>
...
-                   selectableValues: selectableValues.map { RegionPickerItem(identifier: $0.identifier, element: $0) },
+        super.init(preselectedValue: Self.pickerElement(from: preselectedValue),
+                   selectableValues: selectableValues.map(Self.pickerElement),
                    style: style)
     }
 
+    public func updateData(_ regions: [Region]) {
+        assert(!regions.isEmpty)
+        selectableValues = regions.map(Self.pickerElement)
+        value = selectableValues[0]
```
kaphacius, line 27: this is a bit dangerous :)

Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
descorp: not a big fun of embedded IFs, but ok..

Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
descorp: 👍

Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
kaphacius: does it have to be an embedded if?
maybe something like
```
        if viewModel.optionalFields.contains(field),
           let itemTitle = item.title {
            item.title = [itemTitle, optionalMessage].joined(separator: " ")
        } else {
            item.validator = LengthValidator(minimumLength: 2, maximumLength: 70)
        }
```

Adyen/UI/Form/Items/Value/FormValueItem.swift
descorp: Not really, because than it become immutable and you can't override or change it, and whole idea of `CompoundItem` is that items could change asynchronously

Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
descorp: I will go with `countrySelectItem` 😄

Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
descorp: 👍

Adyen/UI/Form/Items/Basic Items/FormVerticalStackItemView.swift
descorp: 👍

Adyen/UI/Form/Items/Value Pickers/FormRegionPickerItem.swift
descorp: But what choose do I have? 🤔

kaphacius: maybe use `if let firstElement =  regions.first`

mohammedDehairy: I would use 
```
value = selectableValues.first
```

this way its safer

descorp: `first ` is optional. Here we assume that array **is never empty**. If it is empty - this is serious error

descorp: this method is not used

kaphacius (dismissed):

kaphacius (dismissed):

kaphacius (approved):

PR #486, opened 2021-05-05
Fix payment details

descorp:
### Changes

* move "Remember for next time" below PostCode field
* remove `billing address` from CardDetails decoder

### DIFF
--- a/AdyenCard/Components/Card/CardDetails.swift
+++ b/AdyenCard/Components/Card/CardDetails.swift
@@ -92,7 +92,6 @@ public struct CardDetails: PaymentMethodDetails, BillingAddressInformation {
         case encryptedSecurityCode
         case holderName
         case fundingSource
-        case billingAddress
     }
     
 }
--- a/AdyenCard/Components/Card/CardViewController.swift
+++ b/AdyenCard/Components/Card/CardViewController.swift
@@ -71,10 +71,6 @@ internal class CardViewController: FormViewController {
             append(holderNameItem)
         }
 
-        if configuration.showsStorePaymentMethodField {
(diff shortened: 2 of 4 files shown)


kaphacius (dismissed):

kaphacius (approved):

AdyenCard/Components/Card/CardDetails.swift
```diff
@@ -92,7 +92,6 @@ public struct CardDetails: PaymentMethodDetails, BillingAddressInformation {
         case encryptedSecurityCode
         case holderName
         case fundingSource
-        case billingAddress
```
mohammedDehairy, line 95: was this a bug in the released version?

mohammedDehairy (approved):

AdyenCard/Components/Card/CardDetails.swift
descorp: `billingAddress` is duplicated in payment data, but hopefully it is not causing any troubles (at least on TEST)

descorp: Like 

```json

{
  ...
  "paymentMethod" : {
    ...
    "billingAddress" : {
      "postalCode" : "123456",
      "country" : "ZZ",
      "street" : "null",
      "houseNumberOrName" : "null",
      "city" : "null",
      "stateOrProvince" : "null"
    },
  },
  "billingAddress" : {
    "postalCode" : "123456",
    "country" : "ZZ",
    "street" : "null",
    "houseNumberOrName" : "null",
    "city" : "null",
    "stateOrProvince" : "null"
  },
  ...
}
```

PR #488, opened 2021-05-06
chore: Added arabic translations and used RTL language friendly autolayout constraints

mohammedDehairy:
chore: Added arabic translations and used RTL language friendly autolayout constraints

### DIFF
--- a/AdyenDropIn/Presentation/WrapperViewController.swift
+++ b/AdyenDropIn/Presentation/WrapperViewController.swift
@@ -49,8 +49,8 @@ internal final class WrapperViewController: UIViewController {
         child.view.translatesAutoresizingMaskIntoConstraints = false
         let topConstraint = child.view.topAnchor.constraint(equalTo: view.topAnchor, constant: 0)
         NSLayoutConstraint.activate([
-            child.view.leftAnchor.constraint(equalTo: view.leftAnchor),
-            child.view.rightAnchor.constraint(equalTo: view.rightAnchor),
+            child.view.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            child.view.trailingAnchor.constraint(equalTo: view.trailingAnchor),
             child.view.bottomAnchor.constraint(equalTo: view.bottomAnchor),
             topConstraint
         ])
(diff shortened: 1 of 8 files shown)


kaphacius (approved):

AdyenDropIn/Presentation/WrapperViewController.swift
```diff
@@ -49,8 +49,8 @@ internal final class WrapperViewController: UIViewController {
         child.view.translatesAutoresizingMaskIntoConstraints = false
         let topConstraint = child.view.topAnchor.constraint(equalTo: view.topAnchor, constant: 0)
         NSLayoutConstraint.activate([
```
descorp, line 51: this could be `adyen.anchore`

AdyenDropIn/Views/ModalToolbar.swift
```diff
@@ -91,21 +91,21 @@ internal final class ModalToolbar: UIView {
     }
 
     private func setupLayoutForCenteredMode() {
```
descorp, line 93: Does this work for all the cases?
Does `rightCancel` still make sense for Arabic?

mohammedDehairy: it works yes, and for arabic speakers cancel on the left make sense yes, like when UINavigationController put the back button on the left and viewController's are push from right to left.

AdyenDropIn/Presentation/WrapperViewController.swift
mohammedDehairy: could be, and we get the top constraint by searching for it, but I don't like this approach :-).

AdyenDropIn/Views/ModalToolbar.swift
descorp: that is what I mean:
If it says "left", but it is actually on the right - it could be confusing 😅

We can probably enforce `rightCancel` or `leftCancel` to be actually on left on to the right respectily.. or rename rightCancel` or `leftCancel` to something else ?

descorp (approved):

AdyenDropIn/Views/ModalToolbar.swift
mohammedDehairy: fixed by adding `ToolbarMode.natural` and made it the default.

descorp (approved):

kaphacius (approved):

PR #490, opened 2021-05-06
V3: post code

descorp:
### Changes

* add post code field to card component

### DIFF
--- a/Adyen/Assets/ko.lproj/Localizable.strings
+++ b/Adyen/Assets/ko.lproj/Localizable.strings
@@ -37,7 +37,6 @@
 "adyen.phoneNumber.placeholder" = "123–456–789";
 "adyen.card.cvcItem.placeholder.digits" = "%@자리";
 "adyen.emailItem.title" = "이메일 주소";
-"adyen.emailItem.placeHolder" = "이메일 주소";
 "adyen.emailItem.invalid" = "유효하지 않은 이메일 주소";
 "adyen.mbway.confirmPayment" = "MB WAY 앱에서 결제를 확인하십시오";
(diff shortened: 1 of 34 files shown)


Adyen/Components/Base/AddressInfo.swift
```diff
@@ -0,0 +1,95 @@
...
+    /// Required for the US and Canada.
+    public var stateOrProvince: String?
+
+    /// The name of the street.
+    /// The house number should not be included in this field; it should be separately provided via houseNumberOrName.
+    public var street: String?
+
+    /// The name or code of apartment. Optional.
+    /// Will be included into houseNumberOrName.
+    public var apartment: String?
```
mohammedDehairy, line 55: we don't need all those fields if its just postal code, right? actually we don't need a whole `AddressInfo` object, just the postal code as a string.

descorp: we still need it for JSON mapping

Adyen/Assets/ko.lproj/Localizable.strings
```diff
@@ -37,6 +37,7 @@
 "adyen.phoneNumber.placeholder" = "123–456–789";
 "adyen.card.cvcItem.placeholder.digits" = "%@자리";
 "adyen.emailItem.title" = "이메일 주소";
+"adyen.emailItem.placeHolder" = "이메일 주소";
```
mohammedDehairy, line 40: this line change is on some files but not others 🤔

descorp: 👍

kaphacius (approved):

PR #495, opened 2021-05-10
fix: Fix constraints on QRCodeView and PreApplePay view not using safeAreaLayoutGuide

kaphacius:
This PR will fix issue when QRCodeView and PreApplePayView subviews were not using `safeAreaLayoutGuide.leadingAnchor` and `safeAreaLayoutGuide.trailingAnchor`, causing them to look wrong in landscape orientation.

### DIFF
--- a/AdyenActions/UI/View Controllers/QR Code/QRCodeView.swift
+++ b/AdyenActions/UI/View Controllers/QR Code/QRCodeView.swift
@@ -55,7 +55,7 @@ internal final class QRCodeView: UIView, Localizable, Observer {
         logo.translatesAutoresizingMaskIntoConstraints = false
         NSLayoutConstraint.activate([
             logo.bottomAnchor.constraint(equalTo: topAnchor, constant: -7),
-            logo.centerXAnchor.constraint(equalTo: centerXAnchor)
+            logo.centerXAnchor.constraint(equalTo: safeAreaLayoutGuide.centerXAnchor)
         ])
     }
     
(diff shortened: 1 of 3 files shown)


mohammedDehairy (approved):

AdyenActions/UI/View Controllers/QR Code/QRCodeView.swift
```diff
@@ -55,7 +55,7 @@ internal final class QRCodeView: UIView, Localizable, Observer {
         logo.translatesAutoresizingMaskIntoConstraints = false
         NSLayoutConstraint.activate([
             logo.bottomAnchor.constraint(equalTo: topAnchor, constant: -7),
-            logo.centerXAnchor.constraint(equalTo: centerXAnchor)
+            logo.centerXAnchor.constraint(equalTo: safeAreaLayoutGuide.centerXAnchor)
```
descorp, line 58: Somehow I belive that centerX is always center?

kaphacius: i think if there is a view that has safe area on the left, but no safe area on the right (does not span the full width of the phone) - `centerX` might be different.

descorp (approved):

AdyenActions/UI/View Controllers/QR Code/QRCodeView.swift
descorp: Than center of proggress bar and center of instructionLabel would be different 🤔
I would say better align those two

PR #506, opened 2021-05-12
Gift card flow

mohammedDehairy:
Gift card flow


Adyen/Components/Base/PaymentComponentData.swift
```diff
@@ -90,13 +96,20 @@ public struct PaymentComponentData: ShopperInformation, BillingAddressInformatio
...
     /// - Parameters:
     ///   - paymentMethodDetails: The payment method details submitted from the payment component.
+    ///   - amount: The payment amount.
+    ///   - order: The partial payment order if any.
     ///   - storePaymentMethod: Whether the user has chosen to store the payment method.
     ///   - browserInfo: The device default browser info.
-    ///   - addressInfo: The payment's billing address.
-    public init(paymentMethodDetails: PaymentMethodDetails, storePaymentMethod: Bool = false, browserInfo: BrowserInfo? = nil) {
+    public init(paymentMethodDetails: PaymentMethodDetails,
+                amount: Payment.Amount?,
```
descorp, line 104: should be `Payment.Amount? = nil`

Adyen/Components/Base/PaymentComponentData.swift
```diff
@@ -107,6 +120,7 @@ public struct PaymentComponentData: ShopperInformation, BillingAddressInformatio
     public func dataByAddingBrowserInfo(completion: @escaping ((_ newData: PaymentComponentData) -> Void)) {
         BrowserInfo.initialize {
             completion(PaymentComponentData(paymentMethodDetails: self.paymentMethod,
+                                            amount: amount,
```
descorp, line 123: What about `order` ?

Adyen/Components/Base/ReadyToSubmitPaymentComponent.swift
```diff
@@ -0,0 +1,30 @@
...
+// Copyright (c) 2021 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+/// A payment component for which payment data is ready to be submitted.
+/// :nodoc:
+public final class ReadyToSubmitPaymentComponent: PaymentComponent {
```
descorp, line 11: Not sure about the naming 🤔
What makes other Payment Components not "ready to Submit" ?

Adyen/Core/APIClient/Requests/OrderStatusRequest.swift
```diff
@@ -0,0 +1,45 @@
...
+//  Adyen
+//
+//  Created by Mohamed Eldoheiri on 5/6/21.
+//  Copyright © 2021 Adyen. All rights reserved.
+//
+
+import Foundation
+
+/// :nodoc:
+public struct OrderStatusRequest: Request {
```
descorp, line 12: We are not exposing this to Jazzy? 
At least we can add proper comments from [API DOCS](https://docs.adyen.com/api-explorer/#/CheckoutService/v67/post/orders)

Adyen/Core/EmptyPaymentComponent.swift
```diff
@@ -0,0 +1,29 @@
...
+// Copyright (c) 2021 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+/// Provides a placeholder for payment methods that don't need any payment detail to be filled.
+/// :nodoc:
+public final class EmptyPaymentComponent: PaymentComponent {
```
descorp, line 11: This was on DropIn, right?

Adyen/Core/Payment Methods/PaymentMethod.swift
```diff
@@ -44,6 +44,22 @@ public struct DisplayInformation {
     /// The name of the logo resource.
     /// :nodoc:
     public var logoName: String
+
+    /// The trailing disclosure text.
+    /// :nodoc:
+    public var disclosureText: String?
+
+    /// Initializes the Doku Wallet, Doku Alfamart, or Doku Indomaret payment method.
```
descorp, line 52: Not only `Doku` now, right?

Adyen/Core/Payment Methods/PreselectedPaymentMethod.swift
```diff
@@ -0,0 +1,48 @@
+//
+// Copyright (c) 2021 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+/// A payment method thats to be shown as preselected payment method that is ready to be submitted.
```
descorp, line 9: Not sure I fully understand this description 🤔

Adyen/Helpers/UIViewConstraintsHelper.swift
```diff
@@ -48,10 +48,7 @@ extension AdyenScope where Base: UIView {
         containerView.addSubview(base)
         containerView.translatesAutoresizingMaskIntoConstraints = false
         base.translatesAutoresizingMaskIntoConstraints = false
-        anchor(inside: containerView, with: UIEdgeInsets(top: insets.top,
-                                                         left: insets.left,
-                                                         bottom: insets.bottom,
-                                                         right: insets.right))
+        anchor(inside: containerView, with: insets)
```
descorp, line 51: 😅

Adyen/UI/Form/Items/Text/FormTextInputItem.swift
```diff
@@ -8,8 +8,10 @@ import Foundation
 
 /// An item for plain text input
 /// :nodoc:
-public final class FormTextInputItem: FormTextItem {
+public class FormTextInputItem: FormTextItem {
```
descorp, line 11: this supposed to be a final class. Use `FormTextItem ` if you need to override something

Adyen/UI/List/ListCell.swift
```diff
@@ -97,7 +97,7 @@ public final class ListCell: UITableViewCell {
         let constraints = [
             itemView.topAnchor.constraint(equalTo: topAnchor),
             itemView.leadingAnchor.constraint(equalTo: layoutGuide.leadingAnchor),
-            itemView.trailingAnchor.constraint(lessThanOrEqualTo: layoutGuide.trailingAnchor),
+            itemView.trailingAnchor.constraint(equalTo: layoutGuide.trailingAnchor),
```
descorp, line 100: I am wondering why we had it like this in a first place 🤔

Adyen/UI/List/ListFooterView.swift
```diff
@@ -0,0 +1,82 @@
...
+    // MARK: - Title Label
+
+    private let title: String
+
+    // MARK: - UI
+
+    private lazy var stackView: UIStackView = {
+        let stackView = UIStackView(arrangedSubviews: [separatorView, titleContainerView])
+        stackView.axis = .vertical
+        stackView.alignment = .fill
```
descorp, line 41: I belive `alignment = .fill` and `distribution = .fill` is default

Adyen/UI/List/ListItemView.swift
```diff
@@ -136,14 +152,21 @@ public final class ListItemView: UIView, AnyFormItemView {
             imageView.widthAnchor.constraint(equalToConstant: imageSize.width),
             imageView.heightAnchor.constraint(equalToConstant: imageSize.height),
             
-            textStackView.topAnchor.constraint(greaterThanOrEqualTo: layoutMarginsGuide.topAnchor),
-            textStackView.centerYAnchor.constraint(equalTo: centerYAnchor),
-            textStackView.leadingAnchor.constraint(equalTo: imageView.trailingAnchor, constant: 16.0),
-            textStackView.trailingAnchor.constraint(equalTo: layoutMarginsGuide.trailingAnchor),
-            textStackView.bottomAnchor.constraint(lessThanOrEqualTo: layoutMarginsGuide.bottomAnchor),
+            trailingTextLabel.leadingAnchor.constraint(equalTo: titleSubtitleStackView.trailingAnchor),
```
descorp, line 155: this is not going to flip on RTL ?

Adyen/UI/List/ListItemView.swift
```diff
@@ -136,14 +152,21 @@ public final class ListItemView: UIView, AnyFormItemView {
             imageView.widthAnchor.constraint(equalToConstant: imageSize.width),
             imageView.heightAnchor.constraint(equalToConstant: imageSize.height),
             
-            textStackView.topAnchor.constraint(greaterThanOrEqualTo: layoutMarginsGuide.topAnchor),
```
descorp, line 139: Feels like this whole thing could be a stackview

Adyen/UI/List/ListItemView.swift
```diff
@@ -161,13 +184,4 @@ public final class ListItemView: UIView, AnyFormItemView {
         imageView.layer.borderColor = item?.style.image.borderColor?.cgColor ?? UIColor.Adyen.componentSeparator.cgColor
     }
     
-    override public var intrinsicContentSize: CGSize {
```
descorp, line 164: This weird was probably here to fix something in iOS 10 or 11..

Adyen/UI/UI Style/TextStyle.swift
```diff
@@ -30,10 +30,16 @@ public struct TextStyle: ViewStyle {
     /// - Parameter font: The font used to display the text.
     /// - Parameter color: The color of the text.
     /// - Parameter textAlignment: The technique to use for aligning the text.
```
descorp, line 32: Also need to update docs

Adyen/Utilities/Debugging/Assertion.swift
```diff
@@ -17,12 +17,20 @@ public enum AdyenAssertion {
...
-            listner?(message)
+            listner?(message())
             return
         }
+        assertionFailure(message())
+    }
 
-        assertionFailure(message)
+    /// :nodoc:
+    /// Calls `assertionFailure` when not runing Tests.
```
descorp, line 29: Description is wrong. `Calls `assertionFailure` when condition not met`

Adyen/Utilities/Debugging/Assertion.swift
```diff
@@ -17,12 +17,20 @@ public enum AdyenAssertion {
 
     /// :nodoc:
     /// Calls `assertionFailure` when not runing Tests.
-    public static func assert(message: String) {
+    public static func assert(message: @autoclosure () -> String) {
```
descorp, line 20: let's call this one `assertionFailure `

Adyen/Validators/BalanceValidator.swift
```diff
@@ -0,0 +1,77 @@
...
+            return true
+        }
+        return transactionLimit.currencyCode == balance.availableAmount.currencyCode
+    }
+
+    private func calculateExpendableLimit(with balance: Balance) -> Payment.Amount {
+        let result: Payment.Amount
+        let availableAmount = balance.availableAmount
+        if let balanceTransactionLimit = balance.transactionLimit {
+            result = min(balanceTransactionLimit, availableAmount)
```
descorp, line 71: No need for `let result: Payment.Amount`:
you can do `return min(balanceTransactionLimit, availableAmount)` or `return availableAmount`

AdyenCard/Components/Card/CardPublicKeyFetcherProtocol.swift
```diff
@@ -0,0 +1,44 @@
+//
+// Copyright (c) 2021 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Adyen
+import Foundation
+
+internal protocol CardPublicKeyFetcherProtocol: PaymentComponent {
```
descorp, line 10: How about `CardPublicKeyRecipient` or something?

mohammedDehairy: `CardPublicKeyConsumer` I think is better

Adyen/Validators/BalanceValidator.swift
mohammedDehairy: I like one return statement per function :-)

AdyenCard/Components/Card/CardComponentExtensions.swift
```diff
@@ -21,11 +21,6 @@ extension CardComponent {
 
 extension CardComponent {
     
-    internal func isPublicKeyValid(key: String) -> Bool {
```
descorp, line 24: No need to validate public key?

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -244,8 +244,8 @@ internal protocol CardViewControllerDelegate: AnyObject {
 
 }
 
-private extension FormValueItem where ValueType == String {
-    var nonEmptyValue: String? {
+extension FormValueItem where ValueType == String {
```
descorp, line 247: maybe we can move it somewhere more appropriate

AdyenCard/Components/Card/CardComponentExtensions.swift
mohammedDehairy: nop, its fetched from the backend, right? no need IMO

Adyen/UI/List/ListCell.swift
mohammedDehairy: I don't know, but I think its not needed

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -0,0 +1,312 @@
...
+
+        startLoading()
+
+        let data = createPaymentData()
+
+        partialPaymentDelegate?.checkBalance(data, from: self) { [weak self] result in
+            guard let self = self else { return }
+            result
+                .mapError(Error.otherError)
+                .flatMap(self.check(balance:))
```
descorp, line 192: This is nice

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -0,0 +1,312 @@
...
+        }
+        do {
+            return .success(try BalanceValidator().validate(balance: balance, toPay: payment.amount))
+        } catch {
+            return .failure(error)
+        }
+    }
+
+    private func onReadyToPayFullAmount() -> Result<Void, Swift.Error> {
+        AdyenAssertion.assert(message: "readyToSubmitComponentDelegate is nil",
```
descorp, line 231: This will fail if not Drop-In .. 

Maybe `assert ` should do the other way around?
It should assert that condition is met. Otherwise it should be called `failOn`

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -0,0 +1,312 @@
...
+                                    amount: payment?.amount,
+                                    order: order,
+                                    storePaymentMethod: false)
+    }
+}
+
+extension Result {
+
+    @discardableResult
+    func handleSuccess(_ handler: (Success) -> Void) -> Result<Success, Failure> {
```
descorp, line 291: should be internal ?

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -0,0 +1,312 @@
...
+        case let .failure(error):
+            handler(error)
+        default: ()
+        }
+        return self
+    }
+}
+
+/// :nodoc:
+extension GiftCardComponent: TrackableComponent {}
```
descorp, line 312: BTW, `TrackableComponent` could be `/// :nodoc:`

AdyenCard/Components/GiftCardComponent/GiftCardDetails.swift
```diff
@@ -0,0 +1,50 @@
...
+    public let cardNumber: String
+
+    /// The encrypted security code.
+    public let securityCode: String
+
+    /// The gift card brand
+    public let brand: String
+
+    /// :nodoc:
+    public var holderName: String? = "BALANCE EUR 10000"
```
descorp, line 26: Is this default holder name really necessary?

Adyen/Core/EmptyPaymentComponent.swift
mohammedDehairy: not any more, its needed in lower level modules

Adyen/Core/APIClient/Requests/OrderStatusRequest.swift
mohammedDehairy: it shouldn't be exposed to jazzy, yes

AdyenDropIn/Components/PaymentMethodListComponent.swift
```diff
@@ -12,7 +12,7 @@ import UIKit
 internal final class PaymentMethodListComponent: ComponentLoader, PresentableComponent, Localizable {
     
     /// The components that are displayed in the list.
-    internal let components: SectionedComponents
+    internal let components: [ComponentsSection]
```
descorp, line 15: I like the twist 😆

AdyenDropIn/Components/PaymentMethodListComponent.swift
```diff
@@ -55,16 +56,24 @@ internal final class PaymentMethodListComponent: ComponentLoader, PresentableCom
...
-        let regularSection = ListSection(title: regularSectionTitle,
-                                         items: components.regular.map(item(for:)))
+
+        let sections: [ListSection] = components.map {
+            ListSection(title: $0.header?.title,
+                        items: $0.components.map(item(for:)),
+                        footerTitle: $0.footer?.title)
+        }
+
+//        let paidItems = components.paid.map(item(for:))
```
descorp, line 66: commented code?

AdyenDropIn/DropInComponent.swift
```diff
@@ -68,15 +74,68 @@ public final class DropInComponent: NSObject, PresentableComponent {
...
+    ///
+    /// - Parameter order: The partial payment order.
+    /// - Parameter paymentMethods: The new payment methods.
+    public func reload(with order: PartialPaymentOrder,
+                       _ paymentMethods: PaymentMethods) {
+        let paymentMethod1 = OrderPaymentMethod(lastFour: "1234", type: "giftcard", transactionLimit: .init(value: 100, currencyCode: "CAD"), amount: .init(value: 200, currencyCode: "CAD"))
+        let paymentMethod2 = OrderPaymentMethod(lastFour: "3455", type: "giftcard", transactionLimit: .init(value: 100, currencyCode: "CAD"), amount: .init(value: 100, currencyCode: "CAD"))
+        let result = OrderStatusResponse(remainingAmount: .init(value: 7408, currencyCode: "CAD"), paymentMethods: [paymentMethod1, paymentMethod2])
+        handle(result)
+//        let request = OrderStatusRequest(orderData: order.orderData)
```
descorp, line 99: Commented code

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
mohammedDehairy: flipped the condition

AdyenCard/Components/GiftCardComponent/GiftCardDetails.swift
mohammedDehairy: needed for testing

AdyenCard/Components/Card/CardPublicKeyFetcherProtocol.swift
mohammedDehairy: renamed it to `CardPublicKeyConsumer `

Adyen/Utilities/Debugging/Assertion.swift
mohammedDehairy: fixed

Adyen/Utilities/Debugging/Assertion.swift
mohammedDehairy: I like assert more

AdyenDropIn/DropInComponent.swift
```diff
@@ -68,15 +74,68 @@ public final class DropInComponent: NSObject, PresentableComponent {
     public func handle(_ action: Action) {
         actionComponent.handle(action)
     }
+
+    // MARK: - Handling Partial Payments
+
+    /// :nodoc:
+    private lazy var apiClient: APIClientProtocol = {
```
descorp, line 81: Maybe we can make another `manager` object or somehow move this logic outside?
DropIn already almost 400 lines of code..

Adyen/UI/UI Style/TextStyle.swift
mohammedDehairy: done

Adyen/UI/List/ListItemView.swift
mohammedDehairy: I'll test it and see

Adyen/UI/List/ListItemView.swift
mohammedDehairy: yeah, that's fine

Adyen/UI/List/ListFooterView.swift
mohammedDehairy: being explicit is better IMO

Adyen/UI/Form/Items/Text/FormTextInputItem.swift
mohammedDehairy: made it final

Adyen/Core/Payment Methods/PreselectedPaymentMethod.swift
mohammedDehairy: I renamed it and changed the docs, let me know if its still vague

Adyen/Core/Payment Methods/PaymentMethod.swift
mohammedDehairy: fixed

Adyen/Components/Base/ReadyToSubmitPaymentComponent.swift
mohammedDehairy: deleted this one and used `EmptyPaymentComponent`, they are basically the same

Adyen/Components/Base/PaymentComponentData.swift
mohammedDehairy: fixed

AdyenDropIn/Utilities/ComponentManager.swift
```diff
@@ -17,29 +17,62 @@ import Adyen
...
-    internal lazy var components: SectionedComponents = {
+    internal lazy var sections: [ComponentsSection] = {
         // Filter out payment methods without the Ecommerce shopper interaction.
         let storedPaymentMethods = paymentMethods.stored.filter { $0.supportedShopperInteractions.contains(.shopperPresent) }
+
+        // Paid section
+        let amountString: String = remainingAmount.flatMap {
+            AmountFormatter.formatted(amount: $0.value, currencyCode: $0.currencyCode)
+        } ?? "Amount"
+        let footerTitle = "Select payment method for the remaining " + amountString
```
descorp, line 50: This should be translated

Demo/Common/IntegrationExample.swift
```diff
@@ -53,15 +55,19 @@ internal final class IntegrationExample: APIClientAware {
 
     // MARK: - Networking
 
-    internal func requestPaymentMethods() {
-        let request = PaymentMethodsRequest()
+    internal func requestPaymentMethods(order: PartialPaymentOrder? = nil,
+                                        completion: ((PaymentMethods) -> Void)? = nil) {
+        let request = PaymentMethodsRequest(order: order)
```
descorp, line 60: what if order is nil?

Demo/Common/IntegrationExample.swift
```diff
@@ -77,12 +83,11 @@ internal final class IntegrationExample: APIClientAware {
     }
 
     internal func finish(with error: Error) {
-        let isCancelled = (error as? ComponentError) == .cancelled
```
descorp, line 80: no need to cancel?

Demo/Common/IntegrationExampleDropIn.swift
```diff
@@ -53,14 +58,25 @@ extension IntegrationExample {
     private func handle(_ action: Action) {
         (currentComponent as? DropInComponent)?.handle(action)
     }
+
+    internal func handle(_ order: PartialPaymentOrder) {
```
descorp, line 62: this should be in `IntegrationExample.swift`

descorp, line 62: Probably we have to copy those two methods to `IntegrationExampleComponents.swift ` as well

Demo/Common/IntegrationExampleDropIn.swift
```diff
@@ -53,14 +58,25 @@ extension IntegrationExample {
...
+    }
 }
 
 extension IntegrationExample: DropInComponentDelegate {
 
     internal func didSubmit(_ data: PaymentComponentData, for paymentMethod: PaymentMethod, from component: DropInComponent) {
         print("User did start: \(paymentMethod.name)")
-        let request = PaymentsRequest(data: data)
-        apiClient.perform(request, completionHandler: paymentResponseHandler)
+        handle(PartialPaymentOrder(pspReference: "ef2ef", orderData: "ef2f23f", reference: "e2f2f2", remainingAmount: .init(value: 7408, currencyCode: "CAD"), expiresAt: nil), paymentMethods!)
```
descorp, line 77: Mocked data

Demo/Common/IntegrationExamplePartialPayment.swift
```diff
@@ -0,0 +1,111 @@
...
+            }
+        }
+    }
+
+    internal func checkBalance(_ data: PaymentComponentData,
+                               from component: PaymentComponent,
+                               completion: @escaping (Result<Balance, Error>) -> Void) {
+        let balance = Balance(availableAmount: Payment.Amount(value: 10000, currencyCode: "CAD"),
+                              transactionLimit: Payment.Amount(value: 100000, currencyCode: "CAD"))
+        completion(.success(balance))
```
descorp, line 32: Nice! I prefer dedicated file for PartialPayment - it is a huge flow..

Tests/AdyenTests/Adyen Tests/Components/Gift Card/GiftCardComponentTests.swift
```diff
@@ -0,0 +1,414 @@
...
+
+        let didFailExpectation = expectation(description: "Expect delegateMock.onDidFail to be called.")
+        didFailExpectation.assertForOverFulfill = true
+        delegateMock.onDidFail = { error, component in
+            XCTAssertTrue(component === self.sut)
+            XCTAssertNotNil(error as? Dummy)
+            didFailExpectation.fulfill()
+        }
+
+        let onCheckBalanceExpectation = expectation(description: "Expect partialPaymentDelegate.onCheckBalance to be called.")
```
descorp, line 64: Why new line for every single line in this file? 😅

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
mohammedDehairy: it is

AdyenDropIn/DropInComponent.swift
mohammedDehairy: yes for now

AdyenDropIn/Utilities/ComponentManager.swift
mohammedDehairy: yep, it will be

Demo/Common/IntegrationExample.swift
mohammedDehairy: that's fine `PaymentMethodsRequest.order` is optional

Demo/Common/IntegrationExample.swift
mohammedDehairy: yeah, I though why not show the alert when user cancels, I can revert if its needed

Demo/Common/IntegrationExampleDropIn.swift
mohammedDehairy: yep, this is a draft PR :-D

Adyen/Components/Base/ReadyToSubmitPaymentComponent.swift
descorp: I am thinking about a better name. Maybe `InstantPaymentComponent ` ? 😆

Tests/AdyenTests/Adyen Tests/Components/Gift Card/GiftCardComponentTests.swift
mohammedDehairy: easier to read I guess, its soft on the eyes :-)

Adyen/UI/List/ListCell.swift
descorp: Something on iOS 10 or 11, maybe?

Adyen/Utilities/Debugging/Assertion.swift
descorp: My point is - this one do what usually `assertionFailure ` does, and the other method - what usually `assert` do..

Adyen/Utilities/Debugging/Assertion.swift
descorp: oh.. it is actually "Calls assertionFailure when conditions met"

Adyen/Validators/BalanceValidator.swift
descorp: ok, just feels bit redundant to me..

AdyenCard/Components/Card/CardComponentExtensions.swift
descorp: fair enough

AdyenCard/Components/Card/CardPublicKeyFetcherProtocol.swift
descorp: 👍

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
descorp: true, my mistake..

AdyenCard/Components/GiftCardComponent/GiftCardDetails.swift
descorp: I don't see it is used...

Demo/Common/IntegrationExample.swift
descorp: 👍

Demo/Common/IntegrationExampleDropIn.swift
descorp: okey okey 😄

Adyen/Core/Payment Methods/PreselectedPaymentMethod.swift
descorp: Maybe ` A payment method that is ready for instant submission. Could be displayed as preselected payment method`

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -0,0 +1,312 @@
...
+
+    private lazy var formViewController: FormViewController = {
+        let formViewController = FormViewController(style: style)
+        formViewController.localizationParameters = localizationParameters
+        formViewController.delegate = self
+        formViewController.title = paymentMethod.name
+        formViewController.append(errorItem)
+        formViewController.append(numberItem)
+        formViewController.append(securityCodeItem)
+        formViewController.append(button.withPadding(padding: .init(top: 8, left: 0, bottom: -16, right: 0)))
```
kaphacius, line 99: i see this padding numbers quite often, maybe good to make them a constant

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
kaphacius, line 291: i must say, i'm still not on board with this. feels like result should not be switching on `self`. also it is too easy to forget one of the handlers.

AdyenDropIn/Utilities/ComponentManager.swift
```diff
@@ -17,29 +17,62 @@ import Adyen
...
     // MARK: - Internal
     
-    internal lazy var components: SectionedComponents = {
+    internal lazy var sections: [ComponentsSection] = {
         // Filter out payment methods without the Ecommerce shopper interaction.
         let storedPaymentMethods = paymentMethods.stored.filter { $0.supportedShopperInteractions.contains(.shopperPresent) }
+
+        // Paid section
+        let amountString: String = remainingAmount.flatMap {
+            AmountFormatter.formatted(amount: $0.value, currencyCode: $0.currencyCode)
```
kaphacius, line 48: can be `remainingAmount.flatMap(\.formatted)`

Demo/Common/IntegrationExamplePartialPayment.swift
```diff
@@ -0,0 +1,111 @@
...
+import AdyenActions
+import AdyenCard
+import AdyenComponents
+import AdyenDropIn
+import UIKit
+
+extension IntegrationExample: PartialPaymentDelegate {
+
+    internal enum GiftCardError: Error, LocalizedError {
+        case nonBalance
```
kaphacius, line 17: maybe `.noBalance`?

Demo/Common/IntegrationExamplePartialPayment.swift
```diff
@@ -0,0 +1,111 @@
...
+
+    private func handle(result: Result<CreateOrderResponse, Error>,
+                        completion: @escaping (Result<PartialPaymentOrder, Error>) -> Void) {
+        switch result {
+        case let .success(response):
+            self.order = response.order
+            completion(.success(response.order))
+        case let .failure(error):
+            finish(with: error)
+            completion(.failure(error))
```
kaphacius, line 83: looks like we are calling 2 completion handlers here

Demo/Common/Networking/CancelOrderRequest.swift
```diff
@@ -0,0 +1,46 @@
...
+
+    internal var queryParameters: [URLQueryItem] = []
+
+    internal func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+
+        let configurations = ConfigurationConstants.current
+
+        try container.encode(order, forKey: .order)
+        try container.encode(configurations.merchantAccount, forKey: .merchantAccount)
```
kaphacius, line 32: probably okay to call `ConfigurationConstants.current.merchantAccount`, since it is used only once

Demo/Common/Networking/PaymentsRequest.swift
```diff
@@ -26,8 +26,9 @@ internal struct PaymentsRequest: Request {
     
     internal func encode(to encoder: Encoder) throws {
         var container = encoder.container(keyedBy: CodingKeys.self)
-        
+
         let currentConfiguration = ConfigurationConstants.current
+        let amount = data.amount ?? currentConfiguration.amount
```
kaphacius, line 31: does this make sense? can there be a payment without an amount? if yes - no need have a fallback of `ConfigurationConstants.current.amount`. if no - no need for the instance variable to be optional

Tests/AdyenTests/Adyen Tests/Components/Gift Card/GiftCardComponentTests.swift
```diff
@@ -0,0 +1,414 @@
...
+        sut.payment = nil // Missing Payment object
+
+        delegateMock.onDidFail = { error, component in
+            XCTFail("delegateMock.onDidFail shouldn't be reported back to merchant.")
+        }
+
+        let onCheckBalanceExpectation = expectation(description: "Expect partialPaymentDelegate.onCheckBalance to be called.")
+        let balance = Balance(availableAmount: .init(value: 200, currencyCode: "EUR"), transactionLimit: .init(value: 1000, currencyCode: "EUR"))
+        partialPaymentDelegate.onCheckBalance = {
+            $2(.success(balance))
```
kaphacius, line 196: i think if there are three input parameters for a closure it makes sense to have names for them (at least the relevant ones). ` { _, _, completion in ... }`

AdyenCard/Components/Card/CardPublicKeyFetcherProtocol.swift
```diff
@@ -0,0 +1,44 @@
...
+    var cardPublicKeyProvider: AnyCardPublicKeyProvider { get set }
+}
+
+extension CardPublicKeyFetcherProtocol {
+    internal typealias CardKeySuccessHandler = (_ cardPublicKey: String) -> Void
+    internal typealias CardKeyFailureHandler = (_ error: Swift.Error) -> Void
+
+    internal func fetchCardPublicKey(onError: CardKeyFailureHandler? = nil, completion: @escaping CardKeySuccessHandler) {
+        do {
+            try cardPublicKeyProvider.fetch { [weak self] in
```
kaphacius: same here with 2 types of success reported, either in completion handler as a `Result`, or throwing an `Error`, which is handled the same way as `Result.Failure`

Adyen/Utilities/Debugging/Assertion.swift
descorp: This is a good point... `assert` is a verb. `assertionFailure` is more of an action name...

AdyenCard/Components/Card/CardPublicKeyFetcherProtocol.swift
descorp: I think 'Every consumer have a provider' make sense here

Adyen/Validators/BalanceValidator.swift
```diff
@@ -0,0 +1,77 @@
...
+    /// :nodoc:
+    public init() { /* Empty initializer */ }
+
+    /// Validates a Balance against a payment amount to pay.
+    ///
+    /// - Parameters:
+    ///   - balance: The balance to validate.
+    ///   - amount: The amount to pay.
+    /// :nodoc:
+    public func validate(balance: Balance, toPay amount: Payment.Amount) throws -> Bool {
```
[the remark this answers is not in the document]
mohammedDehairy: exception are out of the ordinary flow (hence its name exceptions :-D ), validations are "normal" to return false (that's the semantic definition of validation), so the happy flow is either true or false, non happy exception throw an error.

so you can think of error throwing as if its an assertion of some prerequisite of the function.

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
mohammedDehairy: its a trad of, your point is valid, but also the benefit of seeing the whole flow in one place is worth it IMHO

Demo/Common/IntegrationExamplePartialPayment.swift
mohammedDehairy: finish is not a completion handler, right? its business logic of the merchant to show an error alert or something

Demo/Common/Networking/CancelOrderRequest.swift
mohammedDehairy: too long :-)

Demo/Common/Networking/PaymentsRequest.swift
mohammedDehairy: payment is mandatory for a payment request

Adyen/Components/Stored Payment Method/StoredPaymentMethodComponent.swift
```diff
@@ -34,18 +34,20 @@ public final class StoredPaymentMethodComponent: PaymentComponent, PresentableCo
                                                                        localizationParameters, storedPaymentMethod.name),
                                                 message: displayInformation.title,
                                                 preferredStyle: .alert)
-        
-        let cancelAction = UIAlertAction(title: localizedString(.cancelButton, localizationParameters), style: .cancel) { _ in
+
+        let cancelAction = UIAlertAction(title: localizedString(.cancelButton, localizationParameters), style: .cancel) { [weak self] _ in
+            guard let self = self else { return }
```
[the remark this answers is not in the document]
mohammedDehairy: its needed because it needs self as non-optional in the line after.

Adyen/Components/Stored Payment Method/StoredPaymentMethodComponent.swift
```diff
@@ -34,18 +34,20 @@ public final class StoredPaymentMethodComponent: PaymentComponent, PresentableCo
...
             self.delegate?.didFail(with: ComponentError.cancelled, from: self)
         }
         alertController.addAction(cancelAction)
         
         let submitActionTitle = localizedSubmitButtonTitle(with: payment?.amount,
                                                            style: .immediate,
                                                            localizationParameters)
-        let submitAction = UIAlertAction(title: submitActionTitle, style: .default) { _ in
+        let submitAction = UIAlertAction(title: submitActionTitle, style: .default) { [weak self] _ in
+            guard let self = self else { return }
```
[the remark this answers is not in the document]
mohammedDehairy: its needed because it needs self as non-optional in the line after.

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -287,8 +286,12 @@ public final class GiftCardComponent: PartialPaymentComponent,
 
 extension Result {
 
+    func handle(_ success: (Success) -> Void, _ failure: (Failure) -> Void) {
```
kaphacius, line 289: maybe better to have parameters of this function visible, because this is a bit confusing
```
orderResult.handle(submit(order:)) { delegate?.didFail(with: $0, from: self) }
```

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -266,7 +265,7 @@ public final class GiftCardComponent: PartialPaymentComponent,
     }
 
     private func handle(orderResult: Result<PartialPaymentOrder, Swift.Error>) {
-        orderResult.handleError { delegate?.didFail(with: $0, from: self) }.handleSuccess(submit(order:))
+        orderResult.handle(submit(order:)) { delegate?.didFail(with: $0, from: self) }
```
kaphacius, line 268: if there are 2 closures as a parameter, it might be better to not use trailing closure synthax, since we can't tel what the other closure is doing

AdyenCard/Utilities/CardType/CardBrandProvider.swift
```diff
@@ -105,22 +105,18 @@ internal final class CardBrandProvider: AnyCardBrandProvider {
...
+            guard let self = self else { return }
+
+            switch result {
+            case let .success(publicKey):
+                let binLookupService = BinLookupService(publicKey: publicKey,
+                                                        apiClient: localApiClient)
+                self.privateBinLookupService = binLookupService
+                success(binLookupService)
+            case let .failure(error):
+                failure?(error)
```
kaphacius, line 118: maybe failure can be a non-optional parameter. or, even better, `fetchBinLookupService` can accept only one closure, of type `Result<BinLookupService, Swift.Error>`

AdyenCard/Components/Card/CardPublicKeyConsumer.swift
```diff
@@ -13,32 +13,19 @@ internal protocol CardPublicKeyConsumer: PaymentComponent {
...
         switch result {
         case let .success(key):
             completion(key)
         case let .failure(error):
-            if let onError = onError {
-                onError(error)
-            } else {
-                delegate?.didFail(with: error, from: self)
-            }
+            delegate?.didFail(with: error, from: self)
```
kaphacius: maybe as a parameter to `fetchCardPublicKey` we could pass a handler of type `Result<String, Swift.Error>` and use it for both success and fail scenarios, instead of calling `delegate?.didFail(with: error, from: self)`. this way we can remove inheritance of `PaymentComponent` from `protocol CardPublicKeyConsumer`. i think it is a bit better for interface segregation. also, when `submitEncryptedCardData` is called, both success and fail handlers would be in the same place.
```swift
internal protocol CardPublicKeyConsumer: AnyObject {
    var cardPublicKeyProvider: AnyCardPublicKeyProvider { get set }
}

extension CardPublicKeyConsumer {
    internal typealias CardKeyFetchResultHandler = (_ result: Result<String, Swift.Error>) -> Void

    internal func fetchCardPublicKey(completion: @escaping CardKeyFetchResultHandler) {
        cardPublicKeyProvider.fetch(completion: completion)
    }
}

extension CardComponent {
    
    internal func didSelectSubmitButton() {
        guard cardViewController.validate() else {
            return
        }
        
        cardViewController.startLoading()

        fetchCardPublicKey { [weak self] result in
            guard let weakSelf = self else { return }
            
            switch result {
            case .success(let key):
                weakSelf.submitEncryptedCardData(cardPublicKey: key)
            case .failure(let error):
                weakSelf.delegate?.didFail(with: error, from: weakSelf)
            }
        }
    }
```

mohammedDehairy: what's the point then of `CardPublicKeyConsumer`? :-D , this is a convenient way instead of having the switch statement in every component that conforms to `CardPublicKeyConsumer`, right?

kaphacius: that's kind of my point. `fetchCardPublicKey` is supposed to only fetch the key, and report the result. right now it also does error reporting via `delegate`, which it should not even know about. if you look at apple APIs, completion handler usually is called both in case of success and failure. `CardPublicKeyConsumer` does not need to inherit from `PaymentComponent`, because when inheriting it gets access to a lot of things that it shouldn't (`delegate`, `paymentMethod`, `payment` etc. those are not related to fetching card keys).  I honestly don't think we need this extension, because it consists of one method with almost no logic, and is not reused anywhere. Just calling it directly from the `CardComponent` and handling the result in a switch would be better imho.

Adyen/UI/List/ListItemView.swift
mohammedDehairy: works fine on iOS 11, 12, 13 and 14

Adyen/Assets/Generated/LocalizationKey.swift
```diff
@@ -2,7 +2,7 @@
 // Copyright (c) 2021 Adyen N.V.
 //
 // This file is open source and available under the MIT license. See the LICENSE file for more info.
-// This file is autogenerated. Please do not modify it.
```
descorp, line 5: It is still auto-generated, right? Why remove header?

Adyen/Assets/ro-RO.lproj/Localizable.strings
```diff
@@ -65,6 +65,9 @@
 "adyen.voucher.readInstructions" = "Citiți instrucțiunile";
 "adyen.voucher.saveImage" = "Save as image";
 "adyen.voucher.finish" = "Finish";
+"adyen.pix.instructions" = "";
```
descorp, line 68: I am wandering - if you do not provide this empty keys it will fallback to english automatically, right?

Adyen/Assets/Generated/LocalizationKey.swift
mohammedDehairy: I haven't touched this file, I just updated translations and build with Xcode.

Adyen/Assets/ro-RO.lproj/Localizable.strings
mohammedDehairy: removed those empty strings

Adyen/Assets/Generated/LocalizationKey.swift
mohammedDehairy: I think the swift format and the translations script have conflict, lets fix this in a another PR

Adyen/Core/APIClient/Requests/OrderStatusRequest.swift
```diff
@@ -0,0 +1,45 @@
+//
+//  OrderStatusRequest.swift
+//  Adyen
+//
+//  Created by Mohamed Eldoheiri on 5/6/21.
+//  Copyright © 2021 Adyen. All rights reserved.
```
descorp, line 6: Hm.. the header is incorrect 🤔

Adyen/Core/AlreadyPaidPaymentComponent.swift
```diff
@@ -0,0 +1,23 @@
...
+// Copyright (c) 2021 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+/// Provides a placeholder for payment methods that are already paid, just for display.
+/// :nodoc:
+public final class AlreadyPaidPaymentComponent: PaymentComponent {
```
descorp, line 11: Is this used? I only see it checked on DropIn, but never created...

Adyen/Core/InstantPaymentComponent.swift
```diff
@@ -0,0 +1,34 @@
...
+    public let paymentData: PaymentComponentData?
+
+    /// :nodoc:
+    public let paymentMethod: PaymentMethod
+
+    /// The delegate of the component.
+    public weak var delegate: PaymentComponentDelegate?
+
+    /// :nodoc:
+    public init(paymentMethod: PaymentMethod, paymentData: PaymentComponentData?) {
```
descorp, line 23: Could be `paymentData: PaymentComponentData? = nil`

Adyen/Core/AlreadyPaidPaymentComponent.swift
mohammedDehairy: its used in `OrderPaymentMethod`

Adyen/Core/InstantPaymentComponent.swift
mohammedDehairy: no I don't like default values, its error prone

Adyen/UI/Form/Items/Button/FormButtonItem.swift
```diff
@@ -44,3 +44,13 @@ public final class FormButtonItem: FormItem {
     }
     
 }
+
+/// :nodoc:
+extension FormButtonItem {
+
+    /// :nodoc:
+    public var submitButton: FormContainerItem {
```
descorp, line 52: Maybe better call it `submitButtonPadding` and do 

```swift
myButtonItem.withPadding(padding: submitButtonPadding)
```

Adyen/Core/APIClient/Requests/OrderStatusRequest.swift
mohammedDehairy: fixed

Adyen/Data/PartialPaymentOrder.swift
```diff
@@ -0,0 +1,88 @@
...
+
+    private enum CodingKeys: String, CodingKey {
+        case pspReference
+        case orderData
+        case reference
+        case remainingAmount
+        case expiresAt
+    }
+
+    public static func == (lhs: PartialPaymentOrder, rhs: PartialPaymentOrder) -> Bool {
```
kaphacius, line 81: can this be auto synthesized? seems like fields of `PartialPaymentOrder` should be equatable

Adyen/Components/Base/PartialPaymentComponent.swift
```diff
@@ -0,0 +1,32 @@
...
+    public var errorDescription: String? {
+        switch self {
+        case .zeroRemainingAmount:
+            return "There is no remaining amount to be paid."
+        }
+    }
+}
+
+/// Any component that provides partial payments.
+public protocol PartialPaymentComponent: AnyObject {
```
kaphacius, line 25: Should this inherit from `PaymentComponent`?  even the names are similar :)

Adyen/Components/Base/PaymentComponent.swift
```diff
@@ -60,6 +60,9 @@ public protocol PaymentAwareComponent: Component {
 
     /// The payment information.
     var payment: Payment? { get set }
+
+    /// The partial payment order if any.
+    var order: PartialPaymentOrder? { get set }
```
kaphacius, line 65: should this be in `PartialPaymentComponent`?

Adyen/Core/InstantPaymentComponent.swift
kaphacius, line 23: do we need to pass `paymentData` to the initializer? Looks like it can be constructed (line 31) from already existing fields. can we remove the field altogether?

Adyen/Data/PartialPaymentOrder.swift
```diff
@@ -0,0 +1,88 @@
...
+                orderData: String,
+                reference: String? = nil,
+                remainingAmount: Payment.Amount? = nil,
+                expiresAt: Date? = nil) {
+        self.pspReference = pspReference
+        self.orderData = orderData
+        self.reference = reference
+        self.remainingAmount = remainingAmount
+        self.expiresAt = expiresAt
+        self.compactOrder = CompactOrder(pspReference: pspReference,
```
kaphacius, line 57: i like that all fields are `let`, so this can be constructed without worries :)

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -0,0 +1,304 @@
...
+            case .missingPartialPaymentDelegate:
+                return "Please provide a `PartialPaymentDelegate` object"
+            case let .otherError(error):
+                return error.localizedDescription
+            }
+        }
+    }
+
+    /// :nodoc:
+    private var giftCardPaymentMethod: GiftCardPaymentMethod
```
kaphacius, line 49: `let` maybe?

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -0,0 +1,304 @@
...
+
+    // MARK: - Presentable Component Protocol
+
+    /// :nodoc:
+    public var viewController: UIViewController { securedViewController }
+
+    /// :nodoc:
+    public var requiresModalPresentation: Bool { true }
+
+    private lazy var securedViewController = SecuredViewController(child: formViewController, style: style)
```
kaphacius, line 89: can we make `viewController` a lazy var and initialize it with `SecuredViewController(child: formViewController, style: style)`?

AdyenDropIn/Components/PreselectedPaymentMethodComponent.swift
```diff
@@ -43,7 +43,12 @@ internal final class PreselectedPaymentMethodComponent: ComponentLoader,
     ///
     /// - Parameter components: The components to display in the list.
     /// - Parameter style: The component's UI style.
-    internal init(component: PaymentComponent, title: String, style: FormComponentStyle, listItemStyle: ListItemStyle) {
+    /// - Parameter component: The pre-selected component.
+    /// - Parameter title: The title.
+    /// - Parameter listItemStyle: The list item UI style.
+    internal init(component: PaymentComponent,
+                  title: String,
+                  style: FormComponentStyle, listItemStyle: ListItemStyle) {
```
kaphacius, line 51: `listItemStyle` should be on the next line

descorp (approved):

PR #509, opened 2021-05-17
Add callback on app-to-app redirect

descorp:
### Changes

* call `didOpenExternalApplication` on redirecting app-to-app

### DIFF
--- a/AdyenActions/Components/Redirect/BrowserComponent.swift
+++ b/AdyenActions/Components/Redirect/BrowserComponent.swift
@@ -10,6 +10,7 @@ import UIKit
 
 internal protocol BrowserComponentDelegate: AnyObject {
     func didCancel()
+    func didOpenExternalApplication()
 }
 
 /// A component that opens a URL in web browsed and presents it.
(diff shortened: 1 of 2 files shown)


AdyenActions/Components/Redirect/BrowserComponent.swift
```diff
@@ -72,6 +73,8 @@ extension BrowserComponent: UIAdaptivePresentationControllerDelegate {
         DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
             if UIApplication.shared.applicationState == .active {
                 self.delegate?.didCancel()
+            } else {
+                self.delegate?.didOpenExternalApplication()
```
mohammedDehairy, line 77: that's a bit of an assumption, right?

descorp: In a way. If user click "Cancel" and dismissed the app in less than 0.5s - same behaviour will be triggered..
Not sure if similar behaviour could be triggered by Push notification or any other system event 🤔

In general - between this and RedirectComponent that always false-fire "didFailed" on Vipps - It looks like a compromise.

mohammedDehairy (approved):

PR #512, opened 2021-05-21
feature: expose CardComponentDelegate to dropin users

mohammedDehairy:
feature: expose CardComponentDelegate to dropin users

### DIFF
--- a/AdyenDropIn/DropInComponent.swift
+++ b/AdyenDropIn/DropInComponent.swift
@@ -28,12 +28,6 @@ public final class DropInComponent: NSObject, PresentableComponent {
     /// The payment methods to display.
     public private(set) var paymentMethods: PaymentMethods
     
-    /// The delegate of the drop in component.
-    public weak var delegate: DropInComponentDelegate?
-
-    /// The partial payment flow delegate.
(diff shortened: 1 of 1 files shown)


AdyenDropIn/DropInComponent.swift
```diff
@@ -127,7 +127,7 @@ public final class DropInComponent: NSObject, PresentableComponent {
     
     // MARK: - Private
 
-    private lazy var componentManager = createComponentManager(order, nil)
+    private lazy var componentManager = createComponentManager(nil, nil)
```
descorp, line 130: I was looking at this line as well :D 
we can make it `private var componentManager: ComponentManager` and move initiation to init

AdyenDropIn/DropInComponent.swift
```diff
@@ -353,25 +354,6 @@ extension DropInComponent: ReadyToSubmitPaymentComponentDelegate {
     }
 }
 
-extension DropInComponent: PartialPaymentDelegate {
```
descorp, line 356: This is nice :)

kaphacius (approved):

descorp (approved):

PR #513, opened 2021-05-21
Massive refactoring

descorp:
### Changes

* rename FullFormAddressItem -> FormAddressItem
* rename AddressInfo -> PostalAddress
* rename Switch -> Toggle
* make `Amount` a standalone object
* make countryCode and currencyCode non-optional and assert validation on init
* new file for ShopperInformation
* new file for DisplayInformation

### Folder structure

Screenshot 2021-05-21 at 18 44 23

PR #515, opened 2021-05-26
fixes for Gift card flow

mohammedDehairy:
fixes for Gift card flow


Adyen/Validators/BalanceChecker.swift
```diff
@@ -37,8 +37,13 @@ public struct BalanceChecker {
         public let isBalanceEnough: Bool
 
         /// :nodoc:
-        /// The remaining amount in the balance after payment. it is negative in case the `isBalanceEnough` is false.
-        public let remainingAmount: Payment.Amount
+        /// The remaining amount in the balance after payment.
+        /// it is at minimum zero when the available amount is less than the amount to pay.
```
descorp, line 41: `it is at minimum zero when the available amount is less than the amount to pay.` -> `Zero, if available amount is less than the amount to pay` ?

AdyenCard/Components/GiftCardComponent/GiftCardDetails.swift
```diff
@@ -14,32 +14,33 @@ public struct GiftCardDetails: PaymentMethodDetails {
     public let type: String
 
     /// The encrypted card number.
-    public let cardNumber: String
+    public let encryptedCardNumber: String
```
descorp, line 17: 👍

AdyenDropIn/Models/PaymentMethodsConfiguration.swift
```diff
@@ -36,8 +47,10 @@ public extension DropInComponent {
         public var payment: Adyen.Payment?
         
         /// Initializes the drop in configuration.
-        public init(clientKey: String) {
+        public init(clientKey: String, environment: Environment) {
```
descorp, line 50: maybe also add a parameter comments?

AdyenDropIn/Models/PaymentMethodsConfiguration.swift
```diff
@@ -27,7 +27,18 @@ public extension DropInComponent {
...
+        public var clientKey: String {
+            didSet {
+                environment.clientKey = clientKey
+            }
+        }
+
+        /// The API environment to use.
+        public var environment: Environment {
+            didSet {
+                environment.clientKey = clientKey
```
descorp, line 39: this could be confusing if you generated a new instance of Environment with `clientKey` and assigned it to `PaymentMethodsConfiguration`.

I would prefer to make it required and constant:

```swift 
  public let clientKey: String
  public let environment: Environment
  
  public init(clientKey: String, environment: Environment) {
      self.clientKey = clientKey            
      self.environment = environment            
      self.environment.clientKey = clientKey
  }
```

AdyenDropIn/DropInComponent.swift
```diff
@@ -33,6 +33,29 @@ public final class DropInComponent: NSObject, PresentableComponent {
...
+            configuration.environment
+        }
+
+        set {
+            configuration.environment = newValue
+        }
+    }
+
+    /// :nodoc:
+    public var clientKey: String? {
```
descorp, line 49: can we make it read only ? 🤔

AdyenDropIn/Models/PaymentMethodsConfiguration.swift
mohammedDehairy: Merchants are not supposed to create environment with a `clientKey`, the fact that environment has the client key is sort of hidden, he client key set in the `clientKey` property is the one that should be used.

Adyen/Validators/BalanceChecker.swift
mohammedDehairy: its important to say that its minimum is zero, your version doesn't say that explicitly

AdyenDropIn/DropInComponent.swift
mohammedDehairy: 👍

AdyenDropIn/Models/PaymentMethodsConfiguration.swift
descorp: All I am saying is that it is not obvious to the merchant that `Environment` have `clientKey` in it..
I would rather treat those two entities as unrelated constants

Adyen/Validators/BalanceChecker.swift
descorp: I doesn't have to be one sentence :D

AdyenDropIn/Models/PaymentMethodsConfiguration.swift
mohammedDehairy: 👍

Adyen/Validators/BalanceChecker.swift
kaphacius: I think balance (in case of money) already implies amount. Maybe `remainingBalance` is a better name. 
```
/// The remaining balance after the payment. <- first sentence
/// It is either a zero, if the available balance is not enough to pay the amount
/// or a positive number if the available balance is enough to pay the amount

File: Adyen/Validators/BalanceChecker.swift [hunk 6]
```diff
@@ -94,6 +102,8 @@ public struct BalanceChecker {
 }
 
 /// :nodoc:
-private func - (lhs: Payment.Amount, rhs: Payment.Amount) -> Payment.Amount {
-    Payment.Amount(value: lhs.value - rhs.value, currencyCode: lhs.currencyCode)
+internal func - (lhs: Payment.Amount?, rhs: Payment.Amount?) -> Payment.Amount {
```
kaphacius, line 105: looks like we don't need a version with optional parameters, looks like it complies with the `lhs: Payment.Amount, rhs: Payment.Amount` signature.

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -183,24 +205,36 @@ public final class GiftCardComponent: PartialPaymentComponent,
 
         startLoading()
 
-        let data = createPaymentData()
+        fetchCardPublicKey(discardError: false) { [weak self] cardPublicKey in
+            guard let self = self else { return }
+            let paymentDataResult = self.createPaymentData(order: self.order, cardPublicKey: cardPublicKey)
```
kaphacius, line 210: maybe this line is not needed, can do `self.createPaymentData(order: self.order, cardPublicKey: cardPublicKey).mapError...`

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -183,24 +205,36 @@ public final class GiftCardComponent: PartialPaymentComponent,
...
                 .mapError(Error.otherError)
                 .flatMap(self.check(balance:))
                 .flatMap { balanceCheckResult in
                     if balanceCheckResult.isBalanceEnough {
-                        return self.onReadyToPayFullAmount(remainingAmount: balanceCheckResult.remainingAmount)
+                        return self.onReadyToPayFullAmount(remainingAmount: balanceCheckResult.remainingBalanceAmount,
+                                                           paymentData: paymentData)
                     } else {
-                        return self.requestOrder()
+                        let newPaymentData = paymentData.replacingAmount(with: balanceCheckResult.amountToPay)
```
kaphacius, line 228: same here, maybe no extra line

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -263,38 +302,59 @@ public final class GiftCardComponent: PartialPaymentComponent,
...
     }
 
-    private func handle(orderResult: Result<PartialPaymentOrder, Swift.Error>) {
-        orderResult.handle(success: submit(order:), failure: { delegate?.didFail(with: $0, from: self) })
+    private func handle(orderResult: Result<PartialPaymentOrder, Swift.Error>,
+                        paymentData: PaymentComponentData) {
+        orderResult.handle(success: {
+            submit(order: $0, paymentData: paymentData)
+        }, failure: {
+            delegate?.didFail(with: $0, from: self)
```
kaphacius, line 327: can we call `handle(error: $0)` here? so that all errors are handled in one place.

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -232,22 +268,25 @@ public final class GiftCardComponent: PartialPaymentComponent,
...
                               condition: _isDropIn && readyToSubmitComponentDelegate == nil)
         stopLoading()
         if let readyToSubmitComponentDelegate = readyToSubmitComponentDelegate {
-            showConfirmation(delegate: readyToSubmitComponentDelegate, remainingAmount: remainingAmount)
+            showConfirmation(delegate: readyToSubmitComponentDelegate,
+                             remainingAmount: remainingAmount,
+                             paymentData: paymentData)
         } else {
-            let paymentData = createPaymentData()
             delegate?.didSubmit(paymentData, from: self)
```
kaphacius, line 282: can we call `self.submit` here?

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -263,38 +302,59 @@ public final class GiftCardComponent: PartialPaymentComponent,
...
-    private func createPaymentData(order: PartialPaymentOrder? = nil) -> PaymentComponentData {
-        let details = GiftCardDetails(paymentMethod: giftCardPaymentMethod,
-                                      cardNumber: numberItem.value,
-                                      securityCode: securityCodeItem.value)
+    // MARK: - PaymentData creation
+
+    private func createPaymentData(order: PartialPaymentOrder?, cardPublicKey: String) -> Result<PaymentComponentData, Swift.Error> {
+        do {
+            let card = Card(number: numberItem.value, securityCode: securityCodeItem.value)
+            let encryptedCard = try CardEncryptor.encrypt(card: card, with: cardPublicKey)
```
kaphacius, line 342: this is probably out of scope of this PR, but the interface of `CardEncryptor.encrypt` is rather unusual. it does throw, but then can return possibly invalid data (optional `number` and `securityCode`), that has to be checked and throws an error in case they are `nil`. shouldn't this be done inside the `CardEncryptor.encrypt`?

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -312,4 +372,10 @@ public extension Result {
 }
 
 /// :nodoc:
-extension GiftCardComponent: TrackableComponent {}
+extension GiftCardComponent: TrackableComponent {
+    /// :nodoc:
+    public func viewDidLoad(viewController: UIViewController) {
+        Analytics.sendEvent(component: paymentMethod.type, flavor: _isDropIn ? .dropin : .components, environment: environment)
```
kaphacius, line 378: you could actually call the default implementation of `TrackableComponent.viewDidLoad` here (kind of like a `super` method), since the signature of `viewDidLoad` is not in the protocol definition. but in #513 it is added there, so this won't work anymore :)

AdyenDropIn/Models/PaymentMethodsConfiguration.swift
kaphacius: i'm also not very on board about this. why is it not a constant? since we are injecting it at construction time, and it is not supposed to change

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
descorp: In #513 is was actually wrong...

kaphacius: yes, i see that you have now removed that :)

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
mohammedDehairy: no I think the line will be too long

Adyen/Validators/BalanceChecker.swift
mohammedDehairy: its needed for tests, I can make the optional version only in test target, but I don't see the benefit tbh

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
mohammedDehairy: 👍

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
mohammedDehairy: 👍 good one

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
mohammedDehairy: 👍

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
mohammedDehairy: yes, it is out of scope :-D

Adyen/Validators/BalanceChecker.swift
kaphacius: i don't think tests should affect the code they are testing. this version of the function has unusual logic i think. it shouldn't be up to subtraction algorithm to unwrap optionals and provide default values.

AdyenDropIn/Models/PaymentMethodsConfiguration.swift
mohammedDehairy: made them immutable

Adyen/Validators/BalanceChecker.swift
descorp: 👍

kaphacius (approved):

kaphacius (approved):

Demo/Common/Networking/CancelOrderRequest.swift
```diff
@@ -40,7 +40,11 @@ internal struct CancelOrderRequest: Request {
 
 internal struct CancelOrderResponse: Response {
 
-    internal let resultCode: PaymentsResponse.ResultCode
+    internal enum ResultCode: String, Decodable {
+        case received = "Received"
```
descorp, line 44: There is only one result code?

Tests/AdyenTests/Adyen Tests/Utilities/BalanceCheckerTests.swift
```diff
@@ -51,28 +58,60 @@ class BalanceValidatorTests: XCTestCase {
...
+        // return false
+        let balance7 = Balance(availableAmount: .init(value: 1000,
+                                                     currencyCode: "EUR"),
+                              transactionLimit: .init(value: 100,
+                                                      currencyCode: "EUR"))
+        let amount7 = Payment.Amount(value: 1200, currencyCode: "EUR")
+        let result7 = try! sut.check(balance: balance7, isEnoughToPay: amount7)
+        XCTAssertEqual(result7.amountToPay, balance7.transactionLimit)
+        XCTAssertFalse(result7.isBalanceEnough)
+        XCTAssertEqual(result7.remainingBalanceAmount, balance7.availableAmount - balance7.transactionLimit)
```
descorp, line 102: this sort of assumption that is not nice for the tests..

I would rather do:
```swift
XCTAssertEqual(result7.remainingBalanceAmount.value, 900)
XCTAssertEqual(result7.remainingBalanceAmount. currencyCode, "EUR")
```

**+** `fileprivate func - (lhs: Payment.Amount?, rhs: Payment.Amount?) -> Payment.Amount` is not needed

AdyenDropIn/DropInComponent.swift
```diff
@@ -33,6 +33,12 @@ public final class DropInComponent: NSObject, PresentableComponent {
     
     /// The title text on the first page of drop in component.
     public let title: String
+
+    /// :nodoc:
+    public let environment: Environment
```
descorp, line 38: Those two comes from `PaymentMethodsConfiguration`, right?

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -62,6 +39,20 @@ public final class GiftCardComponent: PartialPaymentComponent,
     /// The delegate that handles shopper confirmation UI when the balance of the gift card is sufficient to pay.
     public weak var readyToSubmitComponentDelegate: ReadyToSubmitPaymentComponentDelegate?
 
+    /// :nodoc:
+    public var clientKey: String? {
```
descorp, line 43: GiftCardComponent doesn't need `clientKey` on it own - it pass it to `cardPublicKeyProvider`; and it is set on `init`.

We can simply do 
```swift
 public var clientKey: String? { cardPublicKeyProvider.clientKey }
```

---

Maybe it is easier to just set both `clientKey` and `environment` in `init` as we did with DropInComponent?

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -124,7 +125,7 @@ public final class GiftCardComponent: PartialPaymentComponent,
     internal lazy var securityCodeItem: FormTextInputItem = {
         let item = FormTextInputItem(style: style.textField)
         item.title = localizedString(.cardPinTitle, localizationParameters)
-        item.validator = NumericStringValidator(minimumLength: 3)
+        item.validator = NumericStringValidator(minimumLength: 3, maximumLength: 10)
```
descorp, line 128: Just a curiosity: Security Code is always numerical?

Demo/Common/Networking/CancelOrderRequest.swift
mohammedDehairy: they are not related.

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
mohammedDehairy: this is out of scope, will change this later

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
mohammedDehairy: it is numerical.

Issue #77, opened 2019-06-19
Adyen 2.8.3 SWift <-> ObjC


mkharibalaji:
Hi Team,
          I basically have a react native app which is using ObjC.So when i tried to implement applicationDidOpenURL i am not able get Adyen Instance and getting this error "Use of undeclared identifier 'Adyen'"

Any help would be great,

......
......
#import <Adyen/Adyen-Swift.h>

- (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDictionary<NSString *,id> *)options {

  BOOL handledAdyen =[Adyen application:app openURL:url options:options];// Use of undeclared identifier 'Adyen'
  
  return handledAdyen;
}

Thanks & regards,
MK HB

joostvandijk:
Hi @mkharibalaji,

We recommend adding a new Swift class that bridges this call to the Objective-C environment. Something like this should work:
```swift
@objc class AdyenObjectiveCBridge: NSObject {
    
    @objc(applicationDidOpenURL:)
    static func applicationDidOpen(_ url: URL) -> Bool {
        Adyen.applicationDidOpen(url)
    }
    
}
```

You can then call it from Objective-C:
```objc
[AdyenObjectiveCBridge applicationDidOpenURL:...];
```

Best,

Joost

mkharibalaji:
Thanks a lot @joostvandijk  :)

It Worked !!!

PR #520, opened 2021-06-01
chore: Refactor environment and clientKey to be injected into components

kaphacius:
This refactoring puts `Environment` and `clientKey` into a new `APIContext` struct, which is injected via constructor when needed.


Adyen/Core/APIClient/AnyAPIContext.swift
```diff
@@ -0,0 +1,68 @@
...
+    /// The HTTP headers.
+    var headers: [String: String] { get }
+    
+    /// :nodoc:
+    /// The query parameters.
+    var queryParameters: [URLQueryItem] { get }
+    
+    /// :nodoc:
+    /// The client key.
+    var clientKey: String { get }
```
descorp, line 37: I believe `clientKey ` supposed to be part of `APIContext`, not `AnyAPIContext` ?

mohammedDehairy: I believe so as well, clientKey is an implementation details of the specific `APIContext`

Adyen/Core/APIClient/AnyAPIContext.swift
```diff
@@ -0,0 +1,68 @@
...
+    /// The client key that corresponds to the webservice user you will use for initiating the payment.
+    /// See https://docs.adyen.com/user-management/client-side-authentication for more information.
+    public let clientKey: String
+    
+    /// Initializes the APIContext
+    /// - Parameters:
+    ///   - environment: The environment to retrieve internal resources from.
+    ///   - clientKey: The client key that corresponds to the webservice user you will use for initiating the payment.
+    public init(environment: AnyAPIEnvironment, clientKey: String) {
+        self.environment = environment
```
descorp, line 64: We can assert `clientKey` with `ClientKeyValidator` here

AdyenActions/Components/Base/ActionComponent.swift
```diff
@@ -19,7 +19,7 @@ public protocol ActionComponent: Component {
 /// :nodoc:
 public protocol ParameterlessInitializable {
```
mohammedDehairy, line 20: its not `Parameterless` any more :-D, right?

mohammedDehairy:
also I would modify the MIGRATION.md to include this

Adyen/Core/APIClient/AnyAPIContext.swift
mohammedDehairy: I believe so as well

AdyenActions/Components/Base/ActionComponent.swift
kaphacius: how about `APIContextInitializable`?

mohammedDehairy: yeah, sounds good to me.

AdyenCard/Utilities/CardPublicKeyProvider/CardPublicKeyProvider.swift
```diff
@@ -8,7 +8,7 @@ import Adyen
 import Foundation
 
 /// :nodoc:
-internal protocol AnyCardPublicKeyProvider: Component {
+internal protocol AnyCardPublicKeyProvider: APIContextAware {
```
descorp, line 11: Nice :)

AdyenDropIn/Models/PaymentMethodsConfiguration.swift
```diff
@@ -14,7 +14,7 @@ import PassKit
 public extension DropInComponent {
     
     /// Contains the configuration for the drop in component and the embedded payment method components.
-    final class PaymentMethodsConfiguration {
+    final class PaymentMethodsConfiguration: APIContextAware {
```
descorp, line 17: DropInComponent is a component and already have `apiContext` "by definition"
How about move `apiContext` to DropIn's init, and not make  `PaymentMethodsConfiguration`  a  APIContextAware?

Adyen/Core/APIClient/APIContext.swift
```diff
@@ -0,0 +1,50 @@
...
+public protocol APIContextAware: AnyObject {
+    
+    /// :nodoc:
+    /// The API context
+    var apiContext: APIContext { get }
+    
+}
+
+/// Struct that defines API context for retrieving internal resources.
+public struct APIContext {
```
descorp, line 20: without protocol this struct looks "too concrete".. 😅
But it is not like we expecting another remote service in foreseeable future.

Adyen/Utilities/Debugging/FatalErrorUtil.swift
```diff
@@ -0,0 +1,34 @@
...
+// Copyright (c) 2021 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+/// :nodoc:
+/// Used for testing code that causes fatalError
+public func fatalError(
```
descorp, line 11: Will this override standard Swift's `fatalError` for everyone who included `AdyenCore` module?

Adyen/Core/APIClient/Environment.swift
```diff
@@ -7,51 +7,50 @@
 import Foundation
 
 /// :nodoc:
-/// Struct that defines the environment to retrieve resources from.
-public struct Environment: APIEnvironment {
+public protocol AnyAPIEnvironment {
```
descorp, line 10: What is the point of this `protocol`? 🤔

AdyenDropIn/Models/PaymentMethodsConfiguration.swift
kaphacius: that's a good point. but I think it's better to leave all configuration in for `DropInComponent` in `PaymentMethodsConfiguration`.
maybe we can do `public var apiContext: APIContext { configuration.apiContext }` in `DropInComponent` to show that it uses the configuration.

Adyen/Core/APIClient/Environment.swift
kaphacius: Yes, I was thinking of simplifying it to just be a `URL`. That would make things a bit easier. 
The protocol is also used in `internal enum DemoServerEnvironment: String, AnyAPIEnvironment`.
But `static let demoServerEnvironment = DemoServerEnvironment.test` is not used anywhere. i'm not sure what is going on.

Adyen/Core/Components/Base/Component.swift
```diff
@@ -7,16 +7,7 @@
...
-    
-    /// Defines the environment used to make networking requests.
-    var environment: Environment { get set }
-    
-    /// The client key that corresponds to the webservice user you will use for initiating the payment.
-    /// See https://docs.adyen.com/user-management/client-side-authentication for more information.
-    var clientKey: String? { get set }
-    
-}
+public protocol Component: APIContextAware { }
```
descorp, line 10: We no longer pass-through `environment` and `clientKey` to every generic component, right?

For example before it was:
```swift
(item as? Component).environment = theEnvironment 
```

Now *environment* is read-only `{ get }` and we setting it via `init`

---

How about we only apply `APIContextAware` to specific component's that actually needs it?

Adyen/Utilities/Analytics.swift
```diff
@@ -25,10 +25,10 @@ public class Analytics {
         public var flavor: Flavor
 
         /// :nodoc:
-        public var environment: Environment
+        public var environment: AnyAPIEnvironment
```
descorp, line 28: We can use `ApiContext` here or at least we can add overload for 'init' with `ApiContext`
It is more common pattern.

Adyen/Utilities/Debugging/FatalErrorUtil.swift
kaphacius: `Assert` won't work in this case. If we want to stop execution in an `init` - it has to be a `fatalError`.  
Also good catch, I will make this `internal`, so it does not override `fatalError` for users of the SDK.

Adyen/Core/Components/Base/Component.swift
kaphacius: I think we talked about this previously. We want to make this future-proof in case we need to change implementation of a component to use any internal resources. We won't have to change the interface in that case.

Adyen/Utilities/Analytics.swift
kaphacius: Good point. I think `Analytics` can be refactored a little bit as well.

AdyenActions/Components/3DS2/Fingerprint Submitter/ThreeDS2FingerprintSubmitter.swift
```diff
@@ -16,28 +16,26 @@ internal protocol AnyThreeDS2FingerprintSubmitter {
...
     /// :nodoc:
-    private lazy var apiClient: APIClientProtocol = APIClient(environment: environment)
+    private let apiContext: APIContext
 
     /// :nodoc:
-    internal init(apiClient: APIClientProtocol? = nil) {
-        if let apiClient = apiClient {
-            self.apiClient = apiClient
-        }
+    internal init(apiContext: APIContext, apiClient: APIClientProtocol? = nil) {
```
descorp, line 28: Nice :)

AdyenActions/Components/Await/PollingHandlerProvider.swift
```diff
@@ -31,14 +31,14 @@ internal protocol AnyPollingHandlerProvider {
...
-    private let environment: APIEnvironment
+    private let apiContext: APIContext
 
     /// :nodoc:
     private let apiClient: AnyRetryAPIClient?
 
     /// :nodoc:
-    internal init(environment: APIEnvironment, apiClient: AnyRetryAPIClient?) {
-        self.environment = environment
+    internal init(apiContext: APIContext, apiClient: AnyRetryAPIClient?) {
```
descorp, line 40: What is the purpose of this struct? It does only one thing - always creates `PollingComponent` 😄

AdyenActions/Components/QRCode/QRCodeComponent.swift
```diff
@@ -51,9 +55,11 @@ public final class QRCodeComponent: ActionComponent, Localizable, Cancellable {
     /// - Parameter style: The component UI style.
     /// - Parameter pollingComponentBuilder: The payment method specific await action handler provider.
     /// - Parameter timeoutInterval: QR Code expiration timeout
-    internal init(style: QRCodeComponentStyle = QRCodeComponentStyle(),
+    internal init(apiContext: APIContext,
+                  style: QRCodeComponentStyle = QRCodeComponentStyle(),
                   pollingComponentBuilder: AnyPollingHandlerProvider?,
```
descorp, line 60: This class can probably build it's own `PollingComponent`

AdyenCard/Components/Stored Card/StoredCardComponent.swift
```diff
@@ -38,17 +43,15 @@ internal final class StoredCardComponent: PaymentComponent, PresentableComponent
     
     /// :nodoc:
     internal lazy var storedCardAlertManager: StoredCardAlertManager = {
-        Analytics.sendEvent(component: paymentMethod.type, flavor: _isDropIn ? .dropin : .components, environment: environment)
+        Analytics.sendEvent(
+            component: paymentMethod.type,
+            flavor: _isDropIn ? .dropin : .components,
```
descorp, line 48: ` _isDropIn ? .dropin : .components ` this could be a helper for `Component` ? 🤔

AdyenCard/Utilities/CardType/CardBrandProvider.swift
```diff
@@ -103,7 +108,7 @@ internal final class CardBrandProvider: AnyCardBrandProvider {
             return success(binLookupService)
         }
         
-        let localApiClient = self.apiClient ?? APIClient(environment: self.environment)
+        let localApiClient = self.apiClient ?? APIClient(apiContext: apiContext)
```
descorp, line 111: In this class we only use `apiContext: apiContext` to create `APIClient` if it was not provided on `init` - maybe just change init ?

```swift
internal init(cardPublicKeyProvider: AnyCardPublicKeyProvider,
              apiClient: APIClientProtocol,
              fallbackCardTypeProvider: AnyCardBrandProvider = FallbackCardBrandProvider()) { ... }
```

AdyenActions/Components/Await/PollingHandlerProvider.swift
kaphacius: It is an implementation of `AnyPollingHandlerProvider`. We need this protocol for testing. 
We are passing `nil` for `apiClient` in `QRCodeComponent.swift #77`

Adyen/Core/APIClient/Environment.swift
descorp: It should be used in `Demo/Common/Networking/DefaultAPIClient.swift`

AdyenDropIn/Models/PaymentMethodsConfiguration.swift
descorp: 👍

AdyenActions/Components/Await/PollingHandlerProvider.swift
descorp: > We are passing nil for apiClient in QRCodeComponent.swift #77

this is my point - we always passing `nil`..

It could be easier to mock `AnyRetryAPIClient` and pass it is `AnyPollingHandler` rather than mock optional `AnyPollingHandlerProvider` ? 🤔

Tests/AdyenTests/Card Tests/Mocks/PublicKeyProviderMock.swift
```diff
@@ -7,6 +7,8 @@
 @testable import AdyenCard
 
 final class CardPublicKeyProviderMock: AnyCardPublicKeyProvider {
+    
+    let apiContext: APIContext = APIContext(environment: Environment.test, clientKey: "local_DUMMYKEYFORTESTING")
```
descorp, line 11: Could be `Dummy.context` ?

AdyenCard/Utilities/CardType/CardBrandProvider.swift
kaphacius: I like this idea. But we need both `apiClient` and `apiContext` for tests, so they both need to be passed into the init. But we can make `apiClient` non-optional and construct it in `init`.

Tests/AdyenTests/Card Tests/Mocks/PublicKeyProviderMock.swift
kaphacius: good catch 👍

Adyen/Core/APIClient/APIContext.swift
```diff
@@ -0,0 +1,50 @@
...
+    /// See https://docs.adyen.com/user-management/client-side-authentication for more information.
+    public let clientKey: String
+    
+    /// Initializes the APIContext
+    /// - Parameters:
+    ///   - environment: The environment to retrieve internal resources from.
+    ///   - clientKey: The client key that corresponds to the webservice user you will use for initiating the payment.
+    public init(environment: AnyAPIEnvironment, clientKey: String) {
+        guard ClientKeyValidator().isValid(clientKey) else {
+            fatalError("ClientKey is invalid.")
```
mohammedDehairy, line 43: maybe an assertion would one better, because we shouldn't crash the merchant app.

kaphacius: yes i agree

mohammedDehairy (dismissed):

AdyenActions/Components/Await/PollingHandlerProvider.swift
kaphacius: maybe we could do this in another PR

AdyenActions/Components/QRCode/QRCodeComponent.swift
kaphacius: maybe we could do this in another PR

AdyenCard/Components/Stored Card/StoredCardComponent.swift
kaphacius: maybe we could do this in another PR

descorp (approved):

PR #525, opened 2021-06-07
Refactoring

descorp:
### Changes

* rename PaymentMethodsConfiguration -> Configuration
* make Amount substruction public
* add test for DropIn giftcard


MIGRATION.md
```diff
@@ -16,7 +16,8 @@
 - `CardComponent` configuration properies like `showsSecurityCodeField`, `showsStorePaymentMethodField`, `showsSecurityCodeField`, `stored`, and `excludedCardTypes` are moved to `CardComponent.Configuration` object.
 - Property `payment` moved from `PresentableComponent` to `PaymentComponent`.
 - `DropInActionComponent` renamed into `AdyenActionComponent` and moved to `AdyenActions` module. It could be used outside of `AdyenDropIn`.
-- Parameters `clientKey` and `environment` are required for `DropInComponent.PaymentMethodsConfiguration` constructor, and they are immutable.
+- `DropInComponent.PaymentMethodsConfiguration` renamed into `DropInComponent.Configuration.`
+- Parameters `clientKey` and `environment` are required for `DropInComponent.Configuration` constructor, and they are immutable.
```
kaphacius, line 20: I think this line might not be relavant in v4 bc of `APIContext`

descorp: good point..

Adyen/Model/Amount.swift
```diff
@@ -87,4 +87,10 @@ extension Amount: Comparable {
         lhs.value > rhs.value
     }
 
+    public static func - (lhs: Amount, rhs: Amount) -> Amount {
+        AdyenAssertion.assert(message: "Currency should match to compare",
+                              condition: lhs.currencyCode != rhs.currencyCode)
+        return Amount(value: lhs.value - rhs.value, currencyCode: lhs.currencyCode)
+    }
```
mohammedDehairy, line 94: I think this is not needed, its best to leave it internal, until its needed

Adyen.xcodeproj/project.pbxproj
```diff
@@ -4640,7 +4640,7 @@
 				E9021DFC2260D626009CF7F4 /* ComponentSections.swift in Sources */,
 				E7484C2723B9FDF000CCF74F /* SlideInPresentationAnimator.swift in Sources */,
 				5A238C93263AEA0800758FEF /* PreApplePayView.swift in Sources */,
-				E9B36CC722451C8300EAA368 /* PaymentMethodsConfiguration.swift in Sources */,
+				E9B36CC722451C8300EAA368 /* Configuration.swift in Sources */,
```
mohammedDehairy, line 4643: lets call this file `DropInConfiguration.swift` for better discoverability, since we have already a file called `Configuration.swift`

Adyen/Model/Amount.swift
descorp: But we use it both SDK and in Tests. I see no harm in having it public 🤔

Adyen.xcodeproj/project.pbxproj
descorp: 👍

Adyen/Model/Amount.swift
mohammedDehairy: in tests we can use @testable, the harm is that it will be public interface, and we'll bound to maintain it.

descorp: 👍

kaphacius (approved):

mohammedDehairy (approved):

PR #534, opened 2021-06-10
chore: add callback on navigating to other app

descorp:
### Changes

* add callback on navigating to other app


Tests/AdyenTests/Adyen Tests/DropIn/DropInActionTests.swift
```diff
@@ -41,4 +41,29 @@ class DropInActionsTests: XCTestCase {
         waitForExpectations(timeout: 15, handler: nil)
     }
 
+    func testOpenExternalApp() {
+        let config = DropInComponent.Configuration(apiContext: Dummy.context)
+        config.payment = Payment(amount: Amount(value: 100, currencyCode: "CNY"), countryCode: "CN")
+
+        let waitExpectation = expectation(description: "Expect a callback")
+        let mock = MockDropInDelegate()
```
mohammedDehairy, line 49: I would rather call it `DropInDelegateMock`, starting with `Mock` makes it seem like an imperative order or a function name :-), but its a not a big deal, we can rename it later

mohammedDehairy (dismissed):

Tests/AdyenTests/Adyen Tests/DropIn/DropInActionTests.swift
descorp: fixed

mohammedDehairy (dismissed):

README.md
```diff
@@ -172,19 +172,27 @@ This method is invoked when an error occurred during the use of the Drop-in. Dis
 
 ---
 
+```swift
+func didComplete(from component: DropInComponent)
+```
+
+This method is invoked when the action component finishes, without any further steps needed by the application, for example in case of voucher payment methods. The application just needs to dismiss the `DropInComponent`..
```
kaphacius, line 179: double `..` at the end

README.md
```diff
@@ -172,19 +172,27 @@ This method is invoked when an error occurred during the use of the Drop-in. Dis
...
 
 ---
 
 ```swift
-func didComplete(from component: DropInComponent)
+func didOpenExternalApplication(_ component: DropInComponent)
 ```
 
-This method is invoked when the action component finishes, without any further steps needed by the application, for example in case of voucher payment methods. The application just needs to dismiss the `DropInComponent`..
+This optional method is invoked when redirect to external application occurred.
```
kaphacius, line 195: maybe `This optional method is invoked after a redirect to an external application has occurred.`

kaphacius (approved):

README.md
descorp: 👍

README.md
descorp: 👍

mohammedDehairy (approved):

PR #535, opened 2021-06-11
WIP Refactor ComponentManager to create components on demand

kaphacius:
This PR will change the way `ComponentManager` creates components to be on-demand, instead of creating all of the components beforehand, and reusing them later.

### DIFF
--- a/Adyen/Core/Payment Methods/PaymentMethod.swift
+++ b/Adyen/Core/Payment Methods/PaymentMethod.swift
@@ -52,4 +52,14 @@ public protocol StoredPaymentMethod: PaymentMethod {
     /// The supported types of shopper interaction.
     var supportedShopperInteractions: [ShopperInteraction] { get }
     
+    /// Indicates whether the shopper is present.
+    var isShopperPresent: Bool { get }
+    
+}
+
+/// :nodoc:
+public extension StoredPaymentMethod {
+    
+    var isShopperPresent: Bool { supportedShopperInteractions.contains(.shopperPresent) }
+
 }
--- a/Adyen/UI/List/ListItem.swift
+++ b/Adyen/UI/List/ListItem.swift
@@ -28,9 +28,6 @@ public class ListItem: FormItem {
     /// The trailing text of the item.
     public var trailingText: String?
     
-    /// A boolean value indicating whether a disclosure indicator should be shown in the item's cell.
-    public var showsDisclosureIndicator: Bool
-    
     /// The handler to invoke when the item is selected.
     public var selectionHandler: (() -> Void)?
     
@@ -48,7 +45,6 @@ public class ListItem: FormItem {
     ///   - imageURL: A URL to an image to display.
     ///   - trailingText: The trailing text.
     ///   - style: The list item style.
-    ///   - showsDisclosureIndicator: A boolean value indicating whether a disclosure indicator
     ///                               should be shown in the item's cell.
     ///   - selectionHandler: The closure to execute when an item is selected.
     ///   - canModifyIcon: The flag to indicate that image could be tampered.
@@ -56,14 +52,12 @@ public class ListItem: FormItem {
                 imageURL: URL? = nil,
                 trailingText: String? = nil,
                 style: ListItemStyle = ListItemStyle(),
-                showsDisclosureIndicator: Bool = true,
                 selectionHandler: (() -> Void)? = nil,
                 canModifyIcon: Bool = true) {
         self.title = title
         self.imageURL = imageURL
         self.trailingText = trailingText
         self.style = style
-        self.showsDisclosureIndicator = showsDisclosureIndicator
         self.selectionHandler = selectionHandler
         self.canModifyIcon = canModifyIcon
     }
--- a/AdyenActions/AdyenActionComponent.swift
+++ b/AdyenActions/AdyenActionComponent.swift
@@ -102,7 +102,10 @@ public final class AdyenActionComponent: ActionComponent, Localizable {
     }
     
     private func handle(_ action: ThreeDS2ChallengeAction) {
-        guard let threeDS2Component = currentActionComponent as? ThreeDS2Component else { return }
+        guard let threeDS2Component = currentActionComponent as? ThreeDS2Component else {
+            AdyenAssertion.assertionFailure(message: "Trying to handle a 3DS2Challenge, but current component is not a ThreeDS2Component")
+            return
+        }
         threeDS2Component.handle(action)
     }
 
--- a/AdyenActions/AdyenActionComponent.swift
+++ b/AdyenActions/AdyenActionComponent.swift
@@ -25,24 +25,17 @@ public final class AdyenActionComponent: ActionComponent, Localizable {
     /// :nodoc:
     public weak var presentationDelegate: PresentationDelegate?
     
-    /// Indicates the UI configuration of the redirect component.
-    public var redirectComponentStyle: RedirectComponentStyle?
-    
-    /// Indicates the UI configuration of the await component.
-    public var awaitComponentStyle: AwaitComponentStyle?
-
-    /// Indicates the UI configuration of the voucher component.
-    public var voucherComponentStyle: VoucherComponentStyle?
-
-    /// Indicates the UI configuration of the QR code component.
-    public var qrCodeComponentStyle: QRCodeComponentStyle?
+    /// :nodoc:
+    public let style: AdyenActionComponentStyle
 
     /// :nodoc:
     public var localizationParameters: LocalizationParameters?
     
     /// :nodoc:
-    public init(apiContext: APIContext) {
+    public init(apiContext: APIContext,
+                style: AdyenActionComponentStyle = AdyenActionComponentStyle()) {
         self.apiContext = apiContext
+        self.style = style
     }
     
     // MARK: - Performing Actions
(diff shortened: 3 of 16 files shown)


AdyenDropIn/Components/PaymentMethodListComponent.swift
```diff
@@ -46,30 +46,33 @@ internal final class PaymentMethodListComponent: ComponentLoader, PresentableCom
...
+            listItem.imageURL = LogoURLProvider.logoURL(for: paymentMethod, environment: apiContext.environment)
             listItem.trailingText = displayInformation.disclosureText
             listItem.subtitle = displayInformation.subtitle
-            listItem.showsDisclosureIndicator = showsDisclosureIndicator
-            listItem.selectionHandler = { [unowned self, unowned component] in
-                guard !(component is AlreadyPaidPaymentComponent) else { return }
-                self.delegate?.didSelect(component, in: self)
+
+            // Unsure how to handle this
+//            let showsDisclosureIndicator = (component as? PresentableComponent)?.requiresModalPresentation == true
```
descorp, line 61: This shows `>` chevron on list.. we can probably omite this

AdyenDropIn/Components/PaymentMethodListComponent.swift
```diff
@@ -98,9 +101,9 @@ internal final class PaymentMethodListComponent: ComponentLoader, PresentableCom
     /// - Parameter component: The component for which to start a loading animation.
     internal func startLoading(for component: PaymentComponent) {
         let allListItems = listViewController.sections.flatMap(\.items)
-        let allComponents = componentSections.map(\.components).flatMap { $0 }
+        let allComponents = paymentMethodSections.map(\.paymentMethods).flatMap { $0 }
```
descorp, line 104: `allComponents` -> `allPaymentMethods` ?

AdyenDropIn/Models/DropInComponentStyle.swift
```diff
@@ -25,14 +25,8 @@ public extension DropInComponent {
...
-        /// Indicates any `RedirectComponent` UI style.
-        public var redirectComponent: RedirectComponentStyle?
-        
-        /// Indicates the await action component UI style.
-        public var awaitComponent = AwaitComponentStyle()
-
-        /// Indicates the voucher action component UI style.
-        public var voucherComponent = VoucherComponentStyle()
+        /// Indicates the UI configuration of Action Components
+        public var actionComponents = AdyenActionComponent.Style()
```
descorp, line 29: This is nice :)

AdyenActions/AdyenActionComponent.swift
```diff
@@ -130,42 +118,82 @@ public final class AdyenActionComponent: ActionComponent, Localizable {
...
         component.presentationDelegate = presentationDelegate
         component.localizationParameters = localizationParameters
         
         component.handle(action)
-        qrCodeComponent = component
+        currentActionComponent = component
     }
 }
+
+public extension AdyenActionComponent {
```
descorp, line 164: This extension could be in it's own file

AdyenDropIn/Components/PaymentMethodListComponent.swift
kaphacius: I'm not sure if we need that. To me it seems unnecessary, since the disclosure indicator usually means using native navigation, and something sliding from the right. Maybe removing it is okay

AdyenDropIn/Components/PaymentMethodListComponent.swift
kaphacius: good find

AdyenActions/AdyenActionComponent.swift
kaphacius: yeah i was just in a hurry :)

AdyenActions/AdyenActionComponent.swift
```diff
@@ -25,24 +25,17 @@ public final class AdyenActionComponent: ActionComponent, Localizable {
...
+    /// :nodoc:
+    public let style: Style
 
     /// :nodoc:
     public var localizationParameters: LocalizationParameters?
     
     /// :nodoc:
-    public init(apiContext: APIContext) {
+    public init(apiContext: APIContext,
+                style: Style?) {
```
descorp, line 36: How about `style: Style? = nil` ?

AdyenDropIn/Utilities/ComponentManager.swift
```diff
@@ -56,36 +55,34 @@ internal final class ComponentManager {
...
+        let paidSection = PaymentMethodsSection(header: nil,
+                                                paymentMethods: paymentMethods.paid,
+                                                footer: paidFooter)
 
         // Stored section
-        let storedSection = ComponentsSection(components: storedComponents)
+        // Filter out payment methods without the Ecommerce shopper interaction.
+        let storedSection = PaymentMethodsSection(header: nil,
+                                                  paymentMethods: paymentMethods.stored.filter {
+                                                    $0.supportedShopperInteractions.contains(.shopperPresent)
```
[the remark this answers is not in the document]
kaphacius: 👍

Adyen/Core/Payment Methods/PaymentMethod.swift
```diff
@@ -52,4 +52,14 @@ public protocol StoredPaymentMethod: PaymentMethod {
     /// The supported types of shopper interaction.
     var supportedShopperInteractions: [ShopperInteraction] { get }
     
+    /// Indicates whether the shopper is present.
+    var isShopperPresent: Bool { get }
```
kaphacius, line 56: this has to be in the protocol. if i only add it to the protocol extension - we get a segmentation fault from swift compiler 🤷

Adyen/UI/List/ListItem.swift
```diff
@@ -28,9 +28,6 @@ public class ListItem: FormItem {
     /// The trailing text of the item.
     public var trailingText: String?
     
-    /// A boolean value indicating whether a disclosure indicator should be shown in the item's cell.
-    public var showsDisclosureIndicator: Bool
```
mohammedDehairy, line 32: why is this removed, its needed for the gift card elements that are not clickable, so the disclosure needs to be .none in this case

AdyenDropIn/Components/PaymentMethodListComponent.swift
descorp: 👍

AdyenActions/AdyenActionComponent.swift
```diff
@@ -114,10 +102,10 @@ public final class AdyenActionComponent: ActionComponent, Localizable {
     }
     
     private func handle(_ action: ThreeDS2ChallengeAction) {
-        guard let threeDS2Component = threeDS2Component else { return }
+        guard let threeDS2Component = currentActionComponent as? ThreeDS2Component else { return }
```
mohammedDehairy, line 105: we should make an assertionFailure here in case this condition fails

AdyenDropIn/Components/PaymentMethodListComponent.swift
```diff
@@ -97,13 +95,14 @@ internal final class PaymentMethodListComponent: ComponentLoader, PresentableCom
     ///
     /// - Parameter component: The component for which to start a loading animation.
     internal func startLoading(for component: PaymentComponent) {
-        let allListItems = listViewController.sections.flatMap(\.items)
-        let allComponents = componentSections.map(\.components).flatMap { $0 }
+        let allPaymentMethods = paymentMethodSections.map(\.paymentMethods).flatMap { $0 }
         
-        guard let index = allComponents.firstIndex(where: { $0 === component }) else {
+        guard let index = allPaymentMethods.firstIndex(where: { $0.type == component.paymentMethod.type }) else {
```
mohammedDehairy, line 100: maybe there are multiple payment methods with the same type, like for example gift card PM's I think have the same type `giftcard`, but not sure though, can you check that, or try to find something else to identify the payment method, like adding a UUID to the payment method, generated in the initializer?

AdyenDropIn/DropInComponent.swift
```diff
@@ -257,7 +256,8 @@ public final class DropInComponent: NSObject, PresentableComponent {
 extension DropInComponent: PaymentMethodListComponentDelegate {
     
     /// :nodoc:
-    internal func didSelect(_ component: PaymentComponent, in paymentMethodListComponent: PaymentMethodListComponent) {
+    internal func didSelect(_ paymentMethod: PaymentMethod, in paymentMethodListComponent: PaymentMethodListComponent) {
+        guard let component = componentManager.component(for: paymentMethod) else { return }
```
mohammedDehairy, line 260: assertion failure in case the condition fails?

Adyen/UI/List/ListItem.swift
descorp: This is turns `chevron` on and off..

mohammedDehairy: yes, its needed to hide the chevron for gift card items

descorp: but if we remove it - there will be no chevron on Drop-in, and it is fine [as we discussed above](https://github.com/Adyen/adyen-ios/pull/535#discussion_r649767666)

mohammedDehairy: but its in the design, right? if the designers are okay with it, then its fine.

kaphacius: i haven't talked to design yet. i will

AdyenActions/AdyenActionComponent.swift
kaphacius: how does `Trying to handle a 3DS2Challenge, but current component is not a ThreeDS2Component` sound?

mohammedDehairy: sounds good to me

AdyenDropIn/Components/PaymentMethodListComponent.swift
kaphacius: how about adding a function to compare payment methods?
```swift
    func isEqual(to other: PaymentMethod) -> Bool {
        switch (self as? StoredPaymentMethod, other as? StoredPaymentMethod) {
        case (.some(let lhs), .some(let rhs)):
            return lhs.identifier == rhs.identifier
        case (.none, .none):
            return self.name == other.name && self.type == other.type
        case (_, _):
            return false
        }
    }
```

kaphacius:
This is parked for now.

PR #544, opened 2021-06-15
chore: move NavigationStyle to Drop-in

descorp:
### Changes

* move NavigationStyle to Drop-in module
* make UIFont.Adyen public

### DIFF
--- a/Adyen/UI/Styles/CoreFonts.swift
+++ b/Adyen/UI/Styles/CoreFonts.swift
@@ -10,9 +10,10 @@ import UIKit
 extension UIFont {
     
     /// :nodoc:
-    internal enum AdyenCore {
-        
-        internal static var barTitle: UIFont {
+    public enum Adyen {
+
+        /// :nodoc:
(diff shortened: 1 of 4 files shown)


Adyen/UI/Styles/CoreFonts.swift
```diff
@@ -10,9 +10,10 @@ import UIKit
 extension UIFont {
     
     /// :nodoc:
-    internal enum AdyenCore {
-        
-        internal static var barTitle: UIFont {
+    public enum Adyen {
```
mohammedDehairy, line 13: why is this public now?

Adyen/UI/Styles/CoreFonts.swift
```diff
@@ -10,9 +10,10 @@ import UIKit
...
+
+        /// :nodoc:
+        public static var barTitle: UIFont {
```
mohammedDehairy, line 16: also this was internal

mohammedDehairy (approved):

PR #557, opened 2021-07-02
added validation to verify that the 3DS2 versions in podspec, cartfile, and SPM and ThreeDS2SdkVersion.swift all match

mohammedDehairy:
added validation to verify that the 3DS2 versions in podspec, cartfile, and SPM and ThreeDS2SdkVersion.swift all match

### DIFF
--- a/Scripts/validate-3ds2-SDK-version.sh
+++ b/Scripts/validate-3ds2-SDK-version.sh
@@ -1,0 +1,25 @@
+#!/bin/sh
+
+cardDetails3DS2SDKVersion=$(sed -rn 's/^.*public.*let.*threeDS2SdkVersion.*:.*String.*=.*"([0-9]+.[0-9].[0-9]+)".*$/\1/p' AdyenCard/Components/Card/ThreeDS2SdkVersion.swift)
+cocoapods3DS2SDKVersion=$(sed -rn "s/^.*dependency.*\'Adyen3DS2'.*,.*'([0-9]+.[0-9].[0-9]+)'.*$/\1/p" Adyen.podspec)
+carthage3DS2SDKVersion=$(sed -rn 's/^.*github.*"adyen\/adyen-3ds2-ios".*==.*([0-9]+.[0-9].[0-9]+).*$/\1/p' Cartfile)
+swiftPackageManager3DS2SDKVersion=$(pcregrep -M --om-separator='.' --only-matching=1 --only-matching=2 --only-matching=3 '^[\n\s]*name:[\n\s]*"Adyen3DS2"[\n\s]*,[\n\s]*url[\n\s]*:[\n\s]*"https:\/\/github.com\/Adyen\/adyen\-3ds2\-ios"[\n\s]*,[\n\s]*\.exact\(Version\(([0-9]+)[\n\s]*,[\n\s]*([0-9]+)[\n\s]*,[\n\s]+([0-9]+)\)\)[\n\s]*$' Package.swift)
+
+echo '\n'
+echo '-----------------------------------------------------'
+echo "ThreeDS2SdkVersion.swift: $cardDetails3DS2SDKVersion"
+echo "Adyen.podspec:            $cocoapods3DS2SDKVersion"
+echo "Cartfile:                 $carthage3DS2SDKVersion"
+echo "Package.swift:            $swiftPackageManager3DS2SDKVersion"
+echo '------------------------------------------------------\n'
+
+declare -a array=( $cardDetails3DS2SDKVersion $cocoapods3DS2SDKVersion $carthage3DS2SDKVersion $swiftPackageManager3DS2SDKVersion )
+
+declare -a uniq=($(echo "${array[@]}" | tr ' ' '\n' | sort -u | tr '\n' ' '))
+
+if [ ${#uniq[@]} -ne 1 ]; then
+  RED='\033[0;31m'
+  NC='\033[0m'
+  echo "[ERROR] : ${RED}3DS2 SDK version in Package.swift, Cartfile, Adyen.podspec, and ThreeDS2SdkVersion.swift files donot match, please fix the conflict and try to commit again.${NC}"
+  exit 1
+fi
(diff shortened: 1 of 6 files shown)


Scripts/validate-3ds2-SDK-version.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
+
+cardDetails3DS2SDKVersion=$(sed -rn 's/^.*public.*let.*threeDS2SdkVersion.*:.*String.*=.*"([0-9]+.[0-9].[0-9]+)".*$/\1/p' AdyenCard/Components/Card/ThreeDS2SdkVersion.swift)
+cocoapods3DS2SDKVersion=$(sed -rn "s/^.*dependency.*\'Adyen3DS2'.*,.*'([0-9]+.[0-9].[0-9]+)'.*$/\1/p" Adyen.podspec)
+carthage3DS2SDKVersion=$(sed -rn 's/^.*github.*"adyen\/adyen-3ds2-ios".*==.*([0-9]+.[0-9].[0-9]+).*$/\1/p' Cartfile)
+swiftPackageManager3DS2SDKVersion=$(pcregrep -M --om-separator='.' --only-matching=1 --only-matching=2 --only-matching=3 '^[\n\s]*name:[\n\s]*"Adyen3DS2"[\n\s]*,[\n\s]*url[\n\s]*:[\n\s]*"https:\/\/github.com\/Adyen\/adyen\-3ds2\-ios"[\n\s]*,[\n\s]*\.exact\(Version\(([0-9]+)[\n\s]*,[\n\s]*([0-9]+)[\n\s]*,[\n\s]+([0-9]+)\)\)[\n\s]*$' Package.swift)
```
kaphacius, line 6: i don't think `pcregrep` is a part of standard mac os terminal. my machine does not have it, so the script did not work

Scripts/validate-3ds2-SDK-version.sh
```diff
@@ -0,0 +1,25 @@
...
+
+echo '\n'
+echo '-----------------------------------------------------'
+echo "ThreeDS2SdkVersion.swift: $cardDetails3DS2SDKVersion"
+echo "Adyen.podspec:            $cocoapods3DS2SDKVersion"
+echo "Cartfile:                 $carthage3DS2SDKVersion"
+echo "Package.swift:            $swiftPackageManager3DS2SDKVersion"
+echo '------------------------------------------------------\n'
+
+declare -a array=( $cardDetails3DS2SDKVersion $cocoapods3DS2SDKVersion $carthage3DS2SDKVersion $swiftPackageManager3DS2SDKVersion )
```
kaphacius, line 16: we need to check if the array has exactly 4 values. when testing the script `pcregrep` was not available, but there was an error shown. maybe something like this:
```zsh
RED='\033[0;31m'
NC='\033[0m'

if [ ${#array[@]} != 4 ]; then
	echo "[ERROR] : ${RED}One or more SDK version checks failed. Please check Package.swift, Cartfile, Adyen.podspec, and ThreeDS2SdkVersion.swift${NC}"
	exit 1
fi
```
ideally we would also show it in lines 10-13 as well, depending if there was a value retrieved.

mohammedDehairy: I think you miss understood the script, it's trying to check if all versions are the same and if not throw an error, so we need to check how many unique values in the array, and if not exactly 1 throw an error.

Scripts/validate-3ds2-SDK-version.sh
mohammedDehairy: Hmm 🤔  interesting, it works on my machine

kaphacius: the most developer response ever :D

Scripts/validate-3ds2-SDK-version.sh
kaphacius: i think i understand what it does, but look at this output i got. no version from `Package.swift` was found, yet the script was completed successfully. my suggestion was to check whether all 4 versions were obtained, so it has to be run after line 16.
```zsh
./Scripts/validate-3ds2-SDK-version.sh
./Scripts/validate-3ds2-SDK-version.sh: line 6: pcregrep: command not found

-----------------------------------------------------
ThreeDS2SdkVersion.swift: 2.2.2
Adyen.podspec:            2.2.2
Cartfile:                 2.2.2
Package.swift:
------------------------------------------------------
```

Scripts/validate-3ds2-SDK-version.sh
descorp: I also have it 🤔

Scripts/validate-3ds2-SDK-version.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
```
descorp, line 1: Should we add `validate-3ds2-SDK-version.sh` as a step on CI? At least after merge to `master` ?

Scripts/validate-3ds2-SDK-version.sh
```diff
@@ -0,0 +1,25 @@
...
+echo '------------------------------------------------------\n'
+
+declare -a array=( $cardDetails3DS2SDKVersion $cocoapods3DS2SDKVersion $carthage3DS2SDKVersion $swiftPackageManager3DS2SDKVersion )
+
+declare -a uniq=($(echo "${array[@]}" | tr ' ' '\n' | sort -u | tr '\n' ' '))
+
+if [ ${#uniq[@]} -ne 1 ]; then
+  RED='\033[0;31m'
+  NC='\033[0m'
+  echo "[ERROR] : ${RED}3DS2 SDK version in Package.swift, Cartfile, Adyen.podspec, and ThreeDS2SdkVersion.swift files donot match, please fix the conflict and try to commit again.${NC}"
```
kaphacius, line 23: `do not match`

Scripts/validate-3ds2-SDK-version.sh
mohammedDehairy: After merging is too late, this script will prevent your commit even locally, I wanted prevention as early as possible.

Scripts/validate-3ds2-SDK-version.sh
mohammedDehairy: used sed instead, I think everyone has it

Scripts/validate-3ds2-SDK-version.sh
mohammedDehairy: fixed

kaphacius (dismissed):

kaphacius (approved):

kaphacius (approved):

descorp (approved):

PR #562, opened 2021-07-14
chore: add backticks to await case in advance of ios15 async/await

Schroefdop:
## Summary

Added backticks to the `await` case of the `Action` enum to prevent issues with iOS 15s new async/await keywords


descorp:
Nice catch! Thanks @Schroefdop

descorp (approved):

mohammedDehairy (approved):

PR #563, opened 2021-07-14
v3: add new networks to Apple pay

descorp:
## Changes

* add girocard, elo, electron to Apple Pay


kaphacius (approved):

mohammedDehairy:
👍  once you fix unit tests, I will approve.

mohammedDehairy:
I am wondering if this is necessary to make a release in v3, what if we add those to v4 instead, its not like a blocking bug that merchants really need, right?

PR #580, opened 2021-07-30
fix: luhn check flag condition

erenbesel:
Realized luhn check condition was wrong. Unit tests were missing too. Fixed and confirmed with unit tests


mohammedDehairy (approved):

nauaros (approved):
Nice!

PR #583, opened 2021-08-04
Add didSubmitLastFour to CardComponentDelegate

descorp:
### Changes

* add didSubmitLastFour to CardComponentDelegate
* add Dummy cards for testing
* add helper to populate card and submit button

### DIFF
--- a/AdyenCard/Components/Card/CardComponentDelegate.swift
+++ b/AdyenCard/Components/Card/CardComponentDelegate.swift
@@ -10,7 +10,7 @@ import UIKit
 /// Delegate for observing user's activity on `CardComponent`.
 public protocol CardComponentDelegate: AnyObject {
 
-    /// Called when user enters PAN in `CardComponent`.
+    /// Called when shopper enters PAN in `CardComponent`.
     /// - Parameter value: Up to 6 first digits in entered PAN.
     /// - Parameter component: The `CardComponent` instance.
     func didChangeBIN(_ value: String, component: CardComponent)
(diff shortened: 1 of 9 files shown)


AdyenCard/Components/Card/CardComponentDelegate.swift
```diff
@@ -19,4 +19,9 @@ public protocol CardComponentDelegate: AnyObject {
     /// - Parameter value: Array of card types matching entered value. Null - if no data entered.
     /// - Parameter component: The `CardComponent` instance.
     func didChangeCardBrand(_ value: [CardBrand]?, component: CardComponent)
+
+    /// Called when shopper submits a card in `CardComponent`.
+    /// - Parameter value: Last 4 digits of card.
+    /// - Parameter component: The `CardComponent` instance.
+    func didSubmitLastFour(_ value: String, component: CardComponent)
```
mohammedDehairy, line 26: can we add a default implementation for the new delegate call back, since this is technically a breaking change.

merchants will get an error because the new call back isn't implemented, not to mention merchants might not need it.

descorp: 👍

Tests/AdyenTests/Card Tests/CardComponentTests.swift
```diff
@@ -1280,3 +1260,21 @@ extension UIView {
         }
     }
 }
+
+extension XCTestCase {
+
+    func fillCard(on view: UIView, with card: Card) {
+        let cardNumberItemView: FormTextItemView<FormCardNumberItem>? = view.findView(with: "AdyenCard.CardComponent.numberItem")
```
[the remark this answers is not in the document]
descorp: we can do that 🤔

mohammedDehairy (dismissed):

PR #585, opened 2021-08-05
chore: renamed didSubmitLastFour to didSubmit to make it more swifty

mohammedDehairy:
chore: renamed didSubmitLastFour to didSubmit to make it more swifty

### DIFF
--- a/AdyenCard/Components/Card/CardComponentDelegate.swift
+++ b/AdyenCard/Components/Card/CardComponentDelegate.swift
@@ -21,13 +21,13 @@ public protocol CardComponentDelegate: AnyObject {
     func didChangeCardBrand(_ value: [CardBrand]?, component: CardComponent)
 
     /// Called when shopper submits a card in `CardComponent`.
-    /// - Parameter value: Last 4 digits of card.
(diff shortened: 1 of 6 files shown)


kaphacius (approved):

erenbesel (approved):

AdyenCard/Components/Card/CardComponentDelegate.swift
```diff
@@ -21,13 +21,13 @@ public protocol CardComponentDelegate: AnyObject {
     func didChangeCardBrand(_ value: [CardBrand]?, component: CardComponent)
 
     /// Called when shopper submits a card in `CardComponent`.
-    /// - Parameter value: Last 4 digits of card.
+    /// - Parameter lastFour: Last 4 digits of card.
```
descorp, line 24: I think the param name is still `value` ?

mohammedDehairy: lastFour is what is shown to the code user, but value is more internal to the function, right?

nauaros (approved):

erenbesel (approved):

descorp (approved):

PR #590, opened 2021-08-09
Cois 460 optional expiry date

erenbesel:
## Summary
- Expiry date field made optional based on binlookup response. For now, similar to CVC, if the policy is hidden/optional, the field will be optional with an updated title (pending translation).
- Renamed `CVCPolicy` enum to `RequirementPolicy`
- Changed SplitItem stackview alignment to default to let both textfields be in the same vertical alignment.

## Tested scenarios
validation checks when optional/non optional

**Fixed issue**:  [COIOS-460](https://youtrack.is.adyen.com/issue/COIOS-460)


AdyenCard/Form/FormCardExpiryDateItem.swift
```diff
@@ -0,0 +1,63 @@
+//
+// Copyright (c) 2021 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Adyen
+
+/// A form item into which card expiry date is entered, formatted and validated.
+internal final class FormCardExpiryDateItem: FormTextItem, Hidable {
```
kaphacius, line 10: this is nice, extracting the logic in a separate class

AdyenCard/Form/FormCardExpiryDateItem.swift
```diff
@@ -0,0 +1,63 @@
...
+    
+    /// Flag determining this forms state. Validation changes based on this.
+    internal var isOptional: Bool = false {
+        didSet {
+            updateFormState()
+        }
+    }
+    
+    private lazy var expiryDateValidator = CardExpiryDateValidator()
+    private lazy var expiryDateFormatter = CardExpiryDateFormatter()
```
kaphacius, line 26: we probably don't need this line, we can do the assignment in line 35

AdyenCard/Form/FormCardExpiryDateItem.swift
```diff
@@ -0,0 +1,63 @@
...
+    internal var localizationParameters: LocalizationParameters?
+    
+    /// Flag determining this forms state. Validation changes based on this.
+    internal var isOptional: Bool = false {
+        didSet {
+            updateFormState()
+        }
+    }
+    
+    private lazy var expiryDateValidator = CardExpiryDateValidator()
```
kaphacius, line 25: we probably don't need this line, we can do the assignment in line 36

AdyenCard/Form/FormCardSecurityCodeItem.swift
```diff
@@ -69,6 +74,20 @@ extension Array where Element == CardBrand {
         }
     }
     
+    /// At the moment, due to UI concerns, we won't hide cvc/exp date fields even when response is hidden
```
kaphacius, line 77: maybe this `extension Array where Element == CardBrand` can be in a separate file, since now it contains things not related to `FormCardSecurityCodeItem`

AdyenCard/Form/FormCardSecurityCodeItem.swift
```diff
@@ -69,6 +74,20 @@ extension Array where Element == CardBrand {
...
+    /// We will make them optional.
+    var isExpiryDateOptional: Bool {
```
kaphacius, line 79: what if we do this
```swift
    /// Indicates the requirement level of a field.
    public enum RequirementPolicy: String, Decodable {

        /// Field is required.
        case required

        /// Field is optional.
        case optional

        /// Field should not be asked.
        case hidden
        
        var isOptional: Bool {
            self != .required
        }
    }

extension Array where Element == CardBrand {
    var isCVCOptional: Bool {
        isEmpty ? false : lazy.allSatisfy(\.cvcPolicy.isOptional)
    }

    var isExpiryDateOptional: Bool {
        isEmpty ? false : lazy.allSatisfy(\.cvcPolicy.isOptional)
    }
}

File: AdyenCard/Utilities/CardType/FallbackCardBrandProvider.swift [hunk 6]
```diff
@@ -14,7 +14,7 @@ internal final class FallbackBinInfoProvider: AnyBinInfoProvider {
     internal func provide(for bin: String, supportedTypes: [CardType], completion: @escaping (BinLookupResponse) -> Void) {
         let result: [CardBrand] = supportedTypes.adyen.types(forCardNumber: bin).map { brand in
 
-            let cvcPolicy: CardBrand.CVCPolicy
+            let cvcPolicy: CardBrand.RequirementPolicy
             switch brand {
             case .laser,
```
kaphacius, line 19: does it makes sense to move this into `CardType`? seems like this is directly connected to the card type. 
maybe we can also not use `default` in the `switch` if we decide to move this

AdyenCard/Form/FormCardExpiryDateItem.swift
erenbesel: One of them is accessed later, so it is needed. The other one I kept for similarity. That one can be removed.

AdyenCard/Utilities/CardType/FallbackCardBrandProvider.swift
erenbesel: Could you eloborate on this one? Didn't get it

AdyenCard/Form/FormCardExpiryDateItem.swift
kaphacius: alright. can it be a `let` instead of a `lazy var`? :) or maybe a `static let`?

AdyenCard/Form/FormCardSecurityCodeItem.swift
erenbesel: DO you mean move the extension to CardBrand? If so, makes sense :)

AdyenCard/Form/FormCardExpiryDateItem.swift
mohammedDehairy: 👍

AdyenCard/Form/FormCardExpiryDateItem.swift
```diff
@@ -0,0 +1,63 @@
...
+    
+    private func updateFormState() {
+        // when optional, if user enters anything it should be validated as regular entry.
+        if isOptional {
+            title = localizedString(.cardExpiryItemTitleOptional, localizationParameters)
+            validator = NumericStringValidator(exactLength: 0) || expiryDateValidator
+        } else {
+            title = localizedString(.cardExpiryItemTitle, localizationParameters)
+            validator = expiryDateValidator
+        }
```
mohammedDehairy, line 54: Nice!

AdyenCard/Utilities/CardType/CardBrand.swift
```diff
@@ -8,6 +8,19 @@ import Foundation
 
 /// Describes the Card brand.
 public struct CardBrand: Decodable {
+    
+    /// Indicates the requirement level of a field.
+    public enum RequirementPolicy: String, Decodable {
```
mohammedDehairy, line 13: it can be internal I think, its most probably not used by merchants since `cvcPolicy` was internal.

Tests/AdyenTests/Card Tests/CardComponentTests.swift
```diff
@@ -747,32 +747,25 @@ class CardComponentTests: XCTestCase {
...
-            XCTAssertEqual(cityItemView!.titleLabel.text, "City")
-            XCTAssertEqual(provinceOrTerritoryItemView!.titleLabel.text, "Province or Territory")
-            XCTAssertEqual(postalCodeItemView!.titleLabel.text, "Postal code")
-            XCTAssertEqual(headerItemView!.text, "Billing address")
-
-            expectation.fulfill()
-        }
-
-        wait(for: [expectation], timeout: 10)
+        wait(for: .seconds(1))
```
mohammedDehairy, line 750: nice!

AdyenCard/Utilities/CardType/CardBrand.swift
erenbesel: Changing it :)

AdyenCard/Utilities/CardType/FallbackCardBrandProvider.swift
erenbesel: I am keeping it as is since it's not related to this PR. But can be discussed for change on its own topic.

kaphacius (approved):

mohammedDehairy (approved):

AdyenCard/Form/FormCardExpiryDateItemView.swift
```diff
@@ -0,0 +1,11 @@
...
+// Copyright (c) 2021 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Adyen
+import UIKit
+
+/// A view representing a form card security code item.
+internal final class FormCardExpiryDateItemView: FormTextItemView<FormCardExpiryDateItem> {}
```
mohammedDehairy, line 11: why is this needed? the `FormCardExpiryDateItem` can just build a `FormTextItemView<FormCardExpiryDateItem>`, right?

erenbesel: You are right, I saw that but also saw FormTextInputItemView and assumed we separate them for their own.

mohammedDehairy: `FormTextInputItemView` is also not needed :-D , I don't know why we have it.

Adyen/Assets/en-US.lproj/Localizable.strings
```diff
@@ -21,6 +21,7 @@
 "adyen.card.numberItem.placeholder" = "1234 5678 9012 3456";
 "adyen.card.numberItem.invalid" = "Invalid card number";
 "adyen.card.expiryItem.title" = "Expiry date";
+"adyen.card.expiryItem.title.optional" = "Expiry date (optional)";
```
descorp, line 24: We already have line for `Expiry date` and line for `(optional)` and we previously used combination of them.
why change this?
Also you need to request translation for "CVC/CVV (optional)"

AdyenCard/Form/FormCardExpiryDateItemView.swift
descorp: The only purpose of `FormTextInputItemView ` - it is a final non-generic class

Adyen/Assets/en-US.lproj/Localizable.strings
erenbesel: We already had CVC/CVV (optional) translation in and it is used in FormCardSecurityCodeItem. That's why I did the same. 
But also, some languages may differ when writing texts together. The word optional may be different when next to ExpDate etc.

descorp: ok, you right.

descorp (dismissed):

AdyenCard/Form/FormCardSecurityCodeItem.swift
```diff
@@ -32,14 +36,15 @@ internal final class FormCardSecurityCodeItem: FormTextItem {
...
-        self.isCVCOptional = isCVCOptional
+    internal func updateFormState() {
+        // when optional, if user enters anything it should be validated as regular entry.
+        if isOptional {
+            title = localizedString(.cardCvcItemTitleOptional, localizationParameters)
+            validator = NumericStringValidator(exactLength: 0) || securityCodeValidator
+        } else {
+            title = localizedString(.cardCvcItemTitle, localizationParameters)
+            validator = securityCodeValidator
+        }
```
mohammedDehairy, line 47: Nice!

mohammedDehairy (approved):

descorp (approved):

PR #596, opened 2021-08-12
chore: localization update

erenbesel:
empty strings for keys in a language whose with no value


mohammedDehairy:
can you fix conflicts @erenbesel ?

PR #598, opened 2021-08-17
Create localization target

nauaros:
## Summary
Added a localization target to make it easier for us to update content.

The script run by the target can be found in `Scripts`


mohammedDehairy (approved):

kaphacius:
this is a very cool idea. if this is going to be the preferred way to update localization, should we remove the `Localizable.strings` files from the translations repo? and make the `StringsGenerator.swift` always expect an output path?

kaphacius (approved):

erenbesel (approved):

PR #603, opened 2021-08-23
fix: changed credit card textfield returnKeyType to default

valerio8:
fix: changed credit card textfield returnKeyType to default

We noticed that keyboard type of credit card textfield in drop in component was changed after migratiion from Adyen iOS 3.8.4 to 4.0.0 .
Previously .numberPad keyboard type was displayed. Now the keyboard with .numbersAndPunctuation appears instead of it.

The root issue is the iOS bug which happens if textContentType is set to .creditCardNumber and and returnKeyType is not default.
ReturnKeyType was changed long ago and textContentType was added in 4.0.0 update, that's why it breaks.

My suggestion is to override the returnKeyType for CardNumberItemView with default value. 
Return button is not displayed for .numberPad keyboard so this fix wont affect the UI.

### DIFF
--- a/AdyenCard/Form/FormCardNumberItemView.swift
+++ b/AdyenCard/Form/FormCardNumberItemView.swift
@@ -20,6 +20,7 @@ internal final class FormCardNumberItemView: FormTextItemView<FormCardNumberItem
         super.init(item: item)
         accessory = .customView(cardTypeLogosView)
         textField.textContentType = .creditCardNumber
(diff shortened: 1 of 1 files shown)


descorp (approved):

mohammedDehairy (approved):

descorp:
Good catch @valerio8 

Thanks for contributor! 💚

PR #624, opened 2021-09-13
chore: fix displaying social security field based on config

erenbesel:
Now checks bin lookup response based on config (.auto), instead of always checking binlookup.

### DIFF
--- a/AdyenCard/Components/Card/CardViewController.swift
+++ b/AdyenCard/Components/Card/CardViewController.swift
@@ -143,30 +143,12 @@ internal class CardViewController: FormViewController {
             items.numberItem.showLogos(for: brands.map(\.type))
         }
 
-        let kcpItemsHidden: Bool
-        switch configuration.koreanAuthenticationMode {
-        case .show:
-            kcpItemsHidden = false
-        case .hide:
-            kcpItemsHidden = true
-        case .auto:
-            kcpItemsHidden = !configuration.showAdditionalAuthenticationFields(for: binInfo.issuingCountryCode)
-        }
-        
-        var socialSecurityItemHidden: Bool
-        switch configuration.socialSecurityNumberMode {
-        case .show:
-            socialSecurityItemHidden = false
-        case .hide:
-            socialSecurityItemHidden = true
-        case .auto:
-            socialSecurityItemHidden = !brands.socialSecurityNumberRequired
-        }
+        let kcpItemsHidden = shouldHideKcpItems(with: binInfo.issuingCountryCode)
 
         items.numberItem.validator = CardNumberValidator(isLuhnCheckEnabled: brands.luhnCheckRequired)
         items.additionalAuthPasswordItem.isHidden.wrappedValue = kcpItemsHidden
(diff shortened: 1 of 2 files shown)


descorp (dismissed):

nauaros (approved):

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -143,20 +143,30 @@ internal class CardViewController: FormViewController {
             items.numberItem.showLogos(for: brands.map(\.type))
         }
 
-        let isHidden: Bool
+        let kcpItemsHidden: Bool
```
mohammedDehairy, line 146: `kcpItemsHidden` -> `kcpItemIsHidden`  :-)

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -143,20 +143,30 @@ internal class CardViewController: FormViewController {
...
+            kcpItemsHidden = false
         case .hide:
-            isHidden = true
+            kcpItemsHidden = true
         case .auto:
-            isHidden = !configuration.showAdditionalAuthenticationFields(for: binInfo.issuingCountryCode)
+            kcpItemsHidden = !configuration.showAdditionalAuthenticationFields(for: binInfo.issuingCountryCode)
+        }
+        
+        var socialSecurityItemHidden: Bool
```
mohammedDehairy, line 156: can we move this to a separate `var`? like this:

```Swift
var socialSecurityItemHidden: Bool {
switch configuration.socialSecurityNumberMode {
         case .show:
             socialSecurityItemHidden = false
         case .hide:
             socialSecurityItemHidden = true
         case .auto:
             socialSecurityItemHidden = !brands.socialSecurityNumberRequired
         }
}
```

and also `kcpItemIsHidden`?

AdyenCard/Components/Card/CardViewController.swift
erenbesel: :D If we gotta add 'is', gotta add it to beginning right? 
Since they are local vars, figured it shouldn't matter either one.

AdyenCard/Components/Card/CardViewController.swift
erenbesel: Hmm, okay then I will move them and rename both to 'shouldHide....` syntax. What do you think?

mohammedDehairy: yeah, `shouldHide...` sounds good!

AdyenCard/Components/Card/CardViewController.swift
mohammedDehairy: i thought you forgot the `I`, but its actually `kcp Items Hidden` where Items is plural :-D

AdyenCard/Components/Card/CardViewController.swift
erenbesel: Can't make them vars as they require bin response. Making them functions just for this seems redundant. Shall we just keep it like this then?

mohammedDehairy (approved):

nauaros (approved):

PR #628, opened 2021-09-15
Rename TextField to less common name TextFieldAdyen

Jack25apr:
## Summary
My proposal is to use less common name for custom textfields, i.e. TextField .
Because if you have i.e. **Material** library in your project, they have TextField custom class from UITextField
Adyen also have class named TextField.
I See  XCode error during compiling:
**"'TextField' has different definitions in different modules; first difference is definition in module Adyen.Swift'"**

## Tested scenarios

**Fixed issue**:  627

### DIFF
--- a/Adyen.xcodeproj/project.pbxproj
+++ b/Adyen.xcodeproj/project.pbxproj
@@ -303,7 +303,7 @@
 		E9E3DB062226B3B200697074 /* CoderTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E9E3DB052226B3B200697074 /* CoderTests.swift */; };
 		F90FB7BC2446E8C8005BFE0E /* BrowserInfo.swift in Sources */ = {isa = PBXBuildFile; fileRef = F90FB7BB2446E8C8005BFE0E /* BrowserInfo.swift */; };
 		F90FB7BE2446EBB9005BFE0E /* BrowserInfoTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = F90FB7BD2446EBB9005BFE0E /* BrowserInfoTests.swift */; };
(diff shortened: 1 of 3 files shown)


Jack25apr:
I know that it's mainly nor your problem :) , but what if Apple will make  new SwiftUI struct with very common name TextField...

Jack25apr:
@mohammedDehairy  thanx for reply, 
It fails even without using type, I can use my class vars like
var textFieldA: Adyen.TextField
var textFieldM: Material.TextField

but it fails just on import two modules that have the same class inside, called just TextField.
Let say straight on the AppDelegate, two imports:
import Adyen
import Material

then I got the main error  - "'TextField' has different definitions in different modules..."

mohammedDehairy:
Ok, its a bit weird, since Swift Modules are by definition its own name space.

But I guess we can make an exception for this case, if there is no way around it, we'll discuss it internally and get back to you.

Sorry for the inconsistence!

Jack25apr:
Update:
easiest way to reproduce error default iOS app project with UIKIT AppDelegate and Adyen+Material pods:

//  AppDelegate.swift

import UIKit
**import Material
import Adyen**

@main
class AppDelegate: UIResponder, UIApplicationDelegate
    **class customTextField: Material.TextField { }** // defined subclass on Material lib
    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: 
         [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
        // Override point for customization after application launch.
        return true
    }

even when I'm specifying type using "Module.Class" nottation, error persists.

descorp:
@Jack25apr 

could you check if [this fix](https://github.com/Adyen/adyen-ios/pull/632) works for you?

Jack25apr:
@descorp It works! Hope this fix will be in the next release :)

descorp:
Absolutely! I am going to close this PR then

PR #642, opened 2021-09-24
COIOS-471: Prefill payment components

nauaros:
## Summary

1. Shopper information prefilling for the following payment components:
- Affirm
- MBWay
- Doku
- eContext payment methods

2. On Boleto, changed the prefilling appearance for the billing address.
3. Fixed bug with payment button due to wrong autolayout on container.

**Fixed issue**:  COIOS-471


Adyen/Core/Components/AbstractPersonalInformationComponent/AbstractPersonalInformationComponent.swift
```diff
@@ -210,4 +225,39 @@ open class AbstractPersonalInformationComponent: PaymentComponent, PresentableCo
...
+    /// :nodoc:
+    public func showValidation() {
+        formViewController.showValidation()
+    }
+
+    /// :nodoc:
+    internal func populateFields() {
+        guard let shopperInformation = shopperInformation else { return }
+
+        if let firstName = shopperInformation.shopperName?.firstName {
```
erenbesel, line 237: Do we need if lets for these? 
Trying to understand myself too, what happens if we just go `firstNameItem?.value  = shopperInformation.shopperName?.firstName` ?

nauaros: `value` is non-optional type so we need to unwrap it somehow. I was thinking that maybe something like this may look better instead of having `if let` unwrapping. What do you think?

```swift
shopperInformation.shopperName.map {
    firstNameItem?.value = $0.firstName
    lastNameItem?.value = $0.lastName
}
```

erenbesel: Ah my bad, when I saw items optional, I assumed value is too.
That map looks very nice indeed!

Adyen/UI/Form/Items/Text/FormTextItem.swift
```diff
@@ -46,6 +52,21 @@ open class FormTextItem: FormValueItem<String, FormTextItemStyle>, ValidatableFo
     public func isValid() -> Bool {
         validator?.isValid(value) ?? true
     }
+    
+    /// The formatted text value.
+    @Observable("") internal var formattedValue: String
+
+    // MARK: - Private
+
+    private func publish(newValue: String) {
```
erenbesel, line 61: Shall we rename this to `sanitizeAndPublish` or something as we are altering the `value` in its own `set` ?

nauaros: What do you think about `private func publishSanitized(value: String)`?

Or maybe `private func publishTransformed(value: String)` since it's sanitizing and formatting?

AdyenComponents/Boleto/BoletoComponent.swift
```diff
@@ -196,6 +191,20 @@ public final class BoletoComponent: PaymentComponent, LoadingComponent, Presenta
     }
 }
 
+extension BoletoComponent: ViewControllerDelegate {
+
+    /// :nodoc:
+    public func viewDidLoad(viewController: UIViewController) {}
```
mohammedDehairy, line 197: we should call analytics here, right?

nauaros: Yeah, Boleto is missing it. Let me create a different PR for it

Adyen/Formatters/Formatter.swift
```diff
@@ -7,18 +7,20 @@
...
     
     /// Formats the given value.
     ///
     /// - Parameter value: The value to format.
     /// - Returns: A formatted version of the given value.
     func formattedValue(for value: String) -> String
     
+}
+
+public protocol Sanitizer {
```
kaphacius, line 20: perhaps we can come up with a better name for this, because `sanitizing` means something else in this context, removing illegal characters and/or padding.

AdyenComponents/Boleto/BoletoComponent.swift
```diff
@@ -114,55 +111,54 @@ public final class BoletoComponent: PaymentComponent, LoadingComponent, Presenta
...
+
         return fields
     }
-    
+
     /// :nodoc:
     /// Sets the initial values for the form fields based on configuration
     private func prefillFields(for component: FormComponent) {
-        if let shopperName = configuration.shopperInfo.shopperName {
+        if let shopperName = shopperInformation.shopperName {
```
kaphacius, line 136: can we use `map` here

AdyenDropIn/Utilities/ComponentManager.swift
```diff
@@ -184,15 +184,11 @@ internal final class ComponentManager {
...
-                shopperInfo: configuration.shopper,
-                showEmailAddress: true
-            ),
-            apiContext: apiContext
-        )
+        let boletoConfiguration = BoletoComponent.Configuration(boletoPaymentMethod: paymentMethod,
+                                                                payment: configuration.payment,
+                                                                shopperInformation: configuration.shopper,
+                                                                showEmailAddress: true)
+        return BoletoComponent(configuration: boletoConfiguration, apiContext: apiContext)
```
kaphacius, line 191: can we create the `configuration` when passing it into the `BoletoComponent.init` and make this a one-statement function?

Adyen/Core/Components/AbstractPersonalInformationComponent/NameFormItemInjector.swift
```diff
@@ -26,6 +29,7 @@ internal final class NameFormItemInjector: FormItemInjector, Localizable {
     /// :nodoc:
     internal lazy var item: FormTextInputItem = {
         let item = FormTextInputItem(style: style)
+        item.value = value ?? ""
```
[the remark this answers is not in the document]
nauaros: Now, `NameFormItemInjector` holds a value property (optional type) to enable the name's prefilling. Since the item's `value` is a non-optional type, I need to provide a fallback value if the prefilled information doesn't contain a first name.

It's the same default value we're already using in the parent class `FormTextItem`:
```swift
public init(style: FormTextItemStyle) {
    super.init(value: "", style: style)
}
```

Adyen/Core/Components/AbstractPersonalInformationComponent/PhoneFormItemInjector.swift
```diff
@@ -25,13 +28,16 @@ internal final class PhoneFormItemInjector: FormItemInjector, Localizable {
         let item = FormPhoneNumberItem(selectableValues: phoneExtensions,
                                        style: style,
                                        localizationParameters: localizationParameters)
+        item.value = value ?? ""
```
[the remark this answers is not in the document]
nauaros: Same reason as the other comment

Adyen/Core/Components/Base/PresentableComponent.swift
```diff
@@ -74,4 +74,7 @@ extension TrackableComponent {
 
     /// :nodoc:
     public func viewDidAppear(viewController: UIViewController) { /* Empty Implementation */ }
+
+    /// :nodoc:
+    public func viewWillAppear(viewController: UIViewController) {}
```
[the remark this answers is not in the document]
nauaros: Done

AdyenComponents/Boleto/BoletoComponent.swift
nauaros: Sure

AdyenDropIn/Utilities/ComponentManager.swift
nauaros: You mean like this:

```swift
    private func createBoletoComponent(_ paymentMethod: BoletoPaymentMethod) -> BoletoComponent {
        BoletoComponent(configuration: .init(boletoPaymentMethod: paymentMethod,
                                             payment: configuration.payment,
                                             shopperInformation: configuration.shopper,
                                             showEmailAddress: true),
                        apiContext: apiContext)
    }
```

Adyen/Formatters/Formatter.swift
nauaros: Any suggestion here? Cause this is the term we're using all across the code

AdyenDropIn/Utilities/ComponentManager.swift
kaphacius: yes, exactly like this 👍

Adyen/Formatters/Formatter.swift
kaphacius: I think in some cases we are using the term correctly, if you check `NumericFormatter.swift` line 16, we are actually calling `sanitizeValue` from `formattedValue`. But unformatting is something else. Maybe `Unformatter` or `Restorer`?

mohammedDehairy: sanitizing means indeed removing illegal characters, and that's what it does, right?.

kaphacius: actually the comment on lines 21-24 is wrong, but the name of the protocol is good.

nauaros: Updated the documentation to fix the confusion

kaphacius (approved):

erenbesel (approved):

PR #657, opened 2021-10-04
chore: disables some unit tests on iOS 12 and below

mohammedDehairy:
chore: disables some unit tests on iOS 12 and below, since they are failing for weird reasons

### DIFF
--- a/Tests/AdyenTests/Adyen Tests/Components/Apple Pay/ApplePayComponentTests.swift
+++ b/Tests/AdyenTests/Adyen Tests/Components/Apple Pay/ApplePayComponentTests.swift
@@ -291,4 +291,12 @@ enum Available {
             return false
         }
     }
+    
+    static var iOS13: Bool {
+        if #available(iOS 13.0, *) {
+            return true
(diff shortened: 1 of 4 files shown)


Tests/AdyenTests/Adyen Tests/Components/Apple Pay/ApplePayComponentTests.swift
```diff
@@ -291,4 +291,12 @@ enum Available {
             return false
         }
     }
+
```
descorp, line 294: How about we move `Available` enum into it's own file?

erenbesel (approved):

Tests/AdyenTests/Adyen Tests/Components/Apple Pay/ApplePayComponentTests.swift
mohammedDehairy: done 👍

nauaros (approved):

descorp (approved):

PR #660, opened 2021-10-06
Allow ApplePay to run onboarding when no supported network

descorp:
### Changes

* add `allowOnboarding` flag to ApplePayConfig that allows bypassing supported networks check on ApplePayComponent to open ApplePay onboarding flow

### DIFF
--- a/AdyenComponents/Apple Pay/ApplePayComponentExtensions.swift
+++ b/AdyenComponents/Apple Pay/ApplePayComponentExtensions.swift
@@ -64,6 +64,10 @@ extension ApplePayComponent {
         /// A prepopulated billing address.
         public var billingContact: PKContact?
 
+        /// The flag to toggle available cards in the whallet check.
+        /// When `true` - will ignore networks check and allow fallback to "Apple Pay card onboarding".
+        public var allowOnboarding: Bool
+
         /// Initializes the configuration.
         ///
         /// - Parameter summaryItems: The line items for this payment.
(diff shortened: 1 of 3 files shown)


AdyenComponents/Apple Pay/ApplePayComponentExtensions.swift
```diff
@@ -64,6 +64,10 @@ extension ApplePayComponent {
         /// A prepopulated billing address.
         public var billingContact: PKContact?
 
+        /// The flag to toggle available cards in the whallet check.
+        /// When `true` - will ignore networks check and allow fallback to "Apple Pay card onboarding".
```
mohammedDehairy, line 68: the comment doesn't make much sense to me, maybe something like this:

```
/// if true, allow the shopper to add cards to Apple Pay if non exists yet.
/// if false, then Apple Pay is disabled if the shopper doesn't have cards on Apple Pay.
```

AdyenComponents/Apple Pay/ApplePayComponentExtensions.swift
```diff
@@ -74,16 +78,19 @@ extension ApplePayComponent {
         /// A list of fields that you need for a shipping contact in order to process the transaction. Ignored on iOS 10.*.
         /// - Parameter requiredShippingContactFields: The excluded card brands.
         /// - Parameter billingContact: A prepopulated billing address.
+        /// - Parameter allowOnboarding: The flag to toggle available cards in the whallet check.
```
mohammedDehairy, line 81: same here

PR #683, opened 2021-11-01
Delete stored payment methods

mohammedDehairy:
Delete stored payment methods

### DIFF
--- a/Adyen/Core/Payment Methods/PaymentMethod.swift
+++ b/Adyen/Core/Payment Methods/PaymentMethod.swift
@@ -56,5 +56,13 @@ public protocol StoredPaymentMethod: PaymentMethod {
 
 /// :nodoc:
 public func == (lhs: StoredPaymentMethod, rhs: StoredPaymentMethod) -> Bool {
-    lhs.identifier == rhs.identifier && lhs.supportedShopperInteractions == rhs.supportedShopperInteractions
+    lhs.type == rhs.type &&
+        lhs.name == rhs.name &&
+        lhs.identifier == rhs.identifier &&
+        lhs.supportedShopperInteractions == rhs.supportedShopperInteractions
+}
+
+/// :nodoc:
+public func == (lhs: PaymentMethod, rhs: PaymentMethod) -> Bool {
+    lhs.type == rhs.type && lhs.name == rhs.name
 }
--- a/Adyen/Core/Payment Methods/PaymentMethod.swift
+++ b/Adyen/Core/Payment Methods/PaymentMethod.swift
@@ -53,3 +53,8 @@ public protocol StoredPaymentMethod: PaymentMethod {
     var supportedShopperInteractions: [ShopperInteraction] { get }
     
 }
+
+/// :nodoc:
+public func == (lhs: StoredPaymentMethod, rhs: StoredPaymentMethod) -> Bool {
+    lhs.identifier == rhs.identifier && lhs.supportedShopperInteractions == rhs.supportedShopperInteractions
+}
--- a/Adyen/Core/Payment Methods/PaymentMethods.swift
+++ b/Adyen/Core/Payment Methods/PaymentMethods.swift
@@ -18,10 +18,10 @@ public struct PaymentMethods: Decodable {
     public var paid: [PaymentMethod] = []
     
     /// The regular payment methods.
-    public let regular: [PaymentMethod]
+    public var regular: [PaymentMethod]
     
     /// The stored payment methods.
-    public let stored: [StoredPaymentMethod]
+    public var stored: [StoredPaymentMethod]
     
     /// Initializes the PaymentMethods.
     ///
--- a/Adyen/UI/List/CoreDataSource.swift
+++ b/Adyen/UI/List/CoreDataSource.swift
@@ -8,7 +8,7 @@ import UIKit
 
 /// :nodoc:
 /// A `ListViewController` data source.
-internal protocol ListViewControllerDataSource: NSObject, UITableViewDataSource {
+internal protocol ListViewControllerDataSource: UITableViewDataSource {
     
     /// :nodoc:
     /// All sections data.
@@ -20,11 +20,11 @@ internal protocol ListViewControllerDataSource: NSObject, UITableViewDataSource 
     
     /// :nodoc:
     /// Reloads all data.
-    func reload(newSections: [ListSection], tableView: UITableView)
+    func reload(newSections: [ListSection], tableView: UITableView, animated: Bool)
     
     /// :nodoc:
     /// Deletes an item at a certain `IndexPath`.
-    func deleteItem(at indexPath: IndexPath, tableView: UITableView)
+    func deleteItem(at indexPath: IndexPath, tableView: UITableView, animated: Bool)
     
     /// :nodoc:
     /// Creates a new `UITableViewCell` at a certain `IndexPath`.
@@ -96,12 +96,16 @@ internal final class CoreDataSource: NSObject, ListViewControllerDataSource {
         return cell
     }
     
-    internal func reload(newSections: [ListSection], tableView: UITableView) {
+    internal func reload(newSections: [ListSection],
+                         tableView: UITableView,
+                         animated: Bool = false) {
         sections = newSections.filter { $0.items.count > 0 }
         tableView.reloadData()
     }
     
-    internal func deleteItem(at indexPath: IndexPath, tableView: UITableView) {
+    internal func deleteItem(at indexPath: IndexPath,
+                             tableView: UITableView,
+                             animated: Bool = true) {
         sections.deleteItem(at: indexPath)
         tableView.reloadData()
     }
--- a/Adyen/UI/List/CoreDataSource.swift
+++ b/Adyen/UI/List/CoreDataSource.swift
@@ -14,6 +14,8 @@ internal protocol ListViewControllerDataSource: NSObject, UITableViewDataSource 
     
     func reload(newSections: [ListSection], tableView: UITableView)
     
+    func deleteItem(at indexPath: IndexPath, tableView: UITableView)
+    
     func cell(for tableView: UITableView, at indexPath: IndexPath) -> UITableViewCell
     
     func startLoading(for item: ListItem, _ tableView: UITableView)
@@ -62,7 +64,7 @@ internal final class CoreDataSource: NSObject, ListViewControllerDataSource {
             self?.stopLoading(tableView)
         }
         
-        deletionHandler(completion)
+        deletionHandler(indexPath, completion)
     }
     
     // MARK: - ListViewControllerDataSource
@@ -79,6 +81,11 @@ internal final class CoreDataSource: NSObject, ListViewControllerDataSource {
     
     public func reload(newSections: [ListSection], tableView: UITableView) {
         sections = newSections.filter { $0.items.count > 0 }
+        tableView.reloadData()
+    }
+    
+    public func deleteItem(at indexPath: IndexPath, tableView: UITableView) {
+        sections[indexPath.section].deleteItem(index: indexPath.item)
         tableView.reloadData()
     }
     
--- a/Adyen/UI/List/CoreDataSource.swift
+++ b/Adyen/UI/List/CoreDataSource.swift
@@ -4,7 +4,7 @@
 // This file is open source and available under the MIT license. See the LICENSE file for more info.
 //
 
-import Foundation
+import UIKit
 
 internal protocol ListViewControllerDataSource: NSObject, UITableViewDataSource {
     
--- a/Adyen/UI/List/CoreDataSource.swift
+++ b/Adyen/UI/List/CoreDataSource.swift
@@ -8,7 +8,7 @@ import Foundation
 
 internal protocol ListViewControllerDataSource: NSObject, UITableViewDataSource {
     
-    var sections: [ListSection] { get set }
+    var sections: [ListSection] { get }
     
     var cellReuseIdentifier: String { get }
     
@@ -48,7 +48,8 @@ internal final class CoreDataSource: NSObject, ListViewControllerDataSource {
     
     /// :nodoc:
     internal func tableView(_ tableView: UITableView, canEditRowAt indexPath: IndexPath) -> Bool {
-        sections[indexPath.section].editingStyle != .none
+        guard let header = sections[indexPath.section].header else { return false }
+        return header.editingStyle != .none
     }
     
     /// :nodoc:
@@ -79,13 +80,13 @@ internal final class CoreDataSource: NSObject, ListViewControllerDataSource {
         return cell
     }
     
-    public func reload(newSections: [ListSection], tableView: UITableView) {
+    internal func reload(newSections: [ListSection], tableView: UITableView) {
         sections = newSections.filter { $0.items.count > 0 }
         tableView.reloadData()
     }
     
-    public func deleteItem(at indexPath: IndexPath, tableView: UITableView) {
-        sections[indexPath.section].deleteItem(index: indexPath.item)
+    internal func deleteItem(at indexPath: IndexPath, tableView: UITableView) {
+        sections.deleteItem(at: indexPath)
         tableView.reloadData()
     }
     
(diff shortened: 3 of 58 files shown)


Adyen/UI/List/CoreDataSource.swift
```diff
@@ -0,0 +1,127 @@
...
+    func startLoading(for item: ListItem, _ tableView: UITableView)
+    
+    func stopLoading(_ tableView: UITableView)
+}
+
+internal final class CoreDataSource: NSObject, ListViewControllerDataSource {
+    
+    internal var sections: [ListSection] = []
+    
+    internal let cellReuseIdentifier = "Cell"
```
descorp, line 30: I guess it is ok for now, but in future would be nice to support multiple identifiers for "Instant payment method" cells.

Adyen/UI/List/ListSection.swift
```diff
@@ -6,15 +6,20 @@
...
 /// :nodoc:
-public struct ListSection {
+public struct ListSection: Hashable {
     
     /// The title of the section.
     public let header: ListSectionHeader?
 
     /// The items inside the section.
-    public let items: [ListItem]
+    public private(set) var items: [ListItem]
```
descorp, line 22: Nice!

Adyen/UI/List/ListViewController.swift
```diff
@@ -4,6 +4,7 @@
 // This file is open source and available under the MIT license. See the LICENSE file for more info.
 //
 
+import SwiftUI
```
nauaros, line 7: Why are we importing `SwiftUI` here?

mohammedDehairy: nice catch, removed

descorp (dismissed):

Adyen/UI/List/ListHeaderView.swift
```diff
@@ -28,23 +28,84 @@ internal final class ListHeaderView: UIView {
...
+            updateTrailingButtonTitle(with: headerItem)
+        }
+    }
+    
+    private func updateTrailingButtonTitle(with item: ListSectionHeader?) {
+        guard let item = headerItem else { return }
+        let localizedEdit = Bundle.Adyen.localizedEditCopy
+        let localizedDone = Bundle.Adyen.localizedDoneCopy
+        let localizedTitle = isEditing ? localizedDone : localizedEdit
+        switch item.editingStyle {
```
descorp, line 62: What if merchant wants to change this label or use custom translation?

AdyenDropIn/Components/PaymentMethodListComponent.swift
```diff
@@ -38,14 +45,34 @@ internal final class PaymentMethodListComponent: ComponentLoader, PresentableCom
...
     
     /// :nodoc:
     internal var viewController: UIViewController { listViewController }
 
     private let brandProtectedComponents: Set = ["applepay"]
     
-    internal lazy var listViewController: ListViewController = {
+    internal lazy var listViewController: ListViewController = createListViewController()
+    
+    private func createListViewController() -> ListViewController {
```
descorp, line 67: Why not just:

```
lazy var listViewController: ListViewController = { .. }()

File: Adyen/UI/List/ListHeaderView.swift (hunk 4 above)
Comment (mohammedDehairy) [reply Adyen/UI/List/ListHeaderView.swift:62]: as discussed offline, we don't support forced locale at the moment and its not likely merchant would want to override those translations.

File: AdyenDropIn/Components/PaymentMethodListComponent.swift (hunk 5 above)
Comment (mohammedDehairy) [reply AdyenDropIn/Components/PaymentMethodListComponent.swift:67]: for reusability? :-)

File: AdyenDropIn/Components/PaymentMethodListComponent.swift [hunk 6]
```diff
@@ -56,26 +83,38 @@ internal final class PaymentMethodListComponent: ComponentLoader, PresentableCom
...
             listItem.trailingText = displayInformation.disclosureText
             listItem.subtitle = displayInformation.subtitle
-            listItem.selectionHandler = { [unowned self, unowned component] in
+            listItem.selectionHandler = { [weak self, weak component] in
+                guard let self = self, let component = component else { return }
                 guard !(component is AlreadyPaidPaymentComponent) else { return }
                 self.delegate?.didSelect(component, in: self)
             }
             
+            listItem.deletionHandler = { [weak self, weak component] indexPath, completion in
```
descorp, line 92: this handler is big enough, maybe we can move it to a standalone method?

AdyenDropIn/Components/PaymentMethodListComponent.swift
descorp: Do we reuse it?
It is a private function that is only used as for lazy loading 🤔

nauaros (approved):

AdyenDropIn/Components/PaymentMethodListComponent.swift
mohammedDehairy: extracted into a function

AdyenDropIn/Components/PaymentMethodListComponent.swift
mohammedDehairy: it was :-D, I think its nicer this way

descorp (approved):

AdyenDropIn/Components/PaymentMethodListComponent.swift
mohammedDehairy: as discussed offline :-)

nauaros (approved):

Adyen/UI/List/CoreDataSource.swift
```diff
@@ -0,0 +1,143 @@
...
+        }
+        
+        deletionHandler(indexPath, completion)
+    }
+    
+    // MARK: - ListViewControllerDataSource
+    
+    internal func cell(for tableView: UITableView, at indexPath: IndexPath) -> UITableViewCell {
+        guard let cell = tableView.dequeueReusableCell(withIdentifier: cellReuseIdentifier, for: indexPath) as? ListCell else {
+            fatalError("Failed to dequeue cell.")
```
kaphacius, line 91: maybe something less drastic? 
an `assert` and `return UITableViewCell()`

Adyen/UI/List/CoreDataSource.swift
```diff
@@ -0,0 +1,143 @@
...
+    /// :nodoc:
+    /// Starts a loading animation for a given ListItem.
+    func startLoading(for item: ListItem, _ tableView: UITableView)
+    
+    /// :nodoc:
+    /// Stops all loading animations.
+    func stopLoading(_ tableView: UITableView)
+}
+
+internal final class CoreDataSource: NSObject, ListViewControllerDataSource {
```
kaphacius, line 42: Maybe the name could be better. `CoreData` usually means something different in iOS dev context :)

Adyen/UI/List/CoreDataSource.swift
```diff
@@ -0,0 +1,143 @@
...
+        
+        cell.item = sections[indexPath.section].items[indexPath.row]
+        
+        return cell
+    }
+    
+    internal func reload(newSections: [ListSection],
+                         tableView: UITableView,
+                         animated: Bool = false) {
+        sections = newSections.filter { $0.items.count > 0 }
```
kaphacius, line 102: maybe better here to use `{ $0.items.isEmpty == false }`

Adyen/UI/List/CoreDataSource.swift
```diff
@@ -0,0 +1,143 @@
...
+        
+        tableView.isUserInteractionEnabled = false
+        
+        for case let visibleCell as ListCell in tableView.visibleCells where visibleCell.item != item {
+            visibleCell.isEnabled = false
+        }
+    }
+    
+    private func cell(for item: ListItem, tableView: UITableView) -> ListCell? {
+        for case let cell as ListCell in tableView.visibleCells where cell.item == item {
```
kaphacius, line 128: very cool! 👍

Adyen/UI/List/DiffableDataSource.swift
```diff
@@ -0,0 +1,118 @@
...
+    
+    internal func stopLoading(_ tableView: UITableView) {
+        coreDataSource.stopLoading(tableView)
+    }
+    
+}
+
+extension Array where Element == ListSection {
+    internal var isEditable: Bool {
+        first(where: { $0.header?.editingStyle != EditinStyle.none }) != nil
```
kaphacius, line 111: maybe better to use `contains(where: { $0.header?.editingStyle != EditinStyle.none })`
also, not sure about the naming. `isEditable` referring to the `Array` object? maybe it can be a property on `ListSection` type
```swift
var isEditable: Bool {
    header?.editingStyle != EditinStyle.none
}
...
guard sections.contains(\.isEditable) == false else { return }

```

Adyen/UI/List/DiffableDataSource.swift
```diff
@@ -0,0 +1,118 @@
...
+    }
+    
+    internal mutating func deleteItem(at indexPath: IndexPath) {
+        self[indexPath.section].deleteItem(index: indexPath.item)
+        self = self.filter { !$0.items.isEmpty }
```
kaphacius, line 116: probably `$0.items.isEmpty == false` here :)

Adyen/UI/List/ListSection.swift
```diff
@@ -6,15 +6,20 @@
 
 import Foundation
 
+public enum EditinStyle {
```
kaphacius, line 9: `Editing*`

AdyenDropIn/Components/PaymentMethodListComponent.swift
```diff
@@ -8,14 +8,21 @@ import Adyen
 import Foundation
 import UIKit
 
+/// Payment methods list related configurations.
+public struct PaymentMethodListConfiguration {
+    
+    /// Indicates whether to allow shoppers to disable/delete stored payment methods
+    public var allowDisablingStorePaymentMethods: Bool = false
```
kaphacius, line 15: `Stored*`

AdyenDropIn/Components/PaymentMethodListComponent.swift
```diff
@@ -38,14 +45,34 @@ internal final class PaymentMethodListComponent: ComponentLoader, PresentableCom
...
+    private func createListViewController() -> ListViewController {
+        let listViewController = ListViewController(style: style)
+        listViewController.title = localizedString(.paymentMethodsTitle, localizationParameters)
+        listViewController.reload(newSections: createListSections())
+        
+        return listViewController
+    }
+    
+    private func createListSections() -> [ListSection] {
         func item(for component: PaymentComponent) -> ListItem {
```
kaphacius, line 76: can this function be moved outside? a bit unusual to see in this project. then we can also drop `return` at line 99

AdyenDropIn/Components/PaymentMethodListComponent.swift
```diff
@@ -121,4 +165,22 @@ internal protocol PaymentMethodListComponentDelegate: AnyObject {
...
+    func didDelete(_ paymentMethod: StoredPaymentMethod,
+                   in paymentMethodListComponent: PaymentMethodListComponent,
+                   completion: @escaping Completion<Bool>)
+    
+}
+
+private extension Array where Element == ComponentsSection {
+    mutating func deleteItem(at indexPath: IndexPath) {
+        self[indexPath.section].components.remove(at: indexPath.item)
+        self = self.filter { !$0.components.isEmpty }
```
kaphacius, line 184: maybe `{ $0.components.isEmpty == false }`?

AdyenDropIn/DropInComponentDelegate.swift
```diff
@@ -59,6 +59,16 @@ public protocol DropInComponentDelegate: AnyObject {
     
 }
 
+public protocol StoredPaymentMethodsDelegate: AnyObject {
+    /// Invoked when shopper want to delete a stored payment method.
```
kaphacius, line 63: `wants*`, also in line 66

AdyenDropIn/Utilities/ComponentManager.swift
```diff
@@ -61,8 +65,21 @@ internal final class ComponentManager {
                                             footer: paidFooter)
 
         // Stored section
-        let storedSection = ComponentsSection(components: storedComponents)
-
+        let storedSection: ComponentsSection
+        
+        if supportsEditingStoredPaymentMethods {
+            let allowDeleting = configuration.paymentMethodsList.allowDisablingStorePaymentMethods
```
kaphacius, line 71: do deleting and disabling mean the same in this context?

Adyen/Core/Payment Methods/PaymentMethods.swift
```diff
@@ -18,10 +18,10 @@ public struct PaymentMethods: Decodable {
     public var paid: [PaymentMethod] = []
     
     /// The regular payment methods.
-    public let regular: [PaymentMethod]
+    public var regular: [PaymentMethod]
```
kaphacius, line 21: can this still be a `let`? does not look like we are changing it anywhere

kaphacius (commented):
sorry, a little late to the party

Adyen/Core/Payment Methods/PaymentMethod.swift
```diff
@@ -53,3 +53,16 @@ public protocol StoredPaymentMethod: PaymentMethod {
...
+/// :nodoc:
+public func == (lhs: StoredPaymentMethod, rhs: StoredPaymentMethod) -> Bool {
+    lhs.type == rhs.type &&
+        lhs.name == rhs.name &&
+        lhs.identifier == rhs.identifier &&
+        lhs.supportedShopperInteractions == rhs.supportedShopperInteractions
+}
+
+/// :nodoc:
+public func == (lhs: PaymentMethod, rhs: PaymentMethod) -> Bool {
```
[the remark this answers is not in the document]
mohammedDehairy: that's a rather philosophical question, it will never happen in real life that two different subtypes of `PaymentMethod` will have the same name and type.

Adyen/UI/List/ListSection.swift
```diff
@@ -25,17 +30,39 @@ public struct ListSection {
...
     ///   - items: The items inside the section.
     ///   - footer: The section footer.
-    public init(header: ListSectionHeader? = nil, items: [ListItem], footer: ListSectionFooter? = nil) {
+    public init(header: ListSectionHeader? = nil,
+                items: [ListItem],
+                footer: ListSectionFooter? = nil) {
         self.header = header
         self.items = items
         self.footer = footer
+        self.identifier = UUID().uuidString
```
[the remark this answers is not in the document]
mohammedDehairy: sure we can, but I see no difference at all

Adyen/UI/List/ListSection.swift
```diff
@@ -25,17 +30,39 @@ public struct ListSection {
...
+    
+    private let identifier: String
+    
+    internal mutating func deleteItem(index: Int) {
+        guard items.indices.contains(index) else { return }
+        items.remove(at: index)
+    }
+    
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(identifier)
```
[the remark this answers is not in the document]
mohammedDehairy: the identifier is when header and footer are nil, but the identifier alone is not enough, this is safer, in case for example we change it so that the header is changed to being mutable and someone change the header after creation. so right now both approaches are identical, but by adding the header and footer its more future proof. and more accurate tbh.

using only the identifier is kind of a hack.

Adyen/UI/List/ListSection.swift
```diff
@@ -25,17 +30,39 @@ public struct ListSection {
...
+        hasher.combine(header)
+        hasher.combine(footer)
+    }
+    
+    public static func == (lhs: ListSection, rhs: ListSection) -> Bool {
+        lhs.header == rhs.header &&
```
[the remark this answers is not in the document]
mohammedDehairy: same as [here](https://github.com/Adyen/adyen-ios/pull/683/files#r742773808)

AdyenDropIn/Utilities/ComponentManager.swift
mohammedDehairy: yes

Adyen/Core/Payment Methods/PaymentMethod.swift
```diff
@@ -53,3 +53,16 @@ public protocol StoredPaymentMethod: PaymentMethod {
     var supportedShopperInteractions: [ShopperInteraction] { get }
     
 }
+
+/// :nodoc:
+public func == (lhs: StoredPaymentMethod, rhs: StoredPaymentMethod) -> Bool {
+    lhs.type == rhs.type &&
```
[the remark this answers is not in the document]
mohammedDehairy: identifier comes from the backend, so assuming its unique is probably true, but its still an assumption, and I see no harm in adding the type and name.

PR #712, opened 2021-11-18
allow dual branding only after binlookup

erenbesel:


descorp (approved):

nauaros (approved):
Nice!

AdyenCard/Utilities/BinLookup/BinLookupResponse.swift
```diff
@@ -20,13 +20,18 @@ internal struct BinLookupResponse: Response {
...
 
+    /// :nodoc:
+    internal var isCreatedLocally: Bool = false
+    
     /// :nodoc:
     internal init(brands: [CardBrand]? = nil,
                   requestId: String = UUID().uuidString,
-                  issuingCountryCode: String? = "NL") {
+                  issuingCountryCode: String? = "NL",
+                  isCreatedLocally: Bool = true) {
```
mohammedDehairy, line 30: it should be set to false when coming from bin lookup, right? I don't see where its being set to false?

erenbesel: When from binlookup, this init won't be called. And the default value is `false`. Should be ok, no?

mohammedDehairy: ah, I see, its fine, but its a bit unclear that that's the case from first glance :-)

PR #713, opened 2021-11-19
fix: add acceptHeader to browserInfo

descorp:
### Changes

* add `acceptHeader` to BrowserInfo

### DIFF
--- a/Adyen/Model/BrowserInfo.swift
+++ b/Adyen/Model/BrowserInfo.swift
@@ -12,6 +12,9 @@ public struct BrowserInfo: Encodable {
     
     /// The device default user-agent.
     public var userAgent: String?
+
+    /// The Accept request HTTP header indicates which content types, expressed as MIME types, the client is able to understand.
+    public var acceptHeader: String { "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8" }
     
     /// Initializes a `BrowserInfo` instance asynchronously.
     ///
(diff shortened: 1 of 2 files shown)


Adyen/Model/BrowserInfo.swift
```diff
@@ -12,6 +12,9 @@ public struct BrowserInfo: Encodable {
     
     /// The device default user-agent.
     public var userAgent: String?
+
+    /// The Accept request HTTP header indicates which content types, expressed as MIME types, the client is able to understand.
+    public var acceptHeader: String { "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8" }
```
mohammedDehairy, line 17: why is this a computed property?

also where did you get this value?

mohammedDehairy:
spell check is failing because of `html,application` :-), you can add it to `.typokana_ignore` file as an exception.

Adyen/Model/BrowserInfo.swift
mohammedDehairy, line 17: spell check is failing because of `html,application` :-), you can add it to `.typokana_ignore` file as an exception.

Adyen/Model/BrowserInfo.swift
```diff
@@ -32,10 +35,21 @@ public struct BrowserInfo: Encodable {
             completion(BrowserInfo(userAgent: cachedUserAgent))
         }
     }
+
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(acceptHeader, forKey: .acceptHeader)
+        try container.encodeIfPresent(userAgent, forKey: .userAgent)
+    }
```
mohammedDehairy, line 43: why is this needed? the default implementation should work, right?

Adyen/Model/BrowserInfo.swift
descorp: 👍

Adyen/Model/BrowserInfo.swift
descorp: Same as [previous](https://github.com/Adyen/adyen-ios/pull/713#discussion_r754027972) question:
For some reason it crashes 🤷‍♂️

PR #716, opened 2021-11-23
Option to skip payment list on dropin when single regular method

erenbesel:
## Summary
Added a configuration flag to allow skipping payment list on drop in when there is only one non-instant payment method.

### DIFF
--- a/AdyenDropIn/Utilities/ComponentManager.swift
+++ b/AdyenDropIn/Utilities/ComponentManager.swift
@@ -102,6 +102,21 @@ internal final class ComponentManager {
 
     internal lazy var paidComponents = paymentMethods.paid.compactMap(component(for:))
     
+    /// Returns the only regular component that is not an instant payment,
+    /// when no other payment method exists.
+    internal var singleRegularComponent: (PaymentComponent & PresentableComponent)? {
+        guard storedComponents.isEmpty,
+              paidComponents.isEmpty,
+              regularComponents.count == 1,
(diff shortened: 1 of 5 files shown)


AdyenDropIn/Utilities/ComponentManager.swift
```diff
@@ -102,6 +102,21 @@ internal final class ComponentManager {
 
     internal lazy var paidComponents = paymentMethods.paid.compactMap(component(for:))
     
+    /// Returns the only regular component that is not an instant payment,
+    /// when no other payment method exists.
+    internal var singleRegularComponent: (PaymentComponent & PresentableComponent)? {
+        guard storedComponents.isEmpty,
+              paidComponents.isEmpty,
+              regularComponents.count == 1,
+              let regularComponent = regularComponents.first else { return nil }
```
mohammedDehairy, line 111: you can fail early here, like this:

```Swift
guard storedComponents.isEmpty,
               paidComponents.isEmpty,
               regularComponents.count == 1,
               let regularComponent = regularComponents.first as? (PaymentComponent & PresentableComponent) else {
 return nil
}
return regularComponent
```

no need to check `InstantPaymentComponent ` since its not a `PresentableComponent ` :-)

erenbesel: Ahh ok, good to know

mohammedDehairy (approved):

nauaros (approved):

PR #736, opened 2021-12-17
feat: add `allowPreselectedPaymentView` to Drop-in

descorp:
### Changes

- allow skipping `PreselectedPaymentMethod` view

### DIFF
--- a/AdyenDropIn/DropInComponent.swift
+++ b/AdyenDropIn/DropInComponent.swift
@@ -158,7 +158,8 @@ public final class DropInComponent: NSObject, PresentableComponent {
     }
     
     internal lazy var rootComponent: PresentableComponent = {
-        if let preselectedComponents = componentManager.storedComponents.first {
+        if configuration.allowPreselectedPaymentView,
+           let preselectedComponents = componentManager.storedComponents.first {
             return preselectedPaymentMethodComponent(for: preselectedComponents, onCancel: nil)
         } else if configuration.allowsSkippingPaymentList,
                   let singleRegularComponent = componentManager.singleRegularComponent {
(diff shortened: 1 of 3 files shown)


Tests/AdyenTests/Adyen Tests/DropIn/DropInTests.swift
```diff
@@ -168,6 +168,25 @@ class DropInTests: XCTestCase {
...
+
+        let paymentMethods = try! JSONDecoder().decode(PaymentMethods.self, from: DropInTests.paymentMethodsOneClick.data(using: .utf8)!)
+        sut = DropInComponent(paymentMethods: paymentMethods, configuration: config)
+
+        let root = UIViewController()
+        UIApplication.shared.keyWindow?.rootViewController = root
+        root.present(sut.viewController, animated: true, completion: nil)
+
+        let waitExpectation = expectation(description: "Expect DropIn to open")
+        DispatchQueue.main.asyncAfter(deadline: DispatchTime.now() + .seconds(2)) {
```
mohammedDehairy, line 182: you can just use `wait(for: .seconds(2))`

AdyenDropIn/DropInComponent.swift
```diff
@@ -158,7 +158,8 @@ public final class DropInComponent: NSObject, PresentableComponent {
     }
     
     internal lazy var rootComponent: PresentableComponent = {
-        if let preselectedComponents = componentManager.storedComponents.first {
+        if configuration.allowPreselectedPaymentView,
+           let preselectedComponents = componentManager.storedComponents.first {
```
mohammedDehairy, line 162: but then if `configuration.allowPreselectedPaymentView` is false, you need to show all stored payment methods in the second payment method list, right?

descorp: Of cause, all stored payment methods are there:
This only allows to skip `PreselectedPaymentMethodComponent` and goes directly to a `PaymentMethodListComponent`.

Tests/AdyenTests/Adyen Tests/DropIn/DropInTests.swift
descorp: 👍

mohammedDehairy (approved):

nauaros (approved):
Nice!

PR #746, opened 2021-12-23
chore: flag to disable edit actions on textfields

erenbesel:


AdyenCard/Form/FormCardNumberItemView.swift
```diff
@@ -21,6 +21,7 @@ internal final class FormCardNumberItemView: FormTextItemView<FormCardNumberItem
         accessory = .customView(detectedBrandsView)
         textField.textContentType = .creditCardNumber
         textField.returnKeyType = .default
+        textField.allowsEditingActions = false
```
descorp, line 24: maybe also add it to CVV field ?

descorp (approved):

nauaros (approved):

Issue #747, opened 2021-12-28
[BUG] On iOS 15 library in version 2.5.0 closing animation action is caller sooner then payment event


MNiemiec-Apple:
**Describe the bug**
On iOS 15 with library 2.5.0 when customer pay using Apple Pay, closing animation event is caller sooner then payment event.

**Expected behavior**
On iOS 15 library should work the same as on iOS 14.

**Smartphone (please complete the following information):**
 - Device: iPhone 
 - OS: iOS 15.1
 - SDK Version 2.5.0

https://user-images.githubusercontent.com/64201153/147552262-d286c7aa-9286-42e5-ada9-a3656f453008.mp4

I added my fix here
https://github.com/MNiemiec-Apple/adyen-ios/tree/fix-for-apple-pay

descorp:
Hey @MNiemiec-Apple 

I am not sure what was the issue on a video, but from your fork, I can tell that you have added delay before Drop-in `finishAndDismiss`. 
From a UX perspective, some consider `0.5 sec` delay "too long for iOS", so If you are looking to create a PR - I would suggest providing delay length as a parameter.

Unfortunately, v2 is discontinued - we can review your PR and add your fix to a repo, but it is unlikely that we are going to release a new v2 version. 

---

Have you considered updating to a v4?

MNiemiec-Apple:
Hello,
Yes I have to add this delay to get first event that payment was done correctly.
After this fix app is working fine and going to another screen.
I know that v2 it’s not supporting any more and I’ll create PR with parameter.

descorp:
Hey @MNiemiec-Apple 
This was fixed and released on v4.9.0

descorp:
I'm closing this ticket due to lack of activity. Feel free to continue here

PR #748, opened 2021-12-29
Fix for apple pay

MNiemiec-Apple:
## Tested scenarios
Pay with Apple Pay on iOS 14 and iOS 15.

**Fixed issue**:  #747

### DIFF
--- a/Adyen.xcodeproj/project.xcworkspace/contents.xcworkspacedata
+++ b/Adyen.xcodeproj/project.xcworkspace/contents.xcworkspacedata
@@ -2,6 +2,6 @@
 <Workspace
    version = "1.0">
    <FileRef
-      location = "self:Adyen.xcodeproj">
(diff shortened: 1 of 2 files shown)


descorp:
Hey @MNiemiec-Apple 

Indeed there is an issue with our current ApplePayComponent - it get's dismissed without waiting for animation.

In v5 we are going to approached it in bit a [different way](https://github.com/Adyen/adyen-ios/blob/develop/AdyenComponents/Apple%20Pay/ApplePayComponent.swift#L87-L96).
We providing a closure to get notified when the animation is finished and the component could be dismissed:

```swift
  currentComponent.finalizeIfNeeded(with: success) {
      checkoutViewController.dismiss(animated: true)
  }
```

Thanks for your feedback and contribution!
I am going to close this PR, but feel free to raise discussion here or create a new ticket.

PR #765, opened 2022-01-14
ACH direct debit component

erenbesel:
Support for ACH direct debit including additional encryption for it.


Adyen/Utilities/PublicKeyProvider/PublicKeyConsumer.swift
```diff
@@ -0,0 +1,42 @@
...
+
+    /// Provider for fetching the public key.
+    var publicKeyProvider: AnyPublicKeyProvider { get set }
+    
+    /// Convenient way to fetch the client public key with a closure for the success case
+    /// and an option to notify the delegate on the failure case.
+    /// - Parameters:
+    ///   - notifyingDelegateOnFailure: If `true`, notifies the `PaymentComponentDelegate` on failure.
+    ///   - successHandler: The block that is called when fetching was successful. Contains the public key.
+    func fetchCardPublicKey(notifyingDelegateOnFailure: Bool, successHandler: PublicKeySuccessHandler?)
```
kaphacius, line 20: this is a bit of an unusual pattern. can we change this to pass an optional completion `_ result: Result<String, Error>` here? the way this is used right now, it is either passing `true` and a `successHandler` (so the consumer cares about both outcomes) or passing `false` and no handler (consumer does not care about the outcome). this would also simplify the `fetchCardPublicKey` implementation in the extension.
i understand that `PublicKeyConsumer` extends `PaymentComponent`, that's why this pattern was implemented. i wonder if it makes sense to split it from `PaymentComponent` and change the `fetchCardPublicKey`

AdyenEncryption/BankDetailsEncryptor.swift
```diff
@@ -0,0 +1,46 @@
...
+    /// - Parameters:
+    ///   - number: The bank account number
+    ///   - publicKey: The public key to use for encryption (format "Exponent|Modulus").
+    /// - Returns: An encrypted token containing card number data.
+    /// - Throws: `CardEncryptor.Error.encryptionFailed` if the encryption failed,
+    ///  maybe because the card public key is an invalid one, or for any other reason.
+    /// - Throws: `CardEncryptor.Error.emptyValue` when trying to encrypt an empty or invalid card number
+    public static func encrypt(routingNumber: String, with publicKey: String) throws -> String {
+        guard !routingNumber.isEmpty, routingNumber.allSatisfy(\.isNumber) else {
+            throw CardEncryptor.Error.emptyValue
```
kaphacius, line 40: maywe we can change the name of the error to `invalidValue`, same in line 43

AdyenEncryption/CardEncryptor.swift
```diff
@@ -205,7 +195,7 @@ extension CardEncryptor {
         /// Indicates an error when trying to encrypt empty or invalid expiry year.
         case invalidSecureCode
 
-        /// Indicates an error when trying to encrypt empty or invalid expiry year.
+        /// Indicates an error when trying to encrypt empty or invalid card number.
```
kaphacius, line 198: nice catch! can we also change the comment in lines 192, 195?

kaphacius (commented):
Did not make any comments about `BankPayload` since it will be discussed that offline.

Adyen/Utilities/PublicKeyProvider/PublicKeyConsumer.swift
mohammedDehairy, line 20: no need to expose this function in the protocol interface, its supposed to be a helper function for protocol adopters, its not supposed to be implemented by protocol adopters.

Adyen/Utilities/PublicKeyProvider/PublicKeyConsumer.swift
```diff
@@ -0,0 +1,42 @@
...
+}
+
+/// :nodoc:
+extension PublicKeyConsumer {
+    /// :nodoc:
+    public typealias PublicKeySuccessHandler = (_ publicKey: String) -> Void
+
+    public func fetchCardPublicKey(notifyingDelegateOnFailure: Bool, successHandler: PublicKeySuccessHandler? = nil) {
```
mohammedDehairy, line 28: maybe add `/// :nodoc:` ?

Adyen/Utilities/PublicKeyProvider/PublicKeyProvider.swift
```diff
@@ -1,53 +1,48 @@
...
 //
 
-import Adyen
 import AdyenNetworking
 import Foundation
 
-/// :nodoc:
-internal protocol AnyCardPublicKeyProvider: AnyObject {
+/// Interface to fetch the client public key.
+public protocol AnyPublicKeyProvider: AnyObject {
```
mohammedDehairy, line 11: maybe add `/// :nodoc:` merchants are not supposed to use this protocol

Adyen/Utilities/PublicKeyProvider/PublicKeyConsumer.swift
erenbesel: Had the same conversation with Mohamed. Wanted to have a better way since we are making it public now, but ended up this way as if we go with the result, it will be same signature as the one this calls inside.

Adyen/Utilities/PublicKeyProvider/PublicKeyProvider.swift
```diff
@@ -1,53 +1,48 @@
...
     
-    /// :nodoc:
+    /// Fetches the client public key with a closure for success and failure.
     func fetch(completion: @escaping CompletionHandler)
 }
 
-/// :nodoc:
-internal final class CardPublicKeyProvider: AnyCardPublicKeyProvider {
+/// `PublicKeyProvider` is used to fetch the client public key that is needed for encrypting data.
+public final class PublicKeyProvider: AnyPublicKeyProvider {
```
mohammedDehairy, line 21: maybe add /// :nodoc: merchants are not supposed to use `PublicKeyProvider` or care about it :-)

AdyenEncryption/BankDetailsEncryptor.swift
erenbesel: They are public unfortunately.

AdyenEncryption/AnyEncryptor.swift
```diff
@@ -0,0 +1,25 @@
...
+// Copyright (c) 2022 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+/// Defines the interface to encrypt given data with JWE Encryption.
+internal protocol AnyEncryptor {
+    static func encrypt(_ payload: Payload, with publicKey: String) throws -> String
```
mohammedDehairy, line 11: no need to expose this in protocol interface, its supposed to be a helper function and not supposed to be overridden

AdyenEncryption/BankDetailsEncryptor.swift
kaphacius: we probably want to have a different error enum then? since this is not a `CardEnryptor,` but a `BankDetailsEncryptor`

erenbesel: This one is tricky since that enum contains some generic errors too like `encryptionFailed` which both could use. 
I did not want to create a new enum under this class only for the new cases. Wasn't ideal, the best would be to modify this but it is public.

erenbesel: Added a new enum case in the end.

Adyen/Utilities/PublicKeyProvider/PublicKeyProvider.swift
```diff
@@ -1,53 +1,51 @@
...
+    public convenience init(apiContext: APIContext) {
         let scheduler = SimpleScheduler(maximumCount: 2)
         self.init(apiClient: APIClient(apiContext: apiContext).retryAPIClient(with: scheduler),
                   request: ClientKeyRequest(clientKey: apiContext.clientKey))
     }
 
     /// :nodoc:
-    /// For testing only
-    internal init(apiClient: AnyRetryAPIClient, request: ClientKeyRequest) {
+    public init(apiClient: AnyRetryAPIClient, request: ClientKeyRequest) {
```
mohammedDehairy, line 40: this shouldn't be public, this is only for unit tests

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -25,7 +25,7 @@ public final class GiftCardComponent: PartialPaymentComponent,
     private let giftCardPaymentMethod: GiftCardPaymentMethod
 
     /// :nodoc:
-    internal var cardPublicKeyProvider: AnyCardPublicKeyProvider
+    public var publicKeyProvider: AnyPublicKeyProvider
```
mohammedDehairy, line 28: can this be let? since its public now? or at least private(set)?

AdyenComponents/ACH Direct Debit/ACHDirectDebitComponent.swift
```diff
@@ -0,0 +1,283 @@
...
+    public let style: FormComponentStyle
+
+    /// :nodoc:
+    public let requiresModalPresentation: Bool = true
+
+    /// :nodoc:
+    public let shopperInformation: PrefilledShopperInformation?
+    
+    /// :nodoc:
+    public var publicKeyProvider: AnyPublicKeyProvider
```
mohammedDehairy, line 55: can this be let or private(set)? also the in the consumer protocol it doesn't have to be set only get is enough

mohammedDehairy (dismissed):

mohammedDehairy (approved):

kaphacius (approved):

PR #772, opened 2022-01-27
New ApplePay flow

descorp:
### Changes

* always execute Apple Pay callback when token present
* adopt Component live-cycle to a new flow
* fix PreApplePayComponent: now can be opened multiple time

### DIFF
--- a/AdyenComponents/Apple Pay/ApplePayComponent.swift
+++ b/AdyenComponents/Apple Pay/ApplePayComponent.swift
@@ -12,7 +12,7 @@ import PassKit
 public class ApplePayComponent: NSObject, PresentableComponent, PaymentComponent, Localizable, FinalizableComponent {
 
     /// :nodoc:
-    internal var success: Bool = false
+    internal var state: State = .initiated
     
     /// :nodoc:
     private let payment: Payment
@@ -110,21 +110,9 @@ public class ApplePayComponent: NSObject, PresentableComponent, PaymentComponent
 
     /// Finalizes ApplePay payment after being processed by payment provider.
     /// - Parameter success: The status of the payment.
-    public func didFinalize(with success: Bool) {
-        self.success = success
-        paymentAuthorizationCompletion?(success ? .success : .failure)
-        paymentAuthorizationCompletion = nil
-    }
+    public func didFinalize(with success: Bool) { }
 
     // MARK: - Private
-
-    internal func dismiss(completion: (() -> Void)?) {
-        paymentAuthorizationViewController?.dismiss(animated: true) { [weak self] in
-            guard let self = self else { return }
-            self.paymentAuthorizationViewController = nil
-            completion?()
-        }
-    }
     
     private func getPaymentAuthorizationViewController() -> PKPaymentAuthorizationViewController {
         if paymentAuthorizationViewController == nil {
@@ -132,7 +120,7 @@ public class ApplePayComponent: NSObject, PresentableComponent, PaymentComponent
             let request = configuration.createPaymentRequest(payment: payment, supportedNetworks: supportedNetworks)
             paymentAuthorizationViewController = ApplePayComponent.createPaymentAuthorizationViewController(from: request)
             paymentAuthorizationViewController?.delegate = self
-            paymentAuthorizationCompletion = nil
+            state = .initiated
         }
         return paymentAuthorizationViewController!
     }
@@ -153,6 +141,12 @@ public class ApplePayComponent: NSObject, PresentableComponent, PaymentComponent
 }
 
 extension ApplePayComponent {
+
+    internal enum State {
+        case initiated
+        case paymentInProgress
+        case error(Swift.Error)
+    }
 
     /// Describes the error that can occur during Apple Pay payment.
     public enum Error: Swift.Error, LocalizedError {
@@ -177,6 +171,8 @@ extension ApplePayComponent {
         /// Indicates that the currency code is invalid.
         case invalidCurrencyCode
 
+        case invalidToken
+
         /// :nodoc:
         public var errorDescription: String? {
             switch self {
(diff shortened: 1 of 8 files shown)


AdyenDropIn/Components/PreApplePayComponent.swift
```diff
@@ -98,7 +95,35 @@ extension PreApplePayComponent: PreApplePayViewDelegate {
...
+            component = applePayComponent
+        } else {
+            do {
+                component = try createApplePayComponent()
+                applePayComponent = component
+            } catch {
+                delegate?.didFail(with: error, from: self)
+                return
+            }
+        }
```
mohammedDehairy, line 109: when clicking pay, maybe the `ApplPayComponent` should be always reinitialized.

AdyenComponents/Apple Pay/ApplePayComponent.swift
```diff
@@ -110,29 +110,17 @@ public class ApplePayComponent: NSObject, PresentableComponent, PaymentComponent
 
     /// Finalizes ApplePay payment after being processed by payment provider.
     /// - Parameter success: The status of the payment.
-    public func didFinalize(with success: Bool) {
-        self.success = success
-        paymentAuthorizationCompletion?(success ? .success : .failure)
-        paymentAuthorizationCompletion = nil
-    }
+    public func didFinalize(with success: Bool) { }
```
mohammedDehairy, line 113: what if ApplePayComponent  is used as an individual component, shouldn't it call `paymentAuthorizationCompletion` closure  and reset its state ?

AdyenComponents/Apple Pay/ApplePayComponentExtensions.swift
```diff
@@ -13,17 +13,27 @@ extension ApplePayComponent: PKPaymentAuthorizationViewControllerDelegate {
     
     /// :nodoc:
     public func paymentAuthorizationViewControllerDidFinish(_ controller: PKPaymentAuthorizationViewController) {
-        dismiss { [weak self] in
-            guard let self = self, self.success == false else { return }
+        switch state {
+        case .initiated:
             self.delegate?.didFail(with: ComponentError.cancelled, from: self)
+        case .paymentInProgress:
+            break // viewController should be dissmissed when API call is proccessed
```
mohammedDehairy, line 20: `paymentAuthorizationViewControllerDidFinish` can be called when user wants to cancel, right? does this mean the shopper cannot cancel while the payment is in progress? did you try this scenario?

Demo/Configuration.swift
```diff
@@ -37,7 +37,7 @@ internal enum ConfigurationConstants {
 
     static let demoServerAPIKey = "{YOUR_DEMO_SERVER_API_KEY}"
 
-    static let applePayMerchantIdentifier = "{YOUR_APPLE_PAY_MERCHANT_IDENTIFIER}"
+    static let applePayMerchantIdentifier = "merchant.com.adyen.checkoutsandbox.test"
```
mohammedDehairy, line 40: please revert this :-)

mohammedDehairy:
I think this change breaks ApplePayComponent flow when used as an individual component, and tbh I am not even sure its fully working if used in DropIn

AdyenComponents/Apple Pay/ApplePayComponentExtensions.swift
```diff
@@ -35,6 +45,7 @@ extension ApplePayComponent: PKPaymentAuthorizationViewControllerDelegate {
                                       shippingContact: shippingContact)
         
         submit(data: PaymentComponentData(paymentMethodDetails: details, amount: self.amountToPay, order: order))
+        completion(.success)
     }
 }
```
mohammedDehairy, line 51: I think these change breaks ApplePayComponent flow when used as an individual component, and tbh I am not even sure its fully working if used in DropIn

AdyenDropIn/Components/PreApplePayComponent.swift
descorp: Problem is - the only way to check if ApplePay is possible - initiate new pkViewController.
So we have to try building new ApplPayComponent in constructor to prevent "Dummy" Apple Pay button on Drop-in that doesn't do anything...

But you right: after first one - every next click on `pay` will reinitialize the ApplPayComponent

AdyenComponents/Apple Pay/ApplePayComponent.swift
descorp: We are no longer going to use `paymentAuthorizationCompletion` (it is removed in this PR) on v4.
For v5 we will need to build new flow for Apple Pay to treat it as SDK Action.

In this version ApplePayComponent no longer have (need) functional `didFinalize`.

Demo/Configuration.swift
descorp: Ops.. how did that happened 😅

AdyenComponents/Apple Pay/ApplePayComponent.swift
mohammedDehairy: - `to treat it as SDK Action`

have no idea what you mean :-D, its not an SDK and its not an action, its a payment component.

- `We are no longer going to use paymentAuthorizationCompletion `

also don't know what you're talking about :-)

lets discuss this, I don't feel comfortable merging this code

AdyenComponents/Apple Pay/ApplePayComponentExtensions.swift
descorp: > shopper cannot cancel while the payment is in progress?

This is correct. It is possible for about half a second after starting a payment (in other words - if the shopper is determined to do so), but basically - after one passes the authorization step (Face ID, Passcode, etc) Apple Pay will remain on a screen till of API call is conducted.

PR #773, opened 2022-01-28
feature: jump to next field if brand has panLength

erenbesel:
To allow for jumping to next field if a full number is pasted, delegate trigger is added to brand detection in the CardNumberItemView


descorp (approved):

Adyen/UI/Form/Items/Text/FormTextItemView.swift
```diff
@@ -286,6 +282,14 @@ open class FormTextItemView<ItemType: FormTextItem>: FormValueItemView<String, F
...
         }
     }
+    
+    /// :nodoc:
+    public func notifyDelegateOfMaxLengthIfNeeded() {
+        let maximumLength = item.validator?.maximumLength(for: item.value) ?? .max
+        if item.value.count == maximumLength {
+            delegate?.didReachMaximumLength(in: self)
+        }
+    }
```
mohammedDehairy, line 292: nice 👍

mohammedDehairy: maybe it should be `if item.value.count >= maximumLength`, since by the time bin lookup response comes back the shopper maybe have already entered a number greater than max length, right?

erenbesel: Thats a good catch, let me try both cases.

AdyenCard/Validators/CardNumberValidator.swift
```diff
@@ -34,7 +40,7 @@ public final class CardNumberValidator: Validator {
     
     /// :nodoc:
     public func maximumLength(for value: String) -> Int {
-        19
+        panLength ?? 19
```
nauaros, line 43: Is the 19 the default value for `panLength`? It may be a good idea to set it on the constant declaration.

descorp: it is max PAN length we know of..

nauaros: What I mean is if no `panLength` is passed as a parameter, then the fallback value is always 19. I only see `panLength` being used within the `maximumLength(for value: String)`

erenbesel: If you mean to add it in a constant enum, sure why not :)

nauaros (approved):

mohammedDehairy (approved):

PR #812, opened 2022-02-25
feature: groups all action components configurations into one object per component 

mohammedDehairy:
- groups all action components configurations into one object per component 
- and also did a minor refactoring in VoucherComponent and DocumentComponent do be less error prone.


AdyenActions/Components/Await/AwaitComponent.swift
```diff
@@ -22,35 +22,53 @@ public final class AwaitComponent: ActionComponent, Cancellable {
     /// :nodoc:
     public let requiresModalPresentation: Bool = true
     
-    /// The Component UI style.
-    public let style: AwaitComponentStyle
+    /// The await component configurations.
+    public struct Configuration {
```
erenbesel, line 26: I think all of them would be better conforming to `AnyBasicComponentConfiguration` for consistency.

Tests/AdyenTests/Adyen Tests/Components/BACS Direct Debit/Mocks/DocumentActionViewDelegateMock.swift
```diff
@@ -0,0 +1,25 @@
+//
+//  DocumentActionViewDelegateMock.swift
+//  AdyenUIKitTests
+//
+//  Created by Eren Besel on 12/16/21.
```
descorp, line 5: Unformatted headers 🙄

erenbesel (approved):

AdyenActions/Components/Await/AwaitComponent.swift
```diff
@@ -22,35 +22,53 @@ public final class AwaitComponent: ActionComponent, Cancellable {
...
+        
+        /// The component UI style.
+        public var style: AwaitComponentStyle = .init()
```
kaphacius, line 29: do we need to set the initial value here if it is set in the `init`?

descorp (approved):

kaphacius (approved):

AdyenActions/Components/QRCode/QRCodeComponent.swift
```diff
@@ -197,8 +216,8 @@ extension QRCodeComponent: ActionComponentDelegate {
 /// :nodoc:
 extension QRCodeComponent: QRCodeViewDelegate {
     
-    func copyToPasteboard() {
-        UIPasteboard.general.string = action.map(\.qrCodeData)
+    func copyToPasteboard(with action: QRCodeAction) {
```
kaphacius, line 219: this should probably have access level specifier

mohammedDehairy: can I fix this in a new PR?

AdyenActions/Components/Await/AwaitComponent.swift
mohammedDehairy: it doesn't hurt, does it?

AdyenActions/Components/QRCode/QRCodeComponent.swift
kaphacius: yes, i approved this one :)

AdyenActions/Components/Await/AwaitComponent.swift
kaphacius: it's just settings a value which will immediately be overwritten in `init`. if it had no initial value - compiler would check if every version of `init` actually sets a value for this field.

mohammedDehairy: ok will fix it with the other one in another PR

PR #813, opened 2022-02-28
COIOS-492: Analytics provider

nauaros:
## Summary

Tasks performed:
1. Built Analytics Provider.
2. Logic for the telemetry track.
3. Created `AdyenContext` object.
4. Injected `AdyenContext` object in main components.
	* Card component
	* BCMCComponent
	* 3DS2 component
	* ApplePay component
	* PreApplePayComponent
	* Gift Card component
	* SEPA Direct Debit component
	* BACS Direct Debit component (refactor tracker)
	* ACH Direct Debit component
	* Qiwi Wallet component
	* Affirm component
	* BLIK component
	* Boleto component
	* MBWay component
	* Issuer List component
	* EContext component
	* Doku component
	* AdyenAction component


Adyen/Analytics/AnalyticsProvider/AdyenContext.swift
```diff
@@ -0,0 +1,37 @@
...
+//
+
+import AdyenNetworking
+import Foundation
+
+// TODO: - Document
+public enum AnalyticsConfiguration {
+    case enabled(telemetry: Bool, conversion: Bool)
+    case disabled
+}
```
mohammedDehairy: maybe a struct would be better:

```Swift
public struct AnalyticsConfiguration {
     var telemetryEnabled: Bool
     var conversionEnabled: Bool
 }
```

if merchant chose both `telemetryEnabled` and `conversionEnabled` false, then the `disabled` enum case doesn't make sense, right?

mohammedDehairy: also maybe the `AnalyticsConfiguration` should be a subtype of `AdyenContext`?

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
```diff
@@ -0,0 +1,64 @@
...
+        if !conversion { return completion(nil) }
+
+        if let checkoutAttemptId = checkoutAttemptId {
+            return completion(checkoutAttemptId)
+        }
+
+        let checkoutAttemptIdRequest = CheckoutAttemptIdRequest()
+
+        let uniqueAssetAPIClient = UniqueAssetAPIClient<CheckoutAttemptIdResponse>(apiClient: apiClient)
+        uniqueAssetAPIClient.perform(checkoutAttemptIdRequest) { [weak self] result in
```
mohammedDehairy: you must retain the `uniqueAssetAPIClient` instance and reuse it to make sure the attemptId is fetched once, if you create a new instance every time, then the "locking" mechanism is circumvented :-)

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
```diff
@@ -0,0 +1,83 @@
...
+
+/// : nodoc:
+public protocol AnalyticsProviderProtocol: TelemetryTrackerProtocol {}
+
+/// : nodoc:
+internal final class AnalyticsProvider: AnalyticsProviderProtocol {
+
+    // MARK: - Properties
+
+    internal let apiClient: APIClientProtocol
```
erenbesel, line 52: Seems this is not needed aside from as parameter in init

Adyen/Analytics/AnalyticsProvider/CheckoutAttemptIdRequest.swift
```diff
@@ -0,0 +1,49 @@
...
+
+    internal enum CodingKeys: String, CodingKey {
+        case identifier = "id"
+    }
+}
+
+/// :nodoc:
+internal struct CheckoutAttemptIdRequest: APIRequest {
+
+    /// :nodoc:
```
erenbesel: IIRC we don't need `nodoc` on internal stuff. Also not sure but I think if the enclosing type has nodoc, it applies to properties too.

Adyen/Analytics/AnalyticsProvider/TelemetryData.swift
```diff
@@ -0,0 +1,72 @@
...
+//
+
+import Foundation
+
+internal struct TelemetryData: Encodable {
+
+    // MARK: - Properties
+
+    internal var version: String? {
+        Bundle(for: AnalyticsProvider.self).infoDictionary?["CFBundleShortVersionString"] as? String
```
mohammedDehairy, line 14: this can be lazy var, right? no need to recalculate it with every access

Adyen/Analytics/AnalyticsProvider/TelemetryData.swift
```diff
@@ -0,0 +1,72 @@
...
+
+    internal let channel: String = "iOS"
+
+    internal var locale: String {
+        let languageCode = Locale.current.languageCode ?? ""
+        let regionCode = Locale.current.regionCode ?? ""
+        return "\(languageCode)_\(regionCode)"
+    }
+
+    internal let userAgent: String? = nil
```
erenbesel, line 25: This is constant `nil`. What's the point of having this property?

Adyen/Analytics/AnalyticsProvider/TelemetryData.swift
```diff
@@ -0,0 +1,72 @@
...
+
+    internal var deviceBrand: String {
+        var systemInfo = utsname()
+        uname(&systemInfo)
+        let machineMirror = Mirror(reflecting: systemInfo.machine)
+        return machineMirror.children.reduce("") { identifier, element in
+            guard let value = element.value as? Int8, value != 0 else { return identifier }
+            return identifier + String(UnicodeScalar(UInt8(value)))
+        }
+    }
```
mohammedDehairy: this also can be lazy var, the device brand doesn't change at runtime, right?

Adyen/Analytics/AnalyticsProvider/TelemetryData.swift
```diff
@@ -0,0 +1,72 @@
...
+
+    internal var locale: String {
+        let languageCode = Locale.current.languageCode ?? ""
+        let regionCode = Locale.current.regionCode ?? ""
+        return "\(languageCode)_\(regionCode)"
+    }
+
+    internal let userAgent: String? = nil
+
+    internal var deviceBrand: String {
```
erenbesel: Isn't `UIDevice.current.model` enough for our requirements? This seems a bit much

Adyen/Analytics/AnalyticsProvider/TelemetryTracker.swift
```diff
@@ -0,0 +1,48 @@
+//
+// Copyright (c) 2022 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+/// : nodoc:
+public enum TelemetryFlavor {
```
erenbesel, line 10: Not sold on the name flavor :D

Adyen/Analytics/AnalyticsProvider/TelemetryData.swift
```diff
@@ -0,0 +1,72 @@
...
+        UIDevice.current.systemVersion
+    }
+
+    internal var referrer: String {
+        Bundle.main.bundleIdentifier ?? ""
+    }
+
+    internal var screenWidth: Int {
+        Int(UIScreen.main.bounds.width)
+    }
```
mohammedDehairy, line 43: should you be using `UIScreen.main.nativeBounds` instead of `UIScreen.main.bounds`?:
`UIScreen.main.bounds` change with device orientation, but `UIScreen.main.nativeBounds` is the physical device width regardless of orientation.

Adyen/Analytics/AnalyticsProvider/TelemetryTracker.swift
```diff
@@ -0,0 +1,48 @@
...
+    case components(type: String)
```
erenbesel: Looks like this string is the payment method type. We have a new enum for this, maybe we can use it here too?

Adyen/Analytics/AnalyticsProvider/TelemetryTracker.swift
```diff
@@ -0,0 +1,48 @@
...
+    case dropIn(type: String = "dropin", paymentMethods: [String])
+    case dropInComponent
```
mohammedDehairy, line 17: what's the difference between `dropIn` and `dropInComponent`? can we add a comment to explain the difference?

Adyen/Core/Components/AbstractPersonalInformationComponent/AbstractPersonalInformationComponent+Extensions.swift
```diff
@@ -29,6 +29,12 @@ extension AbstractPersonalInformationComponent: LoadingComponent {
 
 extension AbstractPersonalInformationComponent: TrackableComponent {
 
+    /// :nodoc
+    public func viewDidLoad(viewController: UIViewController) {
+        let flavor: TelemetryFlavor = _isDropIn ? .dropInComponent : .components(type: paymentMethod.type)
+        adyenContext.analyticsProvider.trackTelemetryEvent(flavor: flavor)
+    }
```
mohammedDehairy: should it be `viewDidLoad` or `viewWillAppear`? 🤔

Adyen/Core/Components/AbstractPersonalInformationComponent/AbstractPersonalInformationComponent.swift
```diff
@@ -19,6 +19,9 @@ open class AbstractPersonalInformationComponent: PaymentComponent, PresentableCo
 
     /// :nodoc:
     public let apiContext: APIContext
+
+    /// :nodoc:
+    public let adyenContext: AdyenContext
```
mohammedDehairy, line 24: maybe we should add `var adyenContext: AdyenContext { get }` to the `PaymentComponent` and `ActionComponent` protocols? so that if we miss it in one component we get a compile time error.

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
nauaros: It's being used in `func trackTelemetryEvent(flavor: TelemetryFlavor)` method.

Adyen/Analytics/AnalyticsProvider/TelemetryData.swift
nauaros: The reason is due to the fact that we need to send this parameter in the telemetry request. `userAgent` is being used on the web where it makes sense, but for us it doesn't. That's why it's a constant `nil`.

Adyen/Analytics/AnalyticsProvider/TelemetryData.swift
nauaros: `UIDevice.current.model` only returns values such as "iPhone", "iPad", "iPod", etc. and we'd like to have the specific model. That's why we're using that ugly logic xD

Adyen/Analytics/AnalyticsProvider/TelemetryTracker.swift
nauaros: Yeah hahaha "flavor" is the term we're using in telemetry, that's why we're using it. Do you have any suggestion?

Adyen/Analytics/AnalyticsProvider/TelemetryTracker.swift
nauaros: Yes you're right! It was build right before Mo merged the changes, ~I'll change it to the enum type in the following PR~. Nice catch! Thanks for remembering me

Adyen/Analytics/AnalyticsProvider/TelemetryData.swift
nauaros: Hmm this is interesting. Let me check with the data team to see what we actually need. It may be interesting for us to know if there's any merchant using our product in a landscape orientation. Thanks!

Adyen/Analytics/AnalyticsProvider/TelemetryTracker.swift
nauaros: Yes, definitely! `dropInComponent` are those components used within the DropIn. In those cases, we don't want to send the telemetry event since it's already being sent from the DropIn component with the different available payment methods.

`components` are stand alone components, outside of DropIn.

Adyen/Core/Components/AbstractPersonalInformationComponent/AbstractPersonalInformationComponent+Extensions.swift
nauaros: It's doesn't really matter in reality. I'm following the same behavior we have for the 1px thing. I guess it's better to send it as soon as the component is being loaded.

nauaros: Also, on the web they're sending the event before thew view gets painted.

Adyen/Core/Components/AbstractPersonalInformationComponent/AbstractPersonalInformationComponent.swift
nauaros: Yes! Good point. I'm gonna add this on following current PR in which I'm working on!

Adyen/Analytics/AnalyticsProvider/TelemetryData.swift
nauaros: Changed to be a constant

Adyen/Analytics/AnalyticsProvider/TelemetryData.swift
nauaros: changed to be a constant instead

Adyen/Analytics/AnalyticsProvider/TelemetryData.swift
nauaros: Change to use nativeBounds. We're gonna include an `orientation` property later.

Adyen/Analytics/AnalyticsProvider/TelemetryTracker.swift
nauaros: Updated to use `PaymentMethodType`!

Adyen/Analytics/AnalyticsProvider/TelemetryTracker.swift
nauaros: Added a clarifying comment

Adyen/Core/Components/AbstractPersonalInformationComponent/AbstractPersonalInformationComponent+Extensions.swift
nauaros: Sending the telemetry event when the view will appear.

erenbesel (approved):

mohammedDehairy (approved):

PR #820, opened 2022-03-02
feature: Adds the Session object skeleton, refactors DropInComponentD…

mohammedDehairy:
Session skeleton

 * Adds the Session object skeleton
 *  Refactors DropInComponentDelegate protocol.
 *  Created the DropInSession object specially for using Session with DropInComponent.
 * Moved the basic Action protocols to core module.


descorp (dismissed):

AdyenDropIn/Session/DropInSession.swift
```diff
@@ -0,0 +1,20 @@
...
+import Foundation
+
+public final class DropInSession {
+    internal weak var session: SessionProtocol?
+    
+    public init(session: SessionProtocol) {
+        self.session = session
+    }
+    
+    public func didFail(with error: Error, from dropInComponent: DropInComponent) {
```
descorp: Why would someone call `didFail` on `DropInSession` ?
It could be `internal`, right?

mohammedDehairy: its needed by the `DropInComponentDelegate` protocol, so has to be public :-)

descorp (approved):

erenbesel (approved):

PR #824, opened 2022-03-03
feature: Adds a new PaymentMethodType enum instead of relying on strings for PM types

mohammedDehairy:
feature: Adds a new PaymentMethodType enum instead of relying on strings for PM types


Adyen/Core/Payment Methods/ACHDirectDebitPaymentMethod.swift
```diff
@@ -10,7 +10,7 @@ import Foundation
 public struct ACHDirectDebitPaymentMethod: PaymentMethod {
     
     /// :nodoc:
-    public let type: String
+    public let type: PaymentMethodType
```
descorp, line 13: Maybe we can do `public let type: PaymentMethodType = .achDirectDebit ` everywhere? 

then we can do : 

```
ACHDirectDebitPaymentMethod(name: "test_name")
```

instead of 

```
ACHDirectDebitPaymentMethod(type: .achDirectDebit, name: "test_name")
```

🤔

mohammedDehairy: sure, maybe in another PR?

descorp: ok, why not :)

descorp (approved):

erenbesel (approved):

PR #830, opened 2022-03-11
chore: Implementing the paymentComponentDelegate protocol

mohammedDehairy:
chore: Implementing the paymentComponentDelegate protocol


Adyen/Core/Core Protocols/DropInComponentProtocol.swift
```diff
@@ -0,0 +1,22 @@
+//
+// Copyright (c) 2022 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+/// Any DropIn Component.
+public protocol DropInComponentProtocol: PresentableComponent {
```
erenbesel: Nit pick but how about `AnyDropInComponent`? Would match the description too :D

AdyenSession/Session+PaymentComponentDelegate.swift
```diff
@@ -10,42 +10,80 @@ import Adyen
...
                                       sessionData: sessionContext.data,
                                       data: data)
-        apiClient.perform(request, completionHandler: paymentResponseHandler)
+        apiClient.perform(request) { [weak self] in
+            self?.paymentResponseHandler(result: $0, for: currentComponent)
+        }
     }
     
-    internal func paymentResponseHandler(result: Result<PaymentsResponse, Error>) {
+    internal func paymentResponseHandler(result: Result<PaymentsResponse, Error>,
```
erenbesel: Could we rename this to `handleResult` or similar? Current one looks like a closure :)

Adyen/Core/Core Protocols/DropInComponentProtocol.swift
mohammedDehairy: sure I can rename it to `AnyDropInComponent, hahaha

erenbesel (approved):

Adyen/Core/Core Protocols/AnyDropInComponent.swift
```diff
@@ -0,0 +1,22 @@
+//
+// Copyright (c) 2022 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+/// Any DropIn Component.
+public protocol AnyDropInComponent: PresentableComponent {
```
descorp, line 10: Just a thought: maybe we can come up with bit more generic name?
Like "payment interface" or "checkout interface" ?

AdyenSession/Session+ActionComponentDelegate.swift
```diff
@@ -10,22 +10,33 @@ import Adyen
...
         // TODO: call back the merchant
     }
 
     public func didProvide(_ data: ActionComponentData, from component: ActionComponent) {
-        (component as? PresentableComponent)?.viewController.view.isUserInteractionEnabled = false
+        didProvide(data, currentComponent: component)
+    }
+    
+    internal func didProvide(_ data: ActionComponentData, currentComponent: Component) {
+        (currentComponent as? PresentableComponent)?.viewController.view.isUserInteractionEnabled = false
```
descorp, line 32: Is it a good idea that Sessions mutating component or it's view?
Maybe we can do `willProvide` in delegate, so merchant can apply all of those themself?

AdyenSession/Session+DropInComponentDelegate.swift
```diff
@@ -0,0 +1,36 @@
...
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Adyen
+import Foundation
+
+/// :nodoc:
+extension Session: DropInComponentDelegate {
+    public func didSubmit(_ data: PaymentComponentData, from component: PaymentComponent, in dropInComponent: AnyDropInComponent) {
```
descorp, line 12: probably have to mark with `/// :nodoc:` every method as well

AdyenSession/Session+ActionComponentDelegate.swift
mohammedDehairy: this was in the example integration, I think it make sense yes, why not, the session object is supposed to be the components' autopilot so to speak :-)

AdyenSession/API/Session Setup/SessionSetupRequest.swift
```diff
@@ -21,17 +21,29 @@ internal struct SessionSetupRequest: Request {
...
+    internal let order: PartialPaymentOrder?
+    
     internal typealias ResponseType = SessionSetupResponse
     
     internal typealias ErrorResponseType = APIError
     
-    internal init(sessionId: String, sessionData: String) {
+    internal init(sessionId: String,
+                  sessionData: String,
+                  order: PartialPaymentOrder?) {
```
descorp, line 32: This could be `order: PartialPaymentOrder? = nil`

AdyenSession/Session+DropInComponentDelegate.swift
mohammedDehairy: isn't the `/// :nodoc:` in the extension declaration will do the trick?

AdyenSession/API/Session Setup/SessionSetupRequest.swift
mohammedDehairy: I wanted it to be explicit, to prevent missing the order by mistake

AdyenSession/Session+ActionComponentDelegate.swift
descorp: Now that I am thinking of it - this could be not such a great idea..
Does it affect DropIn? DropIn is a PresentableComponent, but if it show "AwaitComponent" and this will disable user's interaction..

AdyenSession/Session+DropInComponentDelegate.swift
descorp: I am not sure, we need to make POC.. 
I think what will happen: it will hide the fact that `Session` is  `DropInComponentDelegate` (this is actually undesired), but still print each function in the dock..

AdyenSession/API/Session Setup/SessionSetupRequest.swift
descorp: I guess `order: nil` is going to be overwhelming majority of use-cases?

mohammedDehairy: its in one place in the code :-)

descorp (approved):

PR #833, opened 2022-03-18
chore: refactors dropin navigation

mohammedDehairy:
chore: refactors dropin navigation


Adyen/UI/List/CoreListDataSource.swift
```diff
@@ -112,32 +112,26 @@ internal final class CoreListDataSource: NSObject, ListViewControllerDataSource
...
-    
-    private func cell(for item: ListItem, tableView: UITableView) -> ListCell? {
-        for case let cell as ListCell in tableView.visibleCells where cell.item == item {
-            return cell
+        for case let item in sections.flatMap(\.items) where item != loadingItem {
+            item.isEnabled = false
         }
         
-        return nil
+        tableView.reloadData()
```
descorp, line 124: This will reload whole table and cause "flickering", right ?
Same for line 135

Demo/Common/IntegrationExampleComponents.swift
```diff
@@ -128,8 +128,12 @@ extension IntegrationExample {
         }
     }
 
-    private func handle(_ action: Action) {
-        actionComponent.handle(action)
+    internal func handle(_ action: Action) {
+        if let actionComponent = currentComponent as? ActionHandlingComponent {
+            actionComponent.handle(action)
+        } else {
+            actionComponent.handle(action)
```
descorp, line 135: I don't think this is good idea! 
Previously we had 2 separate flows per file: one for drop-In, one for components.

Now we have:
*  super confusing `if X { A } else { A }`;
* drop-in and component flows are mixed in one file;
* it is unobvious how to handle action for Drop-In flow (the fact that instance of Drop-In can handle actions).

Demo/Common/IntegrationExampleDropIn.swift
```diff
@@ -55,10 +55,6 @@ extension IntegrationExample {
         }
     }
 
-    private func handle(_ action: Action) {
-        (currentComponent as? DropInComponent)?.handle(action)
```
descorp, line 59: I would create local variable for DropIn and call here simply. This would be 100% more self-explanatory

```swift
dropInComponent?.handle(action)
```

AdyenDropIn/Presentation/OverlayPresentationController.swift
```diff
@@ -1,34 +1,34 @@
...
 //
 
 import Foundation
 import UIKit
 
-// MARK: - DimmingPresentationController
+// MARK: - OverlayPresentationController
 
 /// Presentation Controller that performs dimming view alongside presentation animation.
-internal final class DimmingPresentationController: UIPresentationController {
```
descorp, line 13: what is wrong with `Dimming`? 😆

AdyenDropIn/DropInComponent.swift
```diff
@@ -220,9 +220,7 @@ public final class DropInComponent: NSObject,
         case let component as PresentableComponent where component.requiresModalPresentation:
             navigationController.present(asModal: component)
         case let component as PresentableComponent where component.viewController is UIAlertController:
-            navigationController.present(component.viewController, customPresentation: false)
-        case let component as PresentableComponent:
-            navigationController.present(component.viewController, customPresentation: true)
+            navigationController.present(component.viewController, animated: true, completion: nil)
```
descorp: This will wrap `UIAlertController` into `HalfScreenViewController` ?

AdyenDropIn/DropInComponent.swift
```diff
@@ -220,9 +220,7 @@ public final class DropInComponent: NSObject,
         case let component as PresentableComponent where component.requiresModalPresentation:
             navigationController.present(asModal: component)
```
descorp: We can rename `present(asModal)` into `push(viewController:)` now
BTW `present(asModal:)` also checks `if component.requiresModalPresentation `. One of the checks is redundant..

AdyenDropIn/Presentation/DropInNavigationController.swift
```diff
@@ -1,106 +1,131 @@
...
     private let cancelHandler: CancelHandler?
     
     private var keyboardRect: CGRect = .zero
-
-    internal var keyboardObserver: Any?
     
-    internal let style: NavigationStyle
+    private let rootNavigationController: ComponentNavigationController
+    
+    internal let chileViewController: HalfPageViewController
```
descorp: `chile` - > `child` 😆

AdyenDropIn/Presentation/HalfPageViewController.swift
```diff
@@ -0,0 +1,177 @@
...
+    private func setupChildLayout() {
+        let childView: UIView = child.view
+        childView.translatesAutoresizingMaskIntoConstraints = false
+
+        let topConstraint = childView.topAnchor.constraint(equalTo: view.topAnchor)
+        let bottomConstraint = childView.bottomAnchor.constraint(equalTo: view.bottomAnchor)
+        let leftConstraint = childView.leadingAnchor.constraint(equalTo: view.leadingAnchor)
+        let rightConstraint = childView.trailingAnchor.constraint(equalTo: view.trailingAnchor)
+
+        NSLayoutConstraint.activate([leftConstraint,
```
descorp, line 152: You can do 

```swift
private var constraints: [NSLayoutConstraint]?

...

let constraints = childView.adyen.anchor(inside: view)
NSLayoutConstraint.activate(constraints)
self.constraints = constraints
```

Adyen/UI/List/CoreListDataSource.swift
mohammedDehairy: it doesn't flicker :-)

Demo/Common/IntegrationExampleComponents.swift
mohammedDehairy: I think this is why the Voucher component was looking completely off, it was presented modally instead of through DropIn, somehow even though the two functions `handle(_ action:)` are private, they are still conflicting, because its still the same class.

AdyenActions/Components/QRCode/QRCodeComponent.swift
```diff
@@ -157,9 +157,23 @@ public final class QRCodeComponent: ActionComponent, Cancellable {
     private func createViewController(with action: QRCodeAction) -> UIViewController {
         let viewController = QRCodeViewController(viewModel: createModel(with: action))
         viewController.qrCodeView.delegate = self
+        viewController.navigationItem.rightBarButtonItem = createRightNavigationBarItem()
         return viewController
     }
     
+    private func createRightNavigationBarItem() -> UIBarButtonItem {
```
erenbesel, line 164: createRightNavigationBarItem and onTrailingButtonTap are duplicated, maybe we can move it to a common protocol

AdyenComponents/BACS Direct Debit/BACSDirectDebitComponent.swift
```diff
@@ -85,6 +68,24 @@ public final class BACSDirectDebitComponent: PaymentComponent, PresentableCompon
...
+    }
+    
+    private func createInputPresenter() -> BACSInputPresenterProtocol {
+        let tracker = BACSDirectDebitComponentTracker(paymentMethod: bacsPaymentMethod,
+                                                      apiContext: apiContext,
+                                                      isDropIn: _isDropIn)
+        let itemsFactory = BACSItemsFactory(styleProvider: configuration.style,
+                                            localizationParameters: configuration.localizationParameters,
+                                            scope: String(describing: self))
+        return BACSInputPresenter(view: inputFormViewController,
```
erenbesel, line 84: The reason we had to make those lazy was due to payment not being set on init but after init for the components. I believe when you call `createInputPresenter()` in init, `payment.amount` will be nil :). Did you confirm it?

mohammedDehairy: good that you mentioned that! I'll look into it.

the problem was that both lazy var's refer to each other and we end up with an infinite recursion :-).

Adyen/Utilities/Debugging/Logger.swift
```diff
@@ -22,6 +23,10 @@ public enum AdyenLogging {
 /// and `Swift.print()` is called inside after checking first if `AdyenLogging.isEnabled` is `true`, and returns if `false`.
 public func adyenPrint(_ items: Any..., separator: String = " ", terminator: String = "\n") {
     guard AdyenLogging.isEnabled else { return }
+    
+    let timestamp = ISO8601DateFormatter().string(from: Date())
```
kaphacius: can we create the `ISO8601DateFormatter` only once?

AdyenActions/Components/Voucher/VoucherComponent.swift
```diff
@@ -7,6 +7,7 @@
 import Adyen
 import AdyenNetworking
 import PassKit
+import SwiftUI
```
kaphacius: are we using SwiftUI in this file?

AdyenDropIn/Presentation/ComponentNavigationController.swift
```diff
@@ -0,0 +1,149 @@
...
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Adyen
+import UIKit
+
+internal final class ComponentNavigationController: UINavigationController {
+    
+    internal var cancelHandler: (Bool, PresentableComponent) -> Void
```
kaphacius: can this be a let?

AdyenDropIn/Presentation/DropInNavigationAnimator.swift
```diff
@@ -0,0 +1,80 @@
...
+        dropInNavigationLayouter?.freezeFrameUpdate()
+        
+        let context = SpringAnimationContext(
+            animationKey: Animation.dropinTransitionPresentation.rawValue,
+            duration: duration,
+            delay: 0.3,
+            dampingRatio: 0.8,
+            velocity: 0.2,
+            options: [.beginFromCurrentState, .curveEaseInOut],
+            animations: { [weak self] in
```
kaphacius: can we capture only `isRightToLeftSlideAnimation` instead of self here?

Adyen/UI/Form/FormViewController.swift
```diff
@@ -183,7 +172,10 @@ open class FormViewController: UIViewController, Localizable, KeyboardObserver,
 
         resignFirstResponder()
 
-        showValidation()
+        /// Delay validation until after keyboard has disappeared
+        DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(400)) { [weak self] in
```
kaphacius, line 176: can we use `keyboardAnimationDurationUserInfoKey` instead of a hardcoded duration here?

Adyen/UI/List/CoreListDataSource.swift
```diff
@@ -112,32 +112,26 @@ internal final class CoreListDataSource: NSObject, ListViewControllerDataSource
...
         
-        for case let visibleCell as ListCell in tableView.visibleCells where visibleCell.item != item {
-            visibleCell.isEnabled = false
-        }
-    }
-    
-    private func cell(for item: ListItem, tableView: UITableView) -> ListCell? {
-        for case let cell as ListCell in tableView.visibleCells where cell.item == item {
-            return cell
+        for case let item in sections.flatMap(\.items) where item != loadingItem {
```
kaphacius: can we use `for item in sections.flatMap(\.items) where item != loadingItem` here?

AdyenDropIn/Presentation/HalfPageViewController.swift
```diff
@@ -0,0 +1,180 @@
...
+import UIKit
+
+internal final class HalfPageViewController: UIViewController {
+
+    // MARK: - Properties
+
+    private var topConstraint: NSLayoutConstraint?
+    private var bottomConstraint: NSLayoutConstraint?
+    private var rightConstraint: NSLayoutConstraint?
+    private var leftConstraint: NSLayoutConstraint?
```
kaphacius, line 17: is it possible to make these non-optional? they are all set during initialization. would be nice to not unwrap them

AdyenDropIn/Presentation/HalfPageViewController.swift
```diff
@@ -0,0 +1,180 @@
...
+        internal let animated: Bool
+    }
+    
+    private var isFrameUpdateFrozen: Bool = false
+    
+    /// Keep only the currently animating context and only one next context, i.e the latest.
+    private var currentlyUpdatingFrameContext: FrameUpdateContext?
+    private var nextUpdatingFrameContext: FrameUpdateContext?
+    
+    internal func freezeFrameUpdate() {
```
kaphacius: can we use name this something more common? like `pauseFrameUpdate` etc

AdyenDropIn/Presentation/HalfPageViewController.swift
```diff
@@ -0,0 +1,180 @@
...
+        guard isUpdatingFrameAllowed() else {
+            /// There is currently an ongoing animation or updating frames is frozen,
+            /// in this case update the `nextUpdatingFrameContext`.
+            nextUpdatingFrameContext = context
+            return
+        }
+        updateFrame(with: context)
+    }
+    
+    private func isUpdatingFrameAllowed() -> Bool {
```
kaphacius: maybe we could make this a computed variable, the logic seems very fitting

AdyenDropIn/Presentation/HalfPageViewController.swift
mohammedDehairy: they're set in a separate function that's called in the initializer, but this doesn't count as `during initialization` unfortunately :-)

Adyen/UI/Form/FormViewController.swift
mohammedDehairy: I don't think we have access to this value here

AdyenDropIn/Presentation/HalfPageViewController.swift
kaphacius: can we change the logic so we set those values in the init? maybe extract it to a static function. it is not a very good design to have those marked as optional, even though they are never optional, and we need to unwrap and make custom logic for them like they are.

Adyen/UI/Form/FormViewController.swift
kaphacius: can we add this to the `KeyboardObserver`? i think it would be much nicer to use the duration provided by the system, instead of a hardcoded one

AdyenDropIn/Presentation/HalfPageViewController.swift
mohammedDehairy: yes, its not a very good design, but its also not clean code to put the layout code in the initializer, its not exactly single responsibility principle or [single level of abstraction principle](http://principles-wiki.net/principles:single_level_of_abstraction).

in short the initializer will look messy and less maintainable. so its a matter of compromise, and I prefer to have it as optional as a compromise.

Adyen/UI/Form/FormViewController.swift
mohammedDehairy: yes, if we put it in the `KeyboardObserver` we would be consuming the last duration the keyboard had the last time it showed, which might have been never (in case this is the first time the keyboard showed), so it would be race condition and unpredictable, so I would rather have a reasonable value here, if its off by a few milliseconds , then its should be perfectly fine!

PR #839, opened 2022-03-28
Session advanced cases

erenbesel:


AdyenSession/AdyenSessionDelegate.swift
```diff
@@ -0,0 +1,78 @@
...
+    ///   - session: The session object.
+    func didFail(with error: Error, from component: Component, session: AdyenSession)
+    
+    /// Returns a handler for handling the payment data submitted by the shopper, that is required for the payments call.
+    /// - Parameters:
+    ///   - component: The current payment component object.
+    ///   - session: The session object.
+    /// - Returns: An instance conforming to the `SessionComponentPaymentsHandler`.
+    /// protocol to take over, or nil to let `AdyenSession` handle it.
+    func handlerForPayments(in component: PaymentComponent, session: AdyenSession) -> AdyenSessionPaymentsHandler?
```
kaphacius, line 49: does it make sense to make this method return non-nil, but make it optional to implement? that feels like a common pattern
same for line 47

erenbesel: Intention here is allowing the merchant to return nil or non-nil at run time. Depending on their internal conditions, they can opt in by returning non nil and opt out by returning nil case by case

AdyenSession/AdyenSession+ActionComponentDelegate.swift
```diff
@@ -25,21 +25,27 @@ extension Session: ActionComponentDelegate {
...
     public func didProvide(_ data: ActionComponentData, from component: ActionComponent) {
-        didProvide(data, currentComponent: component)
+        let handler = delegate?.handlerForAdditionalDetails(in: component, session: self) ?? self
+        handler.didProvide(data, from: component, session: self)
     }
     
-    internal func didProvide(_ data: ActionComponentData, currentComponent: Component) {
-        (currentComponent as? PresentableComponent)?.viewController.view.isUserInteractionEnabled = false
+    public func didOpenExternalApplication(_ component: ActionComponent) {
+        // TODO: call back the merchant
```
mohammedDehairy: this should be forwarded to the delegate, no?

AdyenSession/AdyenSession+PaymentComponentDelegate.swift
```diff
@@ -11,17 +11,39 @@ import Adyen
...
+        handler.didSubmit(data, from: component, session: self)
     }
     
-    internal func didSubmit(_ data: PaymentComponentData, currentComponent: Component) {
+    internal func finish() {
+        // TODO: Handle Finish
+    }
+    
+    internal func finish(with error: Error) {
+        // TODO: Handle Finish
```
mohammedDehairy: this should be forwarded to the delegate, right?

AdyenSession/AdyenSession+PaymentComponentDelegate.swift
```diff
@@ -11,17 +11,39 @@ import Adyen
...
+extension AdyenSession: PaymentComponentDelegate {
     public func didSubmit(_ data: PaymentComponentData, from component: PaymentComponent) {
-        didSubmit(data, currentComponent: component)
+        let handler = delegate?.handlerForPayments(in: component, session: self) ?? self
+        handler.didSubmit(data, from: component, session: self)
     }
     
-    internal func didSubmit(_ data: PaymentComponentData, currentComponent: Component) {
+    internal func finish() {
+        // TODO: Handle Finish
```
mohammedDehairy: this also should be forwarded to the delegate, like `didFinish()` or something?

AdyenSession/AdyenSession.swift
```diff
@@ -64,13 +66,16 @@ public final class Session: SessionProtocol {
...
-    /// Initializes an instance of `Session` asynchronously.
+    /// The delegate object.
+    public weak var delegate: AdyenSessionDelegate?
+    
+    /// Initializes an instance of `SessionComponent` asynchronously.
     /// - Parameter configuration: The session configuration.
     /// - Parameter presentationDelegate: The presentation delegate.
     /// - Parameter completion: The completion closure, that delivers the new instance asynchronously.
     public static func initialize(with configuration: Configuration,
                                   presentationDelegate: PresentationDelegate,
```
mohammedDehairy, line 82: maybe the delegate should be passed here, like the `presentationDelegate` since its required for session to work

AdyenSession/AdyenSession.swift
```diff
@@ -64,13 +66,16 @@ public final class Session: SessionProtocol {
     /// The presentation delegate.
     public private(set) weak var presentationDelegate: PresentationDelegate?
     
-    /// Initializes an instance of `Session` asynchronously.
+    /// The delegate object.
+    public weak var delegate: AdyenSessionDelegate?
```
nauaros: Question: We wanna force the merchant to always set a session delegate right? Cause in that case passing it as a non optional in line 76 makes sense

Adyen/Core/Core Protocols/PaymentComponent.swift
```diff
@@ -39,7 +39,7 @@ extension PaymentComponent {
 /// Describes the methods a delegate of the payment component needs to implement.
 public protocol PaymentComponentDelegate: AnyObject {
     
-    /// Invoked when the payment component finishes, typically by a user submitting their payment details.
+    /// Invoked when the shopper submits the data needed for the payments call.
```
descorp, line 42: This is better :)

Adyen/Core/Core Protocols/PartialPaymentDelegate.swift
```diff
@@ -12,21 +12,26 @@ public protocol PartialPaymentDelegate: AnyObject {
...
+                      component: Component,
                       completion: @escaping (Result<Balance, Error>) -> Void)
 
     /// Invoked when the payment component needs a partial payment order object.
     ///
     /// - Parameters:
+    ///   - component: The current component.
     ///   - completion: The completion closure called when the order object.
-    func requestOrder(_ completion: @escaping (Result<PartialPaymentOrder, Error>) -> Void)
+    func requestOrder(in component: Component,
```
mohammedDehairy: not sure about the `in` 🤔

AdyenSession/AdyenSession+DropInComponentDelegate.swift
```diff
@@ -8,29 +8,35 @@ import Adyen
...
+extension AdyenSession: DropInComponentDelegate {
     public func didSubmit(_ data: PaymentComponentData, from component: PaymentComponent, in dropInComponent: AnyDropInComponent) {
-        didSubmit(data, currentComponent: dropInComponent)
+        let handler = delegate?.handlerForPayments(in: component, session: self) ?? self
+        handler.didSubmit(data, from: component, session: self)
     }
     
     public func didFail(with error: Error, from component: PaymentComponent, in dropInComponent: AnyDropInComponent) {
-        didFail(with: error, currentComponent: dropInComponent)
+        didFail(with: error, currentComponent: component)
```
mohammedDehairy, line 18: should be `dropInComponent`, no? 🤔

AdyenSession/AdyenSession+DropInComponentDelegate.swift
```diff
@@ -8,29 +8,35 @@ import Adyen
...
     }
     
     public func didProvide(_ data: ActionComponentData, from component: ActionComponent, in dropInComponent: AnyDropInComponent) {
-        didProvide(data, currentComponent: dropInComponent)
+        let handler = delegate?.handlerForAdditionalDetails(in: component, session: self) ?? self
+        handler.didProvide(data, from: component, session: self)
```
mohammedDehairy, line 23: > should be dropInComponent, no? 🤔

also here?

AdyenSession/AdyenSession+DropInComponentDelegate.swift
```diff
@@ -8,29 +8,35 @@ import Adyen
...
     }
     
     public func didComplete(from component: ActionComponent, in dropInComponent: AnyDropInComponent) {
         didComplete(currentComponent: dropInComponent)
     }
     
     public func didFail(with error: Error, from component: ActionComponent, in dropInComponent: AnyDropInComponent) {
-        didFail(with: error, currentComponent: dropInComponent)
+        didFail(with: error, currentComponent: component)
```
mohammedDehairy, line 31: > should be dropInComponent, no? 🤔

also here?

AdyenSession/AdyenSession.swift
```diff
@@ -64,32 +69,40 @@ public final class Session: SessionProtocol {
...
```
mohammedDehairy: maybe make it `private(set)`?

Adyen/Core/Core Protocols/PartialPaymentDelegate.swift
erenbesel: Hmm, what would you suggest?

mohammedDehairy: just `func requestOrder(_ component: Component,` is fine I guess

nauaros (approved):

mohammedDehairy (approved):

PR #840, opened 2022-03-29
COIOS-504: Added Atome Component Implementation

neelSharma12:
This PR contains Atome Component Implementation below is the current UI screenshot: 
AtomeComponentUI

Things pending: 
1. I will raise separate PR for UI changes related to billing address. 
2. Currently, this PR contains only unit test cases. I will be raising separate PR for to separate the Unit tests from  UI tests by creating separate target in the SDK.

### DIFF
--- a/AdyenComponents/Atome/AtomeComponent.swift
+++ b/AdyenComponents/Atome/AtomeComponent.swift
@@ -58,7 +58,6 @@ public final class AtomeComponent: AbstractPersonalInformationComponent {
     private func setupItems() {
         personalDetailsHeaderItem.text = localizedString(.boletoPersonalDetails, configuration.localizationParameters)
         phoneItem?.title =  localizedString(.phoneNumberTitle, configuration.localizationParameters)
-        configuration.shopperInformation?.deliveryAddress?.country = "SG"
     }
 
     // MARK: - Public
--- a/AdyenComponents/Atome/AtomeComponent.swift
+++ b/AdyenComponents/Atome/AtomeComponent.swift
@@ -68,12 +68,12 @@ public final class AtomeComponent: AbstractPersonalInformationComponent {
     }
 
     /// :nodoc:
-    override public func createPaymentDetails() -> PaymentMethodDetails {
+    override public func createPaymentDetails() throws -> PaymentMethodDetails {
         guard let firstName = firstNameItem?.value,
               let lastName = lastNameItem?.value,
               let telephoneNumber = phoneItem?.value,
               let billingAddress = addressItem?.value else {
-            fatalError("There seems to be an error in the BasicPersonalInfoFormComponent configuration.")
+                  throw UnknownError(errorDescription: "There seems to be an error in the BasicPersonalInfoFormComponent configuration")
         }
 
         let shopperName = ShopperName(firstName: firstName, lastName: lastName)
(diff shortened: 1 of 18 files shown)


mohammedDehairy (dismissed):

nauaros (dismissed):

AdyenComponents/Atome/AtomeComponent.swift
```diff
@@ -0,0 +1,93 @@
...
+        localizedString(.continueTitle, configuration.localizationParameters)
+    }
+
+    /// :nodoc:
+    override public func createPaymentDetails() -> PaymentMethodDetails {
+        guard let firstName = firstNameItem?.value,
+              let lastName = lastNameItem?.value,
+              let telephoneNumber = phoneItem?.value,
+              let billingAddress = addressItem?.value else {
+            fatalError("There seems to be an error in the BasicPersonalInfoFormComponent configuration.")
```
kaphacius: maybe it's better to have this as an `assertionFailure`, rather than `fatalError`

kaphacius (dismissed):
Should we have the `Internal` submodule tracked by git?

AdyenComponents/Atome/AtomeComponent.swift
neelSharma12: @kaphacius assertionFailure() do not guarantee that execution will stop if the assert is triggered and I guess we would like to stop the execution as all this information is required for the payment. Please let me know if I am wrong here.

mohammedDehairy: we should't crash the merchant app, so I am afraid Yurii has a point, `assertionFailure` makes more sense IMO

neelSharma12: Okay then I will change this to `assertionFailure`. Thanks for pointing this out.

descorp: we have `Adyen. assertionFailure()` for this.
But I would make this check on `didSubmit` so we can return it to shopper as validation error

neelSharma12: I have modified the code as per @descorp suggestion.

nauaros (approved):

mohammedDehairy (approved):

erenbesel (approved):

AdyenComponents/Atome/AtomeComponent.swift
```diff
@@ -68,12 +68,12 @@ public final class AtomeComponent: AbstractPersonalInformationComponent {
...
 
     /// :nodoc:
-    override public func createPaymentDetails() -> PaymentMethodDetails {
+    override public func createPaymentDetails() throws -> PaymentMethodDetails {
         guard let firstName = firstNameItem?.value,
               let lastName = lastNameItem?.value,
               let telephoneNumber = phoneItem?.value,
               let billingAddress = addressItem?.value else {
-            fatalError("There seems to be an error in the BasicPersonalInfoFormComponent configuration.")
+                  throw UnknownError(errorDescription: "There seems to be an error in the BasicPersonalInfoFormComponent configuration")
```
descorp, line 76: In another PR we can apply same behaviour for other `AbstractPersonalInformationComponent` based components..

descorp (approved):

AdyenComponents/Atome/AtomeComponent.swift
neelSharma12: sure.

PR #841, opened 2022-03-30
chore: various fixes around the code

mohammedDehairy:
chore: various fixes around the code

### DIFF
--- a/AdyenComponents/BACS Direct Debit/BACSDirectDebitComponent.swift
+++ b/AdyenComponents/BACS Direct Debit/BACSDirectDebitComponent.swift
@@ -60,19 +60,7 @@ public final class BACSDirectDebitComponent: PaymentComponent, PresentableCompon
     
     internal let inputFormViewController: BACSInputFormViewController
     
-    internal lazy var inputPresenter: BACSInputPresenterProtocol? = {
-        let tracker = BACSDirectDebitComponentTracker(paymentMethod: bacsPaymentMethod,
-                                                      apiContext: apiContext,
-                                                      isDropIn: _isDropIn)
-        let itemsFactory = BACSItemsFactory(styleProvider: configuration.style,
-                                            localizationParameters: configuration.localizationParameters,
-                                            scope: String(describing: self))
-        inputFormViewController.presenter = inputPresenter
-        return BACSInputPresenter(view: inputFormViewController,
-                                  router: self,
-                                  tracker: tracker,
-                                  itemsFactory: itemsFactory)
-    }()
+    internal private(set) var inputPresenter: BACSInputPresenterProtocol?
     
     // MARK: - Initializers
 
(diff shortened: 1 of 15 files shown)


AdyenDropIn/DropInComponent.swift
```diff
@@ -215,14 +215,8 @@ public final class DropInComponent: NSObject,
...
         switch component {
-        case let component as PreApplePayComponent:
-            navigationController.present(asModal: component)
-        case let component as PresentableComponent where component.requiresModalPresentation:
-            navigationController.present(asModal: component)
-        case let component as PresentableComponent where component.viewController is UIAlertController:
-            navigationController.present(component.viewController, customPresentation: false)
         case let component as PresentableComponent:
-            navigationController.present(component.viewController, customPresentation: true)
+            navigationController.present(asModal: component)
```
descorp, line 219: Have you tested following ?

- [x] UIAlertController
- [x] ApplePay
- [x] SafariViewController

AdyenDropIn/DropInComponentExtensions.swift
```diff
@@ -122,7 +122,11 @@ extension DropInComponent: FinalizableComponent {
 
     public func didFinalize(with success: Bool, completion: (() -> Void)?) {
         stopLoading()
-        selectedPaymentComponent?.finalizeIfNeeded(with: success, completion: completion)
+        if let selectedPaymentComponent = selectedPaymentComponent {
+            selectedPaymentComponent.finalizeIfNeeded(with: success, completion: completion)
+        } else {
+            completion?()
+        }
```
descorp, line 129: This is good catch..
I am wandering how can we make it bit more simple for integration 🤔

Maybe we can make `finalizeIfNeeded` extension for `Optional<Component>` ?

Demo/Common/IntegrationExampleComponents.swift
```diff
@@ -128,8 +128,12 @@ extension IntegrationExample {
...
     }
 
-    private func handle(_ action: Action) {
-        actionComponent.handle(action)
+    internal func handle(_ action: Action) {
+        if let actionComponent = currentComponent as? ActionHandlingComponent {
+            actionComponent.handle(action)
+        } else {
+            actionComponent.handle(action)
+        }
```
descorp, line 138: I have left my "5 reasons we should not do this" in previous PR 😆

mohammedDehairy: what's the alternative? :-), the old way simply doesn't work, I replied to your comment in the old PR 😁

AdyenDropIn/DropInComponentExtensions.swift
mohammedDehairy: i don't think it will help much, because in order for the extension functions to work, merchants must not unwrap the optional with a "?" otherwise the inner component functions will be invoked and if the component is nil then completion will never be called

descorp: yep, it's not obvious helper 🤔

Demo/Common/IntegrationExampleComponents.swift
descorp: hm.. ok, I see

We can rename `internal lazy var actionComponent: AdyenActionComponent` into `internal lazy var adyenActionComponent: AdyenActionComponent`

OR change local `let` name

```suggestion
        if let dropInAsActionComponent = currentComponent as? ActionHandlingComponent {
            dropInAsActionComponent.handle(action)
        } else {
            actionComponent.handle(action)
        }
```

OR (as we discussed yesterday)

move each integration (DropIn, Component, Session) in a standalone class with it's own set of local variables

AdyenDropIn/DropInComponent.swift
mohammedDehairy: all work Vladimir!

neelSharma12 (approved):

AdyenComponents/BACS Direct Debit/BACSDirectDebitComponent.swift
```diff
@@ -43,34 +43,24 @@ public final class BACSDirectDebitComponent: PaymentComponent, PresentableCompon
     
     /// Component's configuration
     public var configuration: Configuration
+    
+    /// The payment information.
+    public var payment: Payment? {
+        didSet {
+            inputPresenter?.amount = payment?.amount
```
erenbesel, line 50: oh nice

erenbesel (approved):

PR #844, opened 2022-04-06
Apple Pay Payment object

descorp:
### Changes

* resolve duality of `payment` and `paymentSummaryItems`
* fix finalisation issue for ApplePay on iOS 14+
* fix bugs in Pre-ApplePayComponent navigation flow
* update Xcode config to 13.3


Adyen/Assets/cs-CZ.lproj/Localizable.strings
```diff
@@ -75,7 +76,7 @@
 "adyen.amount" = "Částka";
 "adyen.voucher.entity" = "Subjekt";
 "adyen.pix.instructions" = "";
-"adyen.pix.expirationLabel" = "Na zaplacení vám zbývá %@";
+"adyen.pix.expirationLabel" = "Na zaplacení vám zbývá %@";
```
kaphacius, line 79: something happened here? 🤔

Adyen/Assets/Generated/LocalizationKey.swift
```diff
@@ -101,6 +101,8 @@ public struct LocalizationKey {
     public static let emailItemInvalid = LocalizationKey(key: "adyen.emailItem.invalid")
     /// More options
     public static let moreOptions = LocalizationKey(key: "adyen.moreOptions")
+    /// Total
+    public static let applepayTotal = LocalizationKey(key: "adyen.applepay.total")
```
kaphacius: minor thing, maybe this should be `adyen.applePay.total`

AdyenComponents/Apple Pay/ApplePayComponent.swift
```diff
@@ -92,34 +79,32 @@ public class ApplePayComponent: NSObject, PresentableComponent, PaymentComponent
...
     }
 
-    /// Finalizes ApplePay payment after being processed by payment provider.
-    /// - Parameter success: The status of the payment.
     public func didFinalize(with success: Bool, completion: (() -> Void)?) {
         self.resultConfirmed = true
-        finalizeCompletion = completion
-        paymentAuthorizationCompletion?(success ? .success : .failure)
-        paymentAuthorizationCompletion = nil
+        if paymentAuthorizationCompletion == nil {
```
kaphacius: maybe we can use `if let` here, so we don't need to unwrap `paymentAuthorizationCompletion` in line 97

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
```diff
@@ -0,0 +1,149 @@
...
+        return networks
+    }
+
+}
+
+extension PKPaymentNetwork {
+
+    internal var adyenName: String {
+        if self == .masterCard { return "mc" }
+        if #available(iOS 11.2, *), self == .cartesBancaires { return "cartebancaire" }
```
kaphacius, line 145: nice 👍

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
```diff
@@ -0,0 +1,149 @@
...
+extension ApplePayComponent {
+
+    /// Apple Pay component configuration.
+    public struct Configuration {
+
+        /// The context of a current payment. Contains
+        public let applePayPayment: ApplePayPayment
+
+        /// The merchant identifier for apple pay.
+        public var merchantIdentifier: String
```
kaphacius: is it a possible use case that `merchantIdentifier` changes after the configuration is created? maybe the other fields are fine being `var`, because they have default values in the constructor, but `merchantIdenfitier` is a required parameter, so does not look like it will be changed later on. so maybe we can make it a `let` in this case
i'm not sure about the other `var`s, and their use cases.

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
```diff
@@ -0,0 +1,149 @@
...
+            paymentRequest.requiredBillingContactFields = requiredBillingContactFields
+            paymentRequest.requiredShippingContactFields = requiredShippingContactFields
+            paymentRequest.billingContact = billingContact
+            return paymentRequest
+        }
+    }
+
+    // Adyen supports: interac, visa, mc, electron, maestro, amex, jcb, discover, elodebit, elo.
+    // Will support girocard in future versions
+    internal static var defaultNetworks: [PKPaymentNetwork] {
```
kaphacius: can we make this a let, to avoid re-calculation every time it's called

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
```diff
@@ -0,0 +1,149 @@
...
+            networks.append(.eftpos)
+        }
+
+        if #available(iOS 14.0, *) {
+            networks.append(.girocard)
+        }
+
+        if #available(iOS 14.5, *) {
+            networks.append(.mir)
+        }
```
kaphacius, line 121: i'm not sure we need these lines anymore, the scheme is not supported by apple pay at the moment :) 🇺🇦

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
```diff
@@ -0,0 +1,149 @@
...
+            networks = networks.filter { brandsSet.contains($0.adyenName) }
+        }
+        return networks
+    }
+
+}
+
+extension PKPaymentNetwork {
+
+    internal var adyenName: String {
```
kaphacius: maybe `internalName`? don't have a strong opinion on this though

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        return networks
+    }
+
+}
+
+extension ApplePayPaymentMethod {
+
+    internal var supportedNetworks: [PKPaymentNetwork] {
```
kaphacius, line 130: we can't make this a `lazy var`, or a `let`, but maybe we could make this a function to better indicate that there are calculations happening inside. since the data does not change, maybe we could save the result of the function in a new field of `ApplePayComponent`

AdyenComponents/Apple Pay/ApplePayPayment.swift
```diff
@@ -0,0 +1,98 @@
...
+    ///   - summaryItems: The summary items in a payment request—for example, total, tax, discount, or grand total.
+    /// - Throws: `ApplePayComponent.Error.emptySummaryItems` if the summaryItems array is empty.
+    /// - Throws: `ApplePayComponent.Error.negativeGrandTotal` if the grand total is negative.
+    /// - Throws: `ApplePayComponent.Error.invalidSummaryItem` if at least one of the summary items has an invalid amount.
+    /// - Throws: `ApplePayComponent.Error.invalidCountryCode` if the `payment.countryCode` is not a valid ISO country code.
+    /// - Throws: `ApplePayComponent.Error.invalidCurrencyCode` if the `Amount.currencyCode` is not a valid ISO currency code.
+    public init(countryCode: String,
+                currencyCode: String,
+                summaryItems: [PKPaymentSummaryItem]) throws {
+        guard CountryCodeValidator().isValid(countryCode) else {
```
kaphacius, line 41: i wonder if we could make a throwable `isValid` function for `Validator`. and make validators for every condition we have here. we could end up with something like 
```swift
try CurrencyCodeValidator().isValid(currencyCode, onFail: ApplePayComponent.Error.invalidCountryCode)
try CountryCodeValidator().isValid(countryCode, onFail: ApplePayComponent.Error.invalidCurrencyCode)
try NonEmptyValidator(summaryItems, onFail: ApplePayComponent.Error.emptySummaryItems)
```
but that's probably too much :)

AdyenComponents/Apple Pay/ApplePayPayment.swift
```diff
@@ -0,0 +1,98 @@
...
+        guard CurrencyCodeValidator().isValid(currencyCode) else {
+            throw ApplePayComponent.Error.invalidCurrencyCode
+        }
+        guard summaryItems.count > 0 else {
+            throw ApplePayComponent.Error.emptySummaryItems
+        }
+        guard let lastItem = summaryItems.last, lastItem.amount.doubleValue >= 0 else {
+            throw ApplePayComponent.Error.negativeGrandTotal
+        }
+        guard summaryItems.contains(where: { $0.amount.isEqual(to: NSDecimalNumber.notANumber) }) == false else {
```
kaphacius: maybe something like 
```swift
guard summaryItems.map(\.amount).contains(.notANumber) == false else {
    throw ApplePayComponent.Error.invalidSummaryItem
}
```

AdyenComponents/Apple Pay/ApplePayPayment.swift
```diff
@@ -0,0 +1,98 @@
...
+        self.init(payment: payment, summaryItems: [PKPaymentSummaryItem(label: totalString, amount: decimalValue)])
+    }
+
+    private init(payment: Payment, summaryItems: [PKPaymentSummaryItem]) {
+        self.summaryItems = summaryItems
+        self.payment = payment
+    }
+
+    /// The amount for this payment.
+    public var payment: Payment
```
kaphacius: can this be a `let`? since we initialize `summaryItems` based on the `payment`

AdyenDropIn/Presentation/NavigationProtocol.swift
```diff
@@ -0,0 +1,13 @@
...
+// Copyright (c) 2022 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+internal protocol NavigationProtocol: PresentationDelegate {
+
+    func dismiss(completion: (() -> Void)?)
```
kaphacius: maybe `NavigationPresentationDelegate`? or `NavigationPresenter`?

Tests/AdyenTests/Adyen Tests/Components/PresentationDelegateMock.swift
```diff
@@ -2,8 +2,15 @@
 
 import Adyen
 import Foundation
+@testable import AdyenDropIn
 
-final class PresentationDelegateMock: PresentationDelegate {
+final class PresentationDelegateMock: NavigationProtocol {
+
+    var doDismiss: ( ((() -> Void)?) -> Void )?
```
kaphacius, line 9: i think this is getting out of hand 🥲

Adyen/Assets/cs-CZ.lproj/Localizable.strings
descorp: Not sure 😅
But I trust github on this..

AdyenComponents/Apple Pay/ApplePayComponent.swift
```diff
@@ -26,7 +23,17 @@ public class ApplePayComponent: NSObject, PresentableComponent, PaymentComponent
     /// The Apple Pay payment method.
     public var paymentMethod: PaymentMethod { applePayPaymentMethod }
 
-    /// Apple Pay component configuration.
+    public var payment: Payment? {
```
mohammedDehairy, line 29: maybe `private(set)`?

AdyenDropIn/Components/PreApplePayComponent.swift
```diff
@@ -65,24 +56,29 @@ internal final class PreApplePayComponent: PresentableComponent, FinalizableComp
...
-                                                       payment: payment,
                                                        configuration: applePayConfiguration)
         self.applePayComponent.delegate = self
     }
 
     /// :nodoc:
     internal func didFinalize(with success: Bool, completion: (() -> Void)?) {
-        applePayComponent.didFinalize(with: success, completion: completion)
+        applePayComponent.didFinalize(with: success, completion: { [weak self] in
+            let rootViewController = UIApplication.shared.keyWindow?.rootViewController
```
mohammedDehairy: lets not do this again (lets not grab the `UIApplication.shared.keyWindow`) in production code, lets make the `presentationDelegate` required in the initializer instead.

AdyenDropIn/DropInComponentExtensions.swift
```diff
@@ -110,20 +110,25 @@ extension DropInComponent: PreselectedPaymentMethodComponentDelegate {
...
+
 }
 
 extension DropInComponent: FinalizableComponent {
 
     public func didFinalize(with success: Bool, completion: (() -> Void)?) {
         stopLoading()
-        if let selectedPaymentComponent = selectedPaymentComponent {
-            selectedPaymentComponent.finalizeIfNeeded(with: success, completion: completion)
+        if let component = selectedPaymentComponent as? FinalizableComponent {
```
mohammedDehairy: maybe `if let finalizableComponent = selectedPaymentComponent as? FinalizableComponent`?

AdyenDropIn/Presentation/NavigationProtocol.swift
```diff
@@ -0,0 +1,13 @@
+//
+// Copyright (c) 2022 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+internal protocol NavigationProtocol: PresentationDelegate {
```
mohammedDehairy: not sure about the name NavigationProtocol? seems too generic, what's wrong with presentationDelegate? or even better AdyenPresentationDelegate? it is a delegate after all, having the delegate in its name make sense to me

Demo/Common/IntegrationExampleComponents.swift
```diff
@@ -56,13 +56,15 @@ extension IntegrationExample {
     }
 
     internal func presentApplePayComponent() {
-        guard let paymentMethod = paymentMethods?.paymentMethod(ofType: ApplePayPaymentMethod.self) else { return }
-        let config = ApplePayComponent.Configuration(summaryItems: ConfigurationConstants.applePaySummaryItems,
+        guard
+            let paymentMethod = paymentMethods?.paymentMethod(ofType: ApplePayPaymentMethod.self),
+            let applePayPayment = try? ApplePayPayment(payment: payment)
+        else { return }
+        let config = ApplePayComponent.Configuration(payment: applePayPayment,
```
mohammedDehairy, line 63: pleas update the `MIGRATION.md`

Demo/Common/IntegrationExampleDropIn.swift
```diff
@@ -17,9 +17,14 @@ extension IntegrationExample {
...
         guard let paymentMethods = paymentMethods else { return }
+
         let configuration = DropInComponent.Configuration(apiContext: apiContext)
-        configuration.applePay = .init(summaryItems: ConfigurationConstants.applePaySummaryItems,
-                                       merchantIdentifier: ConfigurationConstants.applePayMerchantIdentifier)
+
+        if let applePayPayment = try? ApplePayPayment(payment: payment) {
+            configuration.applePay = .init(payment: applePayPayment,
+                                           merchantIdentifier: ConfigurationConstants.applePayMerchantIdentifier)
+        }
```
mohammedDehairy, line 26: pleas update the `MIGRATION.md`

AdyenComponents/Apple Pay/ApplePayComponent.swift
descorp: 👍

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
descorp: 👍

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
descorp: ~~it is static calculated property - it is always lazy.~~

**UPD**

changed to `internal static var defaultNetworks: [PKPaymentNetwork] = { .. }()`

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
descorp: I'll gladly remove it 😄

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
descorp: Renamed to `txVariantName`

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
descorp: > save the result of the function in a new field of ApplePayComponent

this is good one

AdyenComponents/Apple Pay/ApplePayPayment.swift
descorp: Maybe another PR 😅

AdyenComponents/Apple Pay/ApplePayPayment.swift
descorp: 👍

AdyenComponents/Apple Pay/ApplePayPayment.swift
descorp: it is more complicated then that 😅
Check next commit

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
mohammedDehairy: > save the result of the function in a new field of ApplePayComponent

we shouldn't bother, the complication of saving the result in a property is not worth the one femto second in performance  we gain :-D

AdyenDropIn/Presentation/NavigationProtocol.swift
descorp: My general idea was: "Navigation" = "Presentation" + "Dismission" 🤔

I can do 

```
/// Delegates `ViewController`'s dismission.
internal protocol DismissionDelegate: AnyObject {

    /// Asks the `delegate` to dismiss itself as the `delegate` sees fit.
    func dismiss(completion: (() -> Void)?)
}

internal typealias NavigationDelegate = PresentationDelegate & DismissionDelegate
```

Tests/AdyenTests/Adyen Tests/Components/PresentationDelegateMock.swift
descorp: This is just a mock 😅
I need to test, that if I call `didFinalize` - I got `completion` back

AdyenComponents/Apple Pay/ApplePayComponent.swift
descorp: it is part of `PaymentComponent `.
Check updated version

AdyenDropIn/Components/PreApplePayComponent.swift
descorp: It is not about that. I need to check if `PKPaymentViewController` is currently on a screen.
Do you know a better way?

AdyenDropIn/DropInComponentExtensions.swift
descorp: 👍

AdyenDropIn/Presentation/NavigationProtocol.swift
descorp: It is internal :)
Check updated version

Demo/Common/IntegrationExampleComponents.swift
descorp: 👍

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
descorp: I have persisted it with `paymentRequest` - it is not changing during the component lifecycle

AdyenDropIn/DropInComponent.swift
```diff
@@ -239,6 +239,7 @@ public final class DropInComponent: NSObject,
     internal func userDidCancel(_ component: Component) {
         stopLoading()
         component.cancelIfNeeded()
+        component.finalizeIfNeeded(with: false, completion: nil)
```
mohammedDehairy: `finalizeIfNeeded` is not supposed to be called with cancellation, so maybe make `PreApplePayComponent` `Cancellable`

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
descorp: As discussed with Marc, we are going to keep it for now. 
Worth case scenario - it will be ON in 2036..

Adyen/Assets/Generated/LocalizationKey.swift
descorp: 👍

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
```diff
@@ -15,31 +15,31 @@ extension ApplePayComponent {
...
 
         /// A pre-populated billing address.
-        public var billingContact: PKContact?
+        public let billingContact: PKContact?
 
         /// The flag to toggle onboarding.
         /// If true, allow the shopper to add cards to Apple Pay if non exists yet.
         /// If false, then Apple Pay is disabled if the shopper doesn't have supported cards on Apple Pay wallet.
-        public var allowOnboarding: Bool
+        public let allowOnboarding: Bool
```
mohammedDehairy, line 38: I would keep those as vars with default values, to enable merchant to initialize the object and set only the properties they want .

descorp: For `billingContact` or `requiredShippingContactFields` - this is true.
I can do that in another PR

However, `allowOnboarding` we use on `ApplePayComponent.init?` - it is defining if it is possible or not to create an instance of component.
If we make it `var` - it could be changed at any point of time which could lead to inconsistency.

descorp: For `billingContact` or `requiredShippingContactFields` - this is true.
I can do that in another PR

However, `allowOnboarding` we use on `ApplePayComponent.init?` - it is defining if it is possible or not to create an instance of component.
Changing this afterward does not make much sense and could lead to mistakes 🤔

mohammedDehairy (dismissed):

AdyenDropIn/Components/PreApplePayComponent.swift
```diff
@@ -65,27 +59,30 @@ internal final class PreApplePayComponent: PresentableComponent, FinalizableComp
...
                                                        apiContext: apiContext,
-                                                       payment: payment,
                                                        configuration: applePayConfiguration)
         self.applePayComponent.delegate = self
     }
 
-    /// :nodoc:
+    internal func didCancel() {
+        let rootViewController = UIApplication.shared.keyWindow?.rootViewController
+        if let navigation = presentationDelegate,
```
erenbesel: Why is having the `presentationDelegate` not enough and we have to check for the `rootViewController`?
Could be nice to add this as a comment in the code

descorp: I have changed it a bit - it should be more clear now :)

mohammedDehairy (dismissed):

nauaros (approved):

mohammedDehairy (approved):

PR #847, opened 2022-04-08
fix: PaymentMethods decoding with null values

TParizek:
## Summary
PaymentMethods decoding uses the `contains` function of `KeyedDecodingContainer` to check if the decoding object contains some value. The problem is that `contains` returns true even for null value, which leads to not intended behaviour. For example, if a payment method json object contains `issuers: null`, the SDK will try to parse it with the `IssuerListPaymentMethodDecoder` which is incorrect.

I've created a new extension function `containsValue` that check if there is a non null value associated with a given key.

## Tested scenarios
I've created the `testDecodingPaymentMethodsWithNullValues` test function, which check json parsing with null values.


descorp:
Thanks @TParizek !
This looks really nice 💚

mohammedDehairy (approved):

descorp (approved):

PR #850, opened 2022-04-12
Deprecate live environment

descorp:
## Open PR


Adyen/Core/APIClient/Environment.swift
```diff
@@ -6,42 +6,37 @@
...
     /// :nodoc:
     public static let beta = Environment(baseURL: URL(string: "https://checkoutshopper-beta.adyen.com/")!)
-    
-    /// :nodoc:
+
     /// Adyen's default live environment.
-    public static let live = Environment(baseURL: URL(string: "https://checkoutshopper-live.adyen.com/")!)
-    
-    /// :nodoc:
+    @available(*, deprecated, message: "This property no longer supported. Please select matching environment.")
```
mohammedDehairy: `"This property no longer supported. Please explicitly select the environment matching your region."`

mohammedDehairy (approved):

Adyen/Core/APIClient/Environment.swift
erenbesel: Let's not forget the `is` as in `This property is no longer...`

neelSharma12 (approved):

PR #864, opened 2022-04-20
Improved error handling

neelSharma12:
## Summary
This PR contains refactored code for better error handling. 

## Tested scenarios
I have tested below components: 
1. AffirmComponent
2. BoletoComponent
3. DokuComponent
4. MBWayComponent
5. QiwiWalletComponent

**Fixed issue**: 
[COIOS-537](https://youtrack.is.adyen.com/issue/COIOS-537)

### DIFF
--- a/AdyenComponents/Boleto/BoletoComponent.swift
+++ b/AdyenComponents/Boleto/BoletoComponent.swift
@@ -87,7 +87,15 @@ public final class BoletoComponent: PaymentComponent, LoadingComponent, Presenta
                                       apiContext: apiContext,
                                       fields: formFields,
                                       configuration: configuration,
-                                      onCreatePaymentDetails: { [weak self] in self?.createPaymentDetails() })
+                                      onCreatePaymentDetails: { [weak self] in
+            var paymentMethodDetails: PaymentMethodDetails?
+            do {
+                paymentMethodDetails = try self?.createPaymentDetails()
+            } catch let error {
+                adyenPrint(error)
+            }
+            return paymentMethodDetails
+        })
 
         if let emailItem = component.emailItem {
             bind(sendCopyByEmailItem.publisher, to: emailItem, at: \.isHidden.wrappedValue, with: { !$0 })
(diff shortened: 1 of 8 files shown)


Tests/AdyenTests/Adyen Tests/Components/Affirm/AffirmComponentTests.swift
```diff
@@ -58,7 +58,7 @@ class AffirmComponentTests: XCTestCase {
         sut.deliveryAddressToggleItem.value = false
         
         // When
-        let paymentDetails = sut.createPaymentDetails()
+        let paymentDetails = try sut.createPaymentDetails()
```
erenbesel, line 61: Now that this function can throw, let's test a few failed scenarios with missing data

AdyenComponents/Boleto/BoletoComponent.swift
```diff
@@ -87,7 +87,15 @@ public final class BoletoComponent: PaymentComponent, LoadingComponent, Presenta
...
                                       configuration: configuration,
-                                      onCreatePaymentDetails: { [weak self] in self?.createPaymentDetails() })
+                                      onCreatePaymentDetails: { [weak self] in
+            var paymentMethodDetails: PaymentMethodDetails?
+            do {
+                paymentMethodDetails = try self?.createPaymentDetails()
+            } catch let error {
+                adyenPrint(error)
+            }
+            return paymentMethodDetails
```
descorp, line 97: So in case of error we are going to print message and then use [InstantPaymentDetails(type: paymentMethod.type)](https://github.com/Adyen/adyen-ios/pull/864/files#diff-1b2fc8c8f5c7035f9237a992a77bcb221aba88bf22a1896d8cf5a2bf871c7da1L232) that will (most likely) fail on `\payment`.

I guess we need to refactor Boleto component entirely (in another PR) 🤔

descorp (approved):

AdyenComponents/Boleto/BoletoComponent.swift
```diff
@@ -87,7 +87,15 @@ public final class BoletoComponent: PaymentComponent, LoadingComponent, Presenta
                                       apiContext: apiContext,
                                       fields: formFields,
                                       configuration: configuration,
-                                      onCreatePaymentDetails: { [weak self] in self?.createPaymentDetails() })
+                                      onCreatePaymentDetails: { [weak self] in
+            var paymentMethodDetails: PaymentMethodDetails?
+            do {
+                paymentMethodDetails = try self?.createPaymentDetails()
+            } catch let error {
+                adyenPrint(error)
```
erenbesel, line 95: Hmm, would printing the error be enough? 🤔

descorp: Not really. See [my comment](https://github.com/Adyen/adyen-ios/pull/864#discussion_r854081354) above
In addition - I am not sure that we can even test it. Since this scenario is caused by "Optional handling", rather then an actual business edge case..

erenbesel: Yeah, seems we need some refactoring

mohammedDehairy: I thinK we can make the closure throw, right?

mohammedDehairy (approved):

PR #852, opened 2022-04-13
feature: Add support for changing the store payment method field visibility

eduardourso:
## Summary
There was no support for changing the visibility of the store payment method switch after the CardComponent was presented. 

Internally there was a need to be able to hide/show the store payment method field based on the card brand/type.

162455160-9b2b518b-0a10-40f8-a799-31028b11fec7

PR #875, opened 2022-04-29
Add functionality to update the store payment component value

eduardourso:
## Summary

In addition to the changes introduced on #852, I would like to suggest another update to the CardComponent to add the functionality to change the value of the store payment method component.

This allows consumers of the SDK to control the value of the store payment method component as they seem fit without user input.

In our specific use case, there was a need to start the CreditCardComponent with the store credit card component on by default.


AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -301,7 +302,10 @@ extension FormValueItem where ValueType == String {
 
 extension CardViewController: CardViewControllerProtocol {
     func update(storePaymentMethodFieldVisibility isVisible: Bool) {
-        items.storeDetailsItem.value = false
```
erenbesel, line 304: Thanks again for your PR @eduardourso !
It would be logical to keep this line. If the switch is hidden, we should not store the info for next time.

descorp:
Hey @eduardourso 

Thanks for your contribution!
I am not sure if this is a good/legal UX - as a shopper, I should be aware and provide explicit consent if my Payment is going to be retained.

We will get back to you with more weighted decision..

AdyenCard/Components/Card/CardViewController.swift
eduardourso: sure, that makes sense. I'll revert this change.

eduardourso: @erenbesel I've updated it here, to consider the visibility: [a536d8c](https://github.com/Adyen/adyen-ios/pull/875/commits/a536d8cfb27152031e6c4a52287abf2b63795dfd)

eduardourso:
@descorp I do understand that, but these changes are not enforcing the behavior, it would just add the option for SDK consumers to do it or not, nevertheless, I understand it might involve internal decisions. 

I appreciate the help so far and we are looking forward to hearing back from you.

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -301,7 +302,13 @@ extension FormValueItem where ValueType == String {
...
+        if !isVisible {
+            items.storeDetailsItem.value = false
+        }
         items.storeDetailsItem.isVisible = isVisible
     }
+
+    func update(storePaymentMethodFieldValue isOn: Bool) {
+        items.storeDetailsItem.value = isOn
```
descorp: With this we can always be sure that shopper give explicit consent:

```suggestion
        items.storeDetailsItem.value = items.storeDetailsItem.isVisible && isOn
```

eduardourso: are you referring to this: [a536d8c](https://github.com/Adyen/adyen-ios/pull/875/commits/a536d8cfb27152031e6c4a52287abf2b63795dfd) ?

eduardourso: Or is the idea here to also pass the component visibility?

descorp: Sorry, I have updated `suggested change`
The idea here is to prevent storing shopper's payment without shopper's consent (visual awareness).

eduardourso: done here [aedc3c8](https://github.com/Adyen/adyen-ios/pull/875/commits/aedc3c84eb778539962dcdc06465348e8294da20)

descorp:
Thanks for your contribution @eduardourso 💚
I'll fix the "Spell Check" script and we will merge this PR

descorp:
**Disclaimer** 

If calling `update(storePaymentMethodFieldValue:true)`, the checkbox for storing payment details becomes enabled by default. The shopper can still opt out in the UI. 
It is your responsibility to make sure that this setting is compliant with local regulations.

descorp:
Hey @eduardourso 

Can you trigger "update base branch" ?
Since it is a fork - we can't trigger it ourself.

eduardourso:
@descorp sorry it took so long, just triggered it now.

descorp (approved):

PR #884, opened 2022-05-11
Replace nil bundle in super.init for ViewControllers

descorp:
### Changes

* add bundle to UIViewController init to prevent automatic view inflation in case of matching names


neelSharma12 (approved):

nauaros (approved):
Nice!

PR #888, opened 2022-05-16
highlight detected cards among supported card icons

erenbesel:
## Tested scenarios

**Fixed issue**:

### DIFF
--- a/AdyenCard/Form/FormCardNumberContainerItem.swift
+++ b/AdyenCard/Form/FormCardNumberContainerItem.swift
@@ -55,9 +55,7 @@ internal final class FormCardNumberContainerItem: FormItem, AdyenObserver {
     }
     
     internal func update(brands: [CardBrand]) {
-        // reduce alpha of supported card icons if brand is detected
-        let supportedBrands = brands.filter(\.isSupported)
-        supportedCardLogosItem.alpha = supportedBrands.isEmpty ? 1 : 0.3
+        supportedCardLogosItem.update(brands: brands)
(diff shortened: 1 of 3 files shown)


AdyenCard/Form/FormCardNumberContainerItem.swift
```diff
@@ -71,22 +69,32 @@ internal final class FormCardLogosItem: FormItem, Hidable {
...
-    
     internal let style: FormTextItemStyle
     
-    /// Observable property to update the owner view's alpha.
-    @AdyenObservable(1) internal var alpha: CGFloat
+    @AdyenObservable([]) internal var cardLogos: [CardTypeLogo]
     
     internal init(cardLogos: [CardTypeLogo], style: FormTextItemStyle) {
-        self.cardLogos = cardLogos
         self.style = style
```
neelSharma12, line 77: why are we deleting and adding "self.cardLogos = cardLogos" ?

erenbesel: It's just changing the order, not deleting. Now the `cardLogos` has a default value, all properties without a value must be set first in the init in order to access self so I moved it below.

mohammedDehairy (approved):

AdyenCard/Form/FormCardNumberContainerItem.swift
neelSharma12: make sense :)

neelSharma12 (approved):

PR #894, opened 2022-05-19
Card component validations

neelSharma12:
This PR resolved validation issues on card component's. Below is the youtrack ticket for the same.
https://youtrack.is.adyen.com/issue/COIOS-457

### DIFF
--- a/Tests/AdyenTests/Card Tests/Validators/CardExpiryDateValidatorTests.swift
+++ b/Tests/AdyenTests/Card Tests/Validators/CardExpiryDateValidatorTests.swift
@@ -24,11 +24,8 @@ class CardExpiryDateValidatorTests: XCTestCase {
         // Test date 30 years in the future from reference date 01/01/2020.
         XCTAssertTrue(validator.isValid("0150"))
         
-        // Test date less than 3 months in the past from reference date 01/01/2020.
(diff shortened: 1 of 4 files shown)


Tests/AdyenTests/Card Tests/Validators/CardExpiryDateValidatorTests.swift
```diff
@@ -24,11 +24,8 @@ class CardExpiryDateValidatorTests: XCTestCase {
         // Test date 30 years in the future from reference date 01/01/2020.
         XCTAssertTrue(validator.isValid("0150"))
         
-        // Test date less than 3 months in the past from reference date 01/01/2020.
+        // Test date 3 months in the past from reference date 01/01/2020.
         XCTAssertTrue(validator.isValid("1019"))
-        
-        // Test date 3 months from reference date 01/01/2020.
-        XCTAssertTrue(validator.isValid("0919"))
```
erenbesel, line 31: Instead of deleting we should update the test to reflect the month change. So if we change those numbers again these tests should fail

neelSharma12: As discussed, I have made the change.

erenbesel (approved):

mohammedDehairy (approved):

PR #927, opened 2022-06-21
Made card component address fields optional upon some card brands detection according to merchant configurations

mohammedDehairy:
Made card component address fields optional upon some card brands detection according to merchant configurations

### DIFF
--- a/Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
+++ b/Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
@@ -12,6 +12,12 @@ public final class FormAddressItem: FormValueItem<PostalAddress, AddressStyle>, 
     
     public var isHidden: AdyenObservable<Bool> = AdyenObservable(false)
     
+    public var context: AddressViewModelBuilderContext {
+        didSet {
+            reloadFields()
+        }
+    }
+    
     private var items: [FormItem] = []
     
     private let localizationParameters: LocalizationParameters?
@@ -58,14 +64,15 @@ public final class FormAddressItem: FormValueItem<PostalAddress, AddressStyle>, 
         self.localizationParameters = localizationParameters
         self.supportedCountryCodes = supportedCountryCodes
         self.addressViewModelBuilder = addressViewModelBuilder
+        self.context = .init(countryCode: initialCountry, isOptional: false)
         super.init(value: PostalAddress(), style: style)
 
         self.identifier = identifier
-        update(for: initialCountry)
+        reloadFields()
         
         bind(countrySelectItem.publisher, at: \.identifier, to: self, at: \.value.country)
         observe(countrySelectItem.publisher, eventHandler: { [weak self] event in
-            self?.update(for: event.element.identifier)
+            self?.context.countryCode = event.element.identifier
         })
     }
     
@@ -91,9 +98,9 @@ public final class FormAddressItem: FormValueItem<PostalAddress, AddressStyle>, 
     
     // MARK: - Private
     
-    private func update(for countryCode: String) {
-        let subRegions = RegionRepository.subRegions(for: countryCode)
-        let viewModel = addressViewModelBuilder.build(countryCode: countryCode)
+    private func reloadFields() {
+        let subRegions = RegionRepository.subRegions(for: context.countryCode)
+        let viewModel = addressViewModelBuilder.build(context: context)
         
         items = [FormSpacerItem(),
                  headerItem.addingDefaultMargins(),
--- a/Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
+++ b/Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
@@ -175,21 +175,27 @@ public final class FormAddressItem: FormValueItem<PostalAddress, AddressStyle>, 
         let item = item
         switch field {
         case .street:
+            item.value = value.street ?? ""
             observers[.street] = bind(item.publisher, to: self, at: \.value.street)
             publisherObservers[.street] = observe(publisher) { item.value = $0.street ?? "" }
         case .houseNumberOrName:
+            item.value = value.houseNumberOrName ?? ""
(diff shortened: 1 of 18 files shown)


Adyen/UI/Form/Items/Address/FormPostalCodeItem.swift
```diff
@@ -0,0 +1,43 @@
...
+        contentType = .postalCode
+    }
+    
+    public func updateOptionalStatus(isOptional: Bool) {
+        // when optional, if user enters anything it should be validated as regular entry.
+        if isOptional {
+            title = localizedString(.postalCodeFieldTitle,
+                                    localizationParameters) + " " + localizedString(.fieldTitleOptional,
+                                                                                    localizationParameters)
+            validator = NumericStringValidator(exactLength: 0) || LengthValidator(minimumLength: 2, maximumLength: 30)
```
descorp: `NumericStringValidator(exactLength: 0)` why? 
you can do just `LengthValidator(minimumLength: 0, maximumLength: 30)`, right ?

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -165,6 +193,15 @@ internal class CardViewController: FormViewController {
...
         issuingCountryCode = binInfo.issuingCountryCode
         items.numberContainerItem.update(brands: brands)
+        
+        updateBillingAddressOptionalStatus(brands: brands)
+    }
+    
+    private func updateBillingAddressOptionalStatus(brands: [CardBrand]) {
+        let optionalBrands = configuration.billingAddress.optionalForBrands
+        let isOptional = optionalBrands.isDisjoint(with: brands.map(\.type)) == false
+        items.billingAddressItem.context.isOptional = isOptional
```
descorp: what if `items.billingAddressItem` was not created (`billingAddressMode == .none`)?
I assume it will be created in memory, but not visible and then appear in `PaymentData` json as empty `billingaddress`?
Maybe we need IF or switch here?

Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
```diff
@@ -156,9 +163,20 @@ public final class FormAddressItem: FormValueItem<PostalAddress, AddressStyle>,
         return item
     }
     
-    private func bind(item: FormRegionPickerItem, to field: AddressField) {
+    private func bind(item: FormRegionPickerItem, to field: AddressField, subRegions: [Region]) {
         observers[field].map(remove)
+        publisherObservers[field].map(remove)
         observers[field] = bind(item.publisher, at: \.identifier, to: self, at: \.value.stateOrProvince)
+        func update(address: PostalAddress) {
```
descorp, line 174: neat 👌

AdyenCard/Components/Card/CardViewController.swift
mohammedDehairy: is mode == .none, it will never show up in the PaymentData :-)

Adyen/UI/Form/Items/Address/FormPostalCodeItem.swift
mohammedDehairy: fixed

AdyenCard/Components/Card/CardViewController.swift
descorp: > it will never show up in the PaymentData

this is good.

However `updateBillingAddressOptionalStatus(brands:)` does trigger creation and apply changes to both `items.billingAddressItem` and `items.postalCodeItem` multiple time, even though they are not visible on a screen?

I see there a minor memory issue; also because we are using bindings - I am a bit causes that this might have unpredictable side effects in future 🤔

AdyenCard/Components/Card/CardComponentConfiguration.swift
```diff
@@ -6,6 +6,24 @@
 
 @_spi(AdyenInternal) import Adyen
 
+/// Billing address fields configurations
+public struct BillingAddressConfiguration {
```
descorp, line 10: This is nice idea, BTW :)

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -165,6 +193,15 @@ internal class CardViewController: FormViewController {
...
+        items.postalCodeItem.updateOptionalStatus(isOptional: isOptional)
```
descorp:
```suggestion
    private func updateBillingAddressOptionalStatus(brands: [CardBrand]) {
        guard configuration.billingAddress.mode != .none else { return }
        let optionalBrands = configuration.billingAddressOptionalForBrands
        let isOptional = optionalBrands.isDisjoint(with: brands.map(\.type)) == false
        if configuration.billingAddress.mode == .full {
             items.billingAddressItem.context.isOptional = isOptional
        } else {
             items.postalCodeItem.updateOptionalStatus(isOptional: isOptional)
        }
```

mohammedDehairy: fixed

AdyenCard/Components/Card/CardComponentConfiguration.swift
mohammedDehairy: Thanks! :-)

Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
```diff
@@ -96,6 +96,10 @@ public final class FormAddressItem: FormValueItem<PostalAddress, AddressStyle>,
         return item
     }()
     
+    public func updateOptionalStatus(isOptional: Bool) {
```
descorp, line 99: This is nice 👌

descorp (dismissed):

Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
mohammedDehairy: ❤️

neelSharma12 (approved):

descorp (approved):

PR #929, opened 2022-06-23
fix: fixes a bug where the amount to be paid shown in UI reflects thetotal amount instead of the remaining amount in partial payments flow

mohammedDehairy:
fix: fixes a bug where the amount to be paid shown in UI reflects thetotal amount instead of the remaining amount in partial payments flow


descorp (approved):

erenbesel (commented):
Wait I fixed this on my current task as we talked before :D or is it something different? My fix was different so let's discuss on Monday

neelSharma12 (approved):

erenbesel (approved):

PR #932, opened 2022-06-28
v4: applepay payment + giftcard compatibility

descorp:
# Open PR

## Changes

* fix missed `payment` for ApplePayComponent
* fix gift-card flow compatibility for ApplePay on DropIn

### DIFF
--- a/AdyenComponents/Apple Pay/ApplePayComponent.swift
+++ b/AdyenComponents/Apple Pay/ApplePayComponent.swift
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2021 Adyen N.V.
+// Copyright (c) 2022 Adyen N.V.
 //
 // This file is open source and available under the MIT license. See the LICENSE file for more info.
 //
@@ -15,7 +15,7 @@ public class ApplePayComponent: NSObject, PresentableComponent, PaymentComponent
     internal var success: Bool = false
     
     /// :nodoc:
-    private let payment: Payment
+    internal let internalPayment: Payment
 
     /// :nodoc:
     internal let applePayPaymentMethod: ApplePayPaymentMethod
@@ -37,6 +37,11 @@ public class ApplePayComponent: NSObject, PresentableComponent, PaymentComponent
     
     /// The delegate of the component.
     public weak var delegate: PaymentComponentDelegate?
+
+    public var payment: Payment? {
+        get { internalPayment }
+        set {} // swiftlint:disable:this unused_setter_value
+    }
     
     /// Initializes the component.
     /// - Warning: didFinalize() must be called before dismissing this component.
@@ -92,9 +97,10 @@ public class ApplePayComponent: NSObject, PresentableComponent, PaymentComponent
         self.apiContext = apiContext
         self.paymentAuthorizationViewController = viewController
         self.applePayPaymentMethod = paymentMethod
-        self.payment = payment
+        self.internalPayment = payment
         super.init()
 
+        self.payment = internalPayment
         viewController.delegate = self
     }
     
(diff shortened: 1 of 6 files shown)


neelSharma12 (dismissed):

AdyenComponents/Apple Pay/ApplePayComponentExtensions.swift
```diff
@@ -115,6 +115,18 @@ extension ApplePayComponent {
             paymentRequest.billingContact = billingContact
             return paymentRequest
         }
+
+        public mutating func update(amount: Amount, localeIdentifier: String?) {
+            var newItems = summaryItems
+            guard let lastItem = newItems.last else { return }
+
+            newItems = newItems.dropLast()
+            let newAmount = AmountFormatter.decimalAmount(amount.value,
```
erenbesel: Can't we use the new amount object as is? Why do we need to specify localeIdentifier explicitly?

AdyenComponents/Apple Pay/ApplePayComponent.swift
```diff
@@ -37,6 +37,11 @@ public class ApplePayComponent: NSObject, PresentableComponent, PaymentComponent
     
     /// The delegate of the component.
     public weak var delegate: PaymentComponentDelegate?
+
+    public var payment: Payment? {
+        get { internalPayment }
+        set {} // swiftlint:disable:this unused_setter_value
```
erenbesel, line 43: What's the reason we are not setting `internalPayment` here?

AdyenComponents/Apple Pay/ApplePayComponentExtensions.swift
descorp: `newAmount` is NSDecimalNumber.
I can rename the variable to `decimalAmount` - make it more clear?

AdyenComponents/Apple Pay/ApplePayComponent.swift
descorp: Because we would also have to update `summaryItems` and recreate a `viewController`.
This change is only feasible on pre-applepay component on DropIn

AdyenComponents/Apple Pay/ApplePayComponentExtensions.swift
erenbesel: Got it. What about `localeIdentifier`? Amount also has this property so do we need to set it from external parameter?

descorp: hm.. this is good question which one we should use here 🤔

**UPD**

I think `localeIdentifier` on amount is probably always empty, since this is `remainingAmount`, that we receive from backend 🤔
but I can add the check JIC

AdyenComponents/Apple Pay/ApplePayComponent.swift
mohammedDehairy: maybe an assertionFailure?

AdyenDropIn/Components/PreApplePayComponent.swift
```diff
@@ -19,8 +19,10 @@ internal final class PreApplePayComponent: Localizable, PresentableComponent, Fi
     /// :nodoc:
     internal let paymentMethod: PaymentMethod
     
-    /// :nodoc:
-    private var payment: Payment? { _payment }
+    internal var payment: Payment? {
+        get { _payment }
+        set {} // swiftlint:disable:this unused_setter_value
```
mohammedDehairy: maybe assertionFailure?

AdyenComponents/Apple Pay/ApplePayComponent.swift
descorp: It will break every integration since v4.0.0 😅
Will park for now..

AdyenComponents/Apple Pay/ApplePayComponentExtensions.swift
descorp: I think `localeIdentifier` on amount is probably always empty, since this is `remainingAmount`, that we receive from backend 🤔

mohammedDehairy (approved):

erenbesel (approved):

PR #933, opened 2022-06-28
Remove forcing uppercase

Jack25apr:
## Tested scenarios

**Fixed issue**:


Jack25apr:
sorry, seems work now.

PR #940, opened 2022-06-30
Payment refactoring

descorp:
# Open PR

## Changes

* split PaymentComponent and PaymentAwareComponent
* move `Payment` to AdyenContext
* simplify `submit ` method
* remove `remainingAmount` from `ComponentManager`
* rename `PaymentAwareComponent` -> `PaymentAware`
* simplify ApplePayPayment struct
* rename `PaymentComponentData.amount` -> `PaymentComponentData.amountToPay`


AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
```diff
@@ -30,7 +30,7 @@ extension CardViewController {
         private let addressViewModelBuilder: AddressViewModelBuilder
 
         internal init(formStyle: FormComponentStyle,
-                      payment: Payment?,
+                      payment: Payment? = nil,
```
mohammedDehairy: let's not use default nil values, because we don't want to miss passing it by mistake :- )

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -46,7 +46,7 @@ internal class CardViewController: FormViewController {
     internal init(configuration: CardComponent.Configuration,
                   shopperInformation: PrefilledShopperInformation?,
                   formStyle: FormComponentStyle,
-                  payment: Payment?,
+                  payment: Payment? = nil,
```
mohammedDehairy: let's not use default nil values, because we don't want to miss passing it by mistake :- )

AdyenCard/Components/Stored Card/StoredCardComponent.swift
```diff
@@ -27,9 +29,11 @@ internal final class StoredCardComponent: PaymentComponent, PaymentAwareComponen
     private let storedCardPaymentMethod: StoredCardPaymentMethod
     
     internal init(storedCardPaymentMethod: StoredCardPaymentMethod,
+                  payment: Payment? = nil,
```
mohammedDehairy: let's not use default nil values, because we don't want to miss passing it by mistake :- )

Adyen/Model/PaymentComponentData.swift
```diff
@@ -22,6 +22,11 @@ public struct PaymentComponentData {
 
     /// The partial payment order if any.
     public let order: PartialPaymentOrder?
+
+    /// The payment amount.
+    public var amount: Amount? {
+        order?.remainingAmount
+    }
```
mohammedDehairy, line 31: lets not removed the amount from `PaymentData`, as we discussed offline :-)

AdyenComponents/Boleto/BoletoComponentExtensions.swift
```diff
@@ -28,14 +30,17 @@ extension BoletoComponent {
         ///   - localizationParameters: Localization parameters.
         ///   - shopperInformation: Pre-filled optional personal information about the shopper
         ///   - showEmailAddress: Indicates whether to show `sendCopyByEmail` checkbox and email text field
-        public init(style: FormComponentStyle,
+        ///   - payment: The payment information.
+        public init(style: FormComponentStyle = FormComponentStyle(),
+                    payment: Payment? = nil,
```
mohammedDehairy: lets not use default nil values, we don't want to encourage  merchant not to provide it :-)

Demo/Common/Networking/PaymentsRequest.swift
```diff
@@ -29,7 +29,7 @@ internal struct PaymentsRequest: APIRequest {
         var container = encoder.container(keyedBy: CodingKeys.self)
 
         let currentConfiguration = ConfigurationConstants.current
-        let amount = data.order?.remainingAmount ?? currentConfiguration.amount
+        let amount = data.amount ?? currentConfiguration.amount
```
mohammedDehairy: maybe don't let the merchant worry about that and just provide the right amount as discussed offline

Tests/AdyenTests/Adyen Tests/Components/Gift Card/GiftCardComponentTests.swift
```diff
@@ -236,44 +233,6 @@ class GiftCardComponentTests: XCTestCase {
         waitForExpectations(timeout: 10, handler: nil)
     }
 
-    func testMissingPaymentObject() throws {
```
mohammedDehairy, line 239: why is this removed? payment object is still optional, right?

AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
descorp: In this case it is internal, but sure - good point!

AdyenCard/Components/Card/CardViewController.swift
descorp: 👍

Adyen/Model/PaymentComponentData.swift
descorp: 👍

Demo/Common/Networking/PaymentsRequest.swift
descorp: hm.. in this case this is not a mobile integration, but a "backend code" 🤔
Not sure how can we do this..

Tests/AdyenTests/Adyen Tests/Components/Gift Card/GiftCardComponentTests.swift
descorp: no, not any more :)
DropIn will emit console log / assert failure, but component itself just expects the amount (like ApplePay)

mohammedDehairy: why keep it optional then?

Demo/Common/Networking/PaymentsRequest.swift
mohammedDehairy: don't think I understand what you're saying! we discussed this before, right?

Adyen/Model/PaymentComponentData.swift
```diff
@@ -102,22 +102,34 @@ public struct PaymentComponentData {
...
-                             amount: amount,
                              order: order,
                              storePaymentMethod: storePaymentMethod,
                              browserInfo: browserInfo,
+                             checkoutAttemptId: checkoutAttemptId,
                              installments: installments)
     }
 
     @_spi(AdyenInternal)
     public func replacingAmount(with amount: Amount) -> PaymentComponentData {
```
erenbesel: If amount is removed, this function becomes redundant I think

descorp: amount is back :)

Tests/AdyenTests/Adyen Tests/Components/Gift Card/GiftCardComponentTests.swift
descorp: It [looks like this](https://github.com/Adyen/adyen-ios/pull/940/files#diff-516c91ca2ef64f84384962f06bd375a2c91af3415467d52c9129f6cae0fa3decR55-R58):

```swift 
public convenience init(paymentMethod: GiftCardPaymentMethod,
                            context: AdyenContext,
                            amount: Amount,
                            style: FormComponentStyle = FormComponentStyle())
```

Adyen/Core/Components/Base/BasicComponentConfiguration.swift
```diff
@@ -8,21 +8,33 @@ import Foundation
...
-    var style: FormComponentStyle { get }
+
+    /// The payment information.
+    var payment: Payment? { get }
+}
+
+/// Any presentable payment component basic configuration.
+public protocol AnyPresentableComponentConfiguration: AnyBasicComponentConfiguration {
+
+    associatedtype StyleType
```
mohammedDehairy: I don[t this is a good idea, we can't use `AnyPresentableComponentConfiguration` or any of its derivatives (subprotocols ) in a function parameter or return type or var type, then what's the point?

Adyen/Core/Components/Base/BasicComponentConfiguration.swift
```diff
@@ -8,21 +8,33 @@ import Foundation
 
 /// Any component's most basic configuration.
 public protocol AnyBasicComponentConfiguration: Localizable {
-    
-    /// The form component style.
-    var style: FormComponentStyle { get }
+
+    /// The payment information.
+    var payment: Payment? { get }
```
mohammedDehairy: maybe rename `PaymentAwareComponent` to `PaymentAware` and make `AnyBasicComponentConfiguration` implement it

Adyen/Model/PaymentComponentData.swift
```diff
@@ -13,6 +13,8 @@ import Foundation
  [API Reference](https://docs.adyen.com/api-explorer/#/CheckoutService/latest/post/payments__example_payments-klarna)
  */
 public struct PaymentComponentData {
+
+    internal let internalAmount: Amount?
```
mohammedDehairy: maybe `totalAmount` makes more sense?

Adyen/Model/PaymentComponentData.swift
```diff
@@ -24,7 +26,9 @@ public struct PaymentComponentData {
     public let order: PartialPaymentOrder?
 
     /// The payment amount.
-    public let amount: Amount?
+    public var amount: Amount? {
```
mohammedDehairy: maybe `amountToPay`?

Adyen/Core/Components/Base/BasicComponentConfiguration.swift
descorp: 👍

Adyen/Model/PaymentComponentData.swift
descorp: 👍

Adyen/Core/Components/Base/BasicComponentConfiguration.swift
descorp: > we can't use AnyPresentableComponentConfiguration or any of its derivatives (subprotocols ) in a function parameter or return type or var type

this statement is correct, but I doubt we plan to do any of the things above - hence I don't see it as a counter-argument 🤔

However, I will delete AnyPresentableComponentConfiguration - it does not bring much value.
Main change in this PR - `style: FormComponentStyle` is no longer part of `AnyBasicComponentConfiguration`

Adyen/Model/PaymentComponentData.swift
descorp: Since we introduced `amountToPay` - we can just use `amount` 🤔

Adyen/Core/Components/Base/BasicComponentConfiguration.swift
mohammedDehairy: just try not to use `associatedtype`, thats my point, it doesn't add any value :-)

and for `AnyPresentableComponentConfiguration` why does it not add any value? maybe if you remove the associated type, then it will have value :-D

Adyen/Core/Core Protocols/PaymentAwareComponent.swift
```diff
@@ -7,15 +7,15 @@
 import Foundation
 
 /// Any component with a partial payment property.
-public protocol PartialPaymentOrderAware: Component {
+public protocol PartialPaymentOrderAware: AnyObject {
```
mohammedDehairy: doesn't have to be `AnyObject`, right?

Adyen/Core/Core Protocols/PaymentComponent.swift
```diff
@@ -14,10 +14,6 @@ public protocol PaymentMethodAware: AnyObject {
 
 /// A component that handles the initial phase of getting payment details to initiate a payment.
 public protocol PaymentComponent: Component, PartialPaymentOrderAware, PaymentMethodAware {
-
-    /// The context object for this component.
-    @_spi(AdyenInternal)
-    var context: AdyenContext { get }
```
mohammedDehairy, line 20: why is this removed? any `PaymentComponent` must have a `AdyenContext ` object, right?

descorp: This is duplicate.
`Component` is already `AdyenContextAware`

Adyen/Core/Core Protocols/PaymentAwareComponent.swift
descorp: 👍

Adyen/Core/Core Protocols/PaymentComponent.swift
```diff
@@ -27,23 +23,14 @@ public protocol PaymentComponent: PaymentAwareComponent, PaymentMethodAware {
...
-            updatedData = PaymentComponentData(paymentMethodDetails: data.paymentMethod,
-                                               amount: data.amount,
-                                               order: data.order,
-                                               storePaymentMethod: data.storePaymentMethod,
-                                               browserInfo: data.browserInfo,
-                                               checkoutAttemptId: checkoutAttemptId,
-                                               installments: data.installments)
-
-        }
+        let updatedData = data.replacingCheckoutAttemptID(with: checkoutAttemptId)
```
erenbesel, line 35: nice!

AdyenComponents/Issuer List/IssuerListComponent.swift
```diff
@@ -100,7 +106,9 @@ extension IssuerListComponent: TrackableComponent {}
 extension IssuerListComponent {
     
     /// Configuration for Issuer List type components.
-    public struct Configuration {
+    public struct Configuration: AnyBasicComponentConfiguration {
+
+        public var payment: Payment?
```
erenbesel: I think I missed this part before, why do we add Payment to the config? Doesn't feel like the right place and component already has payment

AdyenDropIn/Models/DropInConfiguration.swift
```diff
@@ -110,9 +113,6 @@ public extension DropInComponent {
         /// Use this property to enforce a custom collection of card types.
         public var allowedCardTypes: [CardType]?
 
-        /// Indicates the card brands excluded from the supported brands.
-        public var excludedCardTypes: Set<CardType> = [.bcmc]
```
erenbesel, line 114: Why is this removed?

descorp: it was not used, because `excludedCardTypes` is internal on `CardComponent.Configuration`

Adyen/Core/Core Protocols/PaymentComponent.swift
descorp: 😎

AdyenComponents/Issuer List/IssuerListComponent.swift
descorp: > component already has payment

This PR was intended to resolve following problems:
- some components can't have optional payment.
- for all components payment must be set before viewController is initiated - this potentially leads to merchant's confusion and mistakes.
- "asynchronous" setting of the payment leads to complicated flows of `didSet { .. }` for embedded components. 

Solution here is to makes `payment` read-only. However, now we need a new way to set the payment :)

IoC sounds like a good alternative + With this approach ApplePay, DropIn and GiftCard aligned with other components - now we always set payment during configuration.

> Doesn't feel like the right place

What are you suggesting?

descorp: Going to move `payment` to a AdyenContext

Adyen/Core/AdyenContext/AdyenContext.swift
```diff
@@ -24,17 +27,28 @@ public final class AdyenContext {
...
     internal init(apiContext: APIContext,
+                  payment: Payment?,
                   analyticsProvider: AnalyticsProviderProtocol) {
         self.apiContext = apiContext
         self.analyticsProvider = analyticsProvider
+        self.payment = payment
+    }
+
+    @_spi(AdyenInternal)
+    public func updated(with payment: Payment?) -> AdyenContext {
```
erenbesel: Maybe similar to others, `replacingPayment(with payment: Payment?)`

AdyenDropIn/Components/PreApplePayComponent.swift
```diff
@@ -92,15 +92,7 @@ internal final class PreApplePayComponent: PresentableComponent,
...
     internal func didSubmit(_ data: PaymentComponentData, from component: PaymentComponent) {
-        let checkoutAttemptId = component.context.analyticsProvider.checkoutAttemptId
-        let updatedData = PaymentComponentData(paymentMethodDetails: data.paymentMethod,
-                                               amount: data.amount,
-                                               order: data.order,
-                                               storePaymentMethod: data.storePaymentMethod,
-                                               browserInfo: data.browserInfo,
-                                               checkoutAttemptId: checkoutAttemptId,
-                                               installments: data.installments)
-        submit(data: updatedData, component: self)
```
mohammedDehairy, line 103: its better if you call `submit(data:component:)`, it won't hurt but will be consistent :-)

AdyenDropIn/DropInComponent.swift
```diff
@@ -265,12 +261,6 @@ public final class DropInComponent: NSObject,
         (component as? PreApplePayComponent)?.presentationDelegate = self
         
         component._isDropIn = true
-        
-        if let payment = configuration.payment, let remainingAmount = component.order?.remainingAmount {
-            component.payment = Payment(amount: remainingAmount, countryCode: payment.countryCode)
-        } else {
-            component.payment = configuration.payment
-        }
```
mohammedDehairy, line 273: where is the code that changes the Payment amount in the AdyenContext in case of partial payment order remaining amount? am I missing something?

Adyen/Core/AdyenContext/AdyenContext.swift
descorp: How about:

```swift
replacing(payment: Payment?)
```

AdyenDropIn/DropInComponent.swift
descorp: It is moved to [ComponentManager](https://github.com/Adyen/adyen-ios/pull/940/files/3d61fcfa7558cbfca94d3410eae0dd859d75eef0#diff-fd23d788c9a3e6dec6b94258dbabae37abd2de363812a97a822efe48678d81e3R48-R54)

AdyenDropIn/Utilities/ComponentManager.swift
```diff
@@ -34,26 +36,30 @@ internal final class ComponentManager {
...
         self.order = order
         self.supportsEditingStoredPaymentMethods = supportsEditingStoredPaymentMethods
         self.presentationDelegate = presentationDelegate
+
+        if let payment = context.payment, let remainingAmount = order?.remainingAmount {
+            let payment = Payment(amount: remainingAmount, countryCode: payment.countryCode)
+            self.context = context.updated(with: payment)
+        } else {
+            self.context = context
+        }
```
mohammedDehairy: I think there is a potential issue here, that the ComponentManager. context will be different than DropInComponent.context, so there is a risk of bugs here.

why not make the `AdyenContext.payment`  `private(set) var`  and in the `updated(with: payment)` just update the payment instead of creating a new instance :-) (AyenContext is a class)

AdyenDropIn/DropInComponent.swift
mohammedDehairy: [will continue the discussion here](https://github.com/Adyen/adyen-ios/pull/940#discussion_r919929556) :-)

AdyenDropIn/Components/PreApplePayComponent.swift
descorp: 👍

AdyenDropIn/Utilities/ComponentManager.swift
descorp: `DropInComponent.context` is holding an original payment and `ComponentManager.context` - the one is in use.
I'd like to keep `DropInComponent.context` untouched because if you opt-out from GiftCard - you need to get back to an original `payment`

descorp: Will make `AdyenContext.payment` private(set)

AdyenDropIn/Utilities/ComponentManager+PaymentComponentBuilder.swift
```diff
@@ -174,8 +174,7 @@ extension ComponentManager: PaymentComponentBuilder {
                                                                    localizationParameters: configuration.localizationParameters)
 
         if let amount = order?.remainingAmount {
-            let localIdentifier = amount.localeIdentifier ?? configuration.localizationParameters?.locale
-            let configuration = applePay.updating(amount: amount, localeIdentifier: localIdentifier)
```
mohammedDehairy: sorry why is this removed?

mohammedDehairy (approved):

Adyen/Model/PaymentComponentData.swift
```diff
@@ -116,8 +121,22 @@ public struct PaymentComponentData {
                              order: order,
                              storePaymentMethod: storePaymentMethod,
                              browserInfo: browserInfo,
+                             checkoutAttemptId: checkoutAttemptId,
                              installments: installments)
     }
+
+    @_spi(AdyenInternal)
+    public func replacingCheckoutAttemptID(with checkoutAttemptId: String?) -> PaymentComponentData {
```
mohammedDehairy, line 129: 'replacingCheckoutAttemptID' is old school :-), can it be:

```
updating(checkoutAttemptId: String?) -> PaymentComponentData

or 

replacing(checkoutAttemptId: String?) -> PaymentComponentData
```

erenbesel (approved):

Adyen/Model/PaymentComponentData.swift
descorp: Will do in another PR 😛

PR #970, opened 2022-08-05
feature: implemented the new native redirect flow

mohammedDehairy:
feature: implemented the new native redirect flow

### DIFF
--- a/Adyen/Model/PaymentComponentData.swift
+++ b/Adyen/Model/PaymentComponentData.swift
@@ -32,6 +32,9 @@ public struct PaymentComponentData {
     
     /// The installments object.
     public let installments: Installments?
+    
+    /// Indicates whether the current SDK version suports native redirect without glue pages.
+    public let supportNativeRedirect: Bool = true
 
     /// Shopper name.
     public var shopperName: ShopperName? {
--- a/AdyenActions/Components/Redirect/NativeRedirectResultRequest.swift
+++ b/AdyenActions/Components/Redirect/NativeRedirectResultRequest.swift
@@ -1,0 +1,47 @@
+//
+// Copyright (c) 2022 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+//
+// Copyright (c) 2022 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import AdyenNetworking
+import Foundation
+@_spi(AdyenInternal) import Adyen
+
+internal struct NativeRedirectResultRequest: Request {
+    internal typealias ResponseType = RedirectDetails
+
+    internal typealias ErrorResponseType = APIError
+
+    internal let path: String = "checkoutshopper/v1/nativeRedirect/redirectResult"
+
+    internal var counter: UInt = 0
+
+    internal let headers: [String: String] = [:]
+
+    internal let queryParameters: [URLQueryItem] = []
+
+    internal let method: HTTPMethod = .post
+
+    internal let redirectData: String?
+
+    internal let returnQueryString: String
+
+    internal init(redirectData: String?, returnQueryString: String) {
+        self.redirectData = redirectData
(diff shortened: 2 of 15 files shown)


Adyen/Model/PaymentComponentData.swift
```diff
@@ -32,6 +32,9 @@ public struct PaymentComponentData {
     
     /// The installments object.
     public let installments: Installments?
+    
+    /// Indicates whether the current SDK version suports native redirect without glue pages.
+    public let supportNativeRedirect: Bool = true
```
descorp, line 37: is it good idea to set it `true` by default ?
On what API version will it be available? 
Can it trigger "unknown field" on LIVE?

AdyenActions/Components/Redirect/NativeRedirectResultRequest.swift
```diff
@@ -0,0 +1,47 @@
+//
+// Copyright (c) 2022 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+//
+// Copyright (c) 2022 Adyen N.V.
```
descorp: two headers

AdyenActions/Components/Redirect/RedirectComponent.swift
```diff
@@ -13,11 +14,23 @@ extension RedirectComponent: AnyRedirectComponent {}
...
         
+        /// Indicates that invalid parameters are passed back from the issuer.
+        case invalidRedirectParameters
+        
+        public var errorDescription: String? {
+            switch self {
+            case .appNotFound:
+                return "No app installed that can handle the payment."
+            case .invalidRedirectParameters:
+                return "Invalid parameters are passed back from the issuer."
```
descorp:
```suggestion
                return "Query parameters returned by the issuer is invalid or empty ."
```

Adyen/Model/PaymentComponentData.swift
mohammedDehairy: if merchants want to disable it, they can just not send it in the payments call

also its not versioned, so its available for all API versions :-)

Tests/Session Tests/SessionTests.swift
```diff
@@ -812,8 +812,8 @@ class SessionTests: XCTestCase {
             didCompleteExpectation.fulfill()
         }
         let actionData = ActionComponentData(
-            details: RedirectDetails(
-                returnURL: URL(string: "https://google.com")!
+            details: try RedirectDetails(
+                returnURL: URL(string: "https://google.com?redirectResult=some")!
```
descorp: We can reuse `URL(string: "https://google.com?redirectResult=some")` as `Dummy.returnURL`

Adyen/Model/PaymentComponentData.swift
descorp: Still don't think it is a good idea to enable experimental feature for everyone.

> if merchants want to disable it, they can just not send it in the payments call

For Sessions it is "ON" out-of-the-box, so to opt-out merchant must be aware of this field's existence.

mohammedDehairy: then merchants can disable the account data property on the backend :-), right?

descorp (dismissed):

AdyenActions/Components/Redirect/RedirectComponent.swift
```diff
@@ -130,13 +155,45 @@ public final class RedirectComponent: ActionComponent {
...
+            switch result {
+            case let .failure(error):
+                self.delegate?.didFail(with: error, from: self)
+            case let .success(response):
+                self.callDidProvide(redirectDetails: response, action)
+            }
+        }
+    }
+
+    private func callDidProvide(redirectDetails: RedirectDetails, _ action: RedirectAction) {
```
erenbesel: Instead of the word `call` how about `notifyDelegateForDidProvide` or something of this sort?

erenbesel (approved):

descorp (approved):

PR #971, opened 2022-08-08
Implemented the new 8 digits bin values

mohammedDehairy:
Implemented the new 8 digits bin values

### DIFF
--- a/AdyenCard/Form/FormCardNumberItem.swift
+++ b/AdyenCard/Form/FormCardNumberItem.swift
@@ -10,7 +10,11 @@ import UIKit
 /// A form item into which a card number is entered.
 internal final class FormCardNumberItem: FormTextItem, AdyenObserver {
     
-    private static let binLength = 12
+    private enum Constants {
+        static let smallBinLength = 6
+        static let largeBinLength = 8
+        static let minimumPANLength = 16
+    }
 
     private let cardNumberFormatter = CardNumberFormatter()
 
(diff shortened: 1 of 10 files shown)


AdyenCard/Form/FormCardNumberItem.swift
```diff
@@ -119,6 +126,22 @@ internal final class FormCardNumberItem: FormTextItem, AdyenObserver {
                                         isEnteredBrandSupported: isBrandSupported,
                                         panLength: currentBrand?.panLength)
         self.currentBrand = currentBrand
+        updateBIN()
+    }
+    
+    private func updateBIN() {
+        guard isValid() else { return }
```
erenbesel: Maybe just one switch for all of it? 
```
switch currendBrand?.panLength {
   case .some(let panLength) where panLenth > Constants.minimumPANLength:
          binValue = String(value.prefix(Constants.largeBinLength))
   default:
         binValue = String(value.prefix(Constants.smallBinLength))
}
```

AdyenCard/Form/FormCardNumberItem.swift
```diff
@@ -24,6 +28,9 @@ internal final class FormCardNumberItem: FormTextItem, AdyenObserver {
     /// The value contains up to 6 first digits of card' PAN.
     @AdyenObservable("") internal var binValue: String
     
+    /// The card' PAN value.
```
erenbesel: The card's :)

AdyenCard/Form/FormCardNumberItem.swift
mohammedDehairy: good one 👍

PR #972, opened 2022-08-08
Session installments

erenbesel:


Adyen/Core/Components/Installment/InstallmentOptions.swift
```diff
@@ -4,19 +4,23 @@
...
+@_spi(AdyenInternal)
+public protocol InstallmentConfigurationAware {
+    var installmentConfiguration: InstallmentConfiguration? { get }
+}
+
 /// Details to configure Installment Options.
 /// These always include regular monthly installments,
 /// but in some countries `revolving` option may be added.
-public struct InstallmentOptions: Equatable {
+public struct InstallmentOptions: Equatable, Decodable, Encodable {
```
mohammedDehairy: maybe `Codable`? :-)

Adyen/Core/Components/Installment/InstallmentOptions.swift
```diff
@@ -4,19 +4,23 @@
...
     
     /// Month options for regular installments.
-    internal let regularInstallmentMonths: [UInt]
+    public let regularInstallmentMonths: [UInt]
     
     /// Determines if revolving installment is an option.
-    internal let includesRevolving: Bool
+    public let includesRevolving: Bool
```
mohammedDehairy, line 25: maybe mark them with `@_spi(AdyenInternal)`?

Adyen/Core/Components/Installment/InstallmentOptions.swift
```diff
@@ -40,21 +44,47 @@ public struct InstallmentOptions: Equatable {
...
-public struct InstallmentConfiguration {
+public struct InstallmentConfiguration: Decodable {
     
     /// The option that apply to all card types, unless included `cardTypeBased` options.
-    internal let defaultOptions: InstallmentOptions?
+    public let defaultOptions: InstallmentOptions?
     
     /// Options that are specific to given card types
-    internal let cardBasedOptions: [CardType: InstallmentOptions]?
+    public let cardBasedOptions: [CardType: InstallmentOptions]?
```
mohammedDehairy, line 85: maybe mark them with `@_spi(AdyenInternal)`?

Adyen/Core/Components/Installment/InstallmentOptions.swift
```diff
@@ -87,4 +117,40 @@ public struct InstallmentConfiguration {
         self.showInstallmentPrice = false
     }
     
+    @_spi(AdyenInternal)
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: DynamicKey.self)
```
mohammedDehairy, line 126: can you use `decoder .unkeyedContainer()` instead of using the `DynamicKey`?

AdyenCard/Components/Card/CardComponent.swift
```diff
@@ -48,12 +48,16 @@ public class CardComponent: PresentableComponent,
     public let supportedCardTypes: [CardType]
 
     /// Card component configuration.
-    public let configuration: Configuration
+    public var configuration: Configuration
```
mohammedDehairy: maybe make it `private(set)`? to prevent merchants from changing it from outside?

Adyen/Core/Components/Installment/InstallmentOptions.swift
erenbesel: `unkeyedContainer` doesn't have properties like `.allKeys` to compare the key names. Don't think it can be done with it.

Adyen/Core/Components/Installment/InstallmentOptions.swift
erenbesel: I realized hiding these serves no purpose and exposing them makes more sense if anybody wanted to check their config objects etc.

Adyen/Core/Components/Installment/InstallmentOptions.swift
mohammedDehairy: yeah, I suppose :-)

Adyen/Core/Components/Installment/InstallmentOptions.swift
mohammedDehairy: not sure about that, we can always expose them later, but we can't make them private after exposing them, so IMHO its best if we make them spi_internal until merchants need them

but that's just me :-)

mohammedDehairy (approved):

nauaros (approved):

PR #974, opened 2022-08-10
8 digits bin values

mohammedDehairy:
8 digits bin values

### DIFF
--- a/AdyenCard/Components/Card/CardComponent.swift
+++ b/AdyenCard/Components/Card/CardComponent.swift
@@ -173,15 +173,27 @@ public class CardComponent: PresentableComponent,
         formViewController.title = paymentMethod.displayInformation(using: configuration.localizationParameters).title
         return formViewController
     }()
+    
+    private let panThrottler = Throttler(minimumDelay: CardComponent.Constant.secondsThrottlingDelay)
+    private let binThrottler = Throttler(minimumDelay: CardComponent.Constant.secondsThrottlingDelay)
 }
 
 extension CardComponent: CardViewControllerDelegate {
     
-    func didChange(bin: String) {
-        self.cardComponentDelegate?.didChangeBIN(bin, component: self)
+    func didChange(pan: String) {
+        panThrottler.throttle { [weak self] in
+            self?.updateBrand(with: pan)
+        }
     }
     
-    func didChange(pan: String) {
+    func didChange(bin: String) {
+        binThrottler.throttle { [weak self] in
+            guard let self = self else { return }
+            self.cardComponentDelegate?.didChangeBIN(bin, component: self)
+        }
+    }
+    
+    private func updateBrand(with pan: String) {
         binInfoProvider.provide(for: pan, supportedTypes: supportedCardTypes) { [weak self] binInfo in
             guard let self = self else { return }
             // update response with sorted brands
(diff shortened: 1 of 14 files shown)


erenbesel (dismissed):

AdyenCard/Components/Card/CardComponent.swift
```diff
@@ -174,13 +173,28 @@ public class CardComponent: PresentableComponent,
...
+    private let panThrottler = Throttler(minimumDelay: CardComponent.Constant.secondsThrottlingDelay)
+    private let binThrottler = Throttler(minimumDelay: CardComponent.Constant.secondsThrottlingDelay)
 }
 
 extension CardComponent: CardViewControllerDelegate {
     
-    func didChangeBIN(_ value: String) {
-        self.cardComponentDelegate?.didChangeBIN(String(value.prefix(Constant.publicBinLength)), component: self)
-        binInfoProvider.provide(for: value, supportedTypes: supportedCardTypes) { [weak self] binInfo in
+    func didChange(pan: String) {
```
descorp: Missing access level. Should be `internal` :)

AdyenCard/Components/Card/CardComponentDelegate.swift
```diff
@@ -10,20 +10,24 @@ import UIKit
 /// Delegate for observing user's activity on `CardComponent`.
 public protocol CardComponentDelegate: AnyObject {
 
-    /// Called when shopper enters PAN in `CardComponent`.
-    /// - Parameter value: Up to 6 first digits in entered PAN.
+    /// Called with the first 6 or 8 digits typed by the shopper in the PAN field in `CardComponent`.
```
descorp, line 13:
```suggestion
    /// Called with the first 6 or 11 digits typed by the shopper in the PAN field in `CardComponent`.
```

mohammedDehairy: it'll never be 11, BIN is max 8 digits :-)

AdyenCard/Components/Card/CardComponent.swift
mohammedDehairy: right :-D

AdyenCard/Components/Card/CardComponentDelegate.swift
descorp: `thresholdBINLength = 11` ?

**UPD**
Ok, this is why it was called "privateBinLength" - it is for internal use only

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -290,18 +292,21 @@ internal class CardViewController: FormViewController {
     }
 
     private func setupViewRelations() {
-        observe(items.numberContainerItem.numberItem.$binValue) { [weak self] in self?.didReceive(bin: $0) }
+        observe(items.numberContainerItem.numberItem.$panValue) { [weak self] in self?.didChange(pan: $0) }
+        observe(items.numberContainerItem.numberItem.$binValue) { [weak self] in self?.didChange(bin: $0) }
 
         items.button.buttonSelectionHandler = { [weak cardDelegate] in
             cardDelegate?.didSelectSubmitButton()
```
descorp, line 299: In another PR we should rename `cardDelegate` -> `componentDelegate`

descorp (approved):

erenbesel (approved):

nauaros (approved):

PR #976, opened 2022-08-12
Make Payment Codable

sebastianvarela:
## Summary
We noticed that the objects inside Payment are Codable but the Payment object is not.

This change is simple, it doesn't break anything and it's useful and ensure that can we handle Payment or PaymentMethods with the same way.

## Tested scenarios
Store payment methods, perform payments with drop-in

**Fixed issue**:

### DIFF
--- a/Adyen/Model/Payment.swift
+++ b/Adyen/Model/Payment.swift
@@ -7,7 +7,7 @@
 import Foundation
 
 /// Describes the current payment.
-public struct Payment {
+public struct Payment: Codable {
     
     /// The amount for this payment.
     public let amount: Amount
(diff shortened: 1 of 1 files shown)


Adyen/Model/Payment.swift
```diff
@@ -31,4 +31,8 @@ public struct Payment {
         self.countryCode = unsafeCountryCode
     }
 
+    private enum CodingKeys: String, CodingKey {
+        case amount
+        case countryCode
+    }
```
mohammedDehairy: Thanks @sebastianvarela for your contribution, its appreciated! :-)

just a small comment, I think `CodingKeys` is redundant, since Swift compiler provide it automatically whenever all members are `Codable`, am I missing something?

sebastianvarela: yes @mohammedDehairy, you are right! we removed... PR updated!

mohammedDehairy (approved):

descorp (approved):

PR #981, opened 2022-08-17
🔬 [Experimental]  use Completion instead of typealiases

descorp:
# WIP

## Changes

* use Completion instead of typealiases

### DIFF
--- a/Adyen/Utilities/Debugging/Assertion.swift
+++ b/Adyen/Utilities/Debugging/Assertion.swift
@@ -6,14 +6,10 @@
 
 import Foundation
 
-/// A typealias for a closure that handles a URL through which the application was opened.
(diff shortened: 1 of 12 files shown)


erenbesel (commented):
Feels like the name `Completion<T>` limits this usage as the trailing closure, a.k.a completion.
I suggest something like `Handler<T>` would be more inclusive IMO.

mohammedDehairy:
@erenbesel 
> Feels like the name `Completion<T>` limits this usage as the trailing closure, a.k.a completion.
> 
> I suggest something like `Handler<T>` would be more inclusive IMO.

The problem is that the T many times is a weak type like String for example, so replacing a descriptive name describing what the string is with a generic name reduces readability IMO

PR #984, opened 2022-08-18
v4: Add `didFinalize` callback for Apple Pay 

descorp:
# Open PR

## Changes

* deprecate `didFinalize(with success: Bool)`
* add `didFinalize(with success: Bool, completion: (() -> Void)?)` as default

### DIFF
--- a/Adyen/Core/Components/Base/Component.swift
+++ b/Adyen/Core/Components/Base/Component.swift
@@ -14,7 +14,7 @@ extension Component {
 
     /// Finalizes the payment if there is any, after being processed by payment provider.
     /// - Parameter success: The status of the payment.
-    @available(*, deprecated, message: "This property will no longer be available.")
+    @available(*, deprecated, message: "Use finalizeIfNeeded(with:, completion:) instead.")
     public func finalizeIfNeeded(with success: Bool) {
         stopLoadingIfNeeded()
         (self as? FinalizableComponent)?.didFinalize(with: success, completion: nil)
@@ -46,7 +46,7 @@ public protocol FinalizableComponent: Component {
 
     /// Finalizes payment after being processed by payment provider.
     /// - Parameter success: The status of the payment.
-    @available(*, deprecated, message: "This property will no longer be available. Use finalizeIfNeeded(with:, completion:) instead.")
+    @available(*, deprecated, message: "Use finalizeIfNeeded(with:, completion:) instead.")
     func didFinalize(with success: Bool)
 
     /// Finalizes payment after being processed by payment provider.
--- a/Adyen/Core/Components/Base/Component.swift
+++ b/Adyen/Core/Components/Base/Component.swift
@@ -14,9 +14,14 @@ extension Component {
 
     /// Finalizes the payment if there is any, after being processed by payment provider.
     /// - Parameter success: The status of the payment.
-    /// :nodoc:
+    @available(*, deprecated, message: "This property will no longer be available.")
     public func finalizeIfNeeded(with success: Bool) {
-        (self as? FinalizableComponent)?.didFinalize(with: success)
+        (self as? FinalizableComponent)?.didFinalize(with: success, completion: {})
+        stopLoadingIfNeeded()
+    }
+
+    public func finalizeIfNeeded(with success: Bool, completion: (() -> Void)?) {
+        (self as? FinalizableComponent)?.didFinalize(with: success, completion: completion)
         stopLoadingIfNeeded()
     }
 
@@ -37,7 +42,14 @@ public protocol FinalizableComponent: Component {
 
     /// Finalizes payment after being processed by payment provider.
     /// - Parameter success: The status of the payment.
+    @available(*, deprecated, message: "This property will no longer be available.")
     func didFinalize(with success: Bool)
+
+    /// Finalizes payment after being processed by payment provider.
+    /// - Parameter success: The status of the payment.
+    /// - Parameter completion: The block to execute after the component finalizes its activity.
+    /// Use of this block is recommended for ApplePayComponent. You may specify nil for this parameter.
+    func didFinalize(with success: Bool, completion: (() -> Void)?)
 }
 
 public extension Component {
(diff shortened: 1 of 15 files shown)


Adyen/Core/Components/Base/PresentableComponentWrapper.swift
```diff
@@ -52,10 +52,17 @@ public final class PresentableComponentWrapper: PresentableComponent,
...
     /// :nodoc:
     public func didFinalize(with success: Bool) {
-        component.finalizeIfNeeded(with: success)
+        component.finalizeIfNeeded(with: success, completion: {})
         stopLoading()
     }
 
+    /// :nodoc:
+    public func didFinalize(with success: Bool, completion: (() -> Void)?) {
+        component.finalizeIfNeeded(with: success, completion: { [weak self] in
```
mohammedDehairy: The completion parameter is not called 🙂

mohammedDehairy (commented):
Also code coveraage is very low, can you raise it a bit 🙂

Adyen/Core/Components/Base/Component.swift
```diff
@@ -14,9 +14,14 @@ extension Component {
 
     /// Finalizes the payment if there is any, after being processed by payment provider.
     /// - Parameter success: The status of the payment.
-    /// :nodoc:
+    @available(*, deprecated, message: "This property will no longer be available.")
```
kaphacius: can we specify which new API to use in the `message` here? same with other deprecations.
maybe something like `message: Use finalizeIfNeeded(with:, completion:) instead`

AdyenComponents/Apple Pay/ApplePayComponent.swift
```diff
@@ -116,21 +116,22 @@ public class ApplePayComponent: NSObject, PresentableComponent, PaymentComponent
 
     /// Finalizes ApplePay payment after being processed by payment provider.
     /// - Parameter success: The status of the payment.
+    @available(*, deprecated, message: "This property will no longer be available.")
     public func didFinalize(with success: Bool) {
-        self.success = success
+        finalizeCompletion = {}
```
erenbesel: Why not set it to nil? setting it to empty block seems weird

Adyen/Core/Components/Base/PresentableComponentWrapper.swift
```diff
@@ -52,10 +52,17 @@ public final class PresentableComponentWrapper: PresentableComponent,
 
     /// :nodoc:
     public func didFinalize(with success: Bool) {
-        component.finalizeIfNeeded(with: success)
+        component.finalizeIfNeeded(with: success, completion: {})
```
erenbesel: Is there a reason for this empty block, can we send `nil` instead?

Adyen/Core/Components/Base/Component.swift
```diff
@@ -14,9 +14,14 @@ extension Component {
...
     public func finalizeIfNeeded(with success: Bool) {
-        (self as? FinalizableComponent)?.didFinalize(with: success)
+        (self as? FinalizableComponent)?.didFinalize(with: success, completion: {})
```
erenbesel: same here

AdyenComponents/Apple Pay/ApplePayComponent.swift
descorp: it's nullability is used as a flag: to differentiate when `paymentAuthorizationViewControllerDidFinish(_ controller: PKPaymentAuthorizationViewController)` were called before or after `didFinalize`

Adyen/Core/Components/Base/PresentableComponentWrapper.swift
descorp: could be `nil` indeed

Adyen/Core/Components/Base/Component.swift
```diff
@@ -14,10 +14,19 @@ extension Component {
 
     /// Finalizes the payment if there is any, after being processed by payment provider.
     /// - Parameter success: The status of the payment.
-    /// :nodoc:
+    @available(*, deprecated, message: "Use finalizeIfNeeded(with:, completion:) instead.")
     public func finalizeIfNeeded(with success: Bool) {
-        (self as? FinalizableComponent)?.didFinalize(with: success)
         stopLoadingIfNeeded()
```
mohammedDehairy: if  this deprecated version is used, the `stopLoadingIfNeeded()` will be called twice here and in the new `version`, maybe remove `stopLoadingIfNeeded()` from here since its called in the new version anyway?

Adyen/Core/Components/Base/PresentableComponentWrapper.swift
```diff
@@ -52,8 +52,14 @@ public final class PresentableComponentWrapper: PresentableComponent,
 
     /// :nodoc:
     public func didFinalize(with success: Bool) {
-        component.finalizeIfNeeded(with: success)
         stopLoading()
```
mohammedDehairy: same here :-)

https://github.com/Adyen/adyen-ios/pull/984/files#r953466896

AdyenComponents/Apple Pay/ApplePayComponent.swift
```diff
@@ -12,7 +12,7 @@ import PassKit
 public class ApplePayComponent: NSObject, PresentableComponent, PaymentComponent, Localizable, FinalizableComponent {
 
     /// :nodoc:
-    internal var success: Bool = false
+    internal var finalizeCompletion: (() -> Void)?
```
mohammedDehairy: maybe I am missed it, but where is `finalizeCompletion` being called?

Adyen/Core/Components/Base/Component.swift
descorp: Good one!

AdyenComponents/Apple Pay/ApplePayComponent.swift
erenbesel: Hmm then do you think this is a clear way to do this? 🤔 
Someone else without this knowledge may just set to `nil`. Creates area for error

descorp: What do you suggest? Bool flag that mimics it?

descorp: applied state :)

mohammedDehairy (approved):

erenbesel (approved):

PR #1003, opened 2022-09-23
chore: change visibility logic for card icons

erenbesel:


nauaros (dismissed):

mohammedDehairy (dismissed):

descorp (dismissed):

AdyenCard/Form/FormCardNumberContainerItem.swift
```diff
@@ -42,11 +42,10 @@ internal final class FormCardNumberContainerItem: FormItem, AdyenObserver {
...
         self.style = style
         
-        observe(numberItem.$isActive) { [weak self] isActive in
+        observe(numberItem.$isActive) { [weak self] _ in
             guard let self = self else { return }
-            // logo item only visible when number item is active or when it's invalid
-            let hidden = !isActive && self.numberItem.isValid()
-            self.supportedCardLogosItem.isHidden.wrappedValue = hidden
+            // logo item should be visible when field is invalid
+            self.supportedCardLogosItem.isHidden.wrappedValue = self.numberItem.isValid()
```
mohammedDehairy, line 48: if the condition for showing the logos is that the card number is invalid, then why are we updating the visibility when the active state is changed, active state and whether the card number is valid are unrelated, right?

can we have  an `isValid` observable in `FormCardNumberItem`? and listen to it to hide/show the logos icons? Note: I am not sure if its easy to do :-)

erenbesel: Good catch. But in that case we would be immediately showing the icons when the field is invalid until the number is filled.
What we want is to show it when the active state changes AND it's invalid.

mohammedDehairy: yeah, make sense, then can we change the comment to reflect this logic?

neelSharma12 (dismissed):

mohammedDehairy (dismissed):

neelSharma12 (approved):

descorp (approved):

PR #1024, opened 2022-10-20
Implemented moving the checkoutAttemptId to the payment method details object

mohammedDehairy:
Implemented moving the checkoutAttemptId to the payment method details object


Adyen/Core/APIClient/Environment.swift
```diff
@@ -17,6 +17,9 @@ public struct Environment: AnyAPIEnvironment {
     
     @_spi(AdyenInternal)
     public static let beta = Environment(baseURL: URL(string: "https://checkoutshopper-beta.adyen.com/")!)
+    
+    @_spi(AdyenInternal)
+    public static let local = Environment(baseURL: URL(string: "http://localhost:8080/")!)
```
descorp, line 22: Is this needed in AdyenCore?

Demo/Common/Networking/SessionRequest.swift
```diff
@@ -45,7 +45,7 @@ internal struct SessionRequest: APIRequest {
         try container.encode(installmentOptions, forKey: .installmentOptions)
         
         // store payment method required fields
-        try container.encode("askForConsent", forKey: .storePaymentMethodMode)
+//        try container.encode("askForConsent", forKey: .storePaymentMethodMode)
```
descorp: commented code?

Demo/Common/Networking/PaymentsRequest.swift
```diff
@@ -47,7 +47,6 @@ internal struct PaymentsRequest: APIRequest {
         try container.encodeIfPresent(data.socialSecurityNumber, forKey: .socialSecurityNumber)
         try container.encode(Locale.current.identifier, forKey: .shopperLocale)
         try container.encodeIfPresent(data.browserInfo, forKey: .browserInfo)
-        try container.encodeIfPresent(data.checkoutAttemptId, forKey: .checkoutAttemptId)
```
descorp, line 50: No need to send `checkoutAttemptId` ?

Adyen/Core/APIClient/Environment.swift
mohammedDehairy: I don't see why not? I needed regularly and its annoying to add and remove it before pushing to remote

Demo/Common/Networking/SessionRequest.swift
mohammedDehairy: yeah, that one is causing an API error, @erenbesel can shed some light on it? why is it causing API error?

Demo/Common/Networking/PaymentsRequest.swift
mohammedDehairy: yes, the whole point of the PR is send the `checkoutAttemptId ` in the payment method details instead of the root of the payments call

Demo/Common/Networking/SessionRequest.swift
```diff
@@ -44,8 +44,6 @@ internal struct SessionRequest: APIRequest {
         
         try container.encode(installmentOptions, forKey: .installmentOptions)
         
-        // store payment method required fields
-        try container.encode("askForConsent", forKey: .storePaymentMethodMode)
```
erenbesel: I assume this field in the example caused an error due to api version being below 70 :)

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
```diff
@@ -26,9 +26,8 @@ public struct AnalyticsConfiguration {
 
 @_spi(AdyenInternal)
 public protocol AnalyticsProviderProtocol: TelemetryTrackerProtocol {
-
-    /// :nodoc:
-    var checkoutAttemptId: String? { get }
+    
+    func fetchCheckoutAttemptId(completion: @escaping (String?) -> Void)
```
erenbesel: So it's the marchant that has to fetch this? Sorry my memory is hazy I don't recall the details

mohammedDehairy: no worries :-), no, the `AnalyticsPrvider` is marked with `@_spi(AdyenInternal)`, right? this is called by us internally

erenbesel: Ah yeah, and the question was because I couldn't see the implementation of `fetchCheckoutAttemptId` in this PR only to realize it existed and you just moved the signature into the protocol  :D

erenbesel (dismissed):

erenbesel (dismissed):

erenbesel (approved):

nauaros (approved):
Nice change!

PR #1029, opened 2022-10-21
u.s. debit manual hiding

erenbesel:


AdyenCard/Form/FormCardNumberContainerItem.swift
```diff
@@ -38,7 +38,14 @@ internal final class FormCardNumberContainerItem: FormItem, AdyenObserver {
...
                   localizationParameters: LocalizationParameters?) {
-        self.cardTypeLogos = cardTypeLogos
+        // these 4 US debit brands are not to be displayed
+        // but should be supported so it's done here for now
+        self.cardTypeLogos = cardTypeLogos.filter { logo in
+            logo.type != .accel &&
+                logo.type != .pulse &&
+                logo.type != .star &&
+                logo.type != .nyce
+        }
```
mohammedDehairy, line 48:
```suggestion
       self.cardTypeLogos = cardTypeLogos.minus([.accel, .pulse, .star, .nyce]) 
```

erenbesel: Ah very nice! Requires `CardTypeLogo` to conform to `hashable`. Should we add it just for this?

PR #1031, opened 2022-10-21
Manually hide 4 US debit icons

erenbesel:


AdyenCard/Form/FormCardNumberItem.swift
```diff
@@ -46,7 +46,14 @@ internal final class FormCardNumberItem: FormTextItem, AdyenObserver {
...
                   localizationParameters: LocalizationParameters? = nil) {
-        self.cardTypeLogos = cardTypeLogos
+        // these 4 US debit brands are not to be displayed
+        // but should be supported so it's done here for now
+        self.cardTypeLogos = cardTypeLogos.filter { logo in
+            logo.type != .accel &&
+                logo.type != .pulse &&
+                logo.type != .star &&
+                logo.type != .nyce
+        }
```
mohammedDehairy, line 67:
```suggestion
        self.cardTypeLogos = cardTypeLogos.filter { [.accel, .pulse, .star, .nyce].contains($0.type) == false }
```

erenbesel: Nice suggestion! But it requires to conform to `Hashable`. Think we should add just for this?

mohammedDehairy: I don't think Hashable is needed, I was able to do the same change on develop and it builds

erenbesel: Oh I replied too fast because your original suggestion was to use `minus` so I assumed this is the same without reading :D

But for this new one I personally don't see it as easier to read. Would like to keep as is :)

mohammedDehairy (approved):

neelSharma12 (approved):

PR #1067, opened 2022-11-14
Added fix for billing address in card component

neelSharma12:
This PR contains fix for billing address issue where the billing Address is coming as nil if the country does not have a state/province.

### DIFF
--- a/Adyen/UI/Form/Items/Address/AddressViewModel.swift
+++ b/Adyen/UI/Form/Items/Address/AddressViewModel.swift
@@ -30,7 +30,7 @@ public struct AddressViewModel {
     internal var labels: [AddressField: LocalizationKey]
     internal var placeholder: [AddressField: LocalizationKey]
     internal var optionalFields: [AddressField]
-    internal var scheme: [AddressFormScheme]
+    public var scheme: [AddressFormScheme]
 
     public init(labels: [AddressField: LocalizationKey],
                 placeholder: [AddressField: LocalizationKey],
--- a/Adyen/UI/Form/Items/Address/AddressViewModel.swift
+++ b/Adyen/UI/Form/Items/Address/AddressViewModel.swift
@@ -30,6 +30,8 @@ public struct AddressViewModel {
     internal var labels: [AddressField: LocalizationKey]
     internal var placeholder: [AddressField: LocalizationKey]
     internal var optionalFields: [AddressField]
+
+    @_spi(AdyenInternal)
     public var scheme: [AddressFormScheme]
 
     public init(labels: [AddressField: LocalizationKey],
--- a/Adyen/UI/Form/Items/Address/AddressViewModel.swift
+++ b/Adyen/UI/Form/Items/Address/AddressViewModel.swift
@@ -20,6 +20,12 @@ public enum AddressField: String, CaseIterable {
 
 @_spi(AdyenInternal)
 public enum AddressFormScheme {
+    public var  children: [AddressField] {
+        switch self {
+        case .item(let item): return [item]
+        case .split(let item1, let item2): return [item1, item2]
+        }
+    }
     case item(AddressField)
     case split(AddressField, AddressField)
 }
@@ -29,7 +35,9 @@ public struct AddressViewModel {
 
     internal var labels: [AddressField: LocalizationKey]
     internal var placeholder: [AddressField: LocalizationKey]
-    internal var optionalFields: [AddressField]
+
+    @_spi(AdyenInternal)
+    public var optionalFields: [AddressField]
 
     @_spi(AdyenInternal)
     public var scheme: [AddressFormScheme]
--- a/Adyen/UI/Form/Items/Address/DefaultAddressViewModelBuilder.swift
+++ b/Adyen/UI/Form/Items/Address/DefaultAddressViewModelBuilder.swift
@@ -14,12 +14,15 @@ public struct AddressViewModelBuilderContext {
 
 @_spi(AdyenInternal)
 public protocol AddressViewModelBuilder {
+    var scheme: [AddressFormScheme]? { get set }
     func build(context: AddressViewModelBuilderContext) -> AddressViewModel
 }
 
 @_spi(AdyenInternal)
-public struct DefaultAddressViewModelBuilder: AddressViewModelBuilder {
-    
+public class DefaultAddressViewModelBuilder: AddressViewModelBuilder {
+
+    public var scheme: [AddressFormScheme]?
+
     public init() {}
 
     // swiftlint:disable function_body_length
@@ -39,6 +42,7 @@ public struct DefaultAddressViewModelBuilder: AddressViewModelBuilder {
                                                        .apartment: .apartmentSuiteFieldPlaceholder],
                                          optionalFields: context.isOptional ? AddressField.allCases : [.apartment],
                                          scheme: AddressField.allCases.filter { $0 != .country }.map { .item($0) })
+        self.scheme = viewModel.scheme
         switch context.countryCode {
         case "BR":
             viewModel.labels[.stateOrProvince] = .stateFieldTitle
@@ -52,10 +56,12 @@ public struct DefaultAddressViewModelBuilder: AddressViewModelBuilder {
             viewModel.placeholder[.street] = .addressFieldPlaceholder
             viewModel.optionalFields = context.isOptional ? AddressField.allCases : [.houseNumberOrName]
             viewModel.scheme = [.street, .houseNumberOrName, .city, .postalCode, .stateOrProvince].map { .item($0) }
+            self.scheme = viewModel.scheme
         case "GB":
             viewModel.labels[.city] = .cityTownFieldTitle
             viewModel.placeholder[.city] = .cityTownFieldPlaceholder
             viewModel.scheme = [.houseNumberOrName, .street, .city, .postalCode].map { .item($0) }
+            self.scheme = viewModel.scheme
         case "US":
             viewModel.labels[.postalCode] = .zipCodeFieldTitle
             viewModel.labels[.houseNumberOrName] = .apartmentSuiteFieldTitle
@@ -70,6 +76,7 @@ public struct DefaultAddressViewModelBuilder: AddressViewModelBuilder {
                                 .item(.houseNumberOrName),
                                 .item(.city),
                                 .split(.stateOrProvince, .postalCode)]
+            self.scheme = viewModel.scheme
         default:
             break
         }
--- a/Adyen/UI/Form/Items/Address/DefaultAddressViewModelBuilder.swift
+++ b/Adyen/UI/Form/Items/Address/DefaultAddressViewModelBuilder.swift
@@ -18,7 +18,7 @@ public protocol AddressViewModelBuilder {
 }
 
 @_spi(AdyenInternal)
-public class DefaultAddressViewModelBuilder: AddressViewModelBuilder {
+public struct DefaultAddressViewModelBuilder: AddressViewModelBuilder {
 
     public init() {}
 
--- a/Adyen/UI/Form/Items/Address/DefaultAddressViewModelBuilder.swift
+++ b/Adyen/UI/Form/Items/Address/DefaultAddressViewModelBuilder.swift
@@ -14,14 +14,11 @@ public struct AddressViewModelBuilderContext {
 
 @_spi(AdyenInternal)
 public protocol AddressViewModelBuilder {
-    var scheme: [AddressFormScheme]? { get set }
     func build(context: AddressViewModelBuilderContext) -> AddressViewModel
 }
 
 @_spi(AdyenInternal)
 public class DefaultAddressViewModelBuilder: AddressViewModelBuilder {
-
-    public var scheme: [AddressFormScheme]?
 
     public init() {}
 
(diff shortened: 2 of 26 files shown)


AdyenCard/Validators/CardAddressValidator.swift
```diff
@@ -0,0 +1,60 @@
...
+        self.address = address
+        self.addressScheme = addressScheme
+    }
+
+    internal var validAddress: PostalAddress? {
+        guard let address = address, isAddressValid(address: address) else { return nil }
+        return address
+    }
+
+    private func isAddressValid(address: PostalAddress) -> Bool {
```
mohammedDehairy: - maybe expose this function instead of using a property?
- Tiny one: maybe `isValid` is a better name?

AdyenCard/Validators/CardAddressValidator.swift
```diff
@@ -0,0 +1,60 @@
...
+// Copyright (c) 2022 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+@_spi(AdyenInternal) import Adyen
+import Foundation
+
+/// Validates a card's address field.
+public final class CardAddressValidator {
```
erenbesel: Card component is not the only place having addresses so let's rename it into AddressValidator.
I'm assuming we will need this for other addresses we have.

AdyenCard/Validators/CardAddressValidator.swift
```diff
@@ -0,0 +1,60 @@
...
+
+    private let configuration: CardComponent.Configuration
```
erenbesel: This class does not seem to need the whole card configuration, only the address mode. It's better not to depend on the card configuration if we want to make this a generic address validator.

AdyenCard/Validators/CardAddressValidator.swift
erenbesel: So the function for `isAddressValid` seems the same but moved to another class. Couldn't figure out how the bug is solved

neelSharma12: Bug is resolved using this ``` return trimmedFieldsValues.compactMap { $0 }.count == ((addressScheme?.count ?? 0) + 1)``` earlier the counts were not matching as we were not using scheme. Instead of using the scheme we were directly comparing the ```compactMap``` and the ```fieldValues``` count.

AdyenCard/Validators/CardAddressValidator.swift
neelSharma12: Yes, You are right. I will change this.

AdyenCard/Validators/CardAddressValidator.swift
neelSharma12: Good point. I will change this.

Adyen/UI/Form/Items/Address/AddressViewModel.swift
```diff
@@ -30,7 +30,7 @@ public struct AddressViewModel {
     internal var labels: [AddressField: LocalizationKey]
     internal var placeholder: [AddressField: LocalizationKey]
     internal var optionalFields: [AddressField]
-    internal var scheme: [AddressFormScheme]
+    public var scheme: [AddressFormScheme]
```
mohammedDehairy, line 43: maybe annotate it with `@_spi(AdyenInternal)`, because merchants shouldn't access it, right? also maybe `private(set)`, we don't want outside code changing it, right?

Adyen/UI/Form/Items/Address/DefaultAddressViewModelBuilder.swift
```diff
@@ -18,8 +18,8 @@ public protocol AddressViewModelBuilder {
 }
 
 @_spi(AdyenInternal)
-public struct DefaultAddressViewModelBuilder: AddressViewModelBuilder {
-    
+public class DefaultAddressViewModelBuilder: AddressViewModelBuilder {
```
mohammedDehairy: maybe revert it back to `struct`? :-)

Adyen/UI/Form/Items/Address/FormFullAddressItem.swift
```diff
@@ -40,6 +40,8 @@ public final class FormAddressItem: FormValueItem<PostalAddress, AddressStyle>,
     
     internal let addressViewModelBuilder: AddressViewModelBuilder
     
+    public var addressViewModel: AddressViewModel
```
mohammedDehairy: maybe annotate it with `@_spi(AdyenInternal)` and also make it `private(set)` ?

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -110,36 +111,18 @@ internal class CardViewController: FormViewController {
...
-        case .none:
-            fieldsValues = []
-        }
-        
-        let trimmedFieldsValues = fieldsValues.map {
-            $0?.trimmingCharacters(in: .whitespaces).adyen.nilIfEmpty
+        guard let address = address,
+              AddressValidator.init(addressMode: configuration.billingAddress.mode,
+                                    address: address,
+                                    addressScheme: items.addressViewModel?.scheme).isValid() else {
```
mohammedDehairy: have you considered making the `address ` parameter passed to the `isValid()` function instead of the initializer? this way we can reuse it? 

also I am wondering why do use the `AddressValidator.init` instead of just `AddressValidator()`

AdyenCard/Validators/AddressValidator.swift
```diff
@@ -0,0 +1,55 @@
...
+// Copyright (c) 2022 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+@_spi(AdyenInternal) import Adyen
+import Foundation
+
+/// Validates a card's address field.
+public final class AddressValidator {
```
mohammedDehairy: this can be `internal`, right?

AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
```diff
@@ -27,7 +27,9 @@ extension CardViewController {
 
         private let defaultCountryCode: String
         
-        private let addressViewModelBuilder: AddressViewModelBuilder
+        internal let addressViewModelBuilder: AddressViewModelBuilder
+
+        internal var addressViewModel: AddressViewModel?
```
mohammedDehairy: I think this has to be a computed property like this:

```Swift
internal var addressViewModel: AddressViewModel { billingAddressItem.addressViewModel } 
```
because the viewModel gets updated, right?


or event better, access the `billingAddressItem` directly from the `CardViewController`

AdyenCard/Components/Card/CardViewController.swift
mohammedDehairy: also maybe this can be simplified like this:

```Swift
return AddressValidator(addressMode: configuration.billingAddress.mode,
                                     address: address,
                                     addressScheme: items.addressViewModel?.scheme).isValid() ? address : nil
```

AdyenCard/Validators/AddressValidator.swift
```diff
@@ -0,0 +1,55 @@
...
+        case .postalCode:
+            fieldsValues = [address.postalCode]
+        case .none:
+            fieldsValues = []
+        }
+
+        let trimmedFieldsValues = fieldsValues.map {
+            $0?.trimmingCharacters(in: .whitespaces).adyen.nilIfEmpty
+        }
+        return trimmedFieldsValues.compactMap { $0 }.count == ((addressScheme?.count ?? 0) + 1)
```
mohammedDehairy: I don't think simply checking if the count is equal to `addressScheme.count` is enough, because it could be that one field that's not in the scheme is not nil and another one that's IN the scheme is nil, then the count would match but the address is invalid according to the scheme, can you test this?

AdyenCard/Validators/AddressValidator.swift
```diff
@@ -0,0 +1,55 @@
...
+    internal func isValid() -> Bool {
+        let fieldsValues: [String?]
+
+        switch addressMode {
+        case .full:
+            fieldsValues = [address.city,
+                            address.country,
+                            address.postalCode,
+                            address.stateOrProvince,
+                            address.street,
```
mohammedDehairy: also I think you need to take into account optional fields, because if an optional field is nil then the address is still valid, right?

Adyen/UI/Form/Items/Address/AddressViewModel.swift
```diff
@@ -30,7 +30,9 @@ public struct AddressViewModel {
     internal var labels: [AddressField: LocalizationKey]
     internal var placeholder: [AddressField: LocalizationKey]
     internal var optionalFields: [AddressField]
-    internal var scheme: [AddressFormScheme]
+
+    @_spi(AdyenInternal)
+    public var scheme: [AddressFormScheme]
```
mohammedDehairy, line 43: can it also be `private(set)`?

AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
```diff
@@ -27,7 +27,7 @@ extension CardViewController {
 
         private let defaultCountryCode: String
         
-        private let addressViewModelBuilder: AddressViewModelBuilder
+        internal let addressViewModelBuilder: AddressViewModelBuilder
```
mohammedDehairy: may I ask why is this `internal`?

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -110,36 +111,14 @@ internal class CardViewController: FormViewController {
...
-            $0?.trimmingCharacters(in: .whitespaces).adyen.nilIfEmpty
-        }
-        return trimmedFieldsValues.compactMap { $0 }.count == fieldsValues.count
+        return AddressValidator().isValid(address: address ?? nil,
+                                          addressMode: configuration.billingAddress.mode,
+                                          addressScheme: items.billingAddressItem.addressViewModel.scheme) ? address : nil
     }
 
-    private var address: PostalAddress? {
+    internal var address: PostalAddress? {
```
mohammedDehairy: also does this need to be `internal`?

AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
neelSharma12: No, I missed to revert. I will do it.

AdyenCard/Components/Card/CardViewController.swift
neelSharma12: No, I missed to revert. I will do it.

Adyen/UI/Form/Items/Address/AddressViewModel.swift
neelSharma12: In ```DefaultAddressViewModelBuilder``` it is throwing error ```Cannot assign to property: 'scheme' setter is inaccessible```

mohammedDehairy: it can be `internal(set)` then :-)

AdyenCard/Validators/AddressValidator.swift
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        let allAddressFieldsInSchemeSet: Set<AddressField> = Set(allAddressFieldsInScheme)
+        let optionalAddressFieldsSet: Set<AddressField> = Set(optionalAddressField)
+
+        let remaingRequiredAddressFields = allAddressFieldsInSchemeSet.subtracting(optionalAddressFieldsSet)
+ 
+        let isAllRequiredAdddressFieldsPresent = checkIfAllFieldsPresent(fieldsValues: fieldsValues,
+                                                                         requiredAddressFields: remaingRequiredAddressFields)
+        return trimmedFieldsValues.compactMap { $0 }.count == ((remaingRequiredAddressFields.count) + 1)
```
mohammedDehairy: why ` + 1`? :-)

AdyenCard/Validators/AddressValidator.swift
```diff
@@ -0,0 +1,56 @@
...
+        || isAllRequiredAdddressFieldsPresent
```
mohammedDehairy: I think checking if `isAllRequiredAdddressFieldsPresent ` is true is enough, right?

AdyenCard/Validators/AddressValidator.swift
mohammedDehairy: Really Nice Neelam!, the code looks really good, would be great if you unit test this validator!

AdyenCard/Validators/AddressValidator.swift
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    private func checkIfAllFieldsPresent(fieldsValues: [String: String?], requiredAddressFields: Set<AddressField>) -> Bool {
+        for addressFields in requiredAddressFields {
+            if fieldsValues[addressFields.rawValue] == nil {
+              return false
+            }
+        }
```
mohammedDehairy:
```suggestion
        requiredAddressFields.allSatisfy({ fieldsValues[$0.rawValue] != nil })
```

AdyenCard/Validators/AddressValidator.swift
```diff
@@ -0,0 +1,56 @@
...
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+@_spi(AdyenInternal) import Adyen
+import Foundation
+
+/// Validates a card's address field.
+internal class AddressValidator {
+
+    internal func isValid(address: PostalAddress?, addressMode: CardComponent.AddressFormType, addressViewModel: AddressViewModel) -> Bool {
```
descorp, line 13: This could be static?

AdyenCard/Validators/AddressValidator.swift
neelSharma12: Cause country is not included in ```remaingRequiredAddressFields```

neelSharma12: yes, I will definitely add tests.

AdyenCard/Validators/AddressValidator.swift
neelSharma12: Then I will have to make private func static.

erenbesel (approved):

descorp (approved):

PR #1075, opened 2022-11-18
remove dual brand defaul selection

erenbesel:


AdyenCard/Form/FormCardNumberItem.swift
```diff
@@ -74,52 +86,58 @@ internal final class FormCardNumberItem: FormTextItem, Observer {
...
             // if there are 2 brands and neither is supported,
             // need to show unsupported text
-            update(currentBrand: nil, defaultSupportedValue: false)
+            update(initialBrand: nil, defaultSupportedValue: false)
         default:
-            update(currentBrand: nil)
+            update(initialBrand: nil)
         }
         
+        isDualBranded = brands.count == 2 && brands.allSatisfy(\.isSupported)
```
mohammedDehairy, line 106: maybe `>= 2`? its highly unlikely that its more than 2 brands, but are we sure its impossible?

erenbesel: We are told so. And if by some bug there are 3, may not be right to call it dual-branded either :)

mohammedDehairy (approved):

neelSharma12 (approved):

PR #1094, opened 2022-12-02
fixed: ListSectionHeaderStyle is not applied to stored payment method and Issuers'list background color

neelSharma12:
# fixed:
<fixed>

* For stored payment methods, the section header now gets the style from ```ListSectionHeaderStyle```.
* When a list of issuers is shown during checkout, the background is no longer transparent.

</fixed>

### DIFF
--- a/AdyenDropIn/Utilities/ComponentManager.swift
+++ b/AdyenDropIn/Utilities/ComponentManager.swift
@@ -82,10 +82,7 @@ internal final class ComponentManager {
                                               components: storedComponents,
                                               footer: nil)
         } else {
-            storedSection = ComponentsSection(header: .init(title: localizedString(.paymentMethodsStoredMethods,
-                                                                                   configuration.localizationParameters),
-                                                            style: configuration.style.listComponent.sectionHeader),
-                                              components: storedComponents)
+            storedSection = ComponentsSection(components: storedComponents)
         }
         
         // Regular section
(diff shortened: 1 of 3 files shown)


AdyenDropIn/Utilities/ComponentManager.swift
```diff
@@ -78,11 +78,14 @@ internal final class ComponentManager {
...
-                                                            style: ListSectionHeaderStyle()),
+                                                            style: configuration.style.listComponent.sectionHeader),
                                               components: storedComponents,
                                               footer: nil)
         } else {
-            storedSection = ComponentsSection(components: storedComponents)
+            storedSection = ComponentsSection(header: .init(title: localizedString(.paymentMethodsStoredMethods,
+                                                                                   configuration.localizationParameters),
+                                                            style: configuration.style.listComponent.sectionHeader),
+                                              components: storedComponents)
```
mohammedDehairy: i think i made a mistake when i told to do this, since in this case there shouldn't be any title for section.
i think you need to revert this :-), sorry!

AdyenDropIn/Utilities/ComponentManager.swift
```diff
@@ -78,11 +78,14 @@ internal final class ComponentManager {
             storedSection = ComponentsSection(header: .init(title: localizedString(.paymentMethodsStoredMethods,
                                                                                    configuration.localizationParameters),
                                                             editingStyle: editingStyle,
-                                                            style: ListSectionHeaderStyle()),
+                                                            style: configuration.style.listComponent.sectionHeader),
```
mohammedDehairy, line 81: 👍

AdyenDropIn/Utilities/ComponentManager.swift
neelSharma12: Okay.

mohammedDehairy (approved):

descorp (approved):

PR #1107, opened 2022-12-21
Update docs and README

descorp:
#  Open PR

## Changes

* Update code snippets in Docs and README

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,9 +112,11 @@ Please read more [here](https://docs.adyen.com/development-resources/client-side
 Use **Environment.test** for environment. When you're ready to accept live payments, change the value to one of our [live environments](https://adyen.github.io/adyen-ios/Docs/Structs/Environment.html)
 
 ```swift
-let apiContext = APIContext(clientKey: clientKey, environment: Environment.test)
-let context = AdyenContext(apiContext: apiContext, analyticsConfiguration: analyticsConfiguration)
-let configuration = DropInComponent.Configuration(context: context)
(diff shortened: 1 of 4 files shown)


neelSharma12 (approved):

README.md
```diff
@@ -112,9 +112,11 @@ Please read more [here](https://docs.adyen.com/development-resources/client-side
 Use **Environment.test** for environment. When you're ready to accept live payments, change the value to one of our [live environments](https://adyen.github.io/adyen-ios/Docs/Structs/Environment.html)
 
 ```swift
-let apiContext = APIContext(clientKey: clientKey, environment: Environment.test)
-let context = AdyenContext(apiContext: apiContext, analyticsConfiguration: analyticsConfiguration)
-let configuration = DropInComponent.Configuration(context: context)
+let apiContext = try! APIContext(environment: componentsEnvironment, clientKey: clientKey)
```
mohammedDehairy, line 115: is it a good idea to use `automatic unwrapping` in the code snippet? 🤔

descorp: It can throw only if merchant does something wrong (using **public key** instead of **client key** etc). 
I don't like force-unwrapping, but otherwise we need to do pseudo-code error handling (try/catch or guard) that is going to make snippet bigger 🤔

Do you have a nice suggestions?

mohammedDehairy: hmm 🤔 , it sounds weird, but some merchants some copy past our example code and blame us when something wrong happens, its up to you :-)

mohammedDehairy (approved):

PR #1140, opened 2023-02-07
Docc script: Fix PIF cache issue on SPM

descorp:
# Open PR

## Changes

* apply [PIF SMP](https://github.com/apple/swift-package-manager/issues/5767#issuecomment-1263262791) fix to "Generate Docc" script

### DIFF
--- a/Scripts/generate_docc_documentation.sh
+++ b/Scripts/generate_docc_documentation.sh
@@ -116,6 +116,13 @@ swift package update
 
 DERIVED_DATA_PATH=.build
 
+MAX_ATTEMP=3
+ATTEMPT=0
+while [ -z $SUCCESS ] && [ "$ATTEMPT" -le "$MAX_ATTEMP" ]; do
+  xcodebuild clean -scheme $PROJECT_NAME -destination 'generic/platform=iOS' | grep -q "CLEAN SUCCEEDED" && SUCCESS=true
+  ATTEMPT=$(($ATTEMPT+1))
+done
+
 # Generate the docc archive.
 xcodebuild docbuild \
  -scheme Adyen \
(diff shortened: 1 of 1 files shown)


Scripts/generate_docc_documentation.sh
```diff
@@ -116,6 +116,13 @@ swift package update
 
 DERIVED_DATA_PATH=.build
 
+MAX_ATTEMP=3
+ATTEMPT=0
+while [ -z $SUCCESS ] && [ "$ATTEMPT" -le "$MAX_ATTEMP" ]; do
+  xcodebuild clean -scheme $PROJECT_NAME -destination 'generic/platform=iOS' | grep -q "CLEAN SUCCEEDED" && SUCCESS=true
+  ATTEMPT=$(($ATTEMPT+1))
+done
+
```
mohammedDehairy, line 125: the fix [here](https://github.com/Adyen/adyen-ios/blob/349c603267269d3b8952432329448bf5ccc70af0/Scripts/test-SPM-integration.sh#L61) is different and it works fine, is there a reason to go for a different approach? since this one is already tested :-)

erenbesel (dismissed):

Scripts/generate_docc_documentation.sh
descorp: I found that "good fix" is not 100% reliable..
This "ugly" one is bit more robust 😄 

Agree, let's try "good one" first

descorp: Fixed [bf2def8](https://github.com/Adyen/adyen-ios/pull/1140/commits/bf2def8c6cb7ab868ad8b2afdfb280daef8353a5)

erenbesel (approved):

mohammedDehairy (approved):

PR #1143, opened 2023-02-07
fix: returning 8 digit pan when bin length is >= 16

erenbesel:
### Fixed
<fixed>

-   For cards, `CardComponentDelegate.didChangeBIN(:component:)` now provides the 8-digit BIN when the card number is 16 or more digits. Previously, it only provided the 8-digit BIN when the card number was 17 or more digits.

</fixed>

### DIFF
--- a/AdyenCard/Form/FormCardNumberItem.swift
+++ b/AdyenCard/Form/FormCardNumberItem.swift
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2022 Adyen N.V.
+// Copyright (c) 2023 Adyen N.V.
 //
 // This file is open source and available under the MIT license. See the LICENSE file for more info.
 //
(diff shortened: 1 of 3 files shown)


AdyenCard/Form/FormCardNumberItem.swift
```diff
@@ -90,7 +90,7 @@ internal final class FormCardNumberItem: FormTextItem, AdyenObserver {
     
     private func updateBINIfNeeded() {
         switch (value, isValid()) {
-        case (_, true) where value.count > Constants.minimumPANLength:
+        case (_, true) where value.count >= Constants.minimumPANLength:
```
descorp, line 93: BTW, `minimumPANLength` should be 8 only PAN if is 16 or longer, otherwise 6

Tests/Card Tests/BCMCComponentTests.swift
```diff
@@ -239,8 +239,8 @@ class BCMCComponentTests: XCTestCase {
         expectationBin.expectedFulfillmentCount = 1
         expectationBin.assertForOverFulfill = true
         let delegateMock = CardComponentDelegateMock(onBINDidChange: { value in
-                                                         XCTAssertTrue("670344".hasPrefix(value))
-                                                         XCTAssertTrue(value.count <= 6)
+                                                         XCTAssertTrue("67034444".hasPrefix(value))
```
erenbesel, line 242: The change means now this card will return 8 digit bin since its length 16. It used to return 6 digits

descorp (approved):

PR #1151, opened 2023-02-14
Fixed the publish-demo-app.yml github workflow

mohammedDehairy:
Fixed the publish-demo-app.yml github workflow

### DIFF
--- a/.github/workflows/publish-demo-app.yml
+++ b/.github/workflows/publish-demo-app.yml
@@ -56,7 +56,7 @@ jobs:
         DEMO_SERVER_API_KEY: ${{ secrets.DEMO_SERVER_TEST_ENV_API_KEY }}
         MERCHANT_ACCOUNT: ${{ secrets.MERCHANT_ACCOUNT }}
         APPLE_DEVELOPMENT_TEAM_ID: ${{ secrets.APPLE_DEVELOPMENT_TEAM_ID }}
-        ENVIRONMENT: ${{ secrets.ENVIRONMENT }}
+        ENVIRONMENT: ${{ env.ENVIRONMENT }}
       run: |
         # client Key
         sed -i '' "s/static let clientKey.*$/static let clientKey = \"$CLIENT_KEY\"/g" ./Demo/Configuration.swift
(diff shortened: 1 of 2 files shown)


erenbesel (dismissed):

neelSharma12 (dismissed):

.github/workflows/publish-demo-app.yml
```diff
@@ -54,12 +54,30 @@ jobs:
       env:
         CLIENT_KEY: ${{ secrets.DEMO_APP_TEST_ENV_CLIENT_KEY }}
         DEMO_SERVER_API_KEY: ${{ secrets.DEMO_SERVER_TEST_ENV_API_KEY }}
+        MERCHANT_ACCOUNT: ${{ secrets.MERCHANT_ACCOUNT }}
+        APPLE_DEVELOPMENT_TEAM_ID: ${{ secrets.APPLE_DEVELOPMENT_TEAM_ID }}
+        ENVIRONMENT: ${{ secrets.ENVIRONMENT }}
```
descorp: This is not a "secret" 😄

descorp (dismissed):

Adyen.xcodeproj/project.pbxproj
```diff
@@ -6983,7 +6983,7 @@
 				CODE_SIGN_ENTITLEMENTS = Demo/Shared.entitlements;
 				CODE_SIGN_IDENTITY = "Apple Development";
 				CODE_SIGN_STYLE = Automatic;
-				CURRENT_PROJECT_VERSION = 4;
+				CURRENT_PROJECT_VERSION = 7;
```
descorp: Do we need to increment this automatically? 🤔

.github/workflows/publish-demo-app.yml
descorp:
```suggestion
        ENVIRONMENT: 'test'
```

mohammedDehairy: its not always test, it can be live also, if we want, right?

Adyen.xcodeproj/project.pbxproj
mohammedDehairy: we can try to increment automatically, but maybe better to do it an another PR

.github/workflows/publish-demo-app.yml
mohammedDehairy: made it an environment variable instead of a secret

erenbesel (approved):

descorp (approved):

PR #1171, opened 2023-03-02
fix: Allow sessions without country code

descorp:
# Open PR

## Changes

<fixed>

* Now Sessions could be used without optional `countryCode` parameter

</fixed>

### DIFF
--- a/AdyenSession/AdyenSession.swift
+++ b/AdyenSession/AdyenSession.swift
@@ -57,7 +57,7 @@ public final class AdyenSession {
         public let identifier: String
         
         /// Country Code
-        public let countryCode: String
+        public let countryCode: String?
         
         /// Shopper Locale
         public let shopperLocale: String?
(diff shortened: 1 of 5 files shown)


erenbesel (dismissed):

erenbesel (commented):
Will approve after checking what we discussed about only changing non public one

AdyenSession/AdyenSession.swift
```diff
@@ -57,7 +57,7 @@ public final class AdyenSession {
         public let identifier: String
         
         /// Country Code
-        public let countryCode: String
+        public let countryCode: String?
```
erenbesel: reminder to check if this one is not needed since it's a public property

descorp: It is needed. 'Context' is a readonly value from API, and API may not return `countryCode` because it is optional.

naxel: @erenbesel @descorp, when are you planning to release it?

descorp: Sorry @naxel 

It takes longer, since this could be a breaking change, and we are debating on how to minimise impact.
This will be part of v5.2.0

PR #1181, opened 2023-03-20
bump 3DS2 SDK version to 2.3.0

jdmoreira:
This is a cherry-pick from https://github.com/Adyen/adyen-ios/commit/34fe5e7da5b2aa8389d2f3b936c4f99d1c43679a
towards `v4`.

We at http://github.com/bokadirekt would like to continue using `v4` for some more time but we have noticed some expired certificates with 2.2.6 version of Adyen3DS2 
screenshot_2023-03-17_at_14 43 26

### DIFF
--- a/Adyen.podspec
+++ b/Adyen.podspec
@@ -56,7 +56,7 @@ Pod::Spec.new do |s|
 
   s.subspec 'Actions' do |plugin|
     plugin.dependency 'Adyen/Core'
-    plugin.dependency 'Adyen3DS2', '2.2.6'
+    plugin.dependency 'Adyen3DS2', '2.3.0'
     plugin.source_files = 'AdyenActions/**/*.swift'
     plugin.exclude_files = 'AdyenActions/**/BundleSPMExtension.swift'
     plugin.resource_bundles = {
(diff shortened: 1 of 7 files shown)


jdmoreira:
@mohammedDehairy could you advise here 😊

erenbesel:
> @mohammedDehairy could you advise here 😊

Hey @jdmoreira , as @mohammedDehairy can confirm, this is a false alarm. You are good to use v4 without any version upgrades on the 3ds2 version :)

jdmoreira:
Ok, interesting.
Even after bumping the sdk we still have an issue with a particular card.

Right now we can see in the logs:
```
Error Domain=com.adyen.Adyen3DS2.ADYProtocolError Code=201 "Message element defined as required is missing from the message." UserInfo={ADYProtocolErrorServerTransactionIdentifierKey=6b503cd3-4285-4dcf-ad00-c6bc57bf1134, ADYProtocolErrorSDKTransactionIdentifierKey=5fab70f3-2eb3-4bdc-9045-4716c7c395f2, ADYProtocolErrorDetailKey=threeDSServerTransID, ADYProtocolErrorACSTransactionIdentifierKey=85846210-c908-4a2a-9fdd-e5de49dd6191, NSLocalizedDescription=Message element defined as required is missing from the message.}
```

We've read https://github.com/Adyen/adyen-3ds2-ios/issues/18
but I find it curious that we can add the card with Adyen on web (possibly even on Android). So would web not be validating `threeDSServerTransID ` then? Feels very inconsistent.

erenbesel:
@jdmoreira I believe that question is better asked on the 3ds2 repo or via our customer support. 
Thanks for the thoughtful PR!

PR #1184, opened 2023-03-31
Added Refactored DropIn with session flow for demo app

neelSharma12:
### Chore
<chore>

-   Refactored DropInWithSession demo app  code.

</chore>


Demo/Common/IntegrationExamples/Protocols/DropInExampleProtocol.swift
```diff
@@ -0,0 +1,65 @@
...
+// Copyright (c) 2023 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+@_spi(AdyenInternal) import Adyen
+import AdyenNetworking
+import AdyenSession
+
+internal protocol DropInExampleProtocol: APIClientAware {
```
descorp: I think the only difference between this and `ComponentExampleProtocol` would be `palApiClient` ?

descorp: Technically it is backend's job to do PAL calls. Maybe we can simply combine `palApiClient` and `apiClient` as one generic `apiClient` ?
Or even move `palApiClient` out of the scope of "simple DropIn" example?

Demo/Common/IntegrationExamples/Protocols/PresenterExampleProtocol.swift
```diff
@@ -0,0 +1,19 @@
+////
+//// Copyright (c) 2023 Adyen N.V.
+////
+//// This file is open source and available under the MIT license. See the LICENSE file for more info.
+////
+//
+
+import UIKit
+
+internal protocol PresenterExampleProtocol: AnyObject {
```
descorp, line 10: This is nice :)

Demo/UIKit/ComponentsViewController.swift
```diff
@@ -30,18 +30,6 @@ internal final class ComponentsViewController: UIViewController, Presenter {
...
-                ComponentsItem(title: "Card", selectionHandler: presentCardComponent),
-                ComponentsItem(title: "iDEAL", selectionHandler: presentIdealComponent),
-                ComponentsItem(title: "Online Banking PL", selectionHandler: presentOnlineBankingPolandComponent),
-                ComponentsItem(title: "SEPA Direct Debit", selectionHandler: presentSEPADirectDebitComponent),
-                ComponentsItem(title: "BACS Direct Debit", selectionHandler: presentBACSDirectDebitComponent),
-                ComponentsItem(title: "MB WAY", selectionHandler: presentMBWayComponent),
-                ComponentsItem(title: "Convenience Stores", selectionHandler: presentConvenienceStore)
-            ],
-            [
-                ComponentsItem(title: "Apple Pay", selectionHandler: presentApplePayComponent)
```
descorp, line 44: Do we need to remove components? 🤔

Demo/Common/IntegrationExamples/DropIn.swift
```diff
@@ -0,0 +1,121 @@
...
+import AdyenSession
+import UIKit
+
+internal final class DropIn: DropInExampleProtocol {
+
+    // MARK: - Properties
+
+    internal var dropInComponent: PresentableComponent?
+    internal weak var presenter: PresenterExampleProtocol?
+    internal var paymentMethods: PaymentMethods?
```
descorp: We don't really need this - `session?.paymentMethods` could be used instead

Demo/Common/IntegrationExamples/DropIn.swift
```diff
@@ -0,0 +1,121 @@
...
+    internal var session: AdyenSession?
+    private lazy var containerView = UIView()
+    private lazy var loadingView = LoadingView(contentView: containerView)
```
descorp: `containerView` and `loadingView ` could be moved to `PresenterExampleProtocol`:

```swift
presenter.showsActivityIndicator = true
```

Demo/Common/IntegrationExamples/DropIn.swift
```diff
@@ -0,0 +1,121 @@
...
+        let configuration = DropInComponent.Configuration()
+
+        if let applePayPayment = try? ApplePayPayment(payment: ConfigurationConstants.current.payment,
+                                                      brand: ConfigurationConstants.appName) {
+            configuration.applePay = .init(payment: applePayPayment,
+                                           merchantIdentifier: ConfigurationConstants.applePayMerchantIdentifier)
+        }
+
+        configuration.actionComponent.threeDS.delegateAuthentication = ConfigurationConstants.delegatedAuthenticationConfigurations
+        configuration.actionComponent.threeDS.requestorAppURL = URL(string: ConfigurationConstants.returnUrl)
```
descorp: We have already set this to Sessions call 🤔
https://github.com/Adyen/adyen-ios/pull/1184/files#diff-05c97209b181bb6a898a57990a209ee18b8ef1e4784b45c72b57652247f82194R50-R54

Demo/Common/IntegrationExamples/Protocols/DropInExampleProtocol.swift
neelSharma12: I have removed ```palApiClient``` from dropIn. Here is the commit: https://github.com/Adyen/adyen-ios/pull/1184

Demo/Common/IntegrationExamples/Protocols/PresenterExampleProtocol.swift
neelSharma12: Thanks:)

Demo/Common/IntegrationExamples/DropIn.swift
neelSharma12: Yes, your are right. I have removed this and using ```session?.paymentMethods``` instead. https://github.com/Adyen/adyen-ios/pull/1184

Demo/Common/IntegrationExamples/DropIn.swift
neelSharma12: yes, I have removed this. 0f782b0d3cbf890eea056cb7060b41c496e6955e

Demo/Common/IntegrationExamples/DropIn.swift
```diff
@@ -19,10 +19,8 @@ internal final class DropIn: DropInExampleProtocol {
 
     internal var dropInComponent: PresentableComponent?
     internal weak var presenter: PresenterExampleProtocol?
-    internal var paymentMethods: PaymentMethods?
+    //internal var paymentMethods: PaymentMethods?
```
descorp: Could be removed :)

neelSharma12: yeah 😄 . removed d5b5eb7f2d00b5eaf1196654aeeda1cddc1beaea

descorp (approved):
You will have to force-merge it anyway 😅

neelSharma12:
> You will have to force-merge it anyway 😅

yes but now by approval I am sure that you have reviewed the PR thoroughly :)

Demo/Common/IntegrationExamples/DropIn.swift
```diff
@@ -0,0 +1,114 @@
...
+@_spi(AdyenInternal) import Adyen
+import AdyenActions
+import AdyenCard
+import AdyenComponents
+import AdyenDropIn
+import AdyenNetworking
+import AdyenSession
+import UIKit
+
+internal final class DropIn: DropInExampleProtocol {
```
erenbesel: Let's name these `example`s too such as `DropInExample`. Dropin could be too broad/confusing.

Demo/Common/IntegrationExamples/Protocols/DropInExampleProtocol.swift
```diff
@@ -0,0 +1,59 @@
...
+        var analyticsConfiguration = AnalyticsConfiguration()
+        analyticsConfiguration.isEnabled = true
+        return AdyenContext(apiContext: ConfigurationConstants.apiContext,
+                            payment: ConfigurationConstants.current.payment,
+                            analyticsConfiguration: analyticsConfiguration)
+    }
+
+    internal func requestAdyenSessionConfiguration(completion: ((AdyenSession.Configuration?, Error?) -> Void)?) {
+        let request = SessionRequest()
+        apiClient.perform(request) { result in
```
erenbesel: `[weak self]` since self is captured

Demo/Common/IntegrationExamples/DropIn.swift
neelSharma12: Nice point :). Made the change here: da093c2e272a314effba7b3c43882271964a0529

Demo/Common/IntegrationExamples/Protocols/DropInExampleProtocol.swift
neelSharma12: Ah! I missed it. Fixed this da093c2e272a314effba7b3c43882271964a0529

Demo/Common/IntegrationExamples/DropIn.swift
```diff
@@ -28,13 +28,16 @@ internal final class DropIn: DropInExampleProtocol {
     // MARK: - Networking
 
     internal func requestInitialData() {
-        requestAdyenSessionConfiguration { adyenSessionConfig, errorResponse in
+        requestAdyenSessionConfiguration { [weak self] adyenSessionConfig, errorResponse in
+            guard let strongSelf = self else {
```
erenbesel: no need to call it strongSelf anymore, `let self = self else` works :)

neelSharma12: Changed here  a840aa9218d8d736818e3e1e33b478ad6b50de8d

Demo/Common/IntegrationExamples/Protocols/DropInExampleProtocol.swift
erenbesel: This still does not have `[weak self]`. line 31 accesses `self`

neelSharma12: This is not a class. ```[weak self]``` is only applied to classes to remove reference cycle.

erenbesel: Ah didn't realize that. But the instance calling this will be a class and thus could fall into the capture. 🤔 

We could add `AnyObject` to this protocol and implement the capture semantics. In our case only classes would inherit this protocol anyway.

neelSharma12: Yes, you have a point. I have made this change 7635bc8fee151fe57becc96e6c6626975f37b3d3

erenbesel (approved):

PR #1186, opened 2023-04-03
Add ability to filter ApplePay by funding source

descorp:
# Open PR

## Changes 

<new>

* For Apple Pay, you can now choose to only accept either debit cards or credit cards.

</new>

### DIFF
--- a/AdyenComponents/Apple Pay/ApplePayConfiguration.swift
+++ b/AdyenComponents/Apple Pay/ApplePayConfiguration.swift
@@ -66,6 +66,9 @@ extension ApplePayComponent {
         /// An optional coupon code that is valid and has been applied to the payment request already.
         public var couponCode: String?
 
+        /// Payment source that the merchant supports. If `nil`, the transaction allows both credit and debit cards.
+        public var merchantCapabilitie: CardFundingSource?
+
         /// Initializes the configuration.
         ///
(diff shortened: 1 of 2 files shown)


erenbesel (commented):
Think it's best not to add anything extra to the growing config object at the moment, until deciding what to do with it.

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
```diff
@@ -101,6 +104,14 @@ extension ApplePayComponent {
             return paymentRequest
         }
 
+        private var merchantCapabilities: PKMerchantCapability {
+            guard let funding = merchantCapability else {
+                return .capability3DS
+            }
+
+            return [.capability3DS, funding == .credit ? .capabilityCredit : .capabilityDebit]
```
erenbesel: was referring to something like:
```
switch self {
        case .debit:
            return .capabilityDebit
        case .credit:
            return .capabilityCredit
        } 
```
        
to prevent side effects if the `CardFundingSource` enum changes

descorp: Fixed in [3785c26](https://github.com/Adyen/adyen-ios/pull/1186/commits/3785c26e8d4315a7a5d226542e91c3d24573ae31)

erenbesel (approved):

PR #1189, opened 2023-04-12
Update ApplePayDetails.swift

sam-sole:
## Summary
- What is the motivation for this change?  - To provide a public shipping address
- What existing problem does this pull request solve? - Currently the shipping address is not available, this PR makes it public after the user has authorised the payment

### DIFF
--- a/AdyenComponents/Apple Pay/ApplePayDetails.swift
+++ b/AdyenComponents/Apple Pay/ApplePayDetails.swift
@@ -29,6 +29,9 @@ public struct ApplePayDetails: PaymentMethodDetails {
     /// The user-selected shipping address for this transaction.
     public let shippingContact: PKContact?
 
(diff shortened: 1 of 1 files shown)


erenbesel:
@sam-sole Thanks a lot for your effort!
The values you are accessing are already available under the `shippingContact` property of this struct. You can access them and put them in any type of your preference. There does not seem to be a need to make a new property consisting of these values.
Additionally, we favor type safety in our instances, having a `[String: String]` dictionary does not fit into this approach.

Thanks again!

sam-sole:
@erenbesel thanks for taking a look. The shippingContact of this struct is not available when accessing the `jsonData` due to the `shippingContact` not being included in the `CodingKeys`, this is the need to make a new property for this as the shipping information is not available when using the `ApplePayComponent`. If a `[String: String]` is not acceptable what would you suggest as an alternative?

PR #1190, opened 2023-04-12
Added Advanced flow in demo app

neelSharma12:
### Chore
<chore>

-   Refactored Advanced Flow in demo app.

</chore>


Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
```diff
@@ -0,0 +1,374 @@
...
+////
+//// This file is open source and available under the MIT license. See the LICENSE file for more info.
+////
+//
+
+import AdyenActions
+import AdyenComponents
+import AdyenDropIn
+
+internal final class AdvancedFlowExample: AdvancedFlowExampleProtocol {
```
erenbesel: I think the name should include dropin similar to regular one. Like `DropInAdvancedFlowExample` etc

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
```diff
@@ -0,0 +1,374 @@
...
+
+    internal var advancedFlowComponent: PresentableComponent?
```
erenbesel: I think this is the `currentComponent` from before. The old name was more clear as to its purpose I think. Suggest to keep it as before.

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
```diff
@@ -0,0 +1,374 @@
...
+        finish(with: error)
+    }
+
+}
+
+extension AdvancedFlowExample: PaymentComponentDelegate {
+
+    internal func didSubmit(_ data: PaymentComponentData, from component: PaymentComponent) {
+        let request = PaymentsRequest(data: data)
+        apiClient.perform(request, completionHandler: paymentResponseHandler)
```
erenbesel: This captures self since `paymentResponseHandler` is a function of `self`, same as currently. let's change it.

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
```diff
@@ -0,0 +1,374 @@
...
+    }
+
+    internal func didProvide(_ data: ActionComponentData, from component: ActionComponent) {
+        (component as? PresentableComponent)?.viewController.view.isUserInteractionEnabled = false
+        let request = PaymentDetailsRequest(
+            details: data.details,
+            paymentData: data.paymentData,
+            merchantAccount: ConfigurationConstants.current.merchantAccount
+        )
+        apiClient.perform(request, completionHandler: paymentResponseHandler)
```
erenbesel: same as above.

Demo/UIKit/ComponentsViewController.swift
```diff
@@ -16,6 +16,12 @@ internal final class ComponentsViewController: UIViewController, PresenterExampl
         dropIn.presenter = self
         return dropIn
     }()
+
+    private lazy var advancedFlow: AdvancedFlowExample = {
```
erenbesel: same as above to include dropin word like `dropinAdvancedFlow`

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
```diff
@@ -0,0 +1,374 @@
...
+        apiClient.perform(request) { [weak self] result in
+            self?.handle(result: result)
+        }
+    }
+
+    private func handle(result: Result<CancelOrderResponse, Error>) {
+        switch result {
+        case let .success(response):
+            if response.resultCode == .received {
+                presenter?.presentAlert(withTitle: "Order Cancelled", message: nil)
```
nauaros: It could nice to move content like this into an `enum`.

Demo/Common/IntegrationExamples/Protocols/AdvancedFlowExampleProtocol.swift
```diff
@@ -0,0 +1,46 @@
...
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+@_spi(AdyenInternal) import Adyen
+import AdyenNetworking
+import AdyenSession
+
+internal protocol AdvancedFlowExampleProtocol: AnyObject, APIClientAware {
+    var context: AdyenContext { get }
+    var palApiClient: APIClientProtocol { get }
```
nauaros, line 13: Why do we call it `palApiClient`?

nauaros (approved):

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
```diff
@@ -0,0 +1,374 @@
...
+import AdyenComponents
+import AdyenDropIn
+
+internal final class AdvancedFlowExample: AdvancedFlowExampleProtocol {
+
+    internal var advancedFlowComponent: PresentableComponent?
+    internal var paymentMethods: PaymentMethods?
+    internal weak var presenter: PresenterExampleProtocol?
+
+    internal lazy var adyenActionComponent: AdyenActionComponent = {
```
descorp: One don't need `AdyenActionComponent` for Drop-In

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
descorp: Let's make it `internal var dropInComponent: DropInComponent?`
Same applies to "DropInExample.swift"

If merchant uses DropInComponent - they know this is a Drop-In. No need for abstractions.

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
```diff
@@ -0,0 +1,374 @@
...
+}
+
+extension AdvancedFlowExample: PresentationDelegate {
+
+    internal func present(component: PresentableComponent) {
+        present(component, delegate: self)
+    }
+}
+
+extension AdvancedFlowExample: ActionComponentDelegate {
```
descorp: This is not needed for DropIn. `didProvide` implementation could be moved to `DropInComponentDelegate`

Demo/Common/IntegrationExamples/Protocols/AdvancedFlowExampleProtocol.swift
descorp, line 13: let's combine `palApiClient` and `apiClient` into one meta "apiClient" that exposed via `APIClientAware`.

From Demo app and Merchant's perspective - there is no such thing as "palApiClient" - everything will go via merchant's backend.

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
```diff
@@ -0,0 +1,374 @@
...
+
+    @objc private func cancelPressed() {
+        advancedFlowComponent?.cancelIfNeeded()
+        presenter?.dismiss(completion: nil)
+    }
+
+}
+
+extension AdvancedFlowExample: StoredPaymentMethodsDelegate {
+    func disable(storedPaymentMethod: StoredPaymentMethod, completion: @escaping (Bool) -> Void) {
```
descorp, line 330: should be private

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
```diff
@@ -0,0 +1,374 @@
...
+                self?.presentAlert(with: errorResponse) { [weak self] in
+                    self?.requestPaymentMethods(order: nil, completion: nil)
+                }
+                return
+            }
+            self?.paymentMethods = paymentMethods
+        }
+    }
+
+    internal func presentAlert(with error: Error, retryHandler: (() -> Void)? = nil) {
```
descorp: Do we need this method? Maybe then we also need to wrap `presentAlert(withTitle:message)` ?

Demo/Common/IntegrationExamples/Protocols/AdvancedFlowExampleProtocol.swift
```diff
@@ -0,0 +1,46 @@
...
+                            analyticsConfiguration: analyticsConfiguration)
+    }
+
+    internal var palApiClient: APIClientProtocol {
+        let context = DemoAPIContext(environment: ConfigurationConstants.classicAPIEnvironment)
+        return DefaultAPIClient(apiContext: context)
+    }
+
+    internal func requestPaymentMethods(order: PartialPaymentOrder?,
+                                        completion: ((PaymentMethods?, Error?) -> Void)?) {
```
descorp, line 34: Maybe use `Result<PaymentMethods,Error>` instead?

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
```diff
@@ -0,0 +1,374 @@
+////
+//// Copyright (c) 2023 Adyen N.V.
+////
+//// This file is open source and available under the MIT license. See the LICENSE file for more info.
+////
+//
```
descorp: Header is commented :)
Too many `//`

Demo/Common/IntegrationExamples/Protocols/AdvancedFlowExampleProtocol.swift
descorp: Because it call PAL API (not Checkout API) to disable stored payment methods.

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
```diff
@@ -0,0 +1,374 @@
...
+    internal var advancedFlowComponent: PresentableComponent?
+    internal var paymentMethods: PaymentMethods?
+    internal weak var presenter: PresenterExampleProtocol?
+
+    internal lazy var adyenActionComponent: AdyenActionComponent = {
+        let handler = AdyenActionComponent(context: context)
+        handler.configuration.threeDS.delegateAuthentication = ConfigurationConstants.delegatedAuthenticationConfigurations
+        handler.configuration.threeDS.requestorAppURL = URL(string: ConfigurationConstants.returnUrl)
+        handler.delegate = self
+        handler.presentationDelegate = self
```
descorp: Those delegate implementations also not needed for Drop-In

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
```diff
@@ -0,0 +1,374 @@
...
+    // MARK: - Networking
+
+    internal func requestInitialData() {
+        requestPaymentMethods(order: nil) { [weak self] paymentMethods, errorResponse in
+            guard paymentMethods != nil else {
+                guard let errorResponse = errorResponse else {
+                    return
+                }
+                self?.presentAlert(with: errorResponse) { [weak self] in
+                    self?.requestPaymentMethods(order: nil, completion: nil)
```
descorp: Without `completion` this retry will just do nothing 🤔

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
```diff
@@ -0,0 +1,374 @@
...
+    }
+
+    // MARK: - Payment response handling
+
+    internal func handle(_ order: PartialPaymentOrder) {
+        requestPaymentMethods(order: order) { [weak self] paymentMethods, errorResponse in
+            guard let paymentMethods = paymentMethods else {
+                if let error = errorResponse {
+                    self?.presentAlert(with: error) { [weak self] in
+                        self?.requestPaymentMethods(order: order, completion: nil)
```
descorp: Without `completion` this retry will just do nothing 🤔

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
```diff
@@ -0,0 +1,374 @@
...
+        return component
+    }
+
+    // MARK: - Payment response handling
+
+    private func paymentResponseHandler(result: Result<PaymentsResponse, Error>) {
+        switch result {
+        case let .success(response):
+            if let action = response.action {
+                handle(action)
```
descorp: Can be just `dropInComponent?.handle(action)`

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
```diff
@@ -0,0 +1,374 @@
...
+                        self?.requestPaymentMethods(order: order, completion: nil)
+                    }
+                }
+                return
+            }
+            self?.handle(order, paymentMethods)
+        }
+    }
+
+    internal func handle(_ action: Action) {
```
descorp, line 122: This is not needed.

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
```diff
@@ -0,0 +1,374 @@
...
+        if let advancedFlowAsActionComponent = advancedFlowComponent as? ActionHandlingComponent {
+            /// In case current component is a `DropInComponent` that implements `ActionHandlingComponent`
+            advancedFlowAsActionComponent.handle(action)
+        } else {
+            /// In case current component is an individual component like `CardComponent`
+            adyenActionComponent.handle(action)
+        }
+    }
```
descorp: This is not needed.

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
```diff
@@ -0,0 +1,374 @@
...
+        print("User did close: \(component.paymentMethod.name)")
+    }
+
+    internal func didFail(with error: Error, from dropInComponent: AnyDropInComponent) {
+        finish(with: error)
+    }
+
+}
+
+extension AdvancedFlowExample: PaymentComponentDelegate {
```
descorp: this is actually not needed. `didSubmit` implementation could be moved to `DropInComponentDelegate`

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
neelSharma12: you mean the string ("Order Cancelled") ?

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
neelSharma12: Nice point :). I have renamed it here da858bb9a858774fa139005a65bb215f94ed0c10

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
neelSharma12: Made the change: da858bb9a858774fa139005a65bb215f94ed0c10

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
neelSharma12: Yes you are right. I have made the change. da858bb9a858774fa139005a65bb215f94ed0c10

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
neelSharma12: Made the change. da858bb9a858774fa139005a65bb215f94ed0c10

Demo/UIKit/ComponentsViewController.swift
neelSharma12: Okay. Updated it in this commit: da858bb9a858774fa139005a65bb215f94ed0c10

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
neelSharma12: Removed it. da858bb9a858774fa139005a65bb215f94ed0c10

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
neelSharma12: Removed it. da858bb9a858774fa139005a65bb215f94ed0c10

Demo/Common/IntegrationExamples/Protocols/AdvancedFlowExampleProtocol.swift
neelSharma12: We will discuss this and take this in separate PR.

Demo/Common/IntegrationExamples/Protocols/AdvancedFlowExampleProtocol.swift
neelSharma12: I need to use optional error in the completion and it does not accept an optional error. So, I cannot use this.

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
neelSharma12: Ohh yes. I do not  how I missed it 😮 . Changed it here. da858bb9a858774fa139005a65bb215f94ed0c10

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
neelSharma12: Removed them. da858bb9a858774fa139005a65bb215f94ed0c10

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
neelSharma12: Added completion block for retry. da858bb9a858774fa139005a65bb215f94ed0c10

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
neelSharma12: yes. Made the change da858bb9a858774fa139005a65bb215f94ed0c10

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
neelSharma12: Yes. Removed it da858bb9a858774fa139005a65bb215f94ed0c10

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
neelSharma12: Changed it : da858bb9a858774fa139005a65bb215f94ed0c10

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
neelSharma12: Made the change: da858bb9a858774fa139005a65bb215f94ed0c10

Demo/UIKit/ComponentsViewController.swift
```diff
@@ -16,6 +16,12 @@ internal final class ComponentsViewController: UIViewController, PresenterExampl
...
+        let advancedFlow = AdvancedFlowExample()
+        advancedFlow.presenter = self
+        return advancedFlow
+    }()
```
descorp, line 24: BTW, we need to add this to SwiftUI project as well..

descorp: #### Side note: 
We might think of making a nice abstraction over it (ex `Example Protocol`) and inject those "examples" per choice, rather than instantiating them as lazy vars.

There is no point in "laziness" since we are going to call "requestInitialData" right after creation.

neelSharma12: Thanks for your valuable comment :) Yes, we will improve this in separate PR :)

Demo/Common/IntegrationExamples/AdvancedFlowExample.swift
neelSharma12: This is a demo app and we want to keep it as simple as possible for merchant integration. So, I think for now we can keep it like this but in future if we feel like changing it we will surely do :)

Demo/SwiftUI/PaymentsViewModel.swift
```diff
@@ -8,10 +8,10 @@ import SwiftUI
...
 internal final class PaymentsViewModel: ObservableObject, Identifiable, PresenterExampleProtocol {
 
-    private lazy var dropIn: DropInExample = {
-        let dropIn = DropInExample()
-        dropIn.presenter = self
-        return dropIn
+    private lazy var dropInAdvancedFlow: DropInAdvancedFlowExample = {
+        let dropInAdvancedFlow = DropInAdvancedFlowExample()
+        dropInAdvancedFlow.presenter = self
+        return dropInAdvancedFlow
```
descorp, line 14: You need both `dropInAdvancedFlow` and `dropIn` 🤔

Demo/SwiftUI/PaymentsViewModel.swift
```diff
@@ -21,7 +21,7 @@ internal final class PaymentsViewModel: ObservableObject, Identifiable, Presente
     // MARK: - DropIn Component
 
     internal func presentDropInComponent() {
-       // todo: add advanced flow
```
descorp, line 24: How was it working without `dropIn.presentDropInComponent()` ?

neelSharma12: I added this `todo: ` in my last PR.

Demo/SwiftUI/PaymentsViewModel.swift
neelSharma12: I do not think so because we did not had session condition here in the old code as well. Also, for components e.g iDEAL, Card etc. we had no session condition.

erenbesel (approved):

Demo/SwiftUI/PaymentsViewModel.swift
descorp: Ok, good point

descorp (approved):

PR #1198, opened 2023-04-19
Update 3ds2sdk release 2.3.1

robertdalmeida:
<changed> Updated 3ds2sdk version to 2.3.1  </changed>

### DIFF
--- a/AdyenActions/Components/3DS2/3DS2 SDK Adapters/AnyChallengeResult.swift
+++ b/AdyenActions/Components/3DS2/3DS2 SDK Adapters/AnyChallengeResult.swift
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2021 Adyen N.V.
+// Copyright (c) 2023 Adyen N.V.
 //
 // This file is open source and available under the MIT license. See the LICENSE file for more info.
 //
@@ -11,7 +11,7 @@ internal protocol AnyChallengeResult {
 
     var sdkTransactionIdentifier: String { get }
 
-    var transactionStatus: String? { get }
+    var transactionStatus: String { get }
 }
 
 extension ADYChallengeResult: AnyChallengeResult {}
--- a/.github/workflows/test-carthage-integration.yml
+++ b/.github/workflows/test-carthage-integration.yml
@@ -12,7 +12,7 @@ on:
 jobs:
 
   carthage:
-    runs-on: macos-11
+    runs-on: macos-12
     steps:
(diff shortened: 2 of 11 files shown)


AdyenActions/Components/3DS2/3DS2 SDK Adapters/AnyChallengeResult.swift
```diff
@@ -11,7 +11,7 @@ internal protocol AnyChallengeResult {
 
     var sdkTransactionIdentifier: String { get }
 
-    var transactionStatus: String? { get }
+    var transactionStatus: String { get }
```
erenbesel, line 14: This looks like a breaking change for v4, we may not be able to change this.

Also could be the reason why some jobs are failing.
Did you run the test target fully and did they all pass?

descorp: It is internal 🤔

robertdalmeida: The type `ADYChallengeResult` is public. Which means that it is a breaking change for checkout.
Discussed with @erenbesel. We can only include a version that doesn't have a breaking change. 

The max version without the breaking change is 2.2.6, which is the existing version.

descorp: But nothing is breaking? Am I missing something?

robertdalmeida: Now that you mentioned it. It is indeed not breaking for the consumer of checkout. 
But only for checkout who is a consumer of the 3ds2sdk. 
@erenbesel WDYT?

erenbesel: Ohh my bad. Haven't noticed it's internal. All good :)
But jobs are failing, please look into them, I imagine some tests may fail due to this change.

descorp: I have fixed Cocoapods. Will check Carthage..

descorp:
Fixed up Carthage :)

descorp (approved):

erenbesel (approved):

neelSharma12 (approved):

PR #1217, opened 2023-05-22
COIOS-603: ACH component improvements

goergisn:
### New
<new>

- For Drop-in, you can now configure the ACH Direct Debit payment method. When creating a Drop-in configuration, set the following in `ach` properties:
  - `showsStorePaymentMethodField`
  - `showsBillingAddress` 
  - `billingAddressCountryCodes`

</new>

### DIFF
--- a/AdyenDropIn/Models/DropInConfiguration.swift
+++ b/AdyenDropIn/Models/DropInConfiguration.swift
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2022 Adyen N.V.
+// Copyright (c) 2023 Adyen N.V.
 //
 // This file is open source and available under the MIT license. See the LICENSE file for more info.
 //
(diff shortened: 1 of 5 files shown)


AdyenDropIn/Models/DropInConfiguration.swift
```diff
@@ -84,6 +87,12 @@ public extension DropInComponent {
         public var showEmailAddress: Bool = true
     }
     
+    /// The ACH configuration
+    struct ACH {
```
goergisn: Alternative solution would be to use the `ACHDirectDebitComponent.Configuration` directly and have some fields overridden in the PaymentComponentBuilder @erenbesel

descorp: We can also do `AnyACHDirectDebitComponentConfiguration` 🤔

descorp: IMHO, this is good for now 🤔

descorp (dismissed):

erenbesel (approved):

PR #1221, opened 2023-05-23
chore: Using default header comment on new files

goergisn:
## Changes

* new files now use our default header comments


descorp (approved):

nauaros (approved):
Nice one!

PR #1222, opened 2023-05-24
feature: Add pin required config for gift card

goergisn:
### New
<new>

- For the gift cards using Drop-in, you can now configure the visibility of the security code field. Set `showsSecurityCodeField` (boolean) in the `giftcard` [payment method configuration](https://docs.adyen.com/online-payments/build-your-integration?platform=iOS&integration=Drop-in&version=5.1.0#configure-drop-in).

</new>


Tests/Components Tests/Gift Card/GiftCardComponentTests.swift
```diff
@@ -572,6 +577,45 @@ class GiftCardComponentTests: XCTestCase {
...
+    }
+    
+    func testMealVoucherHidingSecurityCodeItemView() {
+        
+        // Given
+        let paymentMethod = MealVoucherPaymentMethod(type: .mealVoucherSodexo, name: "Sodexo")
+        sut = GiftCardComponent(partialPaymentMethodType: .mealVoucher(paymentMethod),
+                                context: context,
+                                amount: amountToPay,
+                                showsSecurityCodeField: false,
```
erenbesel, line 606: Think we need the test cases for `showsSecurityCodeField = true` as well so all paths are run for both giftcard and mealvoucher

goergisn: Fixed in 2e405cb4b518bb24fc4f29bb041bc38e7c280830

erenbesel (approved):

neelSharma12 (approved):

PR #1246, opened 2023-06-15
chore: Allow hiding of imageview in listitem

goergisn:
## Changes

* add option to hide `imageView` in the `ListItemView`
* refactoring the `ListItemView` to use a horizontal `UIStackView` for layout

### DIFF
--- a/Adyen/UI/List/ListItem.swift
+++ b/Adyen/UI/List/ListItem.swift
@@ -18,6 +18,7 @@ public class ListItem: FormItem {
         /// Standard icon that can be styled
         case generic
         
+        /// Whether or not the icon should be hidden
         public var isHidden: Bool {
             switch self {
             case .custom: return false
@@ -26,6 +27,7 @@ public class ListItem: FormItem {
             }
         }
         
+        /// Whether or not the icon should be styled/altered
         public var canBeModified: Bool {
             switch self {
             case .custom: return false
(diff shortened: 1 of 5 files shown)


erenbesel (dismissed):

Adyen/UI/List/ListItem.swift
```diff
@@ -37,31 +62,35 @@ public class ListItem: FormItem {
     /// that is set to the `ListItemView.accessibilityIdentifier`.
     public var identifier: String?
 
-    /// The flag to indicate if an icon is a custom image that should not be tempered.
-    public let canModifyIcon: Bool
+    /// Specifies the icon behavior
+    public let iconMode: IconMode
```
descorp: This is breaking change 😅
It is **spi**, so we don't need whole dance with migration guides, but still.
If you planning for this feature to be in v5 - need to deprecate the old one.

Adyen/UI/List/ListItemView.swift
```diff
@@ -67,11 +65,12 @@ public final class ListItemView: UIView, AnyFormItemView {
         }
         
         imageView.imageURL = item?.imageURL
+        imageView.isHidden = item?.iconMode.isHidden == true
```
descorp: why not just `imageView.isHidden = item?.iconMode.isHidden` ?

Adyen/UI/List/ListItemView.swift
```diff
@@ -139,34 +137,37 @@ public final class ListItemView: UIView, AnyFormItemView {
         return stackView
     }()
     
+    private lazy var contentStackView: UIStackView = {
```
descorp, line 141: Nice :)

Adyen/UI/List/ListItemView.swift
goergisn: `item?.iconMode.isHidden` produces an Optional<Bool> (See code above as well)

Adyen/UI/List/ListItem.swift
goergisn: 👌 Fixed here 6e9581df7f7ea068c1c60ff17f36787b5de9f3c1

descorp (dismissed):

neelSharma12 (dismissed):

Adyen/UI/List/ListItem.swift
erenbesel: I don't think spi code requires deprecation. It's the purpose of the spi block. It's meant for our usage between modules. If change is needed, we can do it freely without worrying about public interface change, since this is not "public" interface.

erenbesel (approved):

PR #1251, opened 2023-06-21
chore: Generalized AddressValidator code

goergisn:
## Changes

* simplifying and generalizing address validation code
* adding computed property requiredFields to AddressViewModel
* fixed a hidden bug where the check for nil would not return the expected value


Adyen/UI/Form/Items/Address/AddressViewModel.swift
```diff
@@ -55,6 +55,17 @@ public struct AddressViewModel {
 
 }
 
+@_spi(AdyenInternal)
+public extension AddressViewModel {
+    
+    /// Returns all fields that are not specified as `optionalFields`
+    var requiredFields: Set<AddressField> {
```
neelSharma12, line 62: Can we test these changes for countries like "US" , "CA", "GB" and "BR" because they have different address forms?

goergisn: Good point! Fixed in 652e8de2a5568a7e5056c680cabd59327df34c98

descorp (approved):

erenbesel (approved):

PR #1255, opened 2023-06-26
new: Adding FormSelectableValueItem

goergisn:
## Changes

* adding `FormValidatableValueItem` and `FormValidatableValueItemView` that generalizes validation/display logic
* adding `FormSelectableValueItem` and `FormSelectableValueItemView` to display a generic (formattable) value in a form and allows selection

### DIFF
--- a/Adyen/UI/Form/Items/Text/FormTextItemView.swift
+++ b/Adyen/UI/Form/Items/Text/FormTextItemView.swift
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2022 Adyen N.V.
+// Copyright (c) 2023 Adyen N.V.
 //
 // This file is open source and available under the MIT license. See the LICENSE file for more info.
 //
@@ -32,9 +32,11 @@ public protocol AnyFormTextItemView: AnyFormItemView {
 
 /// A view representing a basic logic of text item.
 @_spi(AdyenInternal)
-open class FormTextItemView<ItemType: FormTextItem>: FormValueItemView<String, FormTextItemStyle, ItemType>,
+open class FormTextItemView<ItemType: FormTextItem>: FormValidatableValueItemView<String, ItemType>,
     UITextFieldDelegate,
     AnyFormTextItemView {
+    
+    override public var accessibilityLabelView: UIView? { textField }
     
     /// Initializes the text item view.
     ///
@@ -57,7 +59,7 @@ open class FormTextItemView<ItemType: FormTextItem>: FormValueItemView<String, F
             self?.alertLabel.text = newValue
         }
     }
-    
+
     override public func reset() {
         item.value = ""
         resetValidationStatus()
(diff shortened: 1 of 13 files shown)


Adyen/UI/Form/Items/Value/FormValueDetailItem.swift
```diff
@@ -0,0 +1,42 @@
...
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+import UIKit
+
+/// An item in a form that holds a value and displays it in a formatted way
+@_spi(AdyenInternal)
+open class FormValueDetailItem<ValueType: Equatable>: FormValueItem<ValueType, FormTextItemStyle>, ValidatableFormItem {
```
descorp: How about name it `FormReadOnlyTextItem` ?

Or maybe just add extra mode/flag for `FormTextItemView` instead of creating new FormItem ?

Adyen/UI/Form/Items/Value/FormValueDetailItem.swift
```diff
@@ -0,0 +1,42 @@
...
+        self.placeholder = placeholder
+        
+        super.init(value: value, style: style)
+    }
+    
+    // MARK: ValidatableFormItem
+    
+    @AdyenObservable(nil) public var validationFailureMessage: String?
+    
+    public func isValid() -> Bool {
```
em0072: imho, the computed property is better suited for this type of checks

Adyen/UI/Form/Items/Value/FormValueDetailItemView.swift
```diff
@@ -0,0 +1,189 @@
...
+            hideAlertLabel(true)
+            unhighlightSeparatorView()
+            titleLabel.textColor = defaultTitleColor
+            valueLabel.accessibilityLabel = item.title
+        }
+    }
+    
+    private func hideAlertLabel(_ hidden: Bool, animated: Bool = true) {
+        guard hidden || alertLabel.text != nil else { return }
+        alertLabel.adyen.hide(animationKey: "hide_alertLabel", hidden: hidden, animated: animated)
```
em0072: Would be quite useful to move "hide_alertLabel" into an enum so you don't have to deal with strings in the middle of the code.

Adyen/UI/Form/Items/Value/FormValueDetailItem.swift
goergisn: Thanks for the input, we considering changing it in a future PR as this would be a bigger change because it is an implementation of a protocol. :)

Adyen/UI/Form/Items/Value/FormValueDetailItem.swift
goergisn: The `FormTextItemView` expects a non optional String as the value of the Item which in this case should be possible to be any type that conforms to Equatable (including being `optional`)
As discussed I refactored the code to introduce a `FormValidatableValueItem` and View that generalizes the validation logic/UI and renamed `FormValueDetailItem` to `FormSelectableValueItem` to reflect the use.

Adyen/UI/Form/Items/Text/FormTextItemView.swift
```diff
@@ -282,24 +260,14 @@ open class FormTextItemView<ItemType: FormTextItem>: FormValueItemView<String, F
...
-        } else {
-            accessory = .none
-        }
-    }
-    
-    private func hideAlertLabel(_ hidden: Bool, animated: Bool = true) {
-        guard hidden || alertLabel.text != nil else { return }
-        alertLabel.adyen.hide(animationKey: "hide_alertLabel", hidden: hidden, animated: animated)
+        if case .customView = accessory { return }
+        accessory = .none
```
descorp, line 270: 👍

descorp (approved):

neelSharma12 (approved):

PR #1258, opened 2023-06-28
Add tests for `didCancel` on Drop-In

descorp:
# Open PR

## Changes

<fixed>

* Now DropInComponentDelegate calls `didCancel` works

</fixed>


AdyenDropIn/DropInComponent.swift
```diff
@@ -227,29 +227,26 @@ public final class DropInComponent: NSObject,
...
-        
         if isRoot {
             self.delegate?.didFail(with: ComponentError.cancelled, from: self)
         } else {
+            userDidCancel(component)
             navigationController.popViewController(animated: true)
         }
     }
 
+    /// This methos is called when
```
goergisn:
```suggestion
    /// This method is called when the user cancelled a component
```

descorp: I removed it.

PR #1259, opened 2023-06-28
feature: Add enforced locale init for LocalizationParameters

descorp:
# Open PR

## Changes

<new>

* Now merchants can enforce specific localisation

</new>

### DIFF
--- a/Adyen/Utilities/LocalizationParameters.swift
+++ b/Adyen/Utilities/LocalizationParameters.swift
@@ -1,33 +1,71 @@
 //
-// Copyright (c) 2022 Adyen N.V.
+// Copyright (c) 2023 Adyen N.V.
 //
 // This file is open source and available under the MIT license. See the LICENSE file for more info.
 //
 
 import Foundation
 
+internal enum LocalizationMode: Equatable {
+    case natural(bundle: Bundle?, tableName: String?, keySeparator: String?, locale: String?)
+    case enforced(locale: String)
+}
+
 /// The localization parameters to control some aspects of how localized strings are fetched,
 /// like the localization table to use and the separator of the key strings.
 public struct LocalizationParameters: Equatable {
 
-    internal let enforcedLocale: String?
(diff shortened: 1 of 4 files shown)


descorp:
Checked for SMP and Cocoapods.

Adyen/Utilities/LocalizationParameters.swift
```diff
@@ -40,5 +43,18 @@ public struct LocalizationParameters: Equatable {
...
         self.keySeparator = keySeparator
         self.locale = locale ?? Locale.current.identifier
+        self.enforcedLocale = nil
+    }
+
+    /// Initializes LocalizationParameters with enforced locale.
+    ///
+    /// - Parameters:
+    ///   - locale: The locale to be enforced.
+    public init(locale: String) {
```
goergisn:
```suggestion
    public init(enforcedLocale: String) {
```

But: Why not add the enforced locale to the end of the already existing init with a default value of nil? so it is not a breaking change :)

descorp: Because those are mutually exclusive.
Also there is no breaking change - we are introducing new init, old one remain untouched.

descorp: in v6 we can turn inner state into enum and use "A" or "B" 🤔

```swift
internal enum LocalisationMode {
    case natural(bundle: Bundle?, tableName: String?, keySeparator: String?, locale: String?)
    case enforced(locale: String)
}


```

goergisn (dismissed):

Adyen/Utilities/LocalizationParameters.swift
descorp: Fixed [e34629f](https://github.com/Adyen/adyen-ios/pull/1259/commits/e34629fcff05e6cb6e663884de456d76a8fd4fee)

neelSharma12 (approved):

PR #1262, opened 2023-06-29
chore: Adding loading state to SearchViewController

goergisn:
## Changes

* Added states (loading, empty, showingItems) to SearchViewController to allow async providing of results
* Extracted EmptyView to be customizable
* Embedded ListViewController in SearchViewController (now that it is very generic)
* Added / Adjusted tests


Adyen/UI/View Controllers/SearchViewController.swift
```diff
@@ -7,38 +7,73 @@
...
-                localizationParameters: LocalizationParameters? = nil) {
-        self.childViewController = childViewController
+    internal let style: ViewStyle
+    private let searchBarPlaceholder: String?
+    private let resultProvider: ResultProvider
+    private let shouldFocusSearchBarOnAppearance: Bool
+    
+    public init(
+        style: ViewStyle,
+        searchBarPlaceholder: String? = nil,
```
descorp: Placeholder can be different for different components?

Adyen/UI/View Controllers/SearchViewController.swift
```diff
@@ -7,38 +7,73 @@
...
     }
+    
+    internal let emptyView: SearchViewControllerEmptyView
 
-    private lazy var searchBar: UISearchBar = {
+    internal lazy var loadingView: UIActivityIndicatorView = {
+        let loadingView = UIActivityIndicatorView(style: .whiteLarge)
+        if #available(iOS 13.0, *) {
+            loadingView.color = .secondarySystemFill
+        }
```
descorp: Maybe add this colors to `Style`?
Are those colors darkmode-friendly?

goergisn: Great point regarding styling, will add that in a separate PR 👌. For now I changed it to use our default loading color.

descorp: TBH, I would prefer to address it here. But up to you

descorp (dismissed):

erenbesel (commented):
Great idea! Would be much better if the view controller didn't change the state and just redraw itself based on it. Let's discuss!

AdyenComponents/Issuer List/IssuerListComponent.swift
```diff
@@ -38,43 +41,57 @@ public final class IssuerListComponent: PaymentComponent, PaymentAware, Presenta
...
     }
     
     private let issuerListPaymentMethod: IssuerListPaymentMethod
 
     // MARK: - Presentable Component Protocol
     
     public var viewController: UIViewController {
-        searchViewController
+        let viewController = searchViewController
+        viewController.title = title
```
descorp, line 53: Nice 👍

Adyen/Utilities/BundleSPMExtension.swift
```diff
@@ -18,7 +18,7 @@ internal extension Bundle {
     static let core: Bundle = .init(for: FormView.self)
 
     /// The bundle in which the framework's resources are located.
-    static let coreInternalResources: Bundle = .module
+    public static let coreInternalResources: Bundle = .module
```
descorp, line 22: We should make `internal extension Bundle { .. }`  as `@_spi(AdyenInternal)` ?

Adyen/UI/View Controllers/SearchViewController/SearchViewController.swift
```diff
@@ -0,0 +1,192 @@
...
+            searchBar.topAnchor.constraint(equalTo: view.layoutMarginsGuide.topAnchor, constant: 0),
+
+            resultsListViewController.view.topAnchor.constraint(equalTo: searchBar.bottomAnchor, constant: 0),
+            resultsListViewController.view.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: 0),
+            resultsListViewController.view.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: 0),
+            resultsListViewController.view.bottomAnchor.constraint(equalTo: view.bottomAnchor, constant: 0),
+            
+            emptyView.leadingAnchor.constraint(equalTo: view.layoutMarginsGuide.leadingAnchor, constant: 0),
+            emptyView.trailingAnchor.constraint(equalTo: view.layoutMarginsGuide.trailingAnchor, constant: 0),
+            emptyView.centerYAnchor.constraint(equalTo: view.layoutMarginsGuide.centerYAnchor, constant: 0)
```
descorp: Can we use `UIStackView`? 🤔

goergisn: Yeah I was looking into it but in this specific case it's easier to do it using autolayout because we're hiding specific views - which indicates to the stack view that it should "remove" the hidden items - which in turn causes other layout issues unfortunately when it comes to e.g. the loading indicator

Adyen/Utilities/BundleSPMExtension.swift
goergisn: We need this to be `@_spi(AdyenInternal) public` because it's used from within the `AdyenComponents` now

Adyen/UI/View Controllers/SearchViewController/SearchViewController.swift
descorp: 🥲

Adyen/Utilities/BundleSPMExtension.swift
descorp: Currently *Adyen/Utilities/BundleSPMExtension.swift* is not under `@_spi(AdyenInternal)`
I just can't set comment on line 14 😅

goergisn: Ahh I see, will change to be consistent with `CoreBundleExtension`

goergisn: Fixed here: f19c9ffed324a42053c0c85a6dab07855e78b89a

goergisn:
> Great idea! Would be much better if the view controller didn't change the state and just redraw itself based on it. Let's discuss!

Fixed with 1f65afa125244fd2402636f612331cdd74b6d298

Adyen/UI/View Controllers/SearchViewController/SearchViewController.swift
```diff
@@ -0,0 +1,168 @@
+//
+// Copyright (c) 2023 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import UIKit
+
+@_spi(AdyenInternal)
+public protocol SearchViewControllerEmptyView: UIView {
```
erenbesel: `SearchResultEmptyView` or `SearchResultsEmptyView` to indicate this is a result view and let's add documentation above

goergisn: Fixed with a8e8581345f4712fcf9fc0b1bf84449432edfcf5

descorp (dismissed):

erenbesel (approved):

descorp (approved):

PR #1267, opened 2023-07-03
chore: Form address item configuration

goergisn:
## Changes

* Adding FormAddressItem.Configuration
  * With an option to hide the title header
* Adding more tests for FormAddressItem

### DIFF
--- a/Adyen/UI/Form/Items/Address/AddressStyle.swift
+++ b/Adyen/UI/Form/Items/Address/AddressStyle.swift
@@ -11,10 +11,12 @@ import UIKit
 public struct AddressStyle: FormValueItemStyle {
 
     /// The section header style.
-    public var title: TextStyle
+    public var title = TextStyle(font: .preferredFont(forTextStyle: .headline),
+                                 color: UIColor.Adyen.componentLabel,
+                                 textAlignment: .natural)
 
     /// The text field style.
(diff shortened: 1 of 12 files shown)


descorp (dismissed):

Adyen/UI/Form/Items/Address/AddressStyle.swift
```diff
@@ -26,8 +24,30 @@ public struct AddressStyle: FormValueItemStyle {
...
     /// The color of form view item's separator line.
     public var separatorColor: UIColor? { textField.separatorColor }
+    
+    /// Initializes the form address item configuration.
+    /// - Parameters:
+    ///   - title: The section header style.
+    ///   - textField: The text field style.
+    ///   - tintColor: The tint color of the view.
+    ///   - backgroundColor: The background color of the view.
+    public init(
```
erenbesel, line 40: This init looks a bit messy :), why change the prev one?
Or maybe we can add an empty init and a regular init and keep the default values as before?

goergisn: Agree - I had to implement a public init as we need to access the init from outside of the module.

Fixed with d6d07e00fc2f7c79d793659de85babcc5cb66942

erenbesel (approved):

PR #1268, opened 2023-07-05
feature: Address lookup

goergisn:
## Changes

<new>

* Adding functionality to provide billing address completion via a lookup handler which is called when the user searches for an address.
* You can use address lookup by setting the `configuration.card.billingAddress.mode` to `.lookup(handler:)`

</new>

https://github.com/Adyen/adyen-ios/assets/4838877/15ea5c5e-f50e-431f-86fc-96432b32aa0b


Adyen/UI/View Controllers/AddressLookup/Search/AddressLookupSearchViewController.swift
```diff
@@ -35,9 +35,38 @@ internal class AddressLookupSearchViewController: SearchViewController {
...
+                listItemStyle.title.color = .Adyen.componentSecondaryLabel
+                listItemStyle.title.textAlignment = .center
+                listItemStyle.subtitle.font = .preferredFont(forTextStyle: .subheadline)
+                listItemStyle.subtitle.color = .Adyen.defaultBlue
+                listItemStyle.subtitle.textAlignment = .center
+                
+                itemsWithManualButton.append(
+                    .init(
+                        title: "Select your address or use",
+                        subtitle: "manual address entry",
```
descorp: Can it be a single title?

Adyen/UI/View Controllers/AddressLookup/Search/AddressLookupSearchViewController+Style.swift
```diff
@@ -0,0 +1,20 @@
+//
+// Copyright (c) 2023 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import UIKit
+
+public struct AddressLookupSearchStyle: ViewStyle {
```
erenbesel, line 10: public type missing documentation

Adyen/UI/View Controllers/AddressLookup/Search/AddressLookupSearchViewController+EmptyView.swift
```diff
@@ -0,0 +1,128 @@
...
+            fatalError("init(coder:) has not been implemented")
+        }
+    }
+}
+
+// MARK: - Updating Interface
+
+private extension AddressLookupSearchViewController.EmptyView {
+    
+    func setupContent() {
```
erenbesel: small thing but this one is identical to search controllers `setupContent()`. Was thinking if we should make a common protocol/func for it

goergisn: Great catch! The views are actually so similar that I'll create a generic version of it 👌

goergisn: Fixed in 1f2519998a83402396d39b279bf7e24a717a8213

Adyen/UI/View Controllers/AddressLookup/Search/AddressLookupSearchViewController+Style.swift
goergisn: Fixed in: 1f2519998a83402396d39b279bf7e24a717a8213

erenbesel (approved):

neelSharma12 (approved):

PR #1269, opened 2023-07-06
COIOS-593: ThreeDS Opaque error handling  - Adding support to V4

robertdalmeida:
## Summary
The corresponding change for V5 of checkout is [here](https://github.com/Adyen/adyen-ios/pull/1198)

Errors from the 3ds sdk reported to the adyen backend '
Why:
Adyen doesn't have visibility of the errors that merchant face during a 3ds challenge, to help the merchant understand errors from the sdk.

What:
We provide some error information back to the merchant as a base64 string. This is passed back to the backend, so that adyen has visibility of the error that occurs and handle it accordingly.

- [x] Fix tests
- [x] Execute tests with the backend

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,8 +25,8 @@ jobs:
         - destination: 'name=iPhone 8,OS=13.7'
           scheme: AdyenUIHost
     steps:
-    - uses: actions/checkout@v2
-    - uses: n1hility/cancel-previous-runs@v2
+    - uses: actions/checkout@v3
+    - uses: n1hility/cancel-previous-runs@v3
       with:
         token: ${{ secrets.MANUAL_ACTION_TOKEN }}
 
--- a/AdyenActions/Components/3DS2/Action handlers/ThreeDS2ClassicActionHandler.swift
+++ b/AdyenActions/Components/3DS2/Action handlers/ThreeDS2ClassicActionHandler.swift
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2023 Adyen N.V.
+// Copyright (c) 2022 Adyen N.V.
 //
 // This file is open source and available under the MIT license. See the LICENSE file for more info.
 //
--- a/AdyenActions/Components/3DS2/Action handlers/ThreeDS2ClassicActionHandler.swift
+++ b/AdyenActions/Components/3DS2/Action handlers/ThreeDS2ClassicActionHandler.swift
@@ -65,15 +65,14 @@ internal class ThreeDS2ClassicActionHandler: AnyThreeDS2ActionHandler, Component
             flavor: _isDropIn ? .dropin : .components,
             environment: apiContext.environment
         )
-        coreActionHandler.handle(fingerprintAction, event: event) { [weak self] result in
-            guard let self = self else { return }
+        coreActionHandler.handle(fingerprintAction, event: event) { result in
             switch result {
             case let .success(encodedFingerprint):
                 let additionalDetails = ThreeDS2Details.fingerprint(encodedFingerprint)
                 let result = ThreeDSActionHandlerResult.details(additionalDetails)
                 completionHandler(.success(result))
             case let .failure(error):
-                completionHandler(.failure(self.onReceiveError(error: error)))
+                completionHandler(.failure(ThreeDS2ActionHandlerError(error: error)))
             }
         }
     }
@@ -98,7 +97,7 @@ internal class ThreeDS2ClassicActionHandler: AnyThreeDS2ActionHandler, Component
             case let .success(result):
                 self.handle(result, completionHandler: completionHandler)
             case let .failure(error):
-                completionHandler(.failure(self.onReceiveError(error: error)))
+                completionHandler(.failure(ThreeDS2ActionHandlerError(error: error)))
             }
         }
     }
@@ -109,19 +108,6 @@ internal class ThreeDS2ClassicActionHandler: AnyThreeDS2ActionHandler, Component
         completionHandler(.success(.details(additionalDetails)))
     }
     
-    private func onReceiveError(error: ThreeDS2CoreActionHandlerError) -> ThreeDS2ActionHandlerError {
-        switch error {
-        case let .cancellationAction(threeDSResult):
-            return .cancellation(ThreeDS2Details.challengeResult(threeDSResult))
-        case .missingTransaction:
-            return .missingTransaction
-        case let .unknown(unknownError):
-            return .unknown(unknownError)
-        case let .underlyingError(underlyingError):
-            return .underlyingError(underlyingError)
-        }
-    }
-
     // MARK: - Private
 
     private let fingerprintEventName = "3ds2fingerprint"
--- a/AdyenActions/Components/3DS2/Action handlers/ThreeDS2ClassicActionHandler.swift
+++ b/AdyenActions/Components/3DS2/Action handlers/ThreeDS2ClassicActionHandler.swift
@@ -59,7 +59,7 @@ internal class ThreeDS2ClassicActionHandler: AnyThreeDS2ActionHandler, Component
     /// - Parameter completionHandler: The completion closure.
     /// :nodoc:
     internal func handle(_ fingerprintAction: ThreeDS2FingerprintAction,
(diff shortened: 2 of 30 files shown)


AdyenActions/Components/3DS2/Action handlers/ThreeDS2ClassicActionHandler.swift
```diff
@@ -85,27 +86,41 @@ internal class ThreeDS2ClassicActionHandler: AnyThreeDS2ActionHandler, Component
...
+        switch error {
+        case let .cancellationAction(threeDSResult):
+            return .cancellation(ThreeDS2Details.challengeResult(threeDSResult))
+        case .missingTransaction:
+            return .missingTransaction
+        case let .unknown(unknownError):
+            return .unknown(unknownError)
+        case let .underlyingError(underlyingError):
+            return .underlyingError(underlyingError)
+        }
```
mohammedDehairy: here you can use the `ThreeDS2ActionHandlerError` initializer you implemented, right? :-)

AdyenActions/Components/3DS2/Action handlers/ThreeDS2CoreActionHandler.swift
```diff
@@ -8,6 +8,20 @@ import Adyen
...
+    func base64Representation() -> String {
+        "base64Representation"
+    }
+}
+
+internal enum ThreeDS2CoreActionHandlerError: Error {
+    case cancellationAction(ThreeDSResult)
+    case underlyingError(Error)
+    case unknown(UnknownError)
+    case missingTransaction
```
mohammedDehairy: whats the reason we have two enum's `ThreeDS2CoreActionHandlerError` and `ThreeDS2ActionHandlerError`, why not just one, i.e just `ThreeDS2ActionHandlerError`, they seem to be identical

AdyenActions/Components/3DS2/Action handlers/ThreeDS2CoreActionHandler.swift
```diff
@@ -144,24 +183,66 @@ internal class ThreeDS2CoreActionHandler: Component {
...
+                    completionHandler: completionHandler)
+            return
+        }
+        switch (error.domain, error.code) {
+        case (ADYRuntimeErrorDomain, Int(ADYRuntimeErrorCode.challengeCancelled.rawValue)):
+            do {
+                let cancellationResult = try ThreeDSResult(authorizationToken: challengeAction.authorisationToken,
+                                                           threeDS2SDKError: error.base64Representation())
+                let cancellationError = ThreeDS2CoreActionHandlerError.cancellationAction(cancellationResult)
+                didFail(with: cancellationError,
```
mohammedDehairy: we should make sure that the merchant will receive the same error as before, otherwise if they suddenly receive `ThreeDS2CoreActionHandlerError.cancellationAction(cancellationResult)` instead of `NSError` for cancellation then this is a breaking change, whatever the merchant used to receive as error should stay the same

AdyenActions/Components/3DS2/Action handlers/ThreeDS2CoreActionHandler.swift
```diff
@@ -144,24 +183,66 @@ internal class ThreeDS2CoreActionHandler: Component {
...
+                        completionHandler: completionHandler)
+            } catch {
+                didFail(with: .unknown(UnknownError(errorDescription: "Unable to create ThreeDSResult on error.")),
```
mohammedDehairy: do the merchant still receives `UnknownError` ? or `.unknown(UnknownError(errorDescription: "Unable to create ThreeDSResult on error."))` because this would be a breaking change :-)

AdyenActions/Components/3DS2/Fingerprint Submitter/ThreeDS2FingerprintSubmitter.swift
```diff
@@ -8,12 +8,16 @@ import Adyen
 import AdyenNetworking
 import Foundation
 
+internal enum ThreeDS2FingerprintSubmitterError: Error {
+    case underlyingError(Error)
+}
```
mohammedDehairy: whats the reason for this enum? why not use the `Error` directly?

AdyenActions/Components/3DS2/Action handlers/ThreeDS2CoreActionHandler.swift
robertdalmeida: image

I've kept an `Error` type at each layer. 

At the CoreAction level, we seem to create a `ThreeDSResult` type, 
but at the `Classic/Compact-Action` level we create the `ThreeDSActionHandlerResult` type.

Which is why we have different error types reflecting that.

AdyenActions/Components/3DS2/Action handlers/ThreeDS2CoreActionHandler.swift
robertdalmeida: The merchant will receive a new error case for cancellation for the existing type `ThreeDS2Component.challengeCancelled`, as below. Not the `CoreActionhandler` one. 
Is there a better way? 

image

AdyenActions/Components/3DS2/Action handlers/ThreeDS2CoreActionHandler.swift
mohammedDehairy: `ThreeDSActionHandlerResult` is fundamentally different than `ThreeDSResult`, but both `ThreeDS2ActionHandlerError` and `ThreeDS2CoreActionHandlerError` look identical :-)

Demo/Configuration.swift
```diff
@@ -23,7 +23,7 @@ internal enum ConfigurationConstants {
 
     static let reference = "Test Order Reference - iOS UIHost"
 
-    static let returnUrl = "ui-host://"
```
descorp, line 26: Nice :)

AdyenActions/Components/3DS2/Action handlers/ThreeDS2CoreActionHandler.swift
robertdalmeida: As we discussed. We will remove the error handling as we will not introduce a cancellationError as it would be a breaking change.

AdyenActions/Components/3DS2/Action handlers/ThreeDS2CoreActionHandler.swift
robertdalmeida: We will not expose the merchant to the cancel error as it would be a breaking change.

AdyenActions/Components/3DS2/Action handlers/ThreeDS2CoreActionHandler.swift
```diff
@@ -8,6 +8,13 @@ import Adyen
 import Adyen3DS2
 import Foundation
 
+// TODO: Remove this once 3ds2 sdk is released.
+extension NSError {
```
goergisn: Will this be removed?

.github/workflows/build.yml
```diff
@@ -25,8 +25,8 @@ jobs:
         - destination: 'name=iPhone 8,OS=13.7'
           scheme: AdyenUIHost
     steps:
-    - uses: actions/checkout@v2
-    - uses: n1hility/cancel-previous-runs@v2
+    - uses: actions/checkout@v3
```
robertdalmeida, line 28: @Adyen/checkout-ios The pipeline fails for v4 branch. Which is why i tried this but this too also fails. Any suggestions here?

AdyenActions/Components/3DS2/Action handlers/ThreeDS2CoreActionHandler.swift
robertdalmeida: Yes, indeed will remove it once i create the 3ds2sdk release today/tomorrow. Thanks. Will close this comment once i do that.

robertdalmeida: Released sdk version 2.3.2. So removed this code.

erenbesel (approved):

goergisn (approved):

neelSharma12 (approved):

PR #1272, opened 2023-07-10
Added settings for card component

neelSharma12:
### Chore
<chore>

- Refactored UI of existing settings screen of the demo app.
-  Added card component settings.

</chore>

Simulator Screen Shot - iPhone 14 - 2023-07-10 at 09 59 33
Simulator Screen Shot - iPhone 14 - 2023-07-10 at 09 59 48

### DIFF
--- a/AdyenCard/Components/Card/CardComponentConfiguration.swift
+++ b/AdyenCard/Components/Card/CardComponentConfiguration.swift
@@ -98,9 +98,6 @@ extension CardComponent {
         /// Do not display address form
         case none
 
-        public var displayName: String {
-            return self.rawValue.capitalized
-        }
     }
 
     /// The mode of input field on Component UI
@@ -115,9 +112,6 @@ extension CardComponent {
         /// Show the field when a specific condition is met.
         case auto
 
-        public var displayName: String {
-            return self.rawValue.capitalized
-        }
     }
 
     /// Card component configuration.
--- a/AdyenCard/Components/Card/CardComponentConfiguration.swift
+++ b/AdyenCard/Components/Card/CardComponentConfiguration.swift
@@ -90,26 +90,34 @@ extension CardComponent {
     public enum AddressFormType: String, Codable, CaseIterable {
 
         /// Display full address form
-        case full = "Full"
+        case full
 
         /// Display simple form with only zip code field
-        case postalCode = "PostalCode"
+        case postalCode
 
         /// Do not display address form
-        case none = "None"
+        case none
+
+        public var displayName: String {
+            return self.rawValue.capitalized
+        }
     }
 
     /// The mode of input field on Component UI
     public enum FieldVisibility: String, Codable, CaseIterable {
 
         /// Always show the field.
-        case show = "Show"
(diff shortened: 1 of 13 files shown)


AdyenCard/Components/Card/CardComponentConfiguration.swift
```diff
@@ -87,29 +87,29 @@ public protocol AnyCardComponentConfiguration {
 extension CardComponent {
 
     /// The mode of address form of card component
-    public enum AddressFormType {
+    public enum AddressFormType: String, Codable, CaseIterable {
 
         /// Display full address form
-        case full
```
descorp, line 93: Not sure if assigning English strings to cases is a good idea 🤔
It is not expected that merchants will ever use them, however we also not providing localisation to it.
It will become part of public API that we need to maintain..

Maybe we can just keep `case full` and then capitalize text on DemoApp side?
```swift
ForEach(CardComponent.AddressFormType.allCases, id: \.self) {
        Text($0.rawValuecapitalized)
}
```

---

I appreciate that making enums compatible to `String` - it makes parsing them on React Native easier 😄

Demo/Common/Configuration/ConfigurationView.swift
```diff
@@ -68,12 +70,24 @@ internal struct ConfigurationView: View {
...
     
     private var merchantAccountSection: some View {
-        TextField(ConfigurationSection.merchantAccount.rawValue, text: $viewModel.merchantAccount)
+        HStack {
+            Text("Merchant Account")
+            TextField(ConfigurationSection.merchantAccount.rawValue, text: $viewModel.merchantAccount)
+                .keyboardType(.numberPad)
+                .multilineTextAlignment(.trailing)
+                .padding(.trailing, 10)
+        }
```
descorp: This could be a standalone `View`

```swift
/// Naming is optional :)
struct PropertyView: View {

   init($title, $placeholder, keyboardType = .numberPad) {
        ...
   }

   internal var body: some View {
        HStack {
            Text($title)
            TextField($value, text: $placeholder)
                .keyboardType(keyboardType)
                .multilineTextAlignment(.trailing)
                .padding(.trailing, 10)
        }
   }   
}
```

AdyenCard/Components/Card/CardComponentConfiguration.swift
goergisn: Maybe instead of using the rawValue we could have a `var displayName: String` on the enum that returns the value we want to show in the interface - would make it more readable :)

neelSharma12: d220da06d51a1b66beebb3f903002cb53f110f60

Demo/Common/Configuration/ConfigurationView.swift
neelSharma12: 3c094e2bae9be3adebf2eeac7e655bd8f927408f

descorp (approved):

goergisn (approved):

PR #1273, opened 2023-07-10
chore: Introducing LinkTextView

goergisn:
## Changes

* Adding LinkTextView that allows easily having selectable links inside of text
* Adding support for multiple links in `FormAttributedLabelItem`


descorp (dismissed):

robertdalmeida:
🚀 Nice component LinkTextView. Will use it in the DA branch once this is merged to develop.

robertdalmeida (dismissed):

Adyen/Helpers/StringHelpers.swift
```diff
@@ -159,6 +159,24 @@ public extension AdyenScope where Base == String {
...
         return lowerIndex...upperIndex
     }
+    
+    /// Returns a list of ``NSRange`` indicating links using following regex pattern: `#(.+?)#`
+    var linkRanges: [NSRange] {
+        let pattern = "#(.+?)#"
+        var ranges: [NSRange] = []
+        do {
+            let regex = try NSRegularExpression(pattern: pattern, options: NSRegularExpression.Options.caseInsensitive)
+            let matches = regex.matches(in: base, options: [], range: NSRange(location: 0, length: base.utf16.count))
```
erenbesel: Any reason to check with utf16 and not utf8?

robertdalmeida: I think the same has already been in use here. Which has been moved to a common place. But an interesting question indeed. 🤔

https://github.com/Adyen/adyen-ios/blob/1f91c010570ac8829a2343019a181077a31ae0e2/Adyen/UI/Form/Items/Basic%20Items/FormAttributedLabelItem.swift#L79

neelSharma12: UTF16 is more efficient than UTF8 for non english and as we support more than 30 languages so it can be a better choice I think 🤔 🤔 What is your opinion on this?

neelSharma12 (dismissed):

neelSharma12 (approved):

Adyen/Helpers/StringHelpers.swift
goergisn: Replaced it with `NSRange(base.startIndex..., in: base)` in 42d3b4dd313ba16883ef73b9e3840aab394e18c0

erenbesel (approved):

PR #1274, opened 2023-07-11
chore: Added api version check for  storePaymentMethod switch

neelSharma12:
### Chore
<chore>

- Added api version check for  storePaymentMethod switch. 

</chore>

Simulator Screen Shot - iPhone 14 - 2023-07-17 at 10 53 59


Demo/Common/Networking/SessionRequest.swift
```diff
@@ -46,6 +47,8 @@ internal struct SessionRequest: APIRequest {
         //        try container.encode(installmentOptions, forKey: .installmentOptions)
 
         if ConfigurationConstants.current.cardComponentConfiguration.showsStorePaymentMethodField {
+            AdyenAssertion.assert(message: "API version should be greater than 69 to apply card component's store payment method field",
```
goergisn: 👍  how about additionally adding a small info text to the the toggle in the settings?

goergisn: Maybe it would also be a good idea to set the default api version to 70 so it supports all the features of the demo app :)

neelSharma12: Yes, I am going to add text next to the switch but I am sure about default version. Let's discuss this first.

neelSharma12: @goergisn I discussed this with Leo and he also said that we should keep the latest version  to make sure that everything will work and Also, because it is important that merchants will use the latest versions of our products. So, I will make v70 default.

Demo/Common/Configuration/CardComponentSettingsView.swift
```diff
@@ -20,8 +20,13 @@ internal struct CardComponentSettingsView: View {
                     Toggle(isOn: $viewModel.showsHolderNameField) {
                         Text("Holder Name")
                     }
+                    
                     Toggle(isOn: $viewModel.showsStorePaymentMethodField) {
-                        Text("Stored Payment Method")
+                        VStack(alignment: .leading) {
+                            Text("Stored Payment Method")
+                            Text("(v70 and above)")
```
goergisn: Maybe tiny improvement to give it more context
```suggestion
                            Text("Requires API version 70 or higher")
```

goergisn (dismissed):

Demo/Common/Networking/SessionRequest.swift
```diff
@@ -46,6 +47,8 @@ internal struct SessionRequest: APIRequest {
         //        try container.encode(installmentOptions, forKey: .installmentOptions)
 
         if ConfigurationConstants.current.cardComponentConfiguration.showsStorePaymentMethodField {
+            AdyenAssertion.assert(message: "API version should be v70 or above to apply card component's store payment method field",
```
descorp, line 50: Maybe just `if` and popup? 🤔

neelSharma12: We discussed this last week and this was the conclusion but we can discuss more if you want 😄

Demo/Common/Configuration/CardComponentSettingsView.swift
neelSharma12: f4620ff8087f903eef8bdf0735b0f91493a2009b

Demo/Common/Networking/SessionRequest.swift
goergisn: I like that the assert gives you more context about what/where the issue is here by "teleporting" you into the right part of the code which encourages exploration of the sample code.

goergisn (approved):

Demo/Common/Networking/SessionRequest.swift
neelSharma12: We can discuss this further and change if needed in another PR.

PR #1281, opened 2023-07-21
Resolve reference loop

descorp:
# Open PR

## Changes

* resolve reference loop

### DIFF
--- a/Demo/UIKit/ComponentsViewController.swift
+++ b/Demo/UIKit/ComponentsViewController.swift
@@ -42,9 +42,9 @@ internal final class ComponentsViewController: UIViewController {
     }()
 
     private lazy var applePayComponentExample: ApplePayComponentExample = {
-        let applePayComponentExample = ApplePayComponentExample()
-        applePayComponentExample.presenter = self
-        return applePayComponentExample
+        let applePayComponent = ApplePayComponentExample()
+        applePayComponent.presenter = self
+        return applePayComponent
     }()
 
     // MARK: - View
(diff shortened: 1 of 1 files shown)


Demo/UIKit/ComponentsViewController.swift
```diff
@@ -42,7 +42,7 @@ internal final class ComponentsViewController: UIViewController {
     }()
 
     private lazy var applePayComponentExample: ApplePayComponentExample = {
-        let applePayComponent = ApplePayComponentExample()
+        let applePayComponentExample = ApplePayComponentExample()
```
goergisn: Uhhh good one!

Maybe better rename to something different than the name of the computed property as it could easily cause issues again in the future.

neelSharma12: 😱 how did this happen? I think it was me who did this 😨 😱

neelSharma12 (dismissed):

Demo/UIKit/ComponentsViewController.swift
descorp: We have missed it on PR
Honest mistake, It happens 😅

descorp: @goergisn 

Fixed [d4b05f0](https://github.com/Adyen/adyen-ios/pull/1281/commits/d4b05f0950607bdb5f7ba29dbf31252a7e933e4b)

erenbesel (approved):

PR #1282, opened 2023-07-21
chore: Fixing flaky tests

goergisn:
## Changes

* Moving some file global objects to a function scope to prevent unwanted sideeffects

### DIFF
--- a/Tests/Components Tests/Boleto/BoletoComponentTests.swift
+++ b/Tests/Components Tests/Boleto/BoletoComponentTests.swift
@@ -161,25 +161,22 @@ class BoletoComponentTests: XCTestCase {
         
         wait(for: .milliseconds(300))
         
-        let emailSwitchItem: FormToggleItemView? = viewController.view.findView(by: "sendCopyToEmailItem") as? FormToggleItemView
-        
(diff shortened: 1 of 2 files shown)


Tests/Components Tests/Boleto/BoletoComponentTests.swift
```diff
@@ -124,18 +120,23 @@ class BoletoComponentTests: XCTestCase {
     }
     
     func testNoEmailSection() {
-        sut = BoletoComponent(paymentMethod: method,
-                              context: context,
-                              configuration: Dummy.getConfiguration(showEmailAddress: false))
         
-        let sutVC = sut.viewController
+        let context = Dummy.context(with: AnalyticsProviderMock())
```
descorp, line 124: so the only meaningful change in `BoletoComponentTests.swift` is local `context` instead of `setUp` and `tearDown`?

goergisn: That's correct

erenbesel (approved):

neelSharma12 (approved):

PR #1287, opened 2023-07-26
chore: Adding amount field to telemetry request

goergisn:
## Summary
- Adding amount value to telemetry requests


Adyen/Analytics/AnalyticsProvider/TelemetryTracker.swift
```diff
@@ -30,24 +30,28 @@ public enum TelemetryFlavor {
 
 @_spi(AdyenInternal)
 public protocol TelemetryTrackerProtocol {
-    func sendTelemetryEvent(flavor: TelemetryFlavor)
+    func sendTelemetryEvent(flavor: TelemetryFlavor, amount: Amount?)
```
descorp: how about simply use `context` on init ?
It is bit odd - only `AdyenContext` uses `AnalyticsProviderProtocol`  but it need `amount` from `AdyenContext`..

Adyen/Core/Core Protocols/PresentableComponent.swift
```diff
@@ -69,8 +69,16 @@ public protocol TrackableComponent: Component {
...
-        context.analyticsProvider.sendTelemetryEvent(flavor: flavor)
+        context.analyticsProvider.sendTelemetryEvent(flavor: flavor, amount: amount)
+    }
+}
+
+@_spi(AdyenInternal)
+extension TrackableComponent where Self: PaymentAware & PaymentMethodAware {
+    
+    public func sendTelemetryEvent() {
+        sendTelemetryEvent(withAmount: payment?.amount)
```
erenbesel: If you added the `PaymentAware` check here, then I think you don't need to re-conform to the function on all the below places

Adyen/Analytics/AnalyticsProvider/TelemetryTracker.swift
goergisn: @descorp Did not think about this, thanks 🙏 
I added an `amount` requirement next to `checkoutAttemptId` in the AnalyticsProvider protocol which resulted in way less changes. Let me know what you think.

Adyen/Core/AdyenContext/AdyenContext.swift
```diff
@@ -43,10 +49,18 @@ public final class AdyenContext: PaymentAware {
...
     }
 
     @_spi(AdyenInternal)
     public func update(payment: Payment?) {
         self.payment = payment
     }
 }
+
+private extension AdyenContext {
+    func updateAnalyticsProvider() {
```
erenbesel: Nitpick but maybe a name like `updateAnalyticsAmount` etc?

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
```diff
@@ -27,8 +27,10 @@ public struct AnalyticsConfiguration {
 @_spi(AdyenInternal)
 public protocol AnalyticsProviderProtocol: TelemetryTrackerProtocol {
     
-    var checkoutAttemptId: String? { get }
+    /// The payment-amount to be tracked
+    var amount: Amount? { get set }
```
descorp: do we need "set" ?

goergisn: https://github.com/Adyen/adyen-ios/pull/1287/files#diff-3a0d5cb797232c8b37393f98b8fa3b114fb17f74fa72d82b016231ee5bf39448R64

This line nees it as the `AdyenContext` receives `any AnalyticsProviderProtocol`

descorp: fair, it should not be possible to change amount on Context, but for now it is the only way 🤔

descorp (dismissed):

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
```diff
@@ -24,12 +24,20 @@ public struct AnalyticsConfiguration {
     public init() { /* Empty implementation */ }
 }
 
+@_spi(AdyenInternal)
+/// Additional fields to be provided with a ``TelemetryRequest``
+public struct AdditionalAnalyticsFields {
```
descorp, line 29: Nice :)

erenbesel (dismissed):

erenbesel (dismissed):

erenbesel (approved):

neelSharma12 (approved):

PR #1294, opened 2023-08-01
feature: Added new initializer for Apple Pay configuration

erenbesel:
<new>

## New
- For Apple Pay, `ApplePayConfiguration` has a new initializer that supports the capabilities of [`PKPaymentRequest`](https://developer.apple.com/documentation/passkit/pkpaymentrequest). Pass `PKPaymentRequest` when initializing Apple Pay. For example:

~~~~
public init(paymentRequest: PKPaymentRequest,  allowOnboarding: Bool = false)
~~~~

</new>


AdyenComponents/Apple Pay/ApplePayComponent.swift
```diff
@@ -61,7 +61,8 @@ public class ApplePayComponent: NSObject, PresentableComponent, PaymentComponent
             throw Error.userCannotMakePayment
         }
 
-        self.paymentRequest = configuration.createPaymentRequest(supportedNetworks: supportedNetworks)
+        var configuration = configuration
```
descorp, line 64: Why do you need `configuration` as variable?

erenbesel: `createPaymentRequest` func is now `mutating`

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
```diff
@@ -78,13 +82,35 @@ extension ApplePayComponent {
...
+        ///   - allowOnboarding: Flag to allow shoppers to add new cards for Apple Pay  if there is none. Default is `false`.
+        /// - Important: When created via this init, rest of the properties of this configuration
+        /// are ignored since the request should already be populated.
+        ///  - Use this initializer to support any new additions to the `PKPaymentRequest` object,
+        ///  such as recurring payments via its `recurringPaymentRequest` property.
+        /// - Throws: An error object describing whether any required field in the request is missing.
+        public init(paymentRequest: PKPaymentRequest,
+                    allowOnboarding: Bool = false) throws {
+            guard !paymentRequest.merchantIdentifier.isEmpty else {
+                throw UnknownError(errorDescription: "Merchant identifier cannot be empty")
```
descorp: Maybe make a proper Error out of it?

AdyenComponents/Apple Pay/ApplePayPayment.swift
```diff
@@ -16,7 +16,7 @@ public struct ApplePayPayment {
     /// The amount for this payment in minor units.
     public private(set) var amountMinorUnits: Int
 
-    /// The public key used for encrypting card details.
```
descorp, line 19: 😬

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
erenbesel: We will need to add it into the Apple Pay error enum which can break merchant code. So for now I opted to go this way

descorp: Is adding a new enum case considered a breaking change? 🤔
This is error `enum`, so I would not expect a major impact, as long as we document it in **Release notes**

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
```diff
@@ -100,7 +126,15 @@ extension ApplePayComponent {
                 paymentRequest.supportsCouponCode = supportsCouponCode
                 paymentRequest.shippingContactEditingMode = allowShippingContactEditing ? .enabled : .storePickup
             }
+            
+            return paymentRequest
+        }
 
+        internal mutating func paymentRequest(with supportedNetworks: [PKPaymentNetwork]) -> PKPaymentRequest {
```
goergisn, line 133: The wording of the function `paymentRequest` implies a simple returning of a payment request with no side effects. What about `createPaymentRequest(with:)` instead? Sounds more substantial.

```suggestion
        internal mutating func createPaymentRequest(with supportedNetworks: [PKPaymentNetwork]) -> PKPaymentRequest {
```

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
```diff
@@ -68,6 +68,10 @@ extension ApplePayComponent {
 
         /// A funding source supported by the merchant. If `nil`, the transaction allows both credit and debit cards.
         public var merchantCapability: CardFundingSource?
+        
+        /// The payment request object needed for Apple Pay. Must contain all the required fileds
+        /// such as `merchantIdentifier`, `summaryItems`, `currencyCode`, and `countryCode`.
+        public private(set) var paymentRequest: PKPaymentRequest?
```
goergisn: Does this only have to be public because of checking in tests?
```suggestion
        internal private(set) var paymentRequest: PKPaymentRequest?
```

descorp (dismissed):

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
erenbesel: Fixed in cc142c422894338fb2a6e72c2cce560a45da754f

goergisn (approved):

descorp (approved):

PR #1304, opened 2023-08-09
new: Allow co-branded bcmc

goergisn:
## Summary

<new>

* Bancontact cards are now supported in the `CardComponent`
* Dual-branded Bancontact cards now allow brand selection in the `BCMCComponent`

</new>

- Adding type to BinLookupRequest for correct card brand sorting
- Using CardPaymentMethod's `brands` to configure the `BCMCComponent`


descorp (dismissed):

AdyenCard/Form/FormCardSecurityCodeItem.swift
```diff
@@ -9,19 +9,41 @@
...
+            switch self {
+            case .required, .optional: return true
+            case .hidden: return false
+            }
+        }
+        
+        internal var isOptional: Bool {
+            switch self {
+            case .optional: return true
+            case .hidden, .required: return false
```
erenbesel: Isn't it considered optional if it's hidden as well?

AdyenCard/Form/FormCardNumberContainerItem.swift
```diff
@@ -55,6 +67,12 @@ internal final class FormCardNumberContainerItem: FormItem, AdyenObserver {
     
     internal func update(brands: [CardBrand]) {
         numberItem.update(brands: brands)
+        
+        guard showsSupportedCardLogos else {
```
erenbesel: exiting here seems dangerous if we want to add some other logic below that could be for both cases. Maybe if else block for this and the below related logic?

AdyenCard/Utilities/BinLookup/BinLookupService.swift
```diff
@@ -27,12 +27,19 @@ internal final class BinLookupService: AnyBinLookupService {
...
         self.publicKey = publicKey
         self.apiClient = apiClient
+        self.binLookupType = binLookupType
     }
     
-    internal func requestCardType(for bin: String, supportedCardTypes: [CardType], caller: @escaping CompletionHandler) {
+    internal func requestCardType(
+        for bin: String,
+        supportedCardTypes: [CardType],
+        caller: @escaping CompletionHandler
```
erenbesel: Not sure why it was called `caller` before but can we rename it to `completion` :D

AdyenCard/Form/FormCardSecurityCodeItem.swift
goergisn: When re-reading it, it is actually confusing - it was meant to check for the actual enum value `optional` hence the naming. I refactored it to us a switch in place as it's only used once with more supporting context 👌

erenbesel (dismissed):

erenbesel (approved):

neelSharma12 (approved):

descorp (approved):

PR #1384, opened 2023-09-21
improvement: Payment List Item Accessibility

goergisn:
## Summary
- Adding `accessibilityLabel`s to `ListItem`

This improves accessibility in following scenarios:
  - Preselected Payment Method
  - Stored Payment Method
  - Other Payment Methods

### DIFF
--- a/Adyen/Core/Models/DisplayInformation.swift
+++ b/Adyen/Core/Models/DisplayInformation.swift
@@ -31,6 +31,7 @@ public struct DisplayInformation: Equatable {
     @_spi(AdyenInternal)
     public let footnoteText: String?
     
+    /// An optional custom `accessibilityLabel` to use.
     @_spi(AdyenInternal)
     public let accessibilityLabel: String?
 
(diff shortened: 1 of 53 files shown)


Adyen/UI/List/ListItem.swift
```diff
@@ -46,14 +49,16 @@ public class ListItem: FormItem {
...
     public init(
         title: String,
         subtitle: String? = nil,
         icon: Icon? = nil,
         trailingText: String? = nil,
         style: ListItemStyle = ListItemStyle(),
         identifier: String? = nil,
-        selectionHandler: (() -> Void)? = nil
+        selectionHandler: (() -> Void)? = nil,
+        accessibilityLabel: String? = nil
```
erenbesel: isn't it better to keep the closure at the end as trailing?

goergisn: yes, good point! will change 👌

nauaros (dismissed):

Adyen/Core/Models/DisplayInformation.swift
```diff
@@ -38,16 +42,19 @@ public struct DisplayInformation: Equatable {
     /// - Parameter logoName: The logo name.
     /// - Parameter disclosureText: The trailing disclosure text.
     /// - Parameter footnoteText: The footnote text if any.
+    /// - Parameter accessibilityLabel: An optional custom `accessibilityLabel` to use. Set this if the title / subtitle might not be sufficient enough to provide a good accessibility
```
descorp: Better split this in two lines

descorp (approved):

PR #1397, opened 2023-09-27
improvement: Issuer list component example

goergisn:
## Summary
- Adds example implementations for IssuerList for session & advanced flow


erenbesel (dismissed):

Demo/Common/Networking/PaymentsRequest.swift
```diff
@@ -145,6 +145,14 @@ internal struct PaymentsResponse: Response {
         case amount
     }
     
+    var isSuccessful: Bool {
```
descorp:
```suggestion
    var isAccepted: Bool {
```

Since, technically, `pending` is not yet a "success"
😁

erenbesel (approved):

PR #1406, opened 2023-10-06
chore: ListItem triggered loading state

goergisn:
## Summary
- Adds a `loadingHandler` to the ListItem that allows triggering `startLoading` without a reference to the `ListViewController`


Adyen/UI/List/ListItem.swift
```diff
@@ -73,6 +76,27 @@ public class ListItem: FormItem {
     public func build(with builder: FormItemViewBuilder) -> AnyFormItemView {
         builder.build(with: self)
     }
+    
+    /// Indicates to the ``ListViewController`` that the specific item is currently loading
```
erenbesel: Descriptions of start and stop loading don't seem to match and is a bit confusing.
Also, couldn't see any calls to `stopLoading()`

Adyen/UI/List/ListItem.swift
```diff
@@ -39,6 +39,9 @@ public class ListItem: FormItem {
     /// The `accessibilityLabel` to be used on the `ListItem` or ``ListCell``
     public let accessibilityLabel: String
     
+    /// The handler to invoke when an item changes its loading state
+    internal var loadingHandler: ((Bool, ListItem) -> Void)?
```
erenbesel, line 45: Is it still not the controller updating item's state? Description is a bit unclear

Adyen/UI/List/ListItem.swift
goergisn: Will update the documentation 👌 

Currently `stopLoading()` is not called - it will be from the AddressLookup PR

Adyen/UI/List/ListItem.swift
goergisn: Fair 🙈 Please see 15de5dec6d2b3e2fc88ddb22613159123451a171 if it's more clear now :)

Adyen/UI/List/ListItem.swift
goergisn: Updated the documentation, see: 15de5dec6d2b3e2fc88ddb22613159123451a171

erenbesel (approved):

PR #1410, opened 2023-10-10
chore: Adding tests for didCancel [INJECTED]

goergisn:


AdyenActions/Components/Redirect/OpenExternalAppDetector.swift
```diff
@@ -0,0 +1,50 @@
...
+// Copyright (c) 2023 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+@_spi(AdyenInternal) import Adyen
+import UIKit
+
+/// Detects if an external app was opened
+public protocol OpenExternalAppDetector {
```
erenbesel: Looks great! I think this could also be `@spi`

goergisn: Actually can be internal 👌

erenbesel (approved):

PR #1413, opened 2023-10-16
`improvement` Enable library evolution

descorp:
# Draft PR

## Changes

<improvement>

- provide default implementations for `@spi` protocols: `AdyenContextAware`, `PaymentMethod`, `PaymentMethodDetails`
- pump iOS version to 13
- enabling BUILD_LIBRARY_FOR_DISTRIBUTION

</improvement>


Adyen/Core/Core Protocols/Details.swift
```diff
@@ -16,5 +16,25 @@ public protocol PaymentMethodDetails: Details {
     var checkoutAttemptId: String? { get set }
 }
 
+public extension PaymentMethodDetails {
+
+    @_spi(AdyenInternal)
+    var checkoutAttemptId: String? {
```
goergisn, line 22: Is there a way to check if the library is built with `BUILD_LIBRARY_FOR_DISTRIBUTION` and only conditionally add the default conformance to make the compiler happy?
This way we would not need to provide this implementation and use objc_getAssociatedObject.

Adyen/Core/Core Protocols/PaymentMethod.swift
```diff
@@ -30,6 +30,14 @@ public protocol PaymentMethod: Decodable {
     func buildComponent(using builder: PaymentComponentBuilder) -> PaymentComponent?
 }
 
+public extension PaymentMethod {
+
+    @_spi(AdyenInternal)
+    func buildComponent(using builder: PaymentComponentBuilder) -> PaymentComponent? {
+        nil
```
goergisn, line 37: Maybe a good place to use AssertionFailure? And (if possible) conditionally add this default conformance :)

descorp: How about just using default implementation?
[edfa105](https://github.com/Adyen/adyen-ios/pull/1413/commits/edfa1058ef897ad6899bbe6e10f6f1671c98f603)

goergisn:
Closing in favor of https://github.com/Adyen/adyen-ios/pull/1482

PR #1431, opened 2023-11-01
fix: Prefil phone number doesn't change countryCode

neelSharma12:
# Open PR

## Changes

<fixed>

* Fixed a problem where, when the phoneNumber in shopperInformation doesn't change countryCode.

</fixed>


Adyen/Model/PrefilledShopperInformation.swift
```diff
@@ -6,6 +6,20 @@
 
 import Foundation
 
+public struct PhoneNumber {
```
descorp, line 9: Maybe add some class description

Adyen/Model/PrefilledShopperInformation.swift
```diff
@@ -41,14 +65,14 @@ public struct PrefilledShopperInformation: ShopperInformation {
     ///   - card: Shopper's card basic information, optional.
     public init(shopperName: ShopperName? = nil,
```
descorp, line 103: Maybe we need another `init` with `phoneNumber`. And old `init` we can deprecate

Adyen/UI/Form/Items/Phone Number/FormPhoneNumberItem.swift
```diff
@@ -27,12 +27,17 @@ public final class FormPhoneNumberItem: FormTextItem {
...
     /// - Parameter style: The `FormTextItemStyle` UI style.
     /// - Parameter localizationParameters: Parameters for custom localization, leave it nil to use the default parameters.
-    public init(selectableValues: [PhoneExtensionPickerItem],
+    public init(phoneNumber: PhoneNumber?,
+                selectableValues: [PhoneExtensionPickerItem],
                 style: FormTextItemStyle,
                 localizationParameters: LocalizationParameters? = nil) {
-        phonePrefixItem = FormPhoneExtensionPickerItem(selectableValues: selectableValues, style: style)
+        // swiftlint:disable:next line_length
+        let preselectedValue = selectableValues.first(where: { $0.element.value == phoneNumber?.phoneExtension }) ?? selectableValues.first(where: { $0.identifier == Locale.current.regionCode }) ?? selectableValues[0]
```
descorp: Minor: Maybe convert this line into readable private func?

Adyen/Model/PrefilledShopperInformation.swift
```diff
@@ -6,6 +6,20 @@
 
 import Foundation
 
+public struct PhoneNumber {
+
+    /// The phone number.
+    public let value: String
+
+    /// The phone extension.
```
descorp:
```suggestion
    /// The phone extension in E.123 international notation (ex. +22).
```

Adyen/Model/PrefilledShopperInformation.swift
descorp, line 10:
```suggestion
/// The phone number in E.123 international telephone number notation.
public struct PhoneNumber {
```

Adyen/Model/PrefilledShopperInformation.swift
goergisn: According to E.123 this would be called the `country code` as far as I understand. So maybe good to also rename it from `phoneExtension` to `countryCode`

descorp: I am concerned if this could be interpreted as ISO 3166-1 "country code" (ex. "NL", "US"), since in other places we  request it..

descorp: We can use ["Country calling codes"](https://en.wikipedia.org/wiki/List_of_country_calling_codes) or ["telephone prefixes"](https://en.wikipedia.org/wiki/List_of_mobile_telephone_prefixes_by_country)

goergisn: Ok I see we already have a `PhoneExtension` that has a `countryCode` so indeed it could be misleading. Maybe `callingCode` would be enough as `internationalCallingCode` might be a bit too long 🙃 
https://en.wikipedia.org/wiki/List_of_mobile_telephone_prefixes_by_country

But yeah after all this is really minor - also fine with having it consistent as phone extension even though it's technically not an extension.

Adyen/Model/PrefilledShopperInformation.swift
neelSharma12: You are right. I have fixed this 9d6daef2e7cb0abbbe6701564d59fa85bc35c9dc

Adyen/Model/PrefilledShopperInformation.swift
neelSharma12: I have renamed the "phoneExtension" to "callingCode" which makes sense :) Thanks for the suggestion guys :)  9d6daef2e7cb0abbbe6701564d59fa85bc35c9dc

Adyen/Model/PrefilledShopperInformation.swift
neelSharma12: I have added the description :) 9d6daef2e7cb0abbbe6701564d59fa85bc35c9dc

Adyen/UI/Form/Items/Phone Number/FormPhoneNumberItem.swift
neelSharma12: This code is executed before the class initialization and we cannot call any method before initialising the class.

Adyen/Model/PrefilledShopperInformation.swift
descorp: Still need to update the docs on line 9.

Adyen/Model/PrefilledShopperInformation.swift
```diff
@@ -39,6 +64,17 @@ public struct PrefilledShopperInformation: ShopperInformation {
     ///   - deliveryAddress: The delivery address of the shopper, optional.
     ///   - socialSecurityNumber: The social security number of the shopper, optional.
     ///   - card: Shopper's card basic information, optional.
+    @available(*, deprecated, message:
```
goergisn, line 67: @neelSharma12 would you mind adding the required changes to the `MIGRATION.md`? :)

Adyen/Model/PrefilledShopperInformation.swift
neelSharma12: af7e655da06c724c19c3990ffd44e3c7dc8cf38a

MIGRATION.md
```diff
@@ -1,7 +1,7 @@
 #  Migration Notes
 
 ## Upcoming release
-- `telephoneNumber` property of `PrefilledShopperInformation` has been deprecated. Use to `phoneNumber` property if needed.
```
goergisn: Thinking whether we should move it to the bottom of the list to keep the order consistent or whether we should sort the list the other way around so the newest release is always on the top 🤔 

What do you think? Also maybe good to have this discussion with @descorp  & @erenbesel :)

neelSharma12: 87d6b7317958c614da3a726848fb9dbd25ff4f8c

Adyen/Model/PrefilledShopperInformation.swift
neelSharma12: 87d6b7317958c614da3a726848fb9dbd25ff4f8c

goergisn (approved):

PR #1443, opened 2023-11-07
chore: Reusing encoded header to prevent signature mismatch

goergisn:
## Summary
- reusing `encodedHeader` to prevent accidental (future) issues by re-encoding the header


AdyenEncryption/JOSE/Encryption Algorithms/JSON Web Encryption/JSONWebEncryption.swift
```diff
@@ -60,21 +58,20 @@ internal struct JSONWebEncryption {
     
     internal let compactRepresentation: String
     
-    internal init(header: Header,
+    internal init(encodedHeader: Data,
```
erenbesel, line 61: Isn't this change supposed to be used somewhere and cause build error anyway?

erenbesel (approved):

PR #1456, opened 2023-11-14
chore: Adding optional AdyenDelegatedAuthentication target

mohammedDehairy:
<new>

## New
- The [Authentication SDK](https://github.com/Adyen/adyen-authentication-ios) for [delegated authentication](https://www.adyen.com/knowledge-hub/delegated-authentication) is now included as an optional module with a version control. To enable it, use your dependency manager to include it in your project. 

</new>


.gitignore
```diff
@@ -28,6 +28,7 @@ DerivedData
 
 # Carthage
 Carthage/
+Cartfile.resolved
```
descorp, line 31: It is needed for Carthage distribution to constrain 3DS and other dependencies.

AdyenDelegatedAuthentication/AdyenDelegatedAuthentication.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  File.swift
+//  
+//
+//  Created by Mohamed Eldoheiri on 14/11/2023.
+//
```
descorp, line 6: Header :)

AdyenActions/Components/3DS2/Action handlers/3DS2+Delegated Authentication/ThreeDS2PlusDACoreActionHandler.swift
```diff
@@ -124,7 +124,7 @@
                 completion(.failure(DelegateAuthenticationError.authenticationFailed(cause: nil)))
                 return
             }
-            delegatedAuthenticationService.authenticate(withBase64URLString: delegatedAuthenticationInput) { result in
+            delegatedAuthenticationService.authenticate(withAuthenticationInput: delegatedAuthenticationInput) { result in
```
descorp, line 127: This naming is so Obj-c 😅

descorp (approved):

AdyenDelegatedAuthentication/AdyenDelegatedAuthentication.swift
mohammedDehairy: there're quite a bit of those in the project actually, maybe it should be a separate PR to fix those :-)

PR #1462, opened 2023-11-16
`chore` Upgrade test pipeline

goergisn:
## Summary
- Switching to Xcode 15.1 + iOS 17.2
- Moving iOS 16.4 to compatibility tests

Depending on [reliable availability of Xcode 15.1](https://github.com/actions/runner-images/issues/9069)


.github/workflows/pr_scan.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     - name: Select latest Xcode
       uses: maxim-lobanov/setup-xcode@v1
       with:
-        xcode-version: '14.3'
+        xcode-version: '15.0.1'
```
descorp: You can do [`latest`](https://github.com/maxim-lobanov/setup-xcode)

.github/workflows/pr_scan.yml
```diff
@@ -46,7 +46,7 @@ jobs:
         project: 'Adyen.xcodeproj'
         params: '-derivedDataPath ./DerivedData -enableCodeCoverage YES -skipPackagePluginValidation'
         scheme: 'AdyenUIHost'
-        destination: 'name=iPhone 14,OS=16.4'
+        destination: 'name=iPhone 14,OS=17.0.1'
```
descorp: GitHub's Xcode 15.0.1 [supposed to](https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md#installed-sdks) come with iOS 17.0.
So we might need to have latest simulator installed with 
```
xcversion simulators --install="iOS ${version}"
```

.github/workflows/build.yml
```diff
@@ -8,30 +8,38 @@ on:
     
 jobs:
   tests:
-    runs-on: macos-12
+    runs-on: ${{ matrix.os }}
```
descorp, line 11: Nice :)

.github/workflows/pr_scan.yml
goergisn: https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md

goergisn: The logs stated that `17.0.1` is available so I just chose the latest

descorp: if it works it works 😆

My main point is - there are two approaches:
1) rely on GitHub - use `latest` Xcode and then not specify iOS version - and hope that this will eventually cover all "modern" setups
2) take control in our hands - specify both Xcode and iOS version - will require manual updates of workflow every now and then

.github/workflows/build.yml
neelSharma12: is this for parallel execution?

.github/workflows/pr_scan.yml
goergisn: Yes I agree with using the latest version - Regarding the iOS version it's a decision between either updating the version manually regularly or updating snapshots regularly 😅

.github/workflows/build.yml
descorp: [Correct](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) :)

descorp (dismissed):

.github/workflows/test_cocoapods_integration.yml
```diff
@@ -6,7 +6,7 @@ on:
 
   push:
     branches:
-    - master
+    - main
```
descorp: technically, we can remove mentioning of both `main` and `master` here - this branch is obsolete and only exist since it contains tags for versions v1-v3

descorp (dismissed):

.github/workflows/build.yml
```diff
@@ -8,30 +8,38 @@ on:
...
     continue-on-error: false
     strategy:
       fail-fast: false
       matrix:
         include:
         
+        - version: '16.4'
+          displayname: 'iPhone 14'
+          os: 'macos-13-xl'
```
descorp, line 20:
```suggestion
          os: 'macos-13-xlarge'
```

BTW, you can do 
```yaml
    runs-on: macos-${{ matrix.os }}-xlarge
    matrix:
        include:
            - os: 13
            - os: 12
```

.github/workflows/build.yml
```diff
@@ -56,9 +65,10 @@ jobs:
...
     - name: Build and test
       run: |
-        xcodebuild build test -project ${project} -scheme "${scheme}" ${params} -destination ${destination} | xcpretty --utf --color && exit ${PIPESTATUS[0]}
+        xcodebuild build test -project ${project} -scheme "${scheme}" ${params} -destination "${destination}" | xcpretty --utf --color && exit ${PIPESTATUS[0]}
       env:
         project: 'Adyen.xcodeproj'
         params: "-skipPackagePluginValidation"
-        scheme: ${{ matrix.scheme }}
+        scheme: 'AdyenUIKitTests'
         destination: "OS=${{ matrix.version }},name=${{ matrix.displayname }}"
```
descorp, line 73: For `device: 'iPhone 13'` and  `displayname: 'iPhone-13` we can use:

```yaml
    matrix:
        include:
            - iphone: 13
            - iphone: 12
            
          env:
                displayname: iPhone-${{ matrix.displayname }}
                device: 'iPhone ${{ matrix.displayname }}'
            
          env:
               destination: "OS=${{ matrix.version }},name= iPhone -${{ matrix.iphone }}"
```

This will potentially allow us to iPad in future 😝

.github/workflows/pr_scan.yml
```diff
@@ -70,6 +71,7 @@ jobs:
 
     - name: SwiftLint
       run: |
+        brew install swiftlint
```
descorp, line 74: "macos-13" runner [supposed to have](https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md#linters) swiftlint

goergisn: It does not tho :/ I added that line because it asked me to :P

goergisn: Apparently the arm version of the runner doesn't have the linter installed 🤷  https://github.com/actions/runner-images/blob/main/images/macos/macos-13-arm64-Readme.md

.github/workflows/build.yml
goergisn: Interesting one - would create a PR that does some more fixes to the `build.yml` and test it in isolation 👌

.github/workflows/build.yml
goergisn: Interesting one - would create a PR that does some more fixes to the `build.yml` and test it in isolation 👌

Tests/Components Tests/ACH Direct Debit/ACHDirectDebitComponentTests.swift
```diff
@@ -285,7 +285,7 @@ class ACHDirectDebitComponentTests: XCTestCase {
         payButtonItemViewButton.didSelectSubmitButton()
         wait(until: payButtonItemViewButton, at: \.item.showsActivityIndicator, is: true)
         
-        wait(for: [expectation], timeout: 30)
+        wait(for: [expectation], timeout: 100)
```
erenbesel, line 288: this one isn't  10, typo?

Tests/Helpers/XCTestCase+RootViewController.swift
```diff
@@ -51,7 +51,8 @@ extension XCTestCase {
             completion?()
             presentationExpectation.fulfill()
         }
-        wait(for: [presentationExpectation], timeout: 1)
+
+        wait(for: [presentationExpectation], timeout: 100)
```
erenbesel, line 55: here as well

Tests/Components Tests/ACH Direct Debit/ACHDirectDebitComponentTests.swift
goergisn: Mhh I can actually try to decrease it again - but on slower machines it actually took around 60seconds sometimes so I set it to 100 - it does not slow it down so we might as well keep it :)

erenbesel (approved):

nauaros (approved):

PR #1463, opened 2023-11-16
`chore` Fix all warnings

goergisn:
## Summary
- Increase `file_limit` to 500
- Move some code to new files / functions / extensions
- Resolved ACL warnings
- Shortened names

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -19,3 +19,6 @@ line_length:
   ignores_function_declarations: true
   ignores_urls: true
   warning: 140
+
+file_length:
+  warning: 500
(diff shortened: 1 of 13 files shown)


.swiftlint.yml
```diff
@@ -19,3 +19,6 @@ line_length:
   ignores_function_declarations: true
   ignores_urls: true
   warning: 140
+
+file_length:
+  warning: 500
```
descorp, line 24: 😆

Adyen/Core/Payment Methods/Abstract/PaymentMethodType.swift
```diff
@@ -180,6 +180,9 @@ public enum PaymentMethodType: RawRepresentable, Hashable, Codable {
         case let .other(value): return value
         }
     }
+}
+
+extension PaymentMethodType {
```
descorp, line 185: What does introduction of extension change? 🤔

descorp (approved):

Adyen/Core/Payment Methods/Abstract/PaymentMethodType.swift
goergisn: Body length

goergisn:
Replacing with https://github.com/Adyen/adyen-ios/pull/1468 which is targeting `develop`

PR #1464, opened 2023-11-16
`chore` Removed useless #available checks

goergisn:
## Summary
- Removed all #available(iOS 12.0, *)  and lower

### DIFF
--- a/AdyenDropIn/Views/PreApplePayView.swift
+++ b/AdyenDropIn/Views/PreApplePayView.swift
@@ -26,15 +26,16 @@ internal final class PreApplePayView: UIView, Localizable {
     
     /// Creates PKPaymentButtonStyle based on Dark or Light Mode.
     private var paymentButtonStyleAuto: PKPaymentButtonStyle {
-        let buttonStyle: PKPaymentButtonStyle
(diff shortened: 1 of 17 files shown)


AdyenDropIn/Views/PreApplePayView.swift
```diff
@@ -26,15 +26,16 @@ internal final class PreApplePayView: UIView, Localizable {
...
     private var paymentButtonStyleAuto: PKPaymentButtonStyle {
-        let buttonStyle: PKPaymentButtonStyle
         if #available(iOS 14.0, *) {
-            buttonStyle = .automatic
-        } else if #available(iOS 12.0, *), traitCollection.userInterfaceStyle == .dark {
-            buttonStyle = .white
-        } else {
-            buttonStyle = .black
+            return .automatic
+        }
```
descorp, line 31: we still have iOS13 right ?
What about `traitCollection.userInterfaceStyle == .dark` ?

goergisn: That's handled by the lines below 🤔

goergisn:
Duplicated by https://github.com/Adyen/adyen-ios/pull/1469

PR #1465, opened 2023-11-17
Fix: refresh dropin pm list after applying gift cards

erenbesel:
<fixed>

## Fixed
- Drop-in now updates the payment method list after the shopper applies a gift card, in case the available payment methods change because of the updated payment amount.

</fixed>


Demo/Common/IntegrationExamples/DropIn/DropInAdvancedFlowExample.swift
```diff
@@ -80,6 +80,7 @@ internal final class DropInAdvancedFlowExample: InitialDataAdvancedFlowProtocol
             } else if let order = response.order,
                       let remainingAmount = order.remainingAmount,
                       remainingAmount.value > 0 {
+                ConfigurationConstants.current.value = remainingAmount.value
```
goergisn: This will update the values in the UserDefaults right?

erenbesel: Good catch, I will do this another way.

Demo/Configuration.swift
```diff
@@ -131,7 +131,7 @@ internal struct DemoAppSettings: Codable {
     private static let defaultsKey = "ConfigurationKey"
     
     internal var countryCode: String
-    internal let value: Int
```
goergisn: This can be reverted right?

descorp (approved):

Demo/Configuration.swift
erenbesel: Yes! done

goergisn (approved):

PR #1472, opened 2023-11-28
`chore` Adding internal telemetry context

goergisn:
## Summary
- Adds internal `TelemetryContext` that specifies the version + platform

### DIFF
--- a/Adyen/Core/AdyenContext/AdyenContext.swift
+++ b/Adyen/Core/AdyenContext/AdyenContext.swift
@@ -19,10 +19,7 @@ public final class AdyenContext: PaymentAware {
     public private(set) var payment: Payment?
 
     @_spi(AdyenInternal)
-    public let analyticsProvider: AnalyticsProviderProtocol
-
-    @_spi(AdyenInternal)
-    public var telemetryContext: TelemetryContext = .init()
+    public var analyticsProvider: AnalyticsProviderProtocol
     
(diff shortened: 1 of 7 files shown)


Adyen/Core/AdyenContext/AdyenContext.swift
```diff
@@ -19,16 +19,20 @@ public final class AdyenContext: PaymentAware {
     public private(set) var payment: Payment?
 
     @_spi(AdyenInternal)
-    public let analyticsProvider: AnalyticsProviderProtocol
-
+    public var analyticsProvider: AnalyticsProviderProtocol
```
descorp, line 22: Is this a good idea?
Maybe better to modify it via `AnalyticsConfiguration` ?

goergisn: You mean moving the `TelemetryContext` to `AnalyticsConfiguration`? 
Could also be an option. No strong feelings on where it fits best 🤔 

What do you think @erenbesel

goergisn: Please check https://github.com/Adyen/adyen-ios/pull/1473 for an alternative implementation

descorp: I mean making `analyticsProvider` **var**

PR #1485, opened 2023-12-06
`release` 5.5.0

goergisn:
## New

- When you offer installment payments, you can now show the number of installments and the amount of each installment in the picker. For example, **3x $20**.
  - For Sessions flow, when you make a `/sessions` request, include [`showInstallmentAmount`](https://docs.adyen.com/api-explorer/Checkout/latest/post/sessions#request-showInstallmentAmount):
**true** and [`installmentOptions`](https://docs.adyen.com/api-explorer/Checkout/latest/post/sessions#request-installmentOptions). 
We use these values to show installment options in the picker.
  - For Advanced flow, in the `InstallmentConfiguration` object, set `showInstallmentAmount` to **true** when you initialize it.

- Support for [library evolution](https://www.swift.org/blog/library-evolution/)

## Changed
      
For cards, the address lookup feature now supports a 2-step lookup process (for example, used by the Google Maps API):
- First, it allows providing suggested addresses (that might be incomplete)
- Then, optionally, it allows making another API request to complete the suggestions.

You must change the callback as follows:

| 5.4.1 or earlier        | 5.5.0                          |
|-------------------------|--------------------------------|
| .lookup(LookupProvider) | .lookup(AddressLookupProvider) |

## Fixed
      
- When the shopper's phone number is included in the `shopperInformation` object, the pre-filled phone number prefix on the payment form automatically updates to match the correct country code.

- Drop-in now updates the payment method list after the shopper applies a gift card, in case the available payment methods change because of the updated payment amount.

- When you define custom localizations in your app bundle using Localizable.strings, they are now applied correctly.

- The Bizum payment method is no longer shown as an available payment method.


Adyen.xcodeproj/project.pbxproj
```diff
@@ -8593,7 +8593,7 @@
 				BUNDLE_LOADER = "$(TEST_HOST)";
 				CLANG_CXX_LANGUAGE_STANDARD = "gnu++20";
 				CODE_SIGN_STYLE = Automatic;
-				CURRENT_PROJECT_VERSION = 11;
+				CURRENT_PROJECT_VERSION = 2;
```
descorp: '2' not '12' ?

AdyenDelegatedAuthentication/AdyenDelegatedAuthentication.swift
```diff
@@ -1,6 +1,6 @@
 //
-//  File.swift
-//  
+//  AdyenDelegatedAuthentication.swift
+//
 //
 //  Created by Mohamed Eldoheiri on 14/11/2023.
 //
```
descorp, line 6: Header should have been updated 🤔

Adyen.xcodeproj/project.pbxproj
goergisn: Yeah not sure what went wrong there 🙃

goergisn: Fixed it

descorp (dismissed):

Adyen.xcodeproj/project.pbxproj
```diff
@@ -8828,6 +8906,21 @@
 /* End XCRemoteSwiftPackageReference section */
 
 /* Begin XCSwiftPackageProductDependency section */
+		81BDD6CD2B21CE6C0022250E /* AdyenAuthentication */ = {
+			isa = XCSwiftPackageProductDependency;
+			package = F94D65E42B0365C30095D61E /* XCRemoteSwiftPackageReference "adyen-authentication-ios" */;
+			productName = AdyenAuthentication;
+		};
+		81BDD6D32B21CF1B0022250E /* AdyenAuthentication */ = {
```
erenbesel, line 8914: Why is there 2 AdyenAuthentication references 🤔

goergisn: Delegated Authentication also imports `AdyenAuthentication`

goergisn: One is for the UIKit target, and the other for the SwiftUI one

erenbesel: Ah got it

descorp (approved):

erenbesel (approved):

PR #1497, opened 2023-12-21
`chore` New analytics initial call

erenbesel:
New analytics replacing the existing checkout attempt Id fetch and initial telemetry with a single call.

The new analytics is on a new webapp so it requires a new `baseURL` which needs a new environment, thus `AnalayticsEnvironment` is added. 

- A new api context with this environment (`analyticsApiContext`) is created internally during the original context's creation.

This new call combines sending the analytics data and returns the checkout attempt id in the response. So this call needs to be done only once per payment flow. So the call is made on component load unless the component is inside dropIn in which case dropIn will have made the call already.


Tests/Adyen Tests/Core/AdyenContextTests.swift
```diff
@@ -21,10 +21,10 @@ class AdyenContextTests: XCTestCase {
             payment: .init(amount: oneEUR, countryCode: "NL")
         )
         
-        XCTAssertEqual(context.analyticsProvider.additionalFields?().amount, oneEUR)
+//        XCTAssertEqual(context.analyticsProvider.additionalFields?().amount, oneEUR)
```
goergisn: 🙈

Tests/Adyen Tests/Core/AdyenContextTests.swift
```diff
@@ -21,10 +21,10 @@ class AdyenContextTests: XCTestCase {
...
         
         context.update(payment: Payment(amount: twoEUR, countryCode: "NL"))
         
-        XCTAssertEqual(context.analyticsProvider.additionalFields?().amount, twoEUR)
+//        XCTAssertEqual(context.analyticsProvider.additionalFields?().amount, twoEUR)
```
goergisn: 🙈

AdyenSession/AdyenSession.swift
```diff
@@ -119,6 +119,8 @@ public final class AdyenSession {
                                            sessionContext: sessionContext)
                 session.delegate = delegate
                 session.presentationDelegate = presentationDelegate
+                // Unfortunately needed by telemetry and this is the only way atm.
```
goergisn: I believe this comment is not needed - also without extra context it's not super straight forward to understand

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
```diff
@@ -66,28 +81,67 @@ internal final class AnalyticsProvider: AnalyticsProviderProtocol {
...
+              let checkoutAttemptId else { return }
+        var request = AdyenAnalyticsRequest(checkoutAttemptId: checkoutAttemptId)
+        
+        request.events = events
+        request.logs = logs
+        request.errors = errors
+        
+        apiClient.perform(request) { [weak self] _ in
+            guard let self else { return }
+            self.clearAll()
```
goergisn: This could lead to losing some events that arrived while the request was going on.

Maybe temporarily hold onto the events (`var eventsBeforeApiCall`) and clear the current ones before making the api call (`clearAll`) and if it fails append the saved ones to the list.

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
```diff
@@ -51,8 +61,13 @@ internal final class AnalyticsProvider: AnalyticsProviderProtocol {
     internal let apiClient: APIClientProtocol
     internal let configuration: AnalyticsConfiguration
     internal private(set) var checkoutAttemptId: String?
-    internal var additionalFields: (() -> AdditionalAnalyticsFields)?
     private let uniqueAssetAPIClient: UniqueAssetAPIClient<CheckoutAttemptIdResponse>
+    
+    private var batchTimer: Timer?
+    
+    private var events: [AdyenAnalytics.Event] = []
```
goergisn: maybe make a struct for events + logs + errors (e.g. `AnalyticsDataPoints`) which makes it easier to handle and reset them

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
```diff
@@ -66,28 +81,67 @@ internal final class AnalyticsProvider: AnalyticsProviderProtocol {
...
+    private func setupTimer() {
+        let timer = Timer(timeInterval: 10, target: self, selector: #selector(sendAll), userInfo: nil, repeats: true)
+        timer.tolerance = 1
+        RunLoop.current.add(timer, forMode: .common)
+    }
+    
+    @objc private func sendAll() {
+        guard configuration.isEnabled,
+              let checkoutAttemptId else { return }
+        var request = AdyenAnalyticsRequest(checkoutAttemptId: checkoutAttemptId)
```
goergisn: Maybe check if there is actual data points to be sent to avoid useless api calls.

Or maybe stop the timer after the api call if there is nothing to send and only start it again after a data point arrives 🤔

Adyen/Core/Components/InstantPaymentComponent.swift
```diff
@@ -56,14 +56,11 @@ public final class InstantPaymentComponent: PaymentComponent {
 
     /// Generate the payment details and invoke PaymentsComponentDelegate method.
     public func initiatePayment() {
-        sendTelemetryEvent()
+        // we can attempt to fetch the checkoutAttemptId but it won't be ready for the payment
```
goergisn:
```suggestion
        // We are not attempting to fetch the checkoutAttemptId here as it won't be ready for the payment and we don't want to block the payment because of an analytics call
```

Or something like that to give it a bit more context

Adyen/Core/Core Protocols/PresentableComponent.swift
```diff
@@ -62,15 +62,26 @@ public extension PresentableComponent {
 
 @_spi(AdyenInternal)
 public protocol TrackableComponent: Component {
-
-    func sendTelemetryEvent()
+    
+    func fetchCheckoutAttemptId()
```
goergisn: The naming feels a bit misleading as it's actually sending analytics events right?
Fetching the `checkoutAttemptId` is only a sideeffect isn't it?

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
```diff
@@ -32,16 +32,26 @@ public struct AnalyticsConfiguration {
...
-    func fetchAndCacheCheckoutAttemptIdIfNeeded()
+    /// Sends the initial data and retrieves the checkout attempt id as a response.
+    func fetchCheckoutAttemptId(with flavor: TelemetryFlavor, additionalFields: AdditionalAnalyticsFields?)
+}
+
+@_spi(AdyenInternal)
+public protocol AnalyticsProviderProtocol: InitialTelemetryProtocol {
     
-    var additionalFields: (() -> AdditionalAnalyticsFields)? { get }
+    var checkoutAttemptId: String? { get }
```
neelSharma12, line 47: Do we need to create `InitialTelemetryProtocol` and `AnalyticsProviderProtocol` protocols? Can't we add  `checkoutAttemptId` in `InitialTelemetryProtocol`?

Adyen/Analytics/Models/AdyenAnalyticsEvent.swift
```diff
@@ -0,0 +1,35 @@
...
+// Copyright (c) 2023 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+extension AdyenAnalytics {
+    
+    /// Represents an event in the analytics scheme that can occur
```
neelSharma12: Should this entire command be inside` /* */` block?

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
erenbesel: Think committing this part was not needed as it was not done. I will work on this part on next PR

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
erenbesel: Will work on this area on next PR, would have been better not to include this code here but yeah.

Adyen/Core/Core Protocols/PresentableComponent.swift
erenbesel: I mean it's a combined call so calling it either may not be enough. `sendInitialCallAndFetchCheckoutAttemptId()` :)

will rename to something like `sendInitialData()`

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
erenbesel: Hmm, there may not be a reason for them to stay separate any more. Will update if that's the case, nice catch :)

Tests/Adyen Tests/Analytics/AnalyticsProviderTests.swift
```diff
@@ -34,18 +35,18 @@ class AnalyticsProviderTests: XCTestCase {
...
-        sut.fetchCheckoutAttemptId { receivedCheckoutAttemptId in
-
-            // Then
-            XCTAssertNotNil(receivedCheckoutAttemptId, "The checkoutAttemptId is nil.")
-            XCTAssertEqual(expectedCheckoutAttemptId, receivedCheckoutAttemptId, "The received checkoutAttemptId is not the expected one.")
-            fetchCheckoutAttemptIdExpection.fulfill()
-        }
-
+        sut.sendInitialAnalytics(with: .components(type: .achDirectDebit), additionalFields: nil)
+        fetchCheckoutAttemptIdExpection.fulfill()
```
goergisn: Looks like this expectation is not really doing anything anymore - We could get rid of it alltogether with the `waitForExpectations`

Tests/Adyen Tests/Analytics/AnalyticsProviderTests.swift
```diff
@@ -34,18 +35,18 @@ class AnalyticsProviderTests: XCTestCase {
...
+        
+        wait(for: .milliseconds(200))
```
goergisn:
```suggestion
        wait(until: sut, at: \.checkoutAttemptId, is: expectedCheckoutAttemptId)
```

Tests/Adyen Tests/Analytics/AnalyticsProviderTests.swift
```diff
@@ -34,18 +35,18 @@ class AnalyticsProviderTests: XCTestCase {
...
+        
+        XCTAssertNotNil(sut.checkoutAttemptId, "The checkoutAttemptId is nil.")
```
goergisn:
```suggestion
```

Tests/Adyen Tests/Analytics/AnalyticsProviderTests.swift
```diff
@@ -34,18 +35,18 @@ class AnalyticsProviderTests: XCTestCase {
...
+        XCTAssertEqual(expectedCheckoutAttemptId, sut.checkoutAttemptId, "The received checkoutAttemptId is not the expected one.")
```
goergisn:
```suggestion
```

Tests/Adyen Tests/Analytics/AnalyticsProviderTests.swift
```diff
@@ -34,18 +35,18 @@ class AnalyticsProviderTests: XCTestCase {
...
+        
         waitForExpectations(timeout: 10)
```
goergisn:
```suggestion
```

Tests/Adyen Tests/Analytics/AnalyticsProviderTests.swift
```diff
@@ -79,43 +75,38 @@ class AnalyticsProviderTests: XCTestCase {
...
-        sut.fetchCheckoutAttemptId { receivedCheckoutAttemptId in
-
-            // Then
-            XCTAssertNotNil(receivedCheckoutAttemptId, "The checkoutAttemptId is nil.")
-            XCTAssertEqual(expectedCheckoutAttemptId, receivedCheckoutAttemptId, "The received checkoutAttemptId is not the expected one.")
-            fetchCheckoutAttemptIdExpection.fulfill()
-        }
-
+        sut.sendInitialAnalytics(with: .components(type: .achDirectDebit), additionalFields: nil)
+        fetchCheckoutAttemptIdExpection.fulfill()
```
goergisn: Same here - the expectation is not needed as far as I can see

Tests/Adyen Tests/Analytics/AnalyticsProviderTests.swift
```diff
@@ -79,43 +75,38 @@ class AnalyticsProviderTests: XCTestCase {
...
+        
+        wait(for: .milliseconds(200))
```
goergisn: wait(until: sut, at: \.checkoutAttemptId, is: expectedCheckoutAttemptId)

Tests/Adyen Tests/Analytics/AnalyticsProviderTests.swift
```diff
@@ -131,13 +122,19 @@ class AnalyticsProviderTests: XCTestCase {
...
         // When
-        sut.fetchCheckoutAttemptId { _ in
-
-            // Then
-            XCTAssertEqual(expectedCheckoutAttemptId, sut.checkoutAttemptId)
-        }
+        sut.sendInitialAnalytics(with: .components(type: .atome), additionalFields: nil)
+        fetchCheckoutAttemptIdExpection.fulfill()
+        
+        wait(for: .milliseconds(200))
```
goergisn: wait(until: sut, at: \.checkoutAttemptId, is: expectedCheckoutAttemptId)

Tests/Adyen Tests/Analytics/AnalyticsProviderTests.swift
```diff
@@ -131,13 +122,19 @@ class AnalyticsProviderTests: XCTestCase {
...
+        let fetchCheckoutAttemptIdExpection = expectation(description: "checkoutAttemptId completion")
 
         // When
-        sut.fetchCheckoutAttemptId { _ in
-
-            // Then
-            XCTAssertEqual(expectedCheckoutAttemptId, sut.checkoutAttemptId)
-        }
+        sut.sendInitialAnalytics(with: .components(type: .atome), additionalFields: nil)
+        fetchCheckoutAttemptIdExpection.fulfill()
```
goergisn: Same here with the expectation

Adyen/Analytics/Models/AdyenAnalytics.swift
```diff
@@ -0,0 +1,34 @@
...
+        internal var metadata: [String: String] = [:]
+    }
+    
+}
+
+internal protocol AdyenAnalyticsCommonFields: Encodable {
+    var commonFields: AdyenAnalytics.CommonFields { get }
+}
+
+//@_spi(AdyenInternal)
```
goergisn: Left over commented out code

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
goergisn: Ahh I see - this code is not even called yet? 👌

AdyenComponents/BACS Direct Debit/Trackers/BACSDirectDebitComponentTracker.swift
```diff
@@ -8,35 +8,35 @@
...
-        self.apiContext = apiContext
-        self.telemetryTracker = telemetryTracker
+        self.context = context
         self.isDropIn = isDropIn
     }
 
     // MARK: - BACSDirectDebitComponentTrackerProtocol
 
-    internal func sendTelemetryEvent() {
+    internal func fetchCheckoutAttemptId() {
```
goergisn: Maybe also rename it here :)

goergisn (dismissed):

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
```diff
@@ -62,32 +73,71 @@ internal final class AnalyticsProvider: AnalyticsProviderProtocol {
...
+    internal func send(error: AdyenAnalytics.Error) {
+        errors.append(error)
+        sendAll()
+    }
+    
+    private func setupTimer() {
+        let timer = Timer(timeInterval: 10, target: self, selector: #selector(sendAll), userInfo: nil, repeats: true)
+        timer.tolerance = 1
+        RunLoop.current.add(timer, forMode: .common)
+    }
```
descorp: Is this used?

erenbesel: No not yet. As I wrote before these parts were not needed yet. I will work on them on next PR

AdyenComponents/Boleto/BoletoComponent.swift
```diff
@@ -171,13 +172,8 @@ extension BoletoComponent: TrackableComponent {}
 
 @_spi(AdyenInternal)
 extension BoletoComponent: ViewControllerDelegate {
-
-    public func viewDidLoad(viewController: UIViewController) {}
```
erenbesel, line 175: All the removed view controller delegate conformances are applied together on PresentableComponent line 73

Adyen/Analytics/Models/AnalyticsEventInfo.swift
```diff
@@ -0,0 +1,34 @@
...
+// Copyright (c) 2023 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+@_spi(AdyenInternal)
+/// Represents an info event  in the analytics scheme that can occur
+/// many times during the checkout flow, such as input field focus/unfocus etc.
```
descorp:
```suggestion
/// Represents an info event in the analytics scheme that can occur
/// multiple times during the checkout flow, such as input field focus/unfocus, etc.
```

Adyen/Analytics/Models/AnalyticsEventLog.swift
```diff
@@ -0,0 +1,39 @@
+//
+// Copyright (c) 2023 Adyen N.V.
```
descorp, line 2: Optional:

```suggestion
// Copyright (c) 2024 Adyen N.V.
```

😄

AdyenDropIn/DropInComponentExtensions.swift
```diff
@@ -22,7 +22,10 @@ extension DropInComponent: PaymentMethodListComponentDelegate {
 
     internal func didLoad(_ paymentMethodListComponent: PaymentMethodListComponent) {
         let paymentMethodTypes = paymentMethods.regular.map(\.type.rawValue)
-        context.analyticsProvider.sendTelemetryEvent(flavor: .dropIn(paymentMethods: paymentMethodTypes))
+        let flavor = AnalyticsFlavor.dropIn(paymentMethods: paymentMethodTypes)
+        let amount = context.payment?.amount
+        let additionalFields = AdditionalAnalyticsFields(amount: amount, sessionId: AnalyticsForSession.sessionId)
+        context.analyticsProvider?.sendInitialAnalytics(with: flavor, additionalFields: additionalFields)
```
descorp: Can we just make `DropInComponent` a `TrackableComponent ` and use default implementation?

AdyenComponents/BACS Direct Debit/Trackers/BACSDirectDebitComponentTracker.swift
```diff
@@ -8,35 +8,37 @@
...
-        let flavor: TelemetryFlavor = isDropIn ? .dropInComponent : .components(type: paymentMethod.type)
-        telemetryTracker.sendTelemetryEvent(flavor: flavor)
+    internal func sendInitialAnalytics() {
+        // initial call is not needed again if inside dropIn
+        guard !isDropIn else { return }
+        let flavor: AnalyticsFlavor = .components(type: paymentMethod.type)
+        let amount = context.payment?.amount
+        let additionalFields = AdditionalAnalyticsFields(amount: amount, sessionId: AnalyticsForSession.sessionId)
+        context.analyticsProvider?.sendInitialAnalytics(with: flavor,
+                                                         additionalFields: additionalFields)
```
descorp, line 41: Can we make `BACSDirectDebitComponentTracker ` implement `TrackableComponent` instead of creating new protocol?

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
```diff
@@ -62,31 +64,33 @@ internal final class AnalyticsProvider: AnalyticsProviderProtocol {
...
-            completion(checkoutAttemptId)
             return
         }
+        
+        let analyticsData = AnalyticsData(flavor: flavor,
+                                          additionalFields: additionalFields,
+                                          context: configuration.context)
 
-        let checkoutAttemptIdRequest = CheckoutAttemptIdRequest()
+        fetchCheckoutAttemptId(with: analyticsData)
```
atmamont: It feels a bit confusing that `send` method calls `fetch` in the end, wdyt? Fetching the id seems to be a side-effect here not the primary action

Adyen/Analytics/Models/AdyenAnalytics.swift
```diff
@@ -0,0 +1,33 @@
...
+    
+    private init() { /* Private empty init */ }
+}
+
+@_spi(AdyenInternal)
+/// A protocol that defines the events that can occur under Checkout Analytics.
+public protocol AnalyticsEvent: Encodable {
+    var timestamp: TimeInterval { get }
+    
+    var component: String { get }
```
atmamont, line 24: Do we have any strict mapping on the backend side or is this just a freeform string?

AdyenDropIn/DropInComponentExtensions.swift
erenbesel: DropIn is not a `PresentableComponent` so the default extension won't work for it.

AdyenComponents/BACS Direct Debit/Trackers/BACSDirectDebitComponentTracker.swift
erenbesel: But this class is not a component :)  BACS is written a little differently like a VIPER structure.

atmamont (dismissed):
This is a biggie, nice job Eren!
I went through some basic parts but wasn't able to grasp the whole scope unfortunately. It would be cool to keep PRs small in the future, this would allow to collect reviews (along with ideas and feedback!) faster

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
erenbesel: Yeah this one is tricky both ways. Can name it `sendData....` then the checkoutAttemptId would be hidden. `sendDataAndFetchCheckoutAttemptId...` :) is the most expressive but it's long

Adyen/Analytics/Models/AdyenAnalytics.swift
erenbesel: You mean `component`? Everything on the models are named from the backend side

AdyenComponents/BACS Direct Debit/Trackers/BACSDirectDebitComponentTracker.swift
descorp: But only responsibility of `BACSDirectDebitComponentTracker ` is to send analytics.
If `BACSDirectDebitComponent` will implements `TrackableComponent` then we can pass it to `BACSInputPresenter` as `var weak TrackableComponent?`. 
This way we keep `VIPER structure` and reuse existing code for sending analytics.

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
erenbesel: Also now the function is really short, we can maybe remove the 2nd function and keep sending the request here as well, wdyt?

Adyen/Analytics/Models/AdyenAnalytics.swift
atmamont: I could be more specific, sorry :) 
I was asking if this could be enum or we need this as a flexible string?

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
atmamont: Yes, seems new changes made everything more complex!

Adyen/Analytics/Models/AdyenAnalytics.swift
erenbesel: Oh so far it will be the txVariant of the payment method which is the `PaymentMethodType` enum rawValue. But this protocol doesn't need to know about that enum, and no need to create another copy enum :)

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
erenbesel: This turned out nice! 63016fdaebcce42e664feb3212e1552a274e8723

atmamont: Neat! 👏

atmamont (approved):

descorp (approved):

neelSharma12 (approved):

PR #1501, opened 2024-01-09
`new` Detached address input

goergisn:
# Changes
<changed>
For all payment methods that include input fields for the shopper's address, the address input fields are now on a separate screen. This declutters the payment screen to improve user experience.
</changed>

## Summary
- Moves address input to a separate screen for ACH, Affirm, ...
- Moving cocoapods + carthage tests to an xl runner
- Fixed demo app settings look for iPad


AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
```diff
@@ -184,3 +208,20 @@ extension CardViewController {
     }
 
 }
+
+internal class ViewControllerPresenting: ViewControllerPresenter {
```
goergisn: A wrapper to have a nonoptional presenter that wraps a weak (and thus optional) presenter.
Ideally we make the presenter of the `FormAddressPickerItem` non optional in the init to force initialization with a presenter that is then held onto weakly.

Any other ideas?

Adyen/UI/Form/Items/Address/Picker/FormAddressPickerItem.swift
```diff
@@ -106,3 +134,99 @@ private extension FormAddressPickerItem {
...
+}
+
+// MARK: - AddressType
+
+public extension FormAddressPickerItem.AddressType {
+    
+    func placeholder(with localizationParameters: LocalizationParameters?) -> String {
+        switch self {
+        case .billing: return localizedString(.billingAddressPlaceholder, localizationParameters)
+        case .delivery: return "DELIVERY" // localizedString(.deliveryAddressPlaceholder, localizationParameters) // TODO: Get translations
```
descorp: Just for visibility :)

AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
descorp: Idea is that `ViewControllerPresenting` will be destroyed upon destruction of `CardViewControllerItemsProvider` and this will break the retain cycle?

goergisn: Issue is that the presenter is a weak reference and thus it's optional when accessed later and that means I'd have to make the presenter parameter on `FormAddressPickerItem.init` optional.

By creating this wrapper I can make the presenter required on init but still have it released correctly.

atmamont: Do you think we can come up with a slightly improved naming? `ViewControllerPresenting` sounds more like a protocol and does not really explains the difference in behavior and the reasoning to create a new class.

Adyen/Assets/pl-PL.lproj/Localizable.strings
```diff
@@ -86,6 +86,7 @@
 "adyen.billingAddressSection.title" = "Adres rozliczeniowy";
 "adyen.billingAddress.placeholder" = "Twój adres rozliczeniowy";
 "adyen.deliveryAddressSection.title" = "Adres dostawy";
+"adyen.deliveryAddress.placeholder" = "Adres dostawy";
```
descorp, line 89:
```suggestion
"adyen.deliveryAddress.placeholder" = "Twój adres adres dostawy";
```

Adyen/Assets/pt-BR.lproj/Localizable.strings
```diff
@@ -86,6 +86,7 @@
 "adyen.billingAddressSection.title" = "Endereço de cobrança";
 "adyen.billingAddress.placeholder" = "Endereço de cobrança";
```
descorp, line 87:
```suggestion
"adyen.billingAddress.placeholder" = "Seu endereço de cobrança";
```

descorp (dismissed):

AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
goergisn: Yes, makes sense thanks - renamed it to `WeakReferenceViewControllerPresenter` and moved it to the `ViewControllerPresenter` file let me know what you think 👌

Adyen/UI/View Controllers/AddressLookup/AddressLookupStyle.swift
```diff
@@ -13,5 +13,11 @@ public struct AddressLookupStyle: ViewStyle {
     public var search: AddressLookupSearchStyle = .init()
     public var form: FormComponentStyle = .init()
     
-    public init() {}
+    public init(
```
neelSharma12, line 20: Should we add comment/description for this public init?

goergisn: Thanks :)

atmamont (dismissed):

AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
atmamont: Makes total sense now 👍

Adyen/Assets/pt-BR.lproj/Localizable.strings
goergisn: Thanks for the suggestion - will fix it in a subsequent PR

Adyen/Assets/pl-PL.lproj/Localizable.strings
goergisn: Thanks for the suggestion - will fix it in a subsequent PR

Adyen/UI/View Controllers/AddressLookup/Form/AddressInputFormViewController+ViewModel.swift
```diff
@@ -36,15 +41,17 @@ extension AddressInputFormViewController {
...
         public init(
+            for addressType: FormAddressPickerItem.AddressType,
             style: FormComponentStyle,
             localizationParameters: LocalizationParameters?,
             initialCountry: String,
             prefillAddress: PostalAddress?,
             supportedCountryCodes: [String]?,
             addressViewModelBuilder: AddressViewModelBuilder = DefaultAddressViewModelBuilder(),
-            handleShowSearch: ShowSearchHandler?,
+            handleShowSearch: ShowSearchHandler? = nil,
```
nauaros, line 51: I guess it's too late to change it since it may be a breaking change, but I believe it's less confusing to use `showSearchHandler` than `handleShowSearch`. Just my perception when I read it for the first time.

nauaros (approved):

neelSharma12 (approved):

Adyen/UI/View Controllers/AddressLookup/Form/AddressInputFormViewController+ViewModel.swift
goergisn: Not too late but felt like the right property name because the `ShowSearchHandler` is a typealias for a closure `public typealias ShowSearchHandler = (_ currentInput: PostalAddress) -> Void` - any thoughts?

PR #1507, opened 2024-01-17
`chore` Improve tintColor tests

goergisn:
## Summary
- Using non-dynamic colors to improve reliability of tint color tests
- Refactoring some tests
- Skipping `testTintColorCustomization` for compatibility tests - will be done via snapshotting

### DIFF
--- a/Tests/Card Tests/Utilities/ThrottlerTests.swift
+++ b/Tests/Card Tests/Utilities/ThrottlerTests.swift
@@ -11,7 +11,7 @@ import XCTest
 class ThrottlerTests: XCTestCase {
 
     func test() {
-        let sut = Throttler(minimumDelay: 1)
+        let sut = Throttler(minimumDelay: 0.3)
         
         let lastBlockExpectation = expectation(description: "wait for last block execution")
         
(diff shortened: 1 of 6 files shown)


Tests/Card Tests/Utilities/ThrottlerTests.swift
```diff
@@ -29,12 +29,11 @@ class ThrottlerTests: XCTestCase {
                 }
             }
             
-            wait(for: .milliseconds(100))
+            wait(for: .milliseconds(5))
```
atmamont, line 32: Quite a significant change here, are these numbers mostly come from trial and error (respecting CI env as well)?

atmamont (approved):

Tests/Card Tests/Utilities/ThrottlerTests.swift
goergisn: This test is actually testing throttling/debouncing behavior - as I reduced the `minimumDelay` i was able to also reduce the time between the calls to `throttle`

atmamont: Oh, I see. Magic number can be tricky. Maybe for better readability we can extract those into a nice named constants in the future

PR #1522, opened 2024-02-15
`chore` Improve Apple Pay flow state naming

atmamont:
## Summary
This change does not change logic. This change improves naming to be more precise. 
We might receive an error after submitting payment details to Apple so name `submitted` fits and reads better than `paid` and doesn't provoke any false assumptions.

## Tested scenarios
No additional testing needed.


erenbesel (approved):
Nice!

goergisn (approved):

PR #1531, opened 2024-02-29
Analytics event handling

erenbesel:
- Creates 2 event data sources (one regular, one syncronous) to help the Analytics provider. 

- Implements a mechanism to trigger sending of the events with a timer and also instantly when needed.


Adyen/Analytics/Models/AnalyticsEventHandler.swift
```diff
@@ -0,0 +1,125 @@
...
+    private func requestWithAllEvents() -> AnalyticsRequest? {
+        guard let checkoutAttemptId else { return nil }
+        
+        return queue.sync {
+            if infos.isEmpty && logs.isEmpty && errors.isEmpty {
+                return nil
+            }
+            
+            // as per this call's limitation, we only send up to the
+            // limit of each event and discard the older ones
```
atmamont: Discard sounds like we just drop them. However, it looks like we just send them later, right?

atmamont: Oh maybe we just discard them indeed, I didn't find any subtraction of events that were sent

erenbesel: Yes indeed we discard since on a success we clear the whole arrays at this point.

atmamont (dismissed):
Looks good, one small thing to clarify regarding what happens with events that didn't make to the batch due to size limitation. Do we have an agreement to drop them?

Adyen/Analytics/Models/AdyenAnalytics.swift
```diff
@@ -30,4 +32,23 @@ public extension AnalyticsEvent {
     var timestamp: TimeInterval {
         Date().timeIntervalSince1970
     }
+    
+    var id: String { UUID().uuidString }
```
goergisn: Idea: This has a different idea everytime it's called. Which might lead to unexpected behavior. Maybe better to assign it on init of the conforming object 🤷‍♂️

Adyen/Analytics/Models/AdyenAnalytics.swift
```diff
@@ -30,4 +32,23 @@ public extension AnalyticsEvent {
...
+}
+
+@_spi(AdyenInternal)
```
goergisn: Maybe not for now but we could already start using the package modifier instead of @_spi

Adyen/Analytics/Models/AnalyticsEventDataSource.swift
```diff
@@ -0,0 +1,46 @@
...
+        errors.append(error)
+    }
+    
+    internal func removeAllEvents() {
+        infos = []
+        logs = []
+        errors = []
+    }
+    
+    internal func wrappedEvents() -> AnalyticsEventWrapper? {
```
atmamont: This var name feels like it could be just named `events` and the type could be just `AnalyticEvents` but I don't feel strong about it. Maybe other reviewers could bring more ideas?

Adyen/Analytics/Models/SyncAnalyticsEventDataSource.swift
```diff
@@ -0,0 +1,72 @@
...
+}
+
+internal struct AnalyticsEventWrapper {
+    internal var infos: [AnalyticsEventInfo]
+    internal var logs: [AnalyticsEventLog]
+    internal var errors: [AnalyticsEventError]
+}
+
+/// Composes handling of events in a thread safe manner.
+internal final class SyncAnalyticsEventDataSource: AnyAnalyticsEventDataSource {
```
atmamont: Oh now I see the short `sync` version reads as a verb, this idea from myself wasn't the greatest :) 
Maybe `Synchronous`?

atmamont: Other than that this class is seriously beautiful! 🤩

Tests/Adyen Tests/Analytics/AnalyticsEventDataSourceTests.swift
```diff
@@ -0,0 +1,74 @@
...
+    override func tearDownWithError() throws {
+        // Put teardown code here. This method is called after the invocation of each test method in the class.
+        sut = nil
+    }
+
+    func testAddingInfo() {
+        XCTAssertEqual(sut.infos.count, 0)
+        
+        sut.add(info: AnalyticsEventInfo(component: "test", type: .rendered))
+        XCTAssertEqual(sut.infos.count, 1)
```
atmamont: Do we need to expose infos logs and errors now since we introduced the wrapper object? Can we make them private?

Tests/Adyen Tests/Analytics/AnalyticsProviderMock.swift
```diff
@@ -26,4 +26,19 @@ class AnalyticsProviderMock: AnalyticsProviderProtocol {
     var initialEventCalled: Bool {
         initialEventCallsCount > 0
     }
+    
+    var infoCount = 0
+    func add(info: Adyen.AnalyticsEventInfo) {
+        infoCount += 1
```
atmamont, line 32: I think for the future tests for SyncDataStore it could be beneficial to capture real incoming events and make count to be a calculated var, but this is something for the next batch

Tests/Adyen Tests/Analytics/AnalyticsProviderTests.swift
```diff
@@ -163,6 +192,129 @@ class AnalyticsProviderTests: XCTestCase {
...
+        apiClient.mockedResults = [checkoutAttemptIdResult, analyticsResult]
+        
+        sut.sendInitialAnalytics(with: .components(type: .achDirectDebit), additionalFields: nil)
+        wait(for: .milliseconds(100))
+        
+        sut.add(log: AnalyticsEventLog(component: "card", type: .submit, subType: .sdk))
+        wait(for: .milliseconds(100))
+        
+        // events return nil if empty
+        XCTAssertNil(sut.eventDataSource.wrappedEvents())
```
atmamont: I would doubt this test actually tests analytics provider behavior. Assuming datasource is faulty and just fails to store any events this test would to be always green! This test checks for rather datasource side-effects and not the sut behavior. Testing side-effects might be fine but here I see we can improve this.
I suggest testing the `apiClient.perform` and check the request data if possible.

And the second this that we test here is "datasource is cleaned up". This could be tested by injecting a datasource mock rather than using a real class and then checking if `removeAllEvents` was called.

Tests/Adyen Tests/Analytics/AnalyticsProviderTests.swift
```diff
@@ -163,6 +192,129 @@ class AnalyticsProviderTests: XCTestCase {
...
+        apiClient.mockedResults = [checkoutAttemptIdResult, analyticsResult]
+        
+        sut.sendInitialAnalytics(with: .components(type: .achDirectDebit), additionalFields: nil)
+        wait(for: .milliseconds(100))
+        
+        sut.add(error: AnalyticsEventError(component: "card", type: .implementation))
+        wait(for: .milliseconds(100))
+        
+        // events return nil if empty
+        XCTAssertNil(sut.eventDataSource.wrappedEvents())
```
atmamont: Same here and in other tests.

atmamont:
Great improvement, Eren! Now it is much easier to read as responsibilities are split in a better way! 💪

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
```diff
@@ -86,6 +86,52 @@ internal final class AnalyticsProvider: AnalyticsProviderProtocol {
...
+    internal func sendEventsIfNeeded() {
+        guard let request = requestWithAllEvents() else { return }
+        
+        apiClient.perform(request) { [weak self] result in
+            guard let self else { return }
+            // clear the current events on successful send
+            switch result {
+            case .success:
+                self.eventDataSource.removeAllEvents()
+                self.startNextTimer()
```
goergisn, line 113: Is this needed? The timer is repeating so it should just continue sending events?

Adyen/Analytics/Models/AnalyticsEventDataSource.swift
goergisn: Maybe `allEvents` to be in line with `removeAllEvents` (could also be a computed var instead of a func)

Adyen/Analytics/Models/SyncAnalyticsEventDataSource.swift
goergisn: `ThreadSafeAnalyticsEventDataSource` (Naming is hard :D)

Tests/Adyen Tests/Analytics/AnalyticsEventDataSourceTests.swift
```diff
@@ -0,0 +1,74 @@
...
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import XCTest
+@_spi(AdyenInternal) @testable import Adyen
+
+final class AnalyticsEventDataSourceTests: XCTestCase {
+    
+    var sut: AnalyticsEventDataSource!
```
goergisn, line 12: In my experience having state inside of tests always leads to weird behavior.
Maybe consider moving this to local state - so you also don't have to setup and tear down

Adyen/Analytics/Models/SyncAnalyticsEventDataSource.swift
atmamont: Funny, this was the very first name we came up with! 🤣

Adyen/Analytics/Models/AdyenAnalytics.swift
erenbesel: The intention was exactly that :), nice catch!

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
erenbesel: Thought it's better to reset the timer after a response since we are clearing the array here and if it triggers right after may be a futile effort.
Both is fine I guess?

.swiftlint.yml
```diff
@@ -22,3 +22,9 @@ line_length:
   
 file_length:
   warning: 500
+  
+variable_name:
+  min_length:
+    warning: 2
```
goergisn, line 28: 👌

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
```diff
@@ -94,4 +141,11 @@ internal final class AnalyticsProvider: AnalyticsProviderProtocol {
             checkoutAttemptId = nil
         }
     }
+    
+    private func startNextTimer() {
+        batchTimer?.invalidate()
```
goergisn, line 146: Idea: maybe move the `invalidate()` to the `willSet` of the `batchTimer` this way we can make sure that it's always invalidated if a new timer is started :)

Tests/Adyen Tests/Analytics/SyncAnalyticsEventDataSourceTests.swift
```diff
@@ -0,0 +1,31 @@
+//
+// Copyright (c) 2024 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import XCTest
+@_spi(AdyenInternal) @testable import Adyen
+
+final class SyncAnalyticsEventDataSourceTests: XCTestCase {
```
goergisn: Is this still needed?

Tests/Adyen Tests/Analytics/AnalyticsProviderTests.swift
```diff
@@ -263,15 +263,56 @@ class AnalyticsProviderTests: XCTestCase {
...
         wait(for: .milliseconds(100))
         
-        sut.add(info: AnalyticsEventInfo(component: "card", type: .rendered))
+        let infoEvent = AnalyticsEventInfo(component: "card", type: .rendered)
+        sut.add(info: infoEvent)
+        
+        let shouldSendExpectation = expectation(description: "send event is called")
+        apiClient.onExecute = { request in
+            if let analyticsRequest = request as? AnalyticsRequest {
+                XCTAssertEqual(analyticsRequest.infos.first?.id, infoEvent.id)
```
atmamont: This block requires some cognitive effort, a small suggestion would be

```swift
let expectedID = infoEvent.id
....
apiClient.onExecute = { request in
    if let analyticsRequest = request as? AnalyticsRequest, 
        let capturedId = analyticsRequest.infos.first?.id {

            XCTAssertEqual(capturedId, expectedId, "Expected to send event with id=\(expectedId), got \(capturedId) instead")
            shouldSendExpectation.fulfill()
    }
}
``` 
It can be very helpful to add meaningful assertion messages as you are killing two birds with one stone: make real assertion clear right away and make this section more readable for someone who uses tests as a short description of business requirements.

Tests/Adyen Tests/Analytics/AnalyticsProviderTests.swift
```diff
@@ -263,15 +263,56 @@ class AnalyticsProviderTests: XCTestCase {
...
+        
+        let shouldSendExpectation = expectation(description: "send event is called")
+        apiClient.onExecute = { request in
+            if let analyticsRequest = request as? AnalyticsRequest {
+                XCTAssertEqual(analyticsRequest.infos.count, 3)
+                shouldSendExpectation.fulfill()
+            }
+        }
+        
+        sut = nil
```
atmamont, line 312: This is really nice!

Tests/Adyen Tests/Analytics/SyncAnalyticsEventDataSourceTests.swift
atmamont: Let's add this class whenever we have tests?

atmamont (dismissed):
I really like how it shapes out!

goergisn (dismissed):

atmamont (dismissed):

Tests/Adyen Tests/Analytics/AnalyticsProviderTests.swift
erenbesel: updated e6d9e4167b750d4e4b1d8a24d821cd81b1f513a2

atmamont (approved):

Tests/Adyen Tests/Analytics/AnalyticsProviderTests.swift
```diff
@@ -163,6 +192,221 @@ class AnalyticsProviderTests: XCTestCase {
...
+        let initialAnalyticsResponse = InitialAnalyticsResponse(checkoutAttemptId: checkoutAttemptIdMockValue)
+        let checkoutAttemptIdResult: Result<Response, Error> = .success(initialAnalyticsResponse)
+        
+        let analyticsResponse = AnalyticsResponse()
+        let analyticsResult: Result<Response, Error> = .success(analyticsResponse)
+        
+        apiClient.mockedResults = [checkoutAttemptIdResult, analyticsResult]
+        
+        sut.sendInitialAnalytics(with: .components(type: .achDirectDebit), additionalFields: nil)
+        wait(for: .milliseconds(100))
```
goergisn, line 332: is there any way around waiting for a "random" amount of time?

Tests/Adyen Tests/Analytics/AnalyticsProviderTests.swift
```diff
@@ -163,6 +192,221 @@ class AnalyticsProviderTests: XCTestCase {
...
```
goergisn, line 361: is there any way around waiting for a "random" amount of time?

goergisn (approved):

Tests/Adyen Tests/Analytics/AnalyticsProviderTests.swift
erenbesel: not sure really, maybe only testing through apiClient.onExecute? we can check next

PR #1532, opened 2024-02-29
[Twint] Checking for TwintSDK to make native and redirect flow possible

goergisn:
## Summary
- checking for `#if canImport(TwintSDK)` as we have to check on `Adyen` whether or not `AdyenTwint` can be used. (We have to check that way because `AdyenTwint` has the `Adyen` dependency and thus the `canImport` check always returns false)

### DIFF
--- a/AdyenActions/Components/SDK/TwintSDKActionComponent.swift
+++ b/AdyenActions/Components/SDK/TwintSDKActionComponent.swift
@@ -67,15 +67,17 @@ import Foundation
         public func handle(_ action: TwintSDKAction) {
             AdyenAssertion.assert(message: "presentationDelegate is nil", condition: presentationDelegate == nil)
             Twint.fetchInstalledAppConfigurations { [weak self] installedApps in
-                if installedApps?.count == 0 {
-                    self?.handleShowError("No or an outdated version of TWINT is installed on this device. Please update or install the TWINT app.")
-                } else if installedApps?.count == 1 {
-                    guard let installedApps else {
-                        return
-                    }
-                    self?.invokeTwintAppWithCode(app: installedApps[0], code: action.sdkData.token)
+                guard let self else { return }
+                
+                guard let installedApps, let firstApp = installedApps.first else {
+                    self.handleShowError("No or an outdated version of TWINT is installed on this device. Please update or install the TWINT app.")
+                    return
+                }
+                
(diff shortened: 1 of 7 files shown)


AdyenActions/Components/SDK/TwintSDKActionComponent.swift
```diff
@@ -1,130 +1,130 @@
...
+            self.context = context
+            self.configuration = configuration
+        }
+
+        /// Handles TwintSDK action.
+        ///
+        /// - Parameter action: The Twint SDK action object.
+        public func handle(_ action: TwintSDKAction) {
+            AdyenAssertion.assert(message: "presentationDelegate is nil", condition: presentationDelegate == nil)
+            Twint.fetchInstalledAppConfigurations { [weak self] installedApps in
```
erenbesel, line 69: guard self since there are multiple `self?` usages

AdyenActions/Components/SDK/TwintSDKActionComponent.swift
```diff
@@ -1,130 +1,130 @@
...
-            self.invokeTwintAppWithCode(app: selectedApp ?? installedApps[0],
-                                        code: code)
-        } cancelHandler: {
-            self.delegate?.didFail(with: ComponentError.cancelled, from: self)
-        }
-        if let viewcontroller = appChooserViewController, let delegate = presentationDelegate {
-            present(viewcontroller, presentationDelegate: delegate)
+        private func presentAppChooser(installedApps: [TWAppConfiguration], code: String) {
+            let appChooserViewController = Twint.controller(for: installedApps) { selectedApp in
+                self.invokeTwintAppWithCode(app: selectedApp ?? installedApps[0],
```
erenbesel: are these closures non escaping? if not, let's add weak

AdyenActions/Components/SDK/TwintSDKActionComponent.swift
goergisn: Added it :)

AdyenActions/Components/SDK/TwintSDKActionComponent.swift
goergisn: I added it 👍

erenbesel (approved):

neelSharma12 (approved):

PR #1534, opened 2024-03-01
[Twint] Alert presentation fix

goergisn:
## Summary
- Made `PresentableComponentWrapper.requiresModalPresentation` variable so a component can present UIViewControllers without wrapping them into a sheet (needed for UIAlertController)
- Cancelling when Twint does not find installed apps

Simulator Screenshot - iPhone 14 - 2024-03-01 at 14 57 01{width=300}

### DIFF
--- a/AdyenActions/Components/SDK/TwintSDKActionComponent.swift
+++ b/AdyenActions/Components/SDK/TwintSDKActionComponent.swift
@@ -93,7 +93,7 @@ import Foundation
         private func presentAppChooser(installedApps: [TWAppConfiguration], code: String) {
             let appChooserViewController = Twint.controller(for: installedApps) { [weak self] selectedApp in
                 self?.invokeTwintAppWithCode(app: selectedApp ?? installedApps[0],
-                                            code: code)
+                                             code: code)
             } cancelHandler: { [weak self] in
                 guard let self else { return }
                 self.delegate?.didFail(with: ComponentError.cancelled, from: self)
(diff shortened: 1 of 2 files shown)


AdyenActions/Components/SDK/TwintSDKActionComponent.swift
```diff
@@ -121,10 +122,15 @@ import Foundation
...
-            alert.addAction(.init(
-                title: localizedString(.dismissButton, configuration.localizationParameters),
-                style: .default
-            )
+            alert.addAction(
+                .init(
+                    title: localizedString(.dismissButton, configuration.localizationParameters),
+                    style: .default, 
+                    handler: { _ in
+                        // TODO: Is this the right thing to do here? - we have to let drop-in know to hide the spinner without dismissing drop-in
```
goergisn, line 130: Any opinion on this part here?
I believe this is the right thing to do to stop drop-in from loading

neelSharma12: you mean to hide the spinner next to  twint PM?

goergisn: Yes exactly. The dismissing of the alert should just hide the spinner but not dismiss drop-in I assume.

neelSharma12: yes we should do this :)

goergisn: Ok I will get rid of the discussion todo 👌

PR #1546, opened 2024-03-12
Dev 64080: Revisit public interface of delegate authentication flow

robertdalmeida:
As per our discussions during the code review we wanted to update the public interface of the ThreeDSComponent.

### DIFF
--- a/AdyenActions/Components/3DS2/ThreeDS2Component.swift
+++ b/AdyenActions/Components/3DS2/ThreeDS2Component.swift
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2023 Adyen N.V.
+// Copyright (c) 2024 Adyen N.V.
 //
 // This file is open source and available under the MIT license. See the LICENSE file for more info.
 //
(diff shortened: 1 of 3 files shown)


AdyenActions/Components/3DS2/ThreeDS2Component.swift
```diff
@@ -22,7 +22,7 @@ public final class ThreeDS2Component: ActionComponent {
     /// The delegate of the component.
     public weak var delegate: ActionComponentDelegate?
 
-    /// Delegates `PresentableComponent`'s presentation.
```
goergisn, line 25: Is there a place where we could assert if the `presentationDelegate` is not set?

e.g. we do following in the `AwaitComponent`

```
let message = "PresentationDelegate is nil. Provide a presentation delegate to AwaitComponent."
AdyenAssertion.assertionFailure(message: message)
```

robertdalmeida: We have on use - in the ThreeDS2PlusDAScreenPresenter.  Isn't that enough? 

https://github.com/Adyen/adyen-ios/blob/3d5f181ee0046da7134a64290780c30b40e38f2a/AdyenActions/Components/3DS2/Action%20handlers/3DS2%2BDelegated%20Authentication/ThreeDS2PlusDAScreenPresenter.swift#L59

atmamont (approved):

AdyenActions/Components/3DS2/ThreeDS2Component.swift
goergisn: Yes, should be good enough 👌

PR #1592, opened 2024-03-26
Remove network image view

goergisn:
# Background
We have a NetworkImageView that idependently loads images from a url. 
This makes it hard to test and creates flaky snapshot tests as the loading might take some time and potentially fail uncontrollably.

## Changes
- Removing `NetworkImageView`
- Introducing ImageLoader to allow loading of network images on any `UIImageView`
- `ImageLoader` can be injected and thus can produce reliable/controllable output for all tests without hitting the actual network


Adyen.xcodeproj/xcshareddata/xcschemes/AdyenUIHostUITests.xcscheme
```diff
@@ -26,7 +26,9 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
-      shouldUseLaunchSchemeArgsEnv = "NO">
+      shouldUseLaunchSchemeArgsEnv = "NO"
+      language = "en-US"
```
atmamont: Looks like a leftover from manual testing, right?

Adyen/Helpers/AdyenCancellable.swift
```diff
@@ -0,0 +1,28 @@
...
+import Foundation
+
+@_spi(AdyenInternal)
+public protocol AdyenCancellable {
+    func cancel()
+}
+
+/// Indicates whether or not a task should be cancelled
+@_spi(AdyenInternal)
+public class AdyenCancellation: AdyenCancellable {
```
atmamont: Name suggestion `AdyenTask`

Adyen/Helpers/ImageLoader/ImageLoader.swift
```diff
@@ -0,0 +1,51 @@
...
+    private let urlSession: URLSession
+    
+    public init(urlSession: URLSession = .init(configuration: .default)) {
+        self.urlSession = urlSession
+    }
+    
+    @discardableResult
+    public func load(url: URL, completion: @escaping ((UIImage?) -> Void)) -> AdyenCancellable {
+        var urlSessionTask: URLSessionTask?
+        let cancellation = AdyenCancellation { urlSessionTask?.cancel() }
```
atmamont: Just `task` would be more readable

Adyen/Helpers/ImageLoader/UIImageView+ImageLoading.swift
```diff
@@ -0,0 +1,22 @@
...
+
+import UIKit
+
+@_spi(AdyenInternal)
+public extension UIImageView {
+    @discardableResult
+    func load(url: URL, using imageLoader: ImageLoading, placeholder: UIImage? = nil) -> AdyenCancellable {
+        imageLoader.load(url: url) { [weak self] image in
+            guard let image else {
+                self?.image = placeholder
```
atmamont: This feels like a part of business-logic actually. 
I can imagine we might need to show a placeholder before loading starts and update it with the loaded image after. However, I am not aware of real business logic but this just seem to be capturing it which is not ideal for this level of abstraction.

Adyen/UI/List/ListItemView.swift
```diff
@@ -10,11 +10,17 @@ import UIKit
 /// Displays a list item.
 @_spi(AdyenInternal)
 public final class ListItemView: UIView, AnyFormItemView {
```
atmamont, line 12: Do we have control over it's lifecycle? Ideally we want to cancel image loading when the view will disappear.

AdyenActions/UI/View Controllers/Document/DocumentActionView.swift
```diff
@@ -75,11 +74,20 @@ internal final class DocumentActionView: UIView {
...
+        viewModel: DocumentActionViewModel,
+        style: DocumentComponentStyle,
+        imageLoader: ImageLoading = ImageLoaderProvider.imageLoader()
+    ) {
         self.viewModel = viewModel
         self.style = style
         super.init(frame: .zero)
         configureViews()
+        
+        imageView.load(
```
atmamont: Ideally we want to trigger load on view appear and not instantiation. However, I'm not familiar with this part of the codebase so please make your own judgement. I won't comment on other similar cases.

Tests/Adyen Tests/Core/ImageLoaderTests.swift
```diff
@@ -0,0 +1,81 @@
...
+        (0..<5).forEach { _ in
+            let cancellable = imageLoader.load(url: URL(string: "https://adyen.com")!) { image in
+                noCompletionExpectation.fulfill()
+            }
+            cancellable.cancel()
+        }
+        
+        wait(for: [noCompletionExpectation, completionExpectation], timeout: 0.1)
+    }
+}
```
atmamont, line 211: More test scenarios:
- instantiating `sut` should not call image load
- getting a network error should call completion with `nil`
- deallocating `sut` before completing image load should not call completion
- image load completion is called on the main thread

Tests/Adyen Tests/Core/ImageLoaderTests.swift
```diff
@@ -0,0 +1,81 @@
...
+
+private extension ImageLoaderTests {
+    
+    func makeSUT() -> ImageLoader {
+        
+        let configuration = URLSessionConfiguration.default
+        configuration.protocolClasses = [ImageResponseURLProtocolMock.self]
```
atmamont: 👍

Tests/Adyen Tests/Mocks/ImageLoaderMock.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+@testable @_spi(AdyenInternal) import Adyen
+
+class ImageLoaderMock: ImageLoading {
+    
+    var imageProvider: (URL) -> UIImage? = { url in
+        url.absoluteString.generateImage()
```
atmamont, line 12: 😮

atmamont (dismissed):
Great stuff, Alex!

Adyen.xcodeproj/xcshareddata/xcschemes/AdyenUIHostUITests.xcscheme
goergisn: Yess 🙈

Adyen/UI/List/ListItemView.swift
goergisn: We could use `didMoveToWindow()` to check whether it's visible 🤔

AdyenActions/UI/View Controllers/Document/DocumentActionView.swift
goergisn: Yeah we could override `didMoveToWindow()` and see whether it became visible/invisible. What do you think?

Adyen/UI/List/ListItemView.swift
atmamont: Then let's go to the level up? This is UIView so it should be managed by either a view controller or table/collection view which have specific methods to let us propagate willShow/willDisappear events.
This goes quite a bit beyond the initial scope though.

AdyenActions/UI/View Controllers/Document/DocumentActionView.swift
atmamont: I am not sure as this method is part of `UIView` lifecycle. I'd love to have full control over the lifecycle and the best we can go with is UIViewController or table/collection view

Adyen/UI/List/ListItemView.swift
goergisn: I agree - but to keep the scope small I'd think of how to best do it in a subsequent PR

atmamont (dismissed):

Tests/Helpers/String+UIImage.swift
```diff
@@ -0,0 +1,26 @@
...
+import UIKit
+
+extension String {
+
+    func generateImage() -> UIImage? {
+        let string = self as NSString
+        let size = string.size()
+
+        return UIGraphicsImageRenderer(size: size).image { context in
+            UIColor.white.setFill()
```
descorp, line 16: Is it "dark mode" compatible? :D

Adyen/Helpers/AdyenDependencies.swift
```diff
@@ -19,7 +19,15 @@ public struct AdyenDependencyValues {
     /// A subscript for updating the `currentValue` of `AdyenDependencyKey` instances.
     public subscript<K>(key: K.Type) -> K.Value where K: AdyenDependencyKey {
         get {
-            (storage[ObjectIdentifier(key)] as? K.Value) ?? key.liveValue
+            let override = (storage[ObjectIdentifier(key)] as? K.Value)
+            
+#if canImport(XCTest)
+            if override != nil {
+                print("No test value override provided for \(key)")
```
erenbesel: Isn't overriding done on didSet? Couldn't fully get the picture here

AdyenActions/UI/View Controllers/Voucher/VoucherView.swift
```diff
@@ -248,4 +247,16 @@ internal final class VoucherView: UIView, Localizable {
         )
     }
     
+    override public func didMoveToWindow() {
```
erenbesel, line 250: What do you think of adding a protocol that all these views conform to and handle this duplicate code block in its extension?

erenbesel (commented):
Great stuff! One last question: Would we achieve a similar result if we just added a mockable urlsession injected to networkImageView?
Feels like that was the main missing part of networkimageview preventing valid testing

AdyenActions/UI/View Controllers/Voucher/VoucherView.swift
goergisn: Good one! Ideally this logic actually does not happen inside of the view but in the viewController. For now I just moved the logic out of the NetworkImageView to the parent view to keep the scope small. I'll look into a better separation of concerns in a future PR.

Adyen/Helpers/AdyenDependencies.swift
goergisn: The idea behind this is to log if the test is run with a `liveValue` which it should not for example with the ImageLoader. Ideally it would actually assert on debug but I'd also extract that scope to a different task.

goergisn:
> Great stuff! One last question: Would we achieve a similar result if we just added a mockable urlsession injected to networkImageView? Feels like that was the main missing part of networkimageview preventing valid testing

Yes, we could have achieved a similar effect with a mockable URLSession.
One of the reasons for an ImageLoader was to extract all the business logic (which there was quite a bit of) of the NetworkImageView and being able to use it on any UIImageView in the future.

atmamont (dismissed):

atmamont (dismissed):

atmamont (approved):

erenbesel (approved):

PR #1593, opened 2024-03-26
Render and issuer list events

erenbesel:
- Send rendering of dropin/components
- Send issuer selection event
- Send issuer search event (with 1 sec debounce)
- Add tests (also updates dummy helpers with mock analytics provider to prevent calling the analytics endpoint)

### DIFF
--- a/Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
+++ b/Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
@@ -126,7 +126,11 @@ internal final class AnalyticsProvider: AnalyticsProviderProtocol {
         
         // as per this call's limitation, we only send up to the
         // limit of each event and discard the older ones
-        var request = AnalyticsRequest(checkoutAttemptId: checkoutAttemptId)
+        let platform = configuration.context.platform.rawValue
+        var request = AnalyticsRequest(
+            checkoutAttemptId: checkoutAttemptId,
+            platform: platform
+        )
         request.infos = events.infos.suffix(Constants.infoLimit)
         request.logs = events.logs.suffix(Constants.logLimit)
         request.errors = events.errors.suffix(Constants.errorLimit)
--- a/Adyen/Analytics/Models/AnalyticsEventInfo.swift
+++ b/Adyen/Analytics/Models/AnalyticsEventInfo.swift
@@ -13,7 +13,7 @@ public struct AnalyticsEventInfo: AnalyticsEvent {
     
     public var id: String = UUID().uuidString
     
-    public var timestamp: TimeInterval = Date().timeIntervalSince1970
+    public var timestamp = Int(Date().timeIntervalSince1970)
     
     public var component: String
     
@@ -25,6 +25,8 @@ public struct AnalyticsEventInfo: AnalyticsEvent {
     
     public var brand: String?
     
+    public var issuer: String?
+    
     public var validationErrorCode: String?
     
     public var validationErrorMessage: String?
@@ -35,5 +37,12 @@ public struct AnalyticsEventInfo: AnalyticsEvent {
         case unfocus = "Unfocus"
         case validationError = "ValidationError"
         case rendered = "Rendered"
+        case displayed = "Displayed"
+        case input = "Input"
+    }
+    
+    public init(component: String, type: InfoType) {
+        self.component = component
+        self.type = type
     }
 }
--- a/Adyen/Analytics/Models/AnalyticsEventInfo.swift
+++ b/Adyen/Analytics/Models/AnalyticsEventInfo.swift
@@ -37,7 +37,6 @@ public struct AnalyticsEventInfo: AnalyticsEvent {
         case unfocus = "Unfocus"
         case validationError = "ValidationError"
         case rendered = "Rendered"
-        case displayed = "Displayed"
         case input = "Input"
     }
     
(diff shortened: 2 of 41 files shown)


AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -294,9 +294,8 @@ extension CardViewController {
     private var billingAddressItem: FormItem? {
         
         switch configuration.billingAddress.mode {
-        case let .lookup(provider):
```
goergisn, line 297: 🙏

goergisn (dismissed):

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
```diff
@@ -126,7 +126,11 @@ internal final class AnalyticsProvider: AnalyticsProviderProtocol {
         
         // as per this call's limitation, we only send up to the
         // limit of each event and discard the older ones
-        var request = AnalyticsRequest(checkoutAttemptId: checkoutAttemptId)
+        let platform = configuration.context.platform.rawValue
+        var request = AnalyticsRequest(
+            checkoutAttemptId: checkoutAttemptId,
+            platform: platform
```
goergisn, line 132: what about making the platform parameter type an enum instead of String? So you don't have to get the `rawValue` and it also reduces potential mistakes

Adyen/Analytics/Models/AnalyticsEventInfo.swift
```diff
@@ -35,5 +37,12 @@ public struct AnalyticsEventInfo: AnalyticsEvent {
         case unfocus = "Unfocus"
         case validationError = "ValidationError"
         case rendered = "Rendered"
+        case displayed = "Displayed"
```
atmamont: What is the difference between rendered and displayed?

AdyenCard/Components/Card/CardComponent.swift
```diff
@@ -128,6 +128,8 @@ public class CardComponent: PresentableComponent,
     public var requiresModalPresentation: Bool { storedCardComponent?.requiresModalPresentation ?? true }
     
     public func stopLoading() {
+        // don't access cardViewController when it's storedCardComponent
```
atmamont: Can we add more context? Like we can extract the guard check to a separate method returning bool and give it a meaningful name. Now this check seems to be pure magic.

Tests/Components Tests/IssuerList/IssuerListComponentTests.swift
```diff
@@ -54,6 +61,29 @@ class IssuerListComponentTests: XCTestCase {
 
         waitForExpectations(timeout: 10)
     }
+    
+    func testSearch() {
```
atmamont: This name doesn't explain what we actually test.
The better way to name it would be something like `test_componentSendsInfo_onSearchInput`
It also feels we are not testing real events we send, only count. This is essential part of the business logic of this scope so I suggest to check for real data that we send.

Tests/Components Tests/IssuerList/IssuerListComponentTests.swift
```diff
@@ -54,6 +61,29 @@ class IssuerListComponentTests: XCTestCase {
...
+        let listViewController = searchViewController.resultsListViewController
+        
+        setupRootViewController(searchViewController)
+        
+        searchViewController.searchBar(searchViewController.searchBar, textDidChange: "3")
+        XCTAssertFalse(listViewController.view.isHidden)
+        XCTAssertEqual(analyticsProviderMock.infoCount, 1)
+        
+        searchViewController.searchBar(searchViewController.searchBar, textDidChange: "34")
+        wait(for: .seconds(1))
```
atmamont, line 84: We can add one more check _before_ this delay to ensure there is still one event sent. Otherwise even if you remove the Throttler, this test still would be green.

Small tip: always make your test red by changing the business logic (comment out the Throttler usage here) before commit

atmamont (commented):
Reads easily, great job Eren!
Some small nitpicks related to tests.

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
erenbesel: Could be good but would either need to expose the existing enum which this struct shouldn't know about or create a new one which feels redundant duplication :D
We are setting directly from the enum so it should be fine

Adyen/Analytics/Models/AnalyticsEventInfo.swift
erenbesel: Turns out this is not needed, it's web only so will remove!

Tests/Components Tests/IssuerList/IssuerListComponentTests.swift
erenbesel: 6bf6afab0029fd741e69b4a44a732cb546211610 Updated name but hard to test the actual event itself via this way. But current test still checks on the right flow and feels enough to me. We can add more later!

Tests/Components Tests/IssuerList/IssuerListComponentTests.swift
erenbesel: Great tip, done!

atmamont (approved):
👍

goergisn (approved):

PR #1594, opened 2024-03-27
Removing all usage of country flags

goergisn:
<changed>

- Flag icons are removed from the UI.
- The text  **Country**  is now replaced with **Country/Region**.

</changed>

## Summary
- [X] Removing all usage of country flags
- [x] Updating `Country` to `Country/Region`


nauaros (dismissed):

erenbesel (dismissed):

erenbesel (dismissed):

atmamont (dismissed):
Beautiful!

Maybe an idea for the future - a separate PR targeting a feature branch with localization updates? So we don't need to scroll to find the actual changes.

neelSharma12 (dismissed):

PR #1597, opened 2024-03-27
Use swiftformat command instead of a bash script to format the project

atmamont:
We aim to provide the same results for `swiftformat` running locally and on CI

This PR:
- removes internal format.sh script usage and relies on `swiftformat` command itself that is running against project folder. By default it uses `.swiftformat.yml` and this is exact behavior that helps us to align with CI.
- splits Lint and Format build phase script into two: SwiftLint and SwiftFormat. It could make CI debug easier
- removes linting and formatting steps from all targets but `AdyenUIHost` (Demo app) and `AdyenSwiftUIHost` and appliy formatting to the whole project.


nauaros (dismissed):

goergisn (approved):

goergisn (dismissed):

erenbesel (dismissed):

goergisn (dismissed):
💚

erenbesel (approved):
Every file updated as well, nice!

goergisn (approved):

PR #1610, opened 2024-04-04
Create unit tests target and extract first batch of pure unit tests

atmamont:
This change aims to improve our tests organization and structure.
So far we have multiple tests targets mostly split by their host app - UIKit or SwiftUI-based.

Unit tests do not need to be based on any specific app and run in the full UI mode unless we want to test either an app business logic or test **integrations** which is not the case for pure unit tests. Having UIKit imported in the unit test immediately helps to mark the test case as something belonging to integration tests.

To keep this change digestible I deliberately limited the scope. I made every change in a separate commit so reviewing by commits is highly recommended.

In the future PRs I will try finding more unit tests and moving them to the new target. The final goal is to identify in the end what do we test via:
- unit tests
- integration tests
- UI tests
- snapshot tests

Then it will become easier to identify our weak points and gaps we want to close.
If you see something is missing (like code coverage settings), please let me know.


goergisn:
Love the idea of having 4 targets and naming them accordingly.
Looks good so far but tests are failing.

atmamont:
> Love the idea of having 4 targets and naming them accordingly. Looks good so far but tests are failing.

Messed up with target settings for one file, on it.

goergisn (approved):

goergisn:
Nice one!

nauaros (approved):

PR #1617, opened 2024-04-05
Focus/unfocus events for card component

erenbesel:
Adds 2 closures to `FormTextItem` that gets triggered when the view of the item begins/ends editing through which we send the relevant analytic events.


AdyenCard/Components/Card/CardComponent.swift
```diff
@@ -192,11 +192,25 @@ public class CardComponent: PresentableComponent,
...
+        }
+        
         return formViewController
     }()
     
     private let panThrottler = Throttler(minimumDelay: CardComponent.Constant.secondsThrottlingDelay)
     private let binThrottler = Throttler(minimumDelay: CardComponent.Constant.secondsThrottlingDelay)
+    
+    private func sendInfoEvent(of type: AnalyticsEventInfo.InfoType, target: AnalyticsEventTarget) {
+        var infoEvent = AnalyticsEventInfo(
```
atmamont, line 207: I would suggest at least extracting the instantiation of events to a separate entity. It could be just an extension for AnalyticsEvent with a set of static functions. It would make any refactoring easier in the future when we for example would shift from three distinct AnalyticEvent types to a single one (assuming this happens sometime in the future). No action needed, just a suggestion for the future improvements.

AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
```diff
@@ -9,7 +9,7 @@ import UIKit
 
 extension CardViewController {
 
-    internal struct ItemsProvider {
+    internal class ItemsProvider {
```
atmamont, line 12: Can this be `final`?

AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
```diff
@@ -159,6 +208,13 @@ extension CardViewController {
             additionalItem.identifier = ViewIdentifierBuilder.build(scopeInstance: scope, postfix: "additionalAuthPasswordItem")
             additionalItem.keyboardType = .numberPad
             additionalItem.isVisible = configuration.koreanAuthenticationMode == .show
+            
+            additionalItem.onDidBeginEditing = { [weak self] in
```
atmamont, line 212: A lot of repetitions, may I suggest extracting these two calls into a helper method? Something similar to what you've done in the tests.

`func setupFocusEvents(for control: UIControl, target: AnalyticsEventTarget)`

Tests/Card Tests/CardComponentEventTests.swift
```diff
@@ -0,0 +1,172 @@
...
+        XCTAssertEqual(secondInfoEvent.target, target)
+    }
+    
+    private func makeSUT(with configuration: CardComponent.Configuration = .init(), analyticsProviderMock: AnalyticsProviderMock) -> CardComponent {
+        let context = Dummy.context(with: analyticsProviderMock)
+        let cardComponent = CardComponent(paymentMethod: method,
+                                          context: context,
+                                          configuration: configuration)
+        
+        setupRootViewController(cardComponent.viewController)
```
atmamont, line 167: The more I read tests the more I feel we can just use 
``` 
let vc = cardComponent.viewController
vc.loadViewIfNeeded()
```
This is how we can prevent this test to be an integration test and convert it to a pure unit test that can be executed in a target that has no host app.

atmamont (dismissed):
Great stuff Eren! 👍

AdyenCard/Components/Card/CardComponent.swift
nauaros: I like the idea

AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
```diff
@@ -91,6 +94,14 @@ extension CardViewController {
...
             let zipCodeItem = FormPostalCodeItem(style: formStyle.textField, localizationParameters: localizationParameters)
             zipCodeItem.identifier = ViewIdentifierBuilder.build(scopeInstance: scope, postfix: "postalCodeItem")
+            
+            zipCodeItem.onDidBeginEditing = { [weak self] in
+                self?.triggerInfoEvent(of: .focus, target: .addressPostalCode)
+            }
+            zipCodeItem.onDidEndEditing = { [weak self] in
+                self?.triggerInfoEvent(of: .unfocus, target: .addressPostalCode)
+            }
+
```
nauaros, line 104: If these events only apply to `FormTextItem`. Having a setup method like this one could be useful:

```swift
func setupEditingEvents(for item: FormTextItem)  {
            item.onDidBeginEditing = { [weak self] in
                self?.triggerInfoEvent(of: .focus, target: .addressPostalCode)
            }
            item.onDidEndEditing = { [weak self] in
                self?.triggerInfoEvent(of: .unfocus, target: .addressPostalCode)
            }
}
```
Of course, only if it's possible

nauaros (dismissed):

atmamont (dismissed):

nauaros (dismissed):

atmamont:
Calling in @goergisn 
He struggled with this test as well, maybe he can add more info.

atmamont (dismissed):

atmamont (approved):

nauaros (approved):

PR #1622, opened 2024-04-10
Fix low coverage and introduce new CI test plan

atmamont:
The main goal of this change is to fix dropped coverage number that was caused by main coverage report being overwritten with snapshot tests report which has much lower coverage.

This change aims to bring some order to the existing test targets, schemes and test plans.
It renames existing targets and testplans so we now have:
- UnitTests
- IntegrationUIKItTests
- IntegrationSwiftUITests
- SnapshotTests
Apart from that it adds new non-default "CI" test plan to the main AdyenUIHost target. This test plan is now used on CI to run tests and collect coverage. 

image

This new test plan includes:
- unit tests
- integration tests
- snapshot tests for English/US locale

We need to decide how often we want to run snapshot tests for multiple locales (we now have two configurations available: English/US and German/DE). My suggestion is to run it once the PR branch is merged to develop. 

The other way would be running two or more test plans separately in one Github job and then combining coverage reports from all of them and passing them to Sonar using the `-Dsonar.coverageReportPaths=report1,report2,report3` command line argument.


erenbesel (approved):

.github/workflows/pr_scan.yml
```diff
@@ -47,20 +47,14 @@ jobs:
         params: '${{env.params}}'
         scheme: 'AdyenUIHost'
         
-    - name: 🧪 Unit Tests
+    - name: 🧪 Tests
       run: |
-        xcodebuild test -project "${{env.project}}" -scheme "${scheme}" ${params} -destination "${{env.destination}}" | xcpretty --utf --color && exit ${PIPESTATUS[0]}
+        xcodebuild test -project "${{env.project}}" -scheme "${scheme}" -testPlan "${testPlan}" ${params} -destination "${{env.destination}}" | xcpretty --utf --color && exit ${PIPESTATUS[0]}
```
goergisn, line 52: Nice! Didn't know about that parameter

Adyen.xcodeproj/xcshareddata/xcschemes/AdyenSwiftUIHost.xcscheme
```diff
@@ -40,8 +40,8 @@
             <BuildableReference
                BuildableIdentifier = "primary"
                BlueprintIdentifier = "F9615FAE252DDA4800FA8EDD"
-               BuildableName = "AdyenSwiftUITests.xctest"
-               BlueprintName = "AdyenSwiftUITests"
+               BuildableName = "IntegrationSwiftUITests.xctest"
```
goergisn, line 43: As far as I know we don't actually execute these tests as they were so unreliable - maybe a good idea to remove them

The only reason we have the SwiftUI target as of right now is to show how to integrate with SiftUI

goergisn (approved):

Adyen.xcodeproj/xcshareddata/xcschemes/AdyenSwiftUIHost.xcscheme
atmamont: Good to know. I'll play around with these tests and deleted them in the next PR if there is no hope :)

PR #1627, opened 2024-04-12
Experiment: Trying to split xcresult files from multiple test runs so slather picks up all of them

atmamont:


atmamont:
Slather doesn't support multiple test runs for multiple targets. It always uses the latest test report and this means this tool does not meet our requirement anymore.

PR #1634, opened 2024-04-17
More validators added for card component analytics

erenbesel:
Expiry date and security code validators conforming to StatusValidator for improved validation (and analytics)


AdyenCard/Validators/CardSecurityCodeValidator.swift
```diff
@@ -42,4 +42,20 @@ public final class CardSecurityCodeValidator: NumericStringValidator, AdyenObser
         minimumLength = length
     }
     
+    public func validate(_ value: String) -> ValidationStatus {
+        if super.isValid(value) {
```
goergisn, line 46: Shouldn't this move a bit down the tree?
I would assume the `isEmpty` check should trigger if the value is empty - right now the parent could override it

AdyenCard/Validators/CardExpiryDateValidator.swift
```diff
@@ -11,7 +11,15 @@ import Foundation
...
 /// Validation will fail when the format is invalid or the date is in the past.
 @_spi(AdyenInternal)
-public final class CardExpiryDateValidator: Validator {
+public final class CardExpiryDateValidator: StatusValidator {
+    
+    private enum Constants {
+        static let maxYearsDifference = 30
+        static let firstMonth = 1
+        static let lastMonth = 12
+        static let minMonthDifference = -3
```
goergisn, line 20: Could we name this more descriptive?

erenbesel: The month one? any suggestions?

AdyenCard/Validators/CardSecurityCodeValidator.swift
erenbesel: You may have a point but that would be changing existing logic. Previously the `isValid` function here was calling super first then doing its own checks.

erenbesel: Yeah but if we change that order it would be changing existing logic. As in currently this class is only calling `super.isValid` for validation so this way we are keeping that logic first then adding the new part

goergisn (approved):

nauaros (approved):

PR #1636, opened 2024-04-19
More validators added for analytics

erenbesel:
Added validators for
- KCP fields
- Brazil SSN
- Postal code


Adyen/Validators/AddressAnalyticsValidationError.swift
```diff
@@ -0,0 +1,35 @@
...
+    /// Postal code is not filled.
+    case postalCodeEmpty
+    
+    /// Postal code is partially filled.
+    case postalCodePartial
+    
+    public var analyticsErrorCode: Int {
+        switch self {
+        case .postalCodeEmpty:
+            return 934
```
atmamont: Do we have a supported list of error codes from the backend? Do you think it might be easier to maintain it and keep all of them together in a separate namespace?

AdyenCard/Validators/CardKCPValidators.swift
```diff
@@ -0,0 +1,74 @@
...
+        static let exactLength = 2
+    }
+    
+    public init() {
+        super.init(exactLength: Constants.exactLength)
+    }
+    
+    public func validate(_ value: String) -> Adyen.ValidationStatus {
+        if super.isValid(value) {
+            return .valid
```
atmamont, line 60: Should these super checks with early returns come first? I imagine this can pass so we return early and we never get a chance to proceed to the real check below. I think a couple of tests wouldn't hurt to cover such scenarios (of course if my comment makes any sense)

AdyenCard/Validators/CardNumberValidator.swift
```diff
@@ -43,14 +43,14 @@ public final class CardNumberValidator: StatusValidator {
...
-        if !isEnteredBrandSupported {
-            return .invalid(CardValidationError.cardUnsupported)
-        }
-        
         if value.isEmpty {
             return .invalid(CardValidationError.cardNumberEmpty)
         }
         
+        if !isEnteredBrandSupported {
+            return .invalid(CardValidationError.cardUnsupported)
```
atmamont, line 51: What do you think of adding "no_return" swiftformat rule?

Tests/IntegrationTests/Card Tests/Validators/CardHolderNameValidatorTests.swift
```diff
@@ -0,0 +1,40 @@
...
+@testable @_spi(AdyenInternal) import AdyenCard
+import XCTest
+
+final class CardHolderNameValidatorTests: XCTestCase {
+
+    func testValidNames() {
+        let validator = CardHolderNameValidator()
+        
+        XCTAssertTrue(validator.isValid("abc"))
+        XCTAssertTrue(validator.isValid("test name 123 ,.;[]'."))
```
atmamont, line 17: This is very interesting rule for the name validation :) Are there any characters that are not allowed?

atmamont (commented):
Looks good! (comments are questions mostly)

Adyen/Validators/BrazilSocialSecurityNumberValidator.swift
```diff
@@ -0,0 +1,71 @@
...
+    public let firstValidator: Validator
+    
+    public let secondValidator: Validator
+    
+    public init() {
+        self.firstValidator = NumericStringValidator(exactLength: Constants.firstMaxLength)
+        self.secondValidator = NumericStringValidator(exactLength: Constants.secondMaxLength)
+    }
+    
+    public func validate(_ value: String) -> Adyen.ValidationStatus {
```
goergisn: looks like autocompletion - the Adyen namespace should not be needed here right?
```suggestion
    public func validate(_ value: String) -> ValidationStatus {
```

Adyen/Validators/BrazilSocialSecurityNumberValidator.swift
```diff
@@ -0,0 +1,71 @@
...
+
+@_spi(AdyenInternal)
+public final class BrazilSocialSecurityNumberValidator: CombinedValidator, StatusValidator {
+    
+    private enum Constants {
+        static let firstMaxLength = 11
+        static let secondMaxLength = 14
+    }
+    
+    public let firstValidator: Validator
```
goergisn, line 17: Maybe not for here but would be nice to be able to have a list of validators and an extension on `[Validator]` that allow use of || and && on the validators in a scalable way.

AdyenCard/Validators/CardHolderNameValidator.swift
```diff
@@ -0,0 +1,28 @@
...
+    
+    public func isValid(_ value: String) -> Bool {
+        validate(value).isValid
+    }
+    
+    public func maximumLength(for value: String) -> Int {
+        .max
+    }
+    
+    public func validate(_ value: String) -> Adyen.ValidationStatus {
```
goergisn:
```suggestion
    public func validate(_ value: String) -> ValidationStatus {
```

AdyenCard/Validators/CardKCPValidators.swift
```diff
@@ -0,0 +1,74 @@
...
+    public let firstValidator: Validator
+    
+    public let secondValidator: Validator
+    
+    public init() {
+        self.firstValidator = NumericStringValidator(exactLength: Constants.exactLength)
+        self.secondValidator = DateValidator(format: DateValidator.Format.kcpFormat)
+    }
+    
+    public func validate(_ value: String) -> Adyen.ValidationStatus {
```
goergisn:
```suggestion
    public func validate(_ value: String) -> ValidationStatus {
```

AdyenCard/Validators/CardKCPValidators.swift
```diff
@@ -0,0 +1,74 @@
...
+    
+    private enum Constants {
+        static let exactLength = 2
+    }
+    
+    public init() {
+        super.init(exactLength: Constants.exactLength)
+    }
+    
+    public func validate(_ value: String) -> Adyen.ValidationStatus {
```
goergisn:
```suggestion
    public func validate(_ value: String) -> ValidationStatus {
```

AdyenCard/Validators/CardKCPValidators.swift
erenbesel: Since we subclass a certain validator for these cases, in order to preserve its existing logic, we need to call super first. If it passes it means it passes the existing validation logic which is not changed. If it doesn't then we check why it didn't.

If we change the validation logic of said classes then we can remove/change the place of calling super.

Tests/IntegrationTests/Card Tests/Validators/CardHolderNameValidatorTests.swift
erenbesel: Apparently not :)

Adyen/Validators/BrazilSocialSecurityNumberValidator.swift
erenbesel: Good catch, updated!

goergisn (dismissed):

goergisn (dismissed):

AdyenCard/Validators/CardKCPValidators.swift
atmamont: Ooof, I imagine this can lead to potential developer errors like the situation I described. Thanks for explaining!

Adyen/Validators/AddressAnalyticsValidationError.swift
atmamont: Any thoughts on this? Now it feels all codes are evenly distributed across multiple files and there is no nice overview of them

erenbesel: Thanks! Updated with `AnalyticsConstants` enum

atmamont (approved):

goergisn (approved):

PR #1645, opened 2024-04-24
Convert more UI-based tests to unit tests

atmamont:


AdyenActions/UI/View Controllers/Voucher/VoucherView.swift
```diff
@@ -247,16 +247,17 @@ internal final class VoucherView: UIView, Localizable {
...
+        super.didMoveToSuperview()
         updateLogo()
     }
     
     private func updateLogo() {
-        if window != nil {
-            imageLoadingTask = logo.load(url: model.logoUrl, using: model.imageLoader)
-        } else {
+        if superview == nil {
+            imageLoadingTask?.cancel()
```
goergisn: Maybe you want to do that in the didSet of the imageLoadingTask? So the previous task always gets cancelled first

goergisn (dismissed):

AdyenActions/UI/View Controllers/Voucher/VoucherView.swift
atmamont: Good point, will do!

goergisn: Good catch btw - was an oversight on my side 🙈

erenbesel (approved):

neelSharma12 (approved):

PR #1660, opened 2024-04-30
Change UPI tabs text

neelSharma12:


AdyenComponents/UPI/UPIComponent.swift
```diff
@@ -94,7 +94,8 @@ public final class UPIComponent: PaymentComponent,
 
     /// The upi selection segment control item to choose the upi flow.
     internal lazy var upiFlowSelectionItem: FormSegmentedControlItem = {
-        let item = FormSegmentedControlItem(items: ["VPA", "QR code"], style: configuration.style.segmentedControlStyle,
+        let item = FormSegmentedControlItem(items: ["Pay by any UPI app", "Other UPI options"],
```
goergisn, line 97: Is this not localized?

neelSharma12: No, I have created the job on smartling for this but there are some issues happening on smartling so once the job is completed I will replace these strings. FYI, there is a separate task for this in subtasks list :)

erenbesel (approved):

AdyenComponents/UPI/UPIComponent.swift
goergisn: It's probably best to wait for the translations as this breaks the tests (probably also snapshot tests)

PR #1664, opened 2024-05-03
Only checking for .active on the `BrowserComponent`

goergisn:
## Summary
- "Reverting" to only checking if the app is `.active` instead of `.active || .inactive`

### Fixed
<fixed>
Fixed an issue that incorrectly cancelled successful redirect payments in some cases.
</fixed>


nauaros (dismissed):

descorp (dismissed):

erenbesel (approved):

atmamont (dismissed):
Noe of our tests are failing so I assume there are no tests covering this flow?
Is it possible to add them?

goergisn:
> Noe of our tests are failing so I assume there are no tests covering this flow?
> 
> Is it possible to add them?

Do you have a suggestion on how to change the application state of UIApplication so this could be tested?

atmamont:
> > Noe of our tests are failing so I assume there are no tests covering this flow?
> > Is it possible to add them?
> 
> Do you have a suggestion on how to change the application state of UIApplication so this could be tested?

The first approach that comes to my mind is:
- extracting `checkIfExternalAppDidOpen` into a separate entity that takes `UIApplication` as dependency
- hiding implementation details behind a new protocol like `ExtenalAppDetecting` with `checkIfExternalAppDidOpen` method (or maybe a simple closure dependency would do?)
- replacing direct `UIApplication.checkIfExternalAppDidOpen` calls with the new dependency calls

Let me know if this works, we can think about other approaches

erenbesel (dismissed):

nauaros (approved):

atmamont (approved):

PR #1673, opened 2024-05-10
Docs: move list of available components to the page bottom

atmamont:
This change aims to improve consistency of our documentation and decrease time to find a solution for new developers not familiar with the SDK.

I noticed we offer DropIn integration example right away on the main GitHub page. However, when it comes to Components, we direct developers to the docs pages and the main one starts with endless list of components that make little sense to someone who is looking for a quick example of how to start using our SDK.

Moving this list to the bottom of the page increases code samples visibility and hopefully will reduce search time for new developers.


nauaros (approved):

nauaros:
Good idea!

descorp (approved):

PR #1674, opened 2024-05-13
Implement "UPI_Intent" flow UI

neelSharma12:


AdyenComponents/UPI/UPIComponent.swift
```diff
@@ -111,14 +120,68 @@ public final class UPIComponent: PaymentComponent,
...
         item.validationFailureMessage = localizedString(.UPIVpaValidationMessage, configuration.localizationParameters)
         item.identifier = ViewIdentifierBuilder.build(scopeInstance: self,
                                                       postfix: ViewIdentifier.virtualPaymentAddressInputItem)
         return item
     }()
 
+    /// The UPI app list item.
+    internal lazy var upiAppsList: [ListItem] = {
+        var upiAppslist = [ListItem]()
+        _ = upiPaymentMethod.apps.map { app -> ListItem in
```
goergisn:
```suggestion
       upiPaymentMethod.apps.forEach { app in
```

Adyen/UI/List/ListItem.swift
```diff
@@ -63,7 +69,9 @@ public class ListItem: FormItem {
         style: ListItemStyle = ListItemStyle(),
         identifier: String? = nil,
         accessibilityLabel: String? = nil,
-        selectionHandler: (() -> Void)? = nil
+        selectionHandler: (() -> Void)? = nil,
+        isSelectable: Bool? = nil,
```
goergisn: Is this property needed? We could decide whether or not it's selectable by checking if a selectionHandler is provided maybe?

Adyen/UI/List/ListItemView.swift
```diff
@@ -49,9 +49,14 @@ public final class ListItemView: UIView, AnyFormItemView {
                 subtitleLabel.adyen.apply(style.subtitle)
                 trailingTextLabel.adyen.apply(style.trailingText)
             }
+            if let isSelectable = item?.isSelectable, isSelectable {
+                self.addGestureRecognizer(itemGestureRecognizer)
```
goergisn: Do you think there is a way to use the `ListCell` instead of adding a gesture recognizer (thinking about the highlighting + accessibility that gets lost by using a gesture recognizer)?

Currently the `ListItem` is intended to only be used by the `ListViewController` and the full functionality might not be needed in this case.
Maybe even using a button that looks like a cell could be an option.

AdyenActions/Components/Await/AwaitComponent.swift
```diff
@@ -73,34 +75,52 @@ public final class AwaitComponent: ActionComponent, Cancellable {
...
+
+    public func didCancel() {
+        paymentMethodSpecificPollingComponent?.didCancel()
+    }
+
     /// Handles await action.
     ///
     /// - Parameter action: The await action object.
-    public func handle(_ action: AwaitAction) {
+    public func handleAwait(_ action: AwaitAction) {
```
erenbesel: This is breaking, we can't change this.

AdyenActions/Actions/AwaitAction.swift
```diff
@@ -27,15 +30,22 @@ public struct AwaitAction: PaymentDataAware, Decodable {
     
     /// The server-generated payment data that should be submitted to the `/payments/details` endpoint.
     public let paymentData: String
-    
-    /// Initializes a redirect action.
+
+    /// The URL to which to redirect the user.
+    public let url: URL?
```
erenbesel: Maybe redirectUrl to be more specific? Not sure

AdyenActions/AdyenActionComponent.swift
```diff
@@ -201,7 +201,7 @@ public final class AdyenActionComponent: ActionComponent, ActionHandlingComponen
         component.presentationDelegate = presentationDelegate
         component.configuration.localizationParameters = configuration.localizationParameters
         
-        component.handle(action)
+        action.url == nil ? component.handleAwait(action) : component.handleRedirectableAwait(action)
```
erenbesel: Isn't there supposed to be a new `action` enum case which the action component will call the relevant handle action call?

AdyenActions/Actions/AwaitAction.swift
neelSharma12: I kept the name url which this is what coming from the backend and I did not want to add coding values.

AdyenActions/AdyenActionComponent.swift
neelSharma12: No, this is what we decided we will implement.

AdyenActions/Components/Await/AwaitComponent.swift
neelSharma12: yes, you are right. I am reverting this.

AdyenActions/Components/Await/AwaitComponent.swift
```diff
@@ -73,34 +75,52 @@ public final class AwaitComponent: ActionComponent, Cancellable {
     
     private let componentName = "await"
     
+    /// Handles redirect await action.
+    ///
+    /// - Parameter action: The await action object.
+    public func handleRedirectableAwait(_ action: AwaitAction) {
```
goergisn: If this function is also just taking an `AwaitAction` the logic to decide what to do could actually happen in the `AwaitComponent` instead of the `AdyenActionComponent` where you do `action.url == nil ? component.handle(action) : component.handleRedirectableAwait(action)` 

What do you think? 
This way we also don't have to add another function.

AdyenComponents/UPI/UPIComponent.swift
```diff
@@ -111,14 +120,70 @@ public final class UPIComponent: PaymentComponent,
...
         item.validationFailureMessage = localizedString(.UPIVpaValidationMessage, configuration.localizationParameters)
         item.identifier = ViewIdentifierBuilder.build(scopeInstance: self,
                                                       postfix: ViewIdentifier.virtualPaymentAddressInputItem)
         return item
     }()
 
+    /// The UPI app list item.
+    internal lazy var upiAppsList: [SelectableFormItem] = {
+        var upiAppslist = [SelectableFormItem]()
+        _ = upiPaymentMethod.apps.map { app -> SelectableFormItem in
```
goergisn: Is there a reason you don't use the result of the map`

If you don't want to use it you could also use:
```suggestion
        upiPaymentMethod.apps.forEach { app in
```

goergisn: It would maybe be also nice to encapsulate the creation of a `SelectableFormItem` from an app into a function so the lazy var content is smaller and easier to read

AdyenComponents/UPI/UPIComponent.swift
```diff
@@ -215,6 +304,22 @@ public final class UPIComponent: PaymentComponent,
         }
     }
 
+    private func changeUPIAppsListVisiblity(shouldHide isHidden: Bool = false) {
+        if isHidden {
```
goergisn, line 309: This is very nitpicky so feel free to ignore but what about simplifying the code to the following
```
guard !upiPaymentMethod.apps.isEmpty else { return }

upiAppsList.forEach {
    $0.isHidden.wrappedValue = isHidden
}
```

AdyenComponents/UPI/UPIComponent.swift
neelSharma12: eab7d68b538ab19f294c4b62fcffe59a625ae3d7

Adyen/UI/Form/Items/SelectableFormItem/SelectableFormItemView.swift
```diff
@@ -0,0 +1,164 @@
...
+            self.checkmarkImageView.isHidden = !isSelected
+        }
+    }
+
+    override public func didMoveToWindow() {
+        super.didMoveToWindow()
+        updateIcon()
+    }
+
+    override public func layoutSubviews() {
```
atmamont, line 127: What is the reason we style imageview every time in `layoutSubviews`? It should be sufficient to do in once when the view is created.

Adyen/UI/Form/Items/SelectableFormItem/SelectableFormItemView.swift
```diff
@@ -0,0 +1,164 @@
...
+        imageView.clipsToBounds = style.imageStyle.clipsToBounds
+        imageView.layer.borderWidth = style.imageStyle.borderWidth
+        imageView.layer.borderColor = style.imageStyle.borderColor?.cgColor
+    }
+
+    private func updateIcon() {
+        if let iconUrl = item.imageUrl, window != nil {
+            imageLoadingTask = imageView.load(url: iconUrl, using: imageLoader)
+        } else {
+            imageLoadingTask = nil
```
atmamont, line 143: Should we cancel loading first?

AdyenComponents/UPI/UPIComponent.swift
```diff
@@ -180,10 +257,20 @@ public final class UPIComponent: PaymentComponent,
         formViewController.view.isUserInteractionEnabled = false
 
         switch UPIFlowType(rawValue: currentSelectedIndex) {
-        case .vpa:
-            let details = UPIComponentDetails(type: Constants.upiCollect,
-                                              virtualPaymentAddress: virtualPaymentAddressItem.value)
+        case .upiApps:
+            let details: UPIComponentDetails
+            let flowType = currentSelectedItem?.identifier == "UPI/VPA" ? Constants.upiCollect : Constants.upiIntent
```
atmamont: "UPI/VPA" -> `Constants.vpaFlowIdentifier`

Adyen/UI/Form/Items/SelectableFormItem/SelectableFormItemView.swift
neelSharma12: This is done in case there is orientation change.

AdyenComponents/UPI/UPIComponent.swift
neelSharma12: good catch. 9d2651333e7dc7e7e43d2d13030010979002b1a9

Adyen/UI/Form/Items/SelectableFormItem/SelectableFormItemView.swift
neelSharma12: edd71e566cf4d9d230b1c2b66769351442f6f61b

Adyen/UI/Form/Items/SelectableFormItem/SelectableFormItemView.swift
atmamont: I see! Does orientation change affect imageView corner radius?

AdyenComponents/UPI/UPIComponent.swift
descorp: I think this should be addressed.

AdyenComponents/UPI/UPIComponent.swift
```diff
@@ -111,14 +118,47 @@ public final class UPIComponent: PaymentComponent,
     /// The  virtual payment address text input item.
     internal lazy var virtualPaymentAddressItem: FormTextInputItem = {
         let item = FormTextInputItem(style: configuration.style.textField)
-        item.title = Constants.virtualPaymentAddress
+        // TODO: Add localised string
+        item.title = "Enter UPI ID / VPA"
```
descorp, line 122: You should create a localization key with english only and use it instead, so when translation is ready - it will automatically  pick it ou.

Adyen/UI/Form/Items/SelectableFormItem/SelectableFormItemView.swift
neelSharma12: Sometimes the corner radius can be distorted on orientation change.

AdyenComponents/UPI/UPIComponent.swift
neelSharma12: They are. I am going to add them in separate PR today.

AdyenComponents/UPI/UPIComponent.swift
descorp: As a sidenote:
We use `!upiPaymentMethod.apps.isEmpty` a lot in the code. Maybe we can create a meaningful property variable for this? Eg. "hasApps"

neelSharma12: This I am going to check but in next PR.

AdyenComponents/UPI/UPIComponent.swift
```diff
@@ -111,14 +118,47 @@ public final class UPIComponent: PaymentComponent,
...
         item.validator = LengthValidator(minimumLength: 1)
         item.validationFailureMessage = localizedString(.UPIVpaValidationMessage, configuration.localizationParameters)
         item.identifier = ViewIdentifierBuilder.build(scopeInstance: self,
                                                       postfix: ViewIdentifier.virtualPaymentAddressInputItem)
         return item
     }()
 
+    /// The UPI app list item.
+    internal lazy var upiAppsList: [SelectableFormItem] = {
```
descorp, line 131:
```suggestion
    internal lazy var appsSelectableFormItems: [SelectableFormItem] = {
```

descorp, line 131: A comment is not needed if your variable has a descriptive name.

```suggestion
    internal lazy var upiAppsSelectableFormItem: [SelectableFormItem] = {
```

AdyenComponents/UPI/UPIComponent.swift
```diff
@@ -166,7 +232,16 @@ public final class UPIComponent: PaymentComponent,
...
+
+        if !upiPaymentMethod.apps.isEmpty {
+            for item in upiAppsList {
+                formViewController.append(item)
+            }
+            formViewController.append(virtualPaymentAddressItem)
+            virtualPaymentAddressItem.isHidden.wrappedValue = true
+        } else {
+            formViewController.append(virtualPaymentAddressItem)
+        }
```
descorp, line 244:
```suggestion
        if !upiPaymentMethod.apps.isEmpty {
            for item in upiAppsList {
                formViewController.append(item)
            }
            virtualPaymentAddressItem.isHidden.wrappedValue = true
        }
        formViewController.append(virtualPaymentAddressItem)
```

Do we need to add `virtualPaymentAddressItem` but hide it? Can we not add it? It can become visible later, right?

neelSharma12: this will add the `virtualPaymentAddressItem` to the top of the list which is not right according to the UI.

descorp: fair, then it can be after IF. Updated suggestion

PR #1701, opened 2024-05-30
Hide default pay button configuration

nauaros:
## Summary

- Define configuration property to show/hide default submit button.
- Add logic to add / remove scroll view from the view hierarchy based on the submit button configuration.
- Apply submit button configuration to all required components.
- Build advanced card component flow to showcase / QA submit button configuration. 

Demo


Adyen/UI/Form/FormViewController.swift
```diff
@@ -203,18 +230,36 @@ open class FormViewController: UIViewController, AdyenObserver, PreferredContent
 
     // MARK: - Private
 
-    private func addFormView() {
-        view.addSubview(formView)
-        view.backgroundColor = style.backgroundColor
-        formView.backgroundColor = style.backgroundColor
-        formView.adyen.anchor(inside: view.safeAreaLayoutGuide)
+    private func addSubviews() {
+        if !scrollDisabled {
```
erenbesel: Maybe better cognitive ease to start with `if isScrollingDisabled` instead of negative

Adyen/UI/Form/FormViewController.swift
```diff
@@ -203,18 +230,36 @@ open class FormViewController: UIViewController, AdyenObserver, PreferredContent
...
+        }
     }
 
-    private lazy var formView: FormView = {
-        let form = FormView()
-        form.translatesAutoresizingMaskIntoConstraints = false
-        return form
-    }()
+    private func setupLayout() {
+        if !scrollDisabled {
```
erenbesel: same here

Adyen/UI/Form/FormViewController.swift
```diff
@@ -11,42 +11,74 @@ import UIKit
...
+
+    private var keyboardObserver = KeyboardObserver()
+    private var scrollDisabled: Bool
+
+    // MARK: - Initializers
 
     /// Initializes the FormViewController.
     ///
     /// - Parameters:
     ///   - style: The `FormViewController` UI style.
```
erenbesel, line 54: missing new parameter

Adyen/UI/Form/FormViewController.swift
```diff
@@ -11,42 +11,74 @@ import UIKit
...
     /// Delegate to handle different viewController events.
     public weak var delegate: ViewControllerDelegate?
-    
-    internal lazy var keyboardObserver = KeyboardObserver()
 
-    // MARK: - Public
+    // MARK: - Private properties
+
+    private var keyboardObserver = KeyboardObserver()
+    private var scrollDisabled: Bool
```
erenbesel: recommend `isScrollingDisabled`

erenbesel (commented):
Great start, just small comments!

Adyen/UI/Form/FormViewController.swift
nauaros: Solved. Thank you!

AdyenCard/Components/Card/CardComponent.swift
```diff
@@ -140,7 +140,13 @@ public class CardComponent: PresentableComponent,
         
         cardViewController.stopLoading()
     }
-    
+
+    public func submit() {
+        guard !configuration.showSubmitButton else { return }
```
nauaros, line 145: I decided to prevent any kind of submit action if the default submit button is enabled. What do you think?

Adyen/UI/Form/FormViewController.swift
```diff
@@ -11,42 +11,75 @@ import UIKit
...
-        fileprivate static let keyboardBottomInset = "keyboardBottomInset"
-        fileprivate static let firstResponder = "firstResponder"
+    private enum Animations {
+        static let keyboardBottomInset = "keyboardBottomInset"
+        static let firstResponder = "firstResponder"
     }
 
+    // MARK: - UI elements
+
+    private let scrollView: UIScrollView = {
```
goergisn, line 21: this should be a lazy var so it only gets instantiated if it's actually needed

Adyen/Core/Components/Base/BasicComponentConfiguration.swift
```diff
@@ -21,18 +21,25 @@ public struct BasicComponentConfiguration: AnyBasicComponentConfiguration {
 
     /// The UI style of the component.
     public var style: FormComponentStyle
-    
+
+    /// A Boolean value that determines whether the payment button is displayed. Defaults to `true`.
+    public var showSubmitButton: Bool
```
atmamont: It is public and open for modifications, are both needed?

Adyen/Core/Components/StoredPaymentMethodComponent.swift
```diff
@@ -81,12 +81,19 @@ extension StoredPaymentMethodComponent {
     /// Configuration for Stored Payment type components.
     public struct Configuration: AnyBasicComponentConfiguration {
 
+        /// A Boolean value that determines whether the payment button is displayed. Defaults to `true`.
+        public var showSubmitButton: Bool
```
atmamont: This is second time we add `showSubmitButton` to the class inherited from `AnyBasicComponentConfiguration`. Can we add it to the root component?
If this is too specific for a parent class, maybe we could introduce a type in-between that would contain this property?

Adyen/UI/Form/FormView.swift
```diff
@@ -1,22 +1,29 @@
 //
-// Copyright (c) 2024 Adyen N.V.
+// Copyright (c) 2017 Adyen N.V.
```
atmamont: Feels like current date is broken :)

Adyen/UI/Form/FormView.swift
```diff
@@ -37,21 +44,25 @@ internal final class FormView: UIScrollView {
...
+
+    private func addSubviews() {
+        addSubview(stackView)
+    }
 
-    private func configureConstraints() {
-        stackView.adyen.anchor(inside: self)
-        stackView.widthAnchor.constraint(equalTo: widthAnchor).isActive = true
+    private func setupLayout() {
+        NSLayoutConstraint.activate([
```
atmamont, line 63: This is the same as it was before, right?
`stackView.adyen.anchor(inside: self)`

Adyen/UI/Form/FormViewController.swift
atmamont: I was thinking maybe we could use a composition instead. Like create a `FormScrollViewController` that has embedded `FormViewController` (inside the `UIScrollView`) so this class would be responsible **ONLY** for scroll view management and not mess up the existing code. But so far my code knowledge is limited so I'm just throwing a possible option.
So then we can define the view controller behavior upon it's instantiation choosing between `FormViewController` and `FormScrollViewController` thus keeping all components not coupled and not knowing anything about each other.
Extracting this to an assembly layer would also make it much easier to produce whatever kind we need in unit tests.

Adyen/UI/Form/FormViewController.swift
```diff
@@ -11,42 +11,76 @@ import UIKit
...
+    // MARK: - Initializers
 
     /// Initializes the FormViewController.
     ///
     /// - Parameters:
+    ///   - scrollEnabled: Boolean value that determines whether the form view contains a scroll view in its view hierarchy.
     ///   - style: The `FormViewController` UI style.
     ///   - localizationParameters: The localization parameters.
     public init(
+        scrollEnabled: Bool,
```
atmamont, line 57: Using composition would make this property redundant. We will extract this logic to an assembly layer.

atmamont (approved):

atmamont:
I will do manual testing on Monday.

Adyen/Core/Components/Base/BasicComponentConfiguration.swift
nauaros: That's actually a good point. I would make it private(set) but still have it publicly accessible to get

Adyen/Core/Components/StoredPaymentMethodComponent.swift
nauaros: `AnyBasicComponentConfiguration` is a root configuration protocol to which all configuration objects conform to. I initially started by defining it there, but later realize that it makes so components (e.g. `DropInComponent`) be aware of a property that's completely unrelated to them. We can add an abstraction for it, but decided to simply add it to the entities that actually make use of it.

Adyen/UI/Form/FormView.swift
nauaros: Sorryyy fixed!

Adyen/UI/Form/FormView.swift
nauaros: Yes, refactoring to use the helper method

Adyen/UI/Form/FormViewController.swift
nauaros: Yeah, the only drawback is the code duplication we may end up having (thinking out loud), but it should work.

atmamont: Let's have a quick whiteboarding session later on. My idea doesn't lead to any code duplication, at least in my mindmap! :)

atmamont:
Should this one have `new` label and a corresponding release notes? Some merchants would be really happy to get this feature.

Adyen/UI/Form/FormView.swift
```diff
@@ -37,21 +48,23 @@ internal final class FormView: UIScrollView {
...
-    }()
+    // MARK: - Private
+
+    private func setup() {
+        addSubviews()
+        setupLayout()
+    }
 
-    // MARK: - Layout
+    private func addSubviews() {
```
descorp, line 58: Do we need this method?

Adyen/UI/Form/FormView.swift
```diff
@@ -7,16 +7,27 @@
 import UIKit
 
 /// Displays a form for the user to enter details.
-internal final class FormView: UIScrollView {
+internal final class FormView: UIView {
```
descorp, line 10: Can we make `FormView` a `UIStackView` instead, so we don't need to bother about sub-views and constrains ?

Adyen/UI/Form/FormView.swift
nauaros: I like splitting into different methods when creating views. But yes, that line can go into any other method.

Adyen/UI/Form/FormView.swift
nauaros: I tried, but sadly we rely on the `minimalSize` of the stack view. Having `FormView` as a `UIStackView` causes an infinite loop when calculating its `intrinsicContentSize`. It's a great suggestion that I will definitely do once we get rid of `minimalSize` calculations across the codebase. Thanks Vlad!

Adyen/UI/Form/FormView.swift
erenbesel: I think it's different just because of the last line `lessThanOrEqualTo`

AdyenCard/Components/Card/CardComponent.swift
```diff
@@ -140,7 +140,13 @@ public class CardComponent: PresentableComponent,
         
         cardViewController.stopLoading()
     }
-    
+
+    public func submit() {
```
erenbesel, line 144: If this is the public function that we will expose to merchants to trigger on their button taps, let's add documentation and maybe a more fitting name.

AdyenCard/Components/Card/CardComponent.swift
descorp: Make sense :)

AdyenCard/Components/Card/CardComponent.swift
descorp, line 144: Is this only available for `CardComponent`? What about other components?

Maybe we should make 2 protocols:
- `SubmitButtonComponent`
- `SubmitButtonComponentConfiguration`

Then make default implementation:

```swift
protocol SubmitButtonComponent: PresentableComponent {
    public func submit()
    public func didSelectSubmitButton()
}

protocol SubmitButtonComponentConfiguration: BasicConfiguration {
    public var showSubmitButton: Bool
}

extension SubmitButtonComponent {
    public func submit() {
         guard 
           let submitConfiguration = (configuration as? SubmitButtonComponentCnfiguration), 
           !submitConfiguration.showSubmitButton
         else { return }
         
         didSelectSubmitButton()
    }
}
```

nauaros: Yes, I'm planning to do the submit logic for all components in a different PR that will be mostly about it

Adyen/UI/Form/FormView.swift
nauaros: Indeed!

AdyenCard/Components/Card/CardComponent.swift
nauaros: I will take this suggestion in consideration when addressing the submit logic in the next PR

Adyen/UI/Form/FormViewController.swift
nauaros: As discussed in a previous mob session, we will build the case to refactor internally how the drop-in is rendered. In a way that no scrolling should be present in the `FormViewController`.

PR #1708, opened 2024-06-06
Detect public api changes

goergisn:
## Summary
- Adds a github action to detect public api changes

### Possible improvements
- [ ] Show a status/emoji that quickly represents whether or not there are actual changes
- [ ] (Speed) Optimization (See TODOs)

### DIFF
--- a/Scripts/compare_public_interface_definition.swift
+++ b/Scripts/compare_public_interface_definition.swift
@@ -198,7 +198,7 @@ func compare() throws {
     
     groupedChanges.keys.sorted().forEach { key in
         fileContent +=  ["## \(key)"]
-        groupedChanges[key]?.forEach {
+        groupedChanges[key]?.sorted(by: { $0.changeDescription < $1.changeDescription }).forEach {
             fileContent +=  ["- \($0.changeType.icon) \($0.changeDescription)"]
         }
     }
--- a/Scripts/compare_public_interface_definition.swift
+++ b/Scripts/compare_public_interface_definition.swift
@@ -128,7 +128,7 @@ func recursiveCompare(element lhs: Element, to rhs: Element, oldFirst: Bool) -> 
     var changes = [Change]()
     
     if oldFirst, (lhs.printedName != rhs.printedName || lhs.spiGroupNames != rhs.spiGroupNames || lhs.children == rhs.children) {
-        changes += [.init(changeType: .change, parentName: lhs.parentPath, changeDescription: "\(lhs) ➡️  \(rhs)")]
+        changes += [.init(changeType: .change, parentName: lhs.parentPath, changeDescription: "`\(lhs)` ➡️  `\(rhs)`")]
     }
     
     changes += lhs.children?.flatMap { lhsElement in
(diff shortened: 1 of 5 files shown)


Scripts/compare_public_interface_definition.swift
```diff
@@ -0,0 +1,264 @@
...
+    
+    enum CodingKeys: String, CodingKey {
+        case printedName
+    }
+}
+
+class Element: Codable, Equatable, CustomDebugStringConvertible {
+    let kind: String
+    let name: String
+    let mangledName: String?
```
nauaros, line 28: What's `mangledName`?

goergisn: it's part of the sdk-dump json and looked like a more unique naming than name - could probably only rely on name but felt more unique 🤔 

Unfortunately I didn't find any documentation regarding the structure generated by the sdk dump option

nauaros: Ah I see

Scripts/compare_public_interface_definition.swift
```diff
@@ -0,0 +1,264 @@
...
+        // TODO: Show what exactly changed (name, spi, conformance, declAttributes, ...) as a bullet list maybe (add a `changeList` property to `Change`)
+        changes += [.init(changeType: .change, parentName: lhs.parentPath, changeDescription: "`\(lhs)` ➡️  `\(rhs)`")]
+    }
+    
+    if lhs.children == rhs.children {
+        return changes
+    }
+    
+    changes += lhs.children?.flatMap { lhsElement in
+        if let rhsChildForName = rhs.children?.first(where: { ($0.mangledName ?? $0.name) == (lhsElement.mangledName ?? lhsElement.name) }) {
```
nauaros: It would be nice to refactor this line. Not that easy to read. It can be a following iteration

nauaros (dismissed):

Scripts/compare_public_interface_definition.swift
goergisn: Good one - I love extensions you know 💚

nauaros (dismissed):

Scripts/generate_public_interface_definition.sh
```diff
@@ -0,0 +1,80 @@
...
+}
+
+function setupComparisonRepo() {
+    rm -rf $COMPARISON_VERSION_DIR_NAME
+    mkdir $COMPARISON_VERSION_DIR_NAME
+    cd $COMPARISON_VERSION_DIR_NAME
+    git clone -b $BRANCH $REPO
+    
+    cd adyen-ios
+    mv Adyen.xcodeproj Adyen.xcode_proj
```
atmamont: Let's add comment here for the reasoning?

atmamont (dismissed):
👏 This will be truly helpful!

Scripts/generate_public_interface_definition.sh
goergisn: Good one - added 👌

atmamont (approved):

PR #1711, opened 2024-06-12
UPI refinements

goergisn:
### `FormTextInputItemView`
- Report `false` of `canBecomeFirstResponder` when item is hidden
- Only `becomeFirstResponder` when the item is not hidden
### `FormViewController`
- Only assign first responder if the item is not hidden (So all items adopt the behavior automatically)
- Exposed `assignInitialFirstResponder` as `@_spi(AdyenInternal)` to allow the `UPIComponent` to re-assign focus to the VPA input field when switching flows 
  - ⚠️ Not sure about this one, maybe @erenbesel you have a better idea?)
### `FormSegmentedControlItem`
- Actually applying the provided text style
### `UPIComponent`
- Simplifying code


Adyen/Core/Payment Methods/UPIPaymentMethod.swift
```diff
@@ -14,7 +14,7 @@ public struct UPIPaymentMethod: PaymentMethod {
     public let name: String
 
     /// The available UPI apps.
-    public let apps: [Issuer]
+    public let apps: [Issuer]?
```
nauaros, line 17: Why is this optional now? Shouldn't there be at least one issuer in order to initialize the payment method?

Adyen/UI/Form/FormViewController.swift
```diff
@@ -228,9 +228,10 @@ open class FormViewController: UIViewController, AdyenObserver, PreferredContent
 
     // MARK: - Other
 
-    private func assignInitialFirstResponder() {
+    @_spi(AdyenInternal)
+    public func assignInitialFirstResponder() {
         guard view.isUserInteractionEnabled else { return }
-        let textItemView = itemManager.topLevelItemViews.first(where: { $0.canBecomeFirstResponder })
+        let textItemView = itemManager.topLevelItemViews.first(where: { $0.canBecomeFirstResponder && !$0.isHidden })
```
nauaros, line 233: nice. good catch

Adyen/Core/Payment Methods/UPIPaymentMethod.swift
goergisn: apps are optional - if the list is empty we show the VPA input field - so there is always one item in the list but the last item is always the VPA input field (not an app)

Tests/SnapshotTests/Components/UPIComponentUITests.swift
```diff
@@ -145,29 +156,30 @@ class UPIComponentUITests: XCTestCase {
...
-
-        wait(for: .milliseconds(300))
         
-        sut.currentSelectedIndex = 0
-        sut.currentSelectedItem = sut.upiAppsList.first
-        sut.currentSelectedItem?.isSelected = true
-        sut.virtualPaymentAddressItem.isHidden.wrappedValue = true
+        sut.upiAppsList.first?.selectionHandler?()
 
+        // TODO: Wait until the first item is selected
```
nauaros: Forgotten or gonna be followed up in a second PR?

nauaros (approved):

Tests/SnapshotTests/Components/UPIComponentUITests.swift
goergisn: Yesss forgotten - removed it 👌

Adyen/UI/Form/FormViewController.swift
```diff
@@ -228,9 +228,10 @@ open class FormViewController: UIViewController, AdyenObserver, PreferredContent
 
     // MARK: - Other
 
-    private func assignInitialFirstResponder() {
+    @_spi(AdyenInternal)
+    public func assignInitialFirstResponder() {
```
atmamont: why do we need to make it public, is this for tests?

atmamont (approved):

Adyen/UI/Form/FormViewController.swift
goergisn: Check out https://github.com/Adyen/adyen-ios/pull/1711#issue-2349160205

It's exposed to re-assign focus to the text field when switching back from the QR code tab

PR #1717, opened 2024-06-14
Api change detection improvements

goergisn:
## Summary
- Modifies the Package.swift to contain all targets
- Only builds the current project + comparison project once
- Automatically considers all available targets (instead of hardcoding them)


Scripts/detect_public_api_changes/compare.sh
```diff
@@ -0,0 +1,157 @@
...
+#
+#  buildProject . # If we're in the current project dir
+#  buildProject ../.. # If we're in the comparison project dir
+function buildProject() {
+
+    # Removing derived data if available
+    rm -rf .build
+
+    # We have to obscure the project file so `xcodebuild` uses the Package.swift to build the module
+    mv Adyen.xcodeproj Adyen.xcode_proj
```
atmamont, line 60: I now think maybe it's better to use underscore in the eof? Wouldn't it be more natural?

Scripts/detect_public_api_changes/compare.sh
```diff
@@ -0,0 +1,157 @@
...
+    
+    # Copying the Package.swift so we can revert the change done by the next step
+    cp Package.swift Package.sw_ift
```
atmamont, line 63: and same here

Scripts/detect_public_api_changes/compare.sh
```diff
@@ -0,0 +1,157 @@
...
+# Main Execution
+# -------------------------------------------------------------------
+
+printRunInfo
+
+echo "↘️  Setting up comparison project"
+setupComparisonRepo
+
+# Move into the comparison project
+cd $COMPARISON_VERSION_DIR_NAME/adyen-ios
```
atmamont, line 140: Can the hardcoded name be avoided with flattening the dir structure after `git clone` so we don't rely on repo name whatsoever?
I'm just trying to think how we could extract this set of tools into a separate repo/github action so it can be easily reused by other teams.

Scripts/detect_public_api_changes/package_file_helper.swift
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env xcrun swift
```
atmamont, line 1: I suggest:
- adding default output listing all supported CLI parameters if none is provided

Scripts/detect_public_api_changes/package_file_helper.swift
```diff
@@ -0,0 +1,70 @@
...
+        
+        if let targetName = scanner.scanUpToString(nameEndTag) {
+            availableTargets.insert(targetName)
+        }
+    }
+    
+    return availableTargets.sorted()
+}
+
+/// Generates a library entry from the name and available target names to be inserted into the `Package.swift` file
```
atmamont, line 31: Nice!
However, for someone who has spent not enough time researching how the api digester works it might be a bit overwhelming to understand how every small step consolidates in the big picture.

atmamont (dismissed):
Niiice improvements! 👏 
All my comments are basically nitpicks to make scaling phase easier.
Please decide yourself if you want to address it now.

Scripts/detect_public_api_changes/compare.sh
goergisn: Yes, agree - in the [swift command-line-tool version](https://github.com/Adyen/adyen-ios/pull/1718) this workaround is not needed anymore so would keep it for now

Scripts/detect_public_api_changes/compare.sh
goergisn: Yes, was also thinking about reusability in the [next version](https://github.com/Adyen/adyen-ios/pull/1718)

Scripts/detect_public_api_changes/package_file_helper.swift
goergisn: I love these ideas :)
Already tackled that in the [next version](https://github.com/Adyen/adyen-ios/pull/1718)

Scripts/detect_public_api_changes/package_file_helper.swift
goergisn: Will write some proper documentation and other findings once/before the [command line version](https://github.com/Adyen/adyen-ios/pull/1718) is done

Scripts/detect_public_api_changes/package_file_helper.swift
atmamont: 🤣 I feel now I didn't spent my time wisely reviewing this one alone

Scripts/detect_public_api_changes/compare.sh
```diff
@@ -0,0 +1,157 @@
...
+    
+    # Copying the Package.swift so we can revert the change done by the next step
+    cp Package.swift Package.sw_ift
+    
+    # Modify the Package.swift file to generate a product/library that contains all targets
+    # so we only have to build it once and can use it to diff all modules
+    $1/Scripts/detect_public_api_changes/package_file_helper.swift Package.swift -add-consolidated-library $ALL_TARGETS_LIBRARY_NAME
+
+    xcodebuild -scheme $ALL_TARGETS_LIBRARY_NAME \
+    -sdk $SDK \
```
OscarSpruit, line 70: Just an FYI surrounding the use of an variable with quotes (f.e. `"$SDK"`) will make sure that when the variable is empty, contains spaces or special characters it doesn't break the script. In general it's better to use quotes, unless you intentionally don't need them

descorp (dismissed):

PR #1728, opened 2024-06-26
UPI Only handle main txvariant

goergisn:
## Summary
- all upi txvariants (`upi_qr`, `upi_collect` & `upi_intent`) except for the main `upi` are ignored by drop-in
- using `VPA` as title when no apps are provided

### DIFF
--- a/Adyen/Core/Payment Methods/Abstract/PaymentMethodType.swift
+++ b/Adyen/Core/Payment Methods/Abstract/PaymentMethodType.swift
@@ -23,19 +23,11 @@ public enum PaymentMethodType: RawRepresentable, Hashable, Codable {
     case applePay
     case payPal
     case bcmc
-    case bcmcMobileQR
     case bcmcMobile
-    case weChatMiniProgram
-    case weChatQR
     case qiwiWallet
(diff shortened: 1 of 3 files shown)


Adyen/Core/Payment Methods/Abstract/PaymentMethodType.swift
```diff
@@ -118,6 +123,9 @@ public enum PaymentMethodType: RawRepresentable, Hashable, Codable {
         case "mealVoucher_FR_natixis": self = .mealVoucherNatixis
         case "mealVoucher_FR_sodexo": self = .mealVoucherSodexo
         case "upi": self = .upi
+        case "upi_collect": self = .upiCollect
```
erenbesel, line 126: So these were decoded as `other(....)` until now and still was shown inside dropin?

goergisn: Correct, yes.
As per requirement we have to filter them out in drop-in now and only consider the main `upi` txvariant as valid.

erenbesel (approved):

PR #1746, opened 2024-07-17
Submit customization on payment components

nauaros:
<new>

## New

- Make submit button configurable based on the payment component's configuration for the following payment components:
  - `AbstractPersonalInformationComponent`
  - `GiftCardComponent`
  - `CardComponent`
  - `ACHDirectDebitComponent`
  - `OnlineBankingComponent`
  - `UPIComponent`
  - `BLIKComponent`
  - `SEPADirectDebitComponent`
  - `CashAppPayComponent`
  - `GiftCardComponent`
- Introduce `SubmitCustomizable` protocol. This protocol provides control over the payment process submission.
- In `UPIComponent`, the selected UPI flow is publicly exposed allowing merchants to change their UI (e.g. submit button) accordingly.
- Provide documentation for the new public entities.


Adyen/Core/Components/AbstractPersonalInformationComponent/AbstractPersonalInformationComponent.swift
```diff
@@ -270,3 +273,28 @@ extension AbstractPersonalInformationComponent: ViewControllerDelegate {
         sendDidLoadEvent()
     }
 }
+
+extension AbstractPersonalInformationComponent: SubmitCustomizable {
+
+    // MARK: - SubmitCustomizable
```
goergisn: Nitpick - shall we move before the extension?

Adyen/Core/Components/AbstractPersonalInformationComponent/AbstractPersonalInformationComponent.swift
```diff
@@ -270,3 +273,28 @@ extension AbstractPersonalInformationComponent: ViewControllerDelegate {
...
+
+    /**
```
goergisn: We are usually using the `/// ` syntax even for multi line comments

Adyen/Core/Core Protocols/SubmitCustomizable.swift
```diff
@@ -0,0 +1,25 @@
...
+// Copyright (c) 2024 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+/// A protocol that provides control over the submission process in a payment component.
+public protocol SubmitCustomizable {
+    /**
```
goergisn: Same here regarding the multi line documentation

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -429,3 +433,27 @@ extension GiftCardComponent: PartialPaymentComponent {}
 
 @_spi(AdyenInternal)
 extension GiftCardComponent: PublicKeyConsumer {}
+
+extension GiftCardComponent: SubmitCustomizable {
+
+    // MARK: - SubmitCustomizable
+
+    /**
```
goergisn: Could be removed because the protocol propagates it

AdyenComponents/UPI/UPIComponent.swift
```diff
@@ -69,7 +68,13 @@ public final class UPIComponent: PaymentComponent,
     
     internal private(set) var currentSelectedItemIdentifier: String?
 
-    internal private(set) var selectedUPIFlow: UPIFlowType = .upiApps
+    /// Represents the selected UPI (Unified Payments Interface) flow for the payment component.
+    /// Determines the specific UPI transaction process to follow.
+    ///
+    /// Possible values are defined by `UPIFlowType`:
```
goergisn: This comment doesn't seem to be necessary - the info should be available on the case documentation of the `UPIFlowType`

Adyen/Core/Core Protocols/SubmitCustomizable.swift
nauaros: Done

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
nauaros: Done

AdyenComponents/UPI/UPIComponent.swift
nauaros: Done

AdyenCashAppPay/CashAppPayComponent.swift
```diff
@@ -255,3 +255,16 @@ extension CashAppPayComponent: TrackableComponent {}
...
 @_spi(AdyenInternal)
 extension CashAppPayComponent: ViewControllerDelegate {}
+
+
+// MARK: - SubmitCustomizable
+
+@available(iOS 13.0, *)
+extension CashAppPayComponent: SubmitCustomizable {
+
+    public func submit() {
```
erenbesel, line 266: if we tied configuration property of the components to a protocol as well we could have done this at one place

AdyenComponents/OnlineBanking/OnlineBankingComponent.swift
```diff
@@ -134,9 +134,12 @@ public final class OnlineBankingComponent: PaymentComponent,
...
         formViewController.append(FormSpacerItem(numberOfSpaces: 4))
         formViewController.append(termsAndConditionsLabelItem.addingDefaultMargins())
-        formViewController.append(FormSpacerItem())
-        formViewController.append(continueButton)
-        formViewController.append(FormSpacerItem(numberOfSpaces: 2))
+
+        if configuration.showSubmitButton {
+            formViewController.append(FormSpacerItem())
+            formViewController.append(continueButton)
+            formViewController.append(FormSpacerItem(numberOfSpaces: 2))
```
erenbesel, line 141: maybe we need one of these spacers outside of the `if`, to preserve the same extra space if there is no button?

AdyenCard/Components/Card/CardComponent.swift
```diff
@@ -308,3 +302,14 @@ private extension CardComponent.Configuration {
         )
     }
 }
+
+// MARK: - SubmitCustomizable
+
+extension CardComponent: SubmitCustomizable {
+
+    public func submit() {
+        guard !configuration.showSubmitButton else { return }
```
goergisn: If we do this check it might be nice to let the merchant know why nothing happens - What do you think about an `AdyenAssertion.assertionFailure(message:)` so the merchant knows that they misconfigured/misused the function?

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -48,7 +48,7 @@ public final class GiftCardComponent: PresentableComponent,
     public let style: FormComponentStyle
 
     /// A boolean value that determines whether the payment button is displayed. Defaults to `true`.
-    public private(set) var showSubmitButton: Bool
+    internal let showSubmitButton: Bool
```
goergisn: Very nitpicky but what about 
```suggestion
    internal let showsSubmitButton: Bool
```

AdyenCard/Components/Card/CardComponent.swift
nauaros: Done

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
nauaros: Replaced everywhere. I saw where using the third form in other property names too. Thanks

AdyenComponents/OnlineBanking/OnlineBankingComponent.swift
nauaros: Ah yeah, good catch. I excluded the single spacer from condition. Thanks Eren!

AdyenCashAppPay/CashAppPayComponent.swift
nauaros: The `showsSubmitButton` property is defined in the configuration object injected in the payment components.

Tests/UnitTests/Mocks/PaymentComponentDelegateMock.swift
```diff
@@ -12,17 +12,37 @@ final class PaymentComponentDelegateMock: PaymentComponentDelegate {
...
-    var onDidSubmit: ((PaymentComponentData, PaymentComponent) -> Void)?
+    // MARK: - onDidSubmit
+
+    var onDidSubmitCallsCount = 0
+    var onDidSubmitCalled: Bool {
+        onDidSubmitCallsCount > 0
+    }
+
+    var onDidSubmitReceivedArguments: (data: PaymentComponentData, component: PaymentComponent)?
+    var onDidSubmitClosure: ((PaymentComponentData, PaymentComponent) -> Void)?
```
erenbesel: Not sure about this rename, old one is more swifty and all the other mocks still uses this style.

Tests/UnitTests/Mocks/PaymentComponentDelegateMock.swift
```diff
@@ -12,17 +12,37 @@ final class PaymentComponentDelegateMock: PaymentComponentDelegate {
...
-        self.onDidSubmit = onDidSubmit
-        self.onDidFail = onDidFail
+        self.onDidSubmitClosure = onDidSubmit
+        self.onDidFailClosure = onDidFail
     }
     
-    var onDidSubmit: ((PaymentComponentData, PaymentComponent) -> Void)?
+    // MARK: - onDidSubmit
+
+    var onDidSubmitCallsCount = 0
```
erenbesel: `didSubmitCallsCount` and `didSubmitCalled`. the `on` implies it's a closure

nauaros: Yeah, I thought on was part of the method name. Already changed.

Tests/UnitTests/Mocks/PaymentComponentDelegateMock.swift
nauaros: Since I'm adding some extra properties in there, I believe adding "closure" makes it more explicit. I can change it if it creates more confusion.

erenbesel: Yeah the `on` part achieves the closure implication. As you reverted it on the other properties only this will have `on` so closure keyword feels unnecessary imo

nauaros: Reverted

Tests/IntegrationTests/Components Tests/Online Banking/OnlineBankingComponentTests.swift
```diff
@@ -66,4 +66,49 @@ class OnlineBankingComponentTests: XCTestCase {
...
+
+        paymentDelegateMock.onDidSubmit = { _, component in
+            expectation.fulfill()
+        }
+
+        // When
+        sut.submit()
+
+        // Then
+        wait(for: [expectation], timeout: 100)
```
erenbesel: this timeout will kill the tests :D

erenbesel (approved):
Tiny comment about timeout: 100 rest is super

Adyen/Core/Components/StoredPaymentMethodComponent.swift
```diff
@@ -81,19 +81,14 @@ extension StoredPaymentMethodComponent {
...
 
-        /// A Boolean value that determines whether the payment button is displayed. Defaults to `true`.
-        public private(set) var showSubmitButton: Bool
-
         public var localizationParameters: LocalizationParameters?
 
         /// Initializes the configuration for Issuer list type components.
         /// - Parameters:
-        ///   - showSubmitButton: Boolean value that determines whether the payment button is displayed.
+        ///   - showsSubmitButton: Boolean value that determines whether the payment button is displayed.
```
goergisn: The function does not have a `showsSubmitButton` parameter anymore

AdyenComponents/SEPA Direct Debit/SEPADirectDebitComponent.swift
```diff
@@ -142,3 +145,17 @@ extension SEPADirectDebitComponent: TrackableComponent {}
...
 @_spi(AdyenInternal)
 extension SEPADirectDebitComponent: ViewControllerDelegate {}
+
+// MARK: - SubmitCustomizable
+
+extension SEPADirectDebitComponent: SubmitCustomizable {
+
+    public func submit() {
+        guard !configuration.showsSubmitButton else {
+            AdyenAssertion.assertionFailure(message: "Default submit button must be hidden in order to call submit.")
```
goergisn, line 155: It always is the same error message right? Maybe we can move it to an extension of `SubmitCustomizable` something along the lines of `func triggerMisconfiguredSubmitButtonAssertion()` or something shorter :D

Tests/IntegrationTests/Actions Tests/ActionComponent/Twint/TwintSDKActionTests+Convenience.swift
```diff
@@ -99,15 +99,15 @@ import XCTest
             onProvide: @escaping (ActionComponentData) -> Void
         ) -> ActionComponentDelegateMock {
         
-            let actonComponentDelegateMock = ActionComponentDelegateMock()
```
goergisn, line 102: 🙈

Tests/IntegrationTests/Components Tests/ACH Direct Debit/ACHDirectDebitComponentTests.swift
```diff
@@ -304,4 +304,65 @@ class ACHDirectDebitComponentTests: XCTestCase {
...
+
+        self.populate(textItemView: nameItemView, with: "test")
+        self.populate(textItemView: accountNumberItemView, with: "123456789")
+        self.populate(textItemView: routingNumberItemView, with: "121000358")
+
+        // When
+        sut.submit()
+
+        // Then
+        wait(for: [expectation], timeout: 100)
```
goergisn: This should not be more than 10 - it does not slow down the tests but if it takes longer there is usually something else wrong - we had cases where CI was really slow but ideally the tests then just fail and don't block the CI for over a minute

nauaros: Mistake from my side, I intended to use 10 seconds, not 100. Fixed

Tests/IntegrationTests/Components Tests/Online Banking/OnlineBankingComponentTests.swift
nauaros: Changed

Adyen/Core/Components/StoredPaymentMethodComponent.swift
nauaros: Good catch. Removed

AdyenComponents/SEPA Direct Debit/SEPADirectDebitComponent.swift
nauaros: Yes, it's always the same check but different components have a different set up for the showsSubmitButton property. The `SubmitCustomizable` property does not know about the property.

AdyenComponents/ACH Direct Debit/ACHDirectDebitComponent.swift
```diff
@@ -275,8 +275,10 @@ public final class ACHDirectDebitComponent: PaymentComponent,
             formViewController.append(storeDetailsItem)
         }
         
-        formViewController.append(FormSpacerItem(numberOfSpaces: 2))
-        formViewController.append(payButton)
+        if configuration.showsSubmitButton {
```
atmamont, line 278: Every instantiation of the `FormViewController` has the same repetitions with a few variations. Any chance this could be extracted to an assembly/factory with a method that has all possible variability options as method parameters?

AdyenComponents/SEPA Direct Debit/SEPADirectDebitComponent.swift
atmamont: From reading the code I am getting the same feeling. It could be a nice whiteboarding exersice.

Demo/UIKit/CustomComponent/CustomComponentPresenter.swift
```diff
@@ -5,25 +5,63 @@
...
 import AdyenNetworking
-
-protocol CustomComponentPresenterProtocol {}
+import AdyenSession
+@testable @_spi(AdyenInternal) import Adyen
+@testable import AdyenCard
+
+protocol CustomComponentPresenterProtocol {
+    var cardViewController: UIViewController { get }
+    func performPayment()
```
atmamont, line 16: `performPayment` doesn't sound as a presenter responsibility, does it? If this triggers payment floe presentation, maybe let's try something like `showPaymentFlow()`?

Demo/UIKit/CustomComponent/CustomComponentPresenter.swift
```diff
@@ -82,4 +133,37 @@ extension CustomComponentPresenter: PaymentComponentDelegate {
...
+            self?.handlePayment(result: result)
+        }
+    }
+}
+
+extension CustomComponentPresenter: PresentationDelegate {
+    
+    func present(component: any PresentableComponent) {
+        let viewController = component.viewController
+        (view as? UIViewController)?.present(viewController, animated: true)
```
atmamont: This casting is a bit confusing :) What would happen if it fails?

Tests/IntegrationTests/Components Tests/ACH Direct Debit/ACHDirectDebitComponentTests.swift
```diff
@@ -304,4 +304,64 @@ class ACHDirectDebitComponentTests: XCTestCase {
...
+
+        // When
+        sut.submit()
+
+        // Then
+        wait(for: [didSubmitExpectation], timeout: 10)
+        XCTAssertEqual(paymentDelegateMock.didSubmitCallsCount, 1)
+    }
+
+    func testSubmitWithDefaultSubmitShownShouldNotCallPaymentDelegateDidSubmit() throws {
```
atmamont, line 345: Is it possible to cover that `guard` check with tests (when `showsSubmitButton = true` and `submit` is called from outside)?

atmamont (approved):
None of my comments are blocking.
Great job Naufal!

Demo/UIKit/CustomComponent/CustomComponentPresenter.swift
nauaros: In this case, `performPayment` triggers the payment flow (payment request). There's no UI involved.

AdyenComponents/ACH Direct Debit/ACHDirectDebitComponent.swift
nauaros: Indeed, every component is responsible for creating and adding items to the form view. Unfortunately, with the current state it's hard to have a single point where this is done. Anyway, we can take a look into it. It would required a major refactor, but definitely up for it.

atmamont: Let's then have a session after I'm back from vacation?
I'm supercurious here as:
- even if we can't do it in one chunk, we can build our strategy on how to approach this kind of refactors
- this would help me to learn the codebase better. 

It does feel like these efforts could help us move faster in the future.

Demo/UIKit/CustomComponent/CustomComponentPresenter.swift
atmamont: I think I am still thinking in viper presenters, sorry. I see we use more MVP approach here.

Demo/UIKit/CustomComponent/CustomComponentPresenter.swift
nauaros: `view` is an abstraction (protocol). Since the `CustomComponentViewController` conforms to it, I decided it's safe to cast. I'm aware it's not the best approach but since this is just demo code and will go away after the task I chose this approach instead of defining more methods to present. To sum up, it won't fail.

nauaros: I changed it to avoid the cast. It should be better, it was a small change.

Tests/IntegrationTests/Components Tests/ACH Direct Debit/ACHDirectDebitComponentTests.swift
nauaros: Do you have any suggestion on how to test the assertion failure?

atmamont: After a quick research I found there is no known way to handle this gracefully. There are some [monstrous suggestions](https://stackoverflow.com/questions/25529625/testing-assertion-in-swift) but it seems to be an overkill.
What could be easier is to make it `func submit() throws` so it can be easily tested. Maybe it would trigger developer attention way earlier when writing the code. They wouldn't be caught by surprise with this assertion later on.

nauaros: I will bring this discussion to the group and we decide on it together. It's something I consider too. Thanks Andrii!

PR #1756, opened 2024-07-26
String only dictionary property for render event on analytics 

erenbesel:
Render event requires sort of a `metaData` type object and it's `[String: String]` containing basic types to make it generic across all platforms.
Initial approach was to have all properties forced into string but last approach uses mirror and only converts the properties that can be converted into `String` and it's done at a single place, `AnalyticsStringDictionaryConvertible`.


AdyenCard/Components/Card/CardAnalyticsConfiguration.swift
```diff
@@ -0,0 +1,74 @@
...
+    private let showKCPType: String
+    private let socialSecurityNumberMode: String
+    private let enableStoredDetails: Bool
+    private let hasInstallmentOptions: Bool
+    private let allowedCardTypes: String?
+    
+    internal init(configuration: CardComponent.Configuration) {
+        self.billingAddressMode = configuration.billingAddress.mode.analyticsDescription
+        self.billingAddressAllowedCountries = configuration.billingAddress.countryCodes?.joined(separator: Constants.stringSeparator)
+        if case .required = configuration.billingAddress.requirementPolicy {
```
nauaros, line 31: Suggestion:

```swift
self.billingAddressRequired = (configuration.billingAddress.requirementPolicy == .required)
```

goergisn: Or maybe even `self.billingAddressRequired = configuration.billingAddress.requirementPolicy.isRequired`
You know I like extensions 😅

Adyen/Analytics/Models/AnalyticsEventInfo.swift
```diff
@@ -44,4 +46,53 @@ public struct AnalyticsEventInfo: AnalyticsEvent {
...
+        case issuer
+        case validationErrorCode
+        case validationErrorMessage
+        case configData
+    }
+}
+
+@_spi(AdyenInternal)
+/// Protocol that requires a string only dictionary for analytics. Only meant for analytics as it uses mirroring.
+public protocol AnalyticsStringDictionaryConvertible: Encodable {
```
goergisn, line 82:
```suggestion
@_spi(AdyenInternal)
public protocol AnalyticsStringDictionaryConvertible: Encodable {
```

Adyen/Analytics/Models/AnalyticsEventInfo.swift
```diff
@@ -44,4 +46,53 @@ public struct AnalyticsEventInfo: AnalyticsEvent {
...
+        case isStoredPaymentMethod
+        case brand
+        case issuer
+        case validationErrorCode
+        case validationErrorMessage
+        case configData
+    }
+}
+
+@_spi(AdyenInternal)
```
goergisn:
```suggestion
```

nauaros (approved):

nauaros (approved):

nauaros (dismissed):

Scripts/test-SPM-integration.sh
```diff
@@ -6,11 +6,13 @@ PROJECT_NAME=TempProject
 
 # Clean up.
 rm -rf $PROJECT_NAME
+rm -rf .swiftpm
```
goergisn: This could be reverted as it didn't have any effect unfortunately

Tests/IntegrationTests/Card Tests/CardComponentEventTests.swift
```diff
@@ -33,8 +33,20 @@ final class CardComponentEventTests: XCTestCase {
         // Then
         XCTAssertEqual(analyticsProviderMock.initialEventCallsCount, 1)
         XCTAssertEqual(analyticsProviderMock.infos.count, 1)
-        let infoType = analyticsProviderMock.infos.first?.type
-        XCTAssertEqual(infoType, .rendered)
+        
+        let info = analyticsProviderMock.infos.first
+        XCTAssertEqual(info?.type, .rendered)
+        
+        let configDataDict = info?.configData?.stringOnlyDictionary
```
goergisn: maybe it would be interesting to also check for the count so we know if "encoding" changed

goergisn (dismissed):

goergisn (approved):

nauaros (approved):

PR #1765, opened 2024-08-01
COIOS-748: Support tokenization for Twint payment

nauaros:
## Summary
As part of the Twint project, this PR introduces tokenization support for the Twint payment method.

Together with the tokenization ability, we added support to perform payments with tokenized Twint payments.

## Motivation
Support tokenization in Twint payments.

## Changelog

- [Added] Add `StoredTwintPaymentMethod` in order to support tokenized payments.

## Release notes
<new>

This update adds support for tokenization in Twint payments, allowing shoppers to securely save their Twint payment details for future transactions.

</new>

<ticket>

COIOS-748

</ticket>


AdyenTwint/TwintComponent.swift
```diff
@@ -35,23 +123,49 @@ public final class TwintComponent: PaymentComponent {
     /// - Parameter configuration: The configuration for the component.
     public init(paymentMethod: TwintPaymentMethod,
                 context: AdyenContext,
-                configuration: Configuration = .init()) {
+                configuration: TwintComponentConfiguration = .init()) {
```
goergisn: Configuration is a typealias

```suggestion
                configuration: Configuration = .init()) {
```

Scripts/public-api-diff/Sources/Helpers/Models/SDKDump+Element+Definition.swift
```diff
@@ -1,8 +1,7 @@
 //
-//  File.swift
-//  
+// Copyright (c) 2024 Adyen N.V.
 //
-//  Created by Alexander Guretzki on 15/07/2024.
```
goergisn: The latest develop excludes these files from being formatted - would you mind reverting the changes on public-api-diff?

AdyenTwint/TwintComponent.swift
nauaros: Changed!

AdyenActions/Actions/TwintSDKData.swift
```diff
@@ -12,16 +12,28 @@ public struct TwintSDKData: Decodable {
...
     
     public init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
         token = try container.decode(String.self, forKey: .token)
+
+        // `isStored` field comes a String instead of Bool due to technical limitations
+        // on the API side. This is the reason it is necessary to parse it from String to
+        // Bool.
+        // This is the ticket where the field was added: https://youtrack.is.adyen.com/issue/FOC-77001
+        let isStoredString = try container.decode(String.self, forKey: .isStored)
```
erenbesel, line 31: Should we check for both types if this can change in the future?

Adyen/Core/Components/PresentableInitiableComponent.swift
```diff
@@ -0,0 +1,12 @@
+//
+// Copyright (c) 2024 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+@_spi(AdyenInternal)
+public protocol PresentableInitiableComponent: PresentableComponent, PaymentInitiable {
```
erenbesel: Let's find a more fitting name before merging even though it's spi :)
`ConditionallyPresentableInstantComponent` maybe?

Tests/SnapshotTests/Components/TwintComponentUITests.swift
```diff
@@ -0,0 +1,210 @@
...
+        sut.delegate = paymentComponentDelegate
+
+        setupRootViewController(sut.viewController)
+
+        let didSubmitExpectation = expectation(description: "PaymentComponentDelegate must be called.")
+        paymentComponentDelegate.onDidSubmit = { data, component in
+            XCTAssertTrue(component === self.sut)
+
+            // Then
+            XCTAssertFalse(data.storePaymentMethod ?? false)
```
erenbesel: This should be `?? true` but I'd suggest using force wrap or XCTUnwrap instead of `??`

AdyenActions/Actions/TwintSDKData.swift
nauaros: I had a discussion with the backend team and the chances of the type changing are nil. They are using String types everywhere.

Adyen/Core/Components/PresentableInitiableComponent.swift
nauaros: @goergisn thoughts on this one?

AdyenTwint/TwintComponent.swift
```diff
@@ -144,7 +144,7 @@ public final class TwintComponent: PaymentComponent,
 
     private var shouldStorePaymentMethod: Bool {
         guard configuration.showsStorePaymentMethodField else {
-            return configuration.storePaymentMethod
+            return false
```
erenbesel: return `nil` if the config is false same as card, ach etc as backend prefers not sending the value if not set explicitly

`return configuration.showsStorePaymentMethodField ? storeDetailsItem.value : nil`

nauaros: Updated

AdyenDropIn/Models/DropInConfiguration.swift
```diff
@@ -253,5 +253,26 @@ public extension DropInComponent {
...
+
+    /// Twint component configuration.
+    struct Twint: AnyTwintComponentConfiguration {
+
+        /// Indicates if the field for storing the payment method should be displayed in the form. Defaults to `true`.
+        public var showsStorePaymentMethodField: Bool
+
+        /// Determines whether to store this payment method. Defaults to `false`.
+        /// Ignored if `showsStorePaymentMethodField` is `true`.
+        public var storePaymentMethod: Bool
```
erenbesel: This flag seems forgotten which can mean this config is not used yet? This should be used on the component manager, get the value from this (dropInConfig.twint) to actual twintComponent.config

nauaros: Removed. Thanks!

AdyenTwint/TwintComponent.swift
```diff
@@ -39,19 +126,49 @@ public final class TwintComponent: PaymentComponent {
...
+@_spi(AdyenInternal)
+extension TwintComponent: PresentableInitiableComponent {
+
+    public var requiresPresentation: Bool {
+        configuration.showsStorePaymentMethodField
+    }
+}
+
+@_spi(AdyenInternal)
+extension TwintComponent: ViewControllerDelegate {}
```
erenbesel: I think this component now also needs to be trackable if it's presented. Let's discuss

AdyenDropIn/Models/DropInConfiguration.swift
```diff
@@ -253,5 +253,19 @@ public extension DropInComponent {
             self.storePaymentMethod = storePaymentMethod
         }
     }
-    
+
+    /// Twint component configuration.
+    struct Twint: AnyTwintComponentConfiguration {
```
erenbesel: I don't see this is added to DropIn's configuration as a property yet right? Needs to be added there and then passed on to the TwintComponentConfiguration so this could be set via dropin

nauaros: Added!

Adyen/Core/Components/PresentableInitiableComponent.swift
atmamont: My first solution from the top of my head to this naming issue would be to take relief of decision-making out of Swift type system completely.
I'd experiment with one and only one `ComponentPresenting` protocol having single `presentationStyle` property that defines it's behavior. Then during runtime `TwintComponent` or it's controller could easily set it to `requiresPresentation` or `silent`.  This would also allow to get rid of fragile logic that I now see appeared in the `DropInComponent func didSelect(_ component: PaymentComponent)` that requires a specific switch order to handle it properly.

Type-checking any given type for conformance to make business logic decision (or presentation  in this case) is a bit of code smell. Specifically for such a strongly typed language as Swift!

AdyenTwint/TwintComponent.swift
nauaros: Done!

Tests/IntegrationTests/DropIn Tests/ComponentManagerTests.swift
```diff
@@ -11,6 +11,10 @@
 #if canImport(AdyenCashAppPay)
     @testable import AdyenCashAppPay
 #endif
+
+#if canImport(AdyenCashAppPay)
```
erenbesel: twint

nauaros: Oh good catch, it was like that all this time. Fixed

Adyen/Core/Payment Methods/Abstract/AnyPaymentMethodDecoder.swift
```diff
@@ -491,17 +491,26 @@ private struct TwintPaymentMethodDecoder: PaymentMethodDecoder {
...
             return AnyPaymentMethod(InstantPaymentMethod(type: .twint, name: "Twint"))
         #endif
     }
 
     func anyPaymentMethod(from paymentMethod: any PaymentMethod) -> AnyPaymentMethod? {
         #if canImport(TwintSDK)
-            (paymentMethod as? TwintPaymentMethod).map { .twint($0) }
-        #else
-            return nil
+            if let method = paymentMethod as? TwintPaymentMethod {
```
atmamont, line 506: What happens if both castings fail?
Should we use switch with type casting cases and send a telemetry error if both fail?

AdyenActions/Actions/TwintSDKData.swift
```diff
@@ -12,16 +12,28 @@ public struct TwintSDKData: Decodable {
...
     }
     
     public init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
         token = try container.decode(String.self, forKey: .token)
+
+        // `isStored` field comes a String instead of Bool due to technical limitations
+        // on the API side. This is the reason it is necessary to parse it from String to
+        // Bool.
+        // This is the ticket where the field was added: https://youtrack.is.adyen.com/issue/FOC-77001
```
atmamont: Should we keep the full link in a public repo? If not, then the ticket number would be enough.

AdyenActions/Components/SDK/Twint+Injectable.swift
```diff
@@ -20,7 +20,12 @@ import Foundation
         @objc func pay(withCode code: String, appConfiguration: TWAppConfiguration, callback: String) -> Error? {
             Twint.pay(withCode: code, appConfiguration: appConfiguration, callback: callback)
         }
-        
+
+        @objc
+        func registerForUOF(withCode code: String, appConfiguration: TWAppConfiguration, callback: String) -> Error? {
```
atmamont, line 25: What does UOF stand for?

AdyenActions/Components/SDK/TwintSDKActionComponent.swift
```diff
@@ -121,12 +120,21 @@ import Foundation
         }
 
         private func invokeTwint(app: TWAppConfiguration, action: TwintSDKAction) {
-
-            let error = twint.pay(
-                withCode: action.sdkData.token,
-                appConfiguration: app,
-                callback: configuration.callbackAppScheme
-            )
+            let error: Error? = {
```
atmamont: It reads more like we store action callback but the name suggests it's an error which is a bit confusing.

AdyenActions/Actions/TwintSDKData.swift
nauaros: Change it to the ticket number only

AdyenActions/Components/SDK/TwintSDKActionComponent.swift
nauaros: What do you think now?

AdyenActions/Components/SDK/Twint+Injectable.swift
nauaros: It means User on File, referencing storing the payment details to allow tokenization. This name comes from the method exposed in the Twint SDK.

AdyenActions/Components/SDK/TwintSDKActionComponent.swift
atmamont: Now it's better.
I might throw another suggestion but it's up to you to decide if it's better or not.
Assuming both `twint.registeForUOF` and `twint.pay` have same signatures, this can be improved a bit:

```
let operation: (String, AppConfiguration, String) -> Error?
    
if action.sdkData.isStored {
    operation = twint.registerForUOF
} else {
    operation = twint.pay
}
    
if let error = operation(action.sdkData.token, app, configuration.callbackAppScheme) {
    handleShowError(error.localizedDescription)
    return false
}
```

AdyenActions/Components/SDK/Twint+Injectable.swift
atmamont: I wonder if `registerForUserOnFile` would look better for anyone who uses this API. Still not enough clarity

nauaros: We're providing a Swift interface to the Obj-C one offered by the Twint SDK, so we're mimicking

atmamont (dismissed):

Adyen/Core/Payment Methods/Abstract/AnyPaymentMethodDecoder.swift
nauaros: Good question!
If that happens, we're on a point where we know it's a valid payment method since we were able to decode it.
With that being said, it means it is a payment method that we don't recognize in the SDK so we treat it like an instant payment method.

You can trace this logic all the way to `AnyPaymentMethodDecoder.swift:134`

Adyen.docc/API Reference.md
```diff
@@ -205,6 +208,13 @@ The Adyen DropIn/Components SDK API Reference.
 - ``CashAppPayPaymentMethod``
 - ``CashAppPayComponent``
 - ``CashAppPayDetails``
+
+### Twint
```
erenbesel: Minor but maybe add the word component to fit the rest

erenbesel (dismissed):

erenbesel (approved):

atmamont (approved):

PR #1768, opened 2024-08-06
Fix mismatching values with web for analytics config

erenbesel:


AdyenCard/Components/Card/CardAnalyticsConfiguration.swift
```diff
@@ -34,13 +33,12 @@ internal struct CardAnalyticsConfiguration: AnalyticsStringDictionaryConvertible
             self.billingAddressRequired = false
         }
         self.showsHolderNameField = configuration.showsHolderNameField
-        self.showsStorePaymentMethodField = configuration.showsStorePaymentMethodField
-        self.showsSecurityCodeField = configuration.showsSecurityCodeField
+        self.hideCVC = configuration.showsSecurityCodeField
```
goergisn: sounds like it should be the reverse?
```suggestion
        self.hideCVC = !configuration.showsSecurityCodeField
```

nauaros (dismissed):

Tests/IntegrationTests/Card Tests/CardComponentEventTests.swift
```diff
@@ -40,13 +40,12 @@ final class CardComponentEventTests: XCTestCase {
         let configDataDict = try XCTUnwrap(info?.configData?.stringOnlyDictionary)
         XCTAssertEqual(configDataDict["socialSecurityNumberMode"], "auto")
         XCTAssertEqual(configDataDict["hasInstallmentOptions"], "false")
-        XCTAssertEqual(configDataDict["showsStorePaymentMethodField"], "true")
         XCTAssertEqual(configDataDict["billingAddressRequired"], "true")
-        XCTAssertEqual(configDataDict["showsSecurityCodeField"], "true")
+        XCTAssertEqual(configDataDict["hideCVC"], "true")
```
goergisn: Maybe then also this needs to change to "false" (Not sure from the limited amount of code)

goergisn (approved):

nauaros (approved):

PR #1786, opened 2024-08-19
Cashapp recurring fix

erenbesel:
Fixes an issue where on session flow, `CashAppPayConfiguration` property `storePaymentMethod` was being ignored even when `showsStorePaymentMethodField` is set to `false`.


AdyenCashAppPay/CashAppPayComponent.swift
```diff
@@ -181,7 +163,25 @@ public final class CashAppPayComponent: PaymentComponent,
...
+    internal func createPaymentActions() -> [PaymentAction] {
+        var actions = [PaymentAction]()
+        if let amount = payment?.amount, amount.value > 0 {
+            let moneyAmount = Money(amount: UInt(amount.value), currency: .USD)
+            let oneTimeAction = PaymentAction.oneTimePayment(scopeID: cashAppPayPaymentMethod.scopeId,
+                                                             money: moneyAmount)
+            actions.append(oneTimeAction)
+        }
     
+        if storePayment == true {
```
goergisn: No need to check for `true` anymore as it's not optional anymore 👍 
```suggestion
        if storePayment {
```

nauaros: suggestion: `if storePayment {`

nauaros (dismissed):

descorp (dismissed):

EyeBrowserr (approved):

PR #1790, opened 2024-08-20
Simplifying QRCodeView

goergisn:
## Summary
- Adjusting some QRCodeView UI
- Making QRCodeView scrollable

| Before (Cut off count down label) | After (Scrollable content) |
| --- | --- |
| Simulator Screenshot - iPhone 15 Pro - 2024-08-20 at 15 35 07 | Simulator Screenshot - iPhone 15 Pro - 2024-08-21 at 12 50 53 |


AdyenActions/UI/View Controllers/QR Code/String+QRCode.swift
```diff
@@ -0,0 +1,27 @@
...
+    
+    /// Generates a QRCode as UIImage representing the string
+    func generateQRCode() -> UIImage? {
+        guard let filter = CIFilter(name: "CIQRCodeGenerator") else {
+            return nil
+        }
+        
+        let data = self.data(using: String.Encoding.ascii)
+        filter.setValue(data, forKey: "inputMessage")
+        let transform = CGAffineTransform(scaleX: 3, y: 3)
```
atmamont, line 19: Can't we just set this as a view constraint instead of manipulating the image itself?

atmamont (dismissed):

AdyenActions/UI/View Controllers/QR Code/String+QRCode.swift
goergisn: Mhh I just took the code as it was. Might want to experiment with it

atmamont (approved):

PR #1793, opened 2024-08-21
Making anchor behave like expected

goergisn:
## Background
The `UIViewConstraintsHelper` offer an `anchor` function to easily add constraints to a view. Because it uses auto layout it was implemented in a way that the values of `UIEdgeInset` were just forwarded as the top/bottom/leading/trailing constants, which caused passing negative numbers for bottom/trailing and positive numbers for top/leading in some places.

## Summary
- `anchor` now expects positive numbers for bottom/trailing if you want to inset from the bottom/trailing anchor

### DIFF
--- a/Adyen/Helpers/UIViewConstraintsHelper.swift
+++ b/Adyen/Helpers/UIViewConstraintsHelper.swift
@@ -72,17 +72,17 @@ extension AdyenScope where Base: UIView {
         }
         if let bottom = edgeInsets.bottom {
             constraints.append(
-                base.bottomAnchor.constraint(
-                    equalTo: anchorSource.anchorSet.bottomAnchor,
-                    constant: -bottom
+                anchorSource.anchorSet.bottomAnchor.constraint(
+                    equalTo: base.bottomAnchor,
+                    constant: bottom
                 )
             )
         }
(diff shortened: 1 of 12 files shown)


Adyen/Helpers/UIViewConstraintsHelper.swift
```diff
@@ -12,48 +12,79 @@ import UIKit
...
+                )
+            )
         }
         if let bottom = edgeInsets.bottom {
-            constraints.append(base.bottomAnchor.constraint(equalTo: anchorSource.anchorSet.bottomAnchor,
-                                                            constant: bottom))
+            constraints.append(
+                base.bottomAnchor.constraint(
+                    equalTo: anchorSource.anchorSet.bottomAnchor,
+                    constant: -bottom
```
erenbesel: It's preference I guess as I'm not sure if I agree with changing the calling of this function :D, but if we agree to go with it then maybe we can reverse the anchors for bottom/trailing so we don't provide a negative anywhere like:
```
constraints.append(
      anchorSource.anchorSet.bottomAnchor.constraint(
          equalTo: base.bottomAnchor, 
          constant: bottom
      )
)
```

erenbesel (approved):

PR #1794, opened 2024-08-21
Remove Completion<T> typealias

goergisn:
## Summary
- The typealias is longer than the signature it replaces -> removing it
- Removing unused CGSize/UIEdgeInsets extensions


erenbesel (commented):
Isn't this a breaking change? Even if it wasn't, it's shorter imo :)

atmamont:
> Isn't this a breaking change? Even if it wasn't, it's shorter imo :)

My concern here is introducing a public API change. Is code style alone a good reason to do so?

PR #1809, opened 2024-09-05
ExpiryTimer fixes

goergisn:
## Background
I noticed that the QRCodeView snapshots are flaky on CI. 
After some research it looks like the expiry label did some time to show up. 
If the CI is slow, it showed the expiry label but if it's fast, the snapshot is taken before the first expiry timer run (1sec) and thus does not show up on the snapshot. 
The reference snapshots did not include the expiry label, which was unfortunate also.

## Summary
- Notifies the caller immediately when the timer is started so we can display the initial value (better user experience + no need to wait about 1 sec for the snapshots)
- Stops the timer once it's expired (Currently it continues running indefinitely)


descorp (dismissed):

AdyenActions/Components/QRCode/ExpirationTimer.swift
```diff
@@ -44,6 +44,9 @@ internal final class ExpirationTimer {
         timer = Timer.scheduledTimer(withTimeInterval: tickInterval, repeats: true) { [weak self] _ in
             self?.onTimerTick()
         }
+        
+        // Notify immediately
+        notify()
```
atmamont, line 46: Let's add a unit test as well?

atmamont (dismissed):

atmamont (dismissed):

descorp (dismissed):

AdyenActions/Components/QRCode/ExpirationTimer.swift
```diff
@@ -29,21 +27,23 @@ internal final class ExpirationTimer {
     ///   - onExpiration: The closure called on timer expiration.
     internal init(
         expirationTimeout: TimeInterval,
-        tickInterval: TimeInterval = 1.0,
```
erenbesel, line 32: Why are we

AdyenActions/Components/QRCode/ExpirationTimer.swift
```diff
@@ -29,21 +27,23 @@ internal final class ExpirationTimer {
...
         onTick: @escaping (TimeInterval) -> Void,
         onExpiration: @escaping () -> Void
     ) {
         self.expirationTimeout = expirationTimeout
-        self.tickInterval = tickInterval
```
erenbesel, line 37: Wouldn't removing this limit flexibilty and testing?

goergisn: I removed it as it was unused and also implied to me that we could speed up the timer, but what it did instead was just notify more often 

but **not** speed up or slow down the timer.

atmamont (approved):

AdyenActions/Components/QRCode/ExpirationTimer.swift
goergisn: As it uses TimeInterval (which implies a duration of seconds) the only valid usecase of this would be for testing but as it was not used so far I removed it. Better to add it later than adding tests for an unused feature imo.
I agree tho that it could be used to speed up tests in the future.

erenbesel (approved):

PR #1815, opened 2024-09-05
Remove the recent SPI addition from some validators

erenbesel:
<changed>

The following protocol and classes that were recently made internal (`@_spi(AdyenInternal)`) are now public again: 
- `Validator`
- `CardNumberValidator`
- `CardExpiryDateValidator`
- `CardSecurityCodeValidator`
- `LengthValidator`
- `NumericStringValidator`

</changed>


goergisn (approved):

atmamont (approved):

AdyenCard/Validators/CardSecurityCodeValidator.swift
```diff
@@ -8,8 +8,7 @@
 import Foundation
 
 /// Validates a card's security code.
-@_spi(AdyenInternal)
-public final class CardSecurityCodeValidator: NumericStringValidator, AdyenObserver, StatusValidator {
+public final class CardSecurityCodeValidator: NumericStringValidator, AdyenObserver {
```
descorp, line 11: Is it possible to move `AdyenObserver` part under an SPI extension?

erenbesel: Hmm, `AdyenObserver` is public anyway, what benefit would that have?

descorp: No direct benefits, just makes public part clear.

`AdyenObserver ` was created back in a days when iOS 13 was not available and we were not able to use `Combine`. Nowadays we can, so this implementation can change..

descorp (approved):

PR #1823, opened 2024-09-09
Adapting script to run public api changes on last release

goergisn:
## Summary
- Compare against the last release when on a `release/...` branch

See: https://github.com/Adyen/adyen-ios/pull/1825#issuecomment-2340142448


atmamont (approved):
I like how you add new emojis every time

Issue #1824, opened 2024-09-10
[Feature] Make AdyenEncryption an isolated package


forgetfulwhale:
Currently if you only require AdyenEncryption for custom card integration with a server layer in between, you are now forced to also use its dependencies "Core" and "Network".

However when examining the code, the only dependency is AdyenCoder.swift->encode(), which is a very basic wrapper around a JSON encoder. Seems overkill to create this large dependency chain.

**Describe the solution you'd like**

This can cause issues in size sensitive applications, such as AppClip (15mb), this inflates the AdyenSDK requirements from < 100kb to > 8mb and becomes unusable, very unnecessary.

One can manually import the whole AdyenEncryption folder and the single file from Core, there is no build or functionality issues but it would be better for CI/CD to have proper package manager support and to reduce the overhead of manual maintenance.

AdyenEncryption should always be 100% standalone.

goergisn:
Hey @forgetfulwhale,
thanks for reporting 🤝

Indeed the `AdyenEncryption` should not have any dependencies on `AdyenCore`.
This will be fixed in the next release 👌 

I'll leave this ticket open for verification after the release

goergisn:
Hi @forgetfulwhale 
I'm happy to let you know that the [latest release (5.12.0)](https://github.com/Adyen/adyen-ios/releases/tag/5.12.0) contains your requested change 👌 
Will close the ticket but feel free to re-open / create a new one if there are any open questions/issues

PR #1827, opened 2024-09-10
Removing `AdyenCore` dependency from `AdyenEncryption`

goergisn:
## Summary
- `AdyenCore` was only needed because of `AdyenCoder` to get access to the `.sortedKeys` behavior

## Change
- Remove `AdyenCore` dependency from `AdyenEncryption` (SPM package)
- Using a simple JSONEncoder extension to provide the functionality

## Release Notes
<changed>

You can now use the `AdyenEncryption` module independently with a reduced size, because it no longer depends on `AdyenCore`.

</changed>

See: https://github.com/Adyen/adyen-ios/issues/1824


atmamont (dismissed):
👍

erenbesel (dismissed):

erenbesel (commented):
Missing changes on the podspec I think

goergisn:
> Missing changes on the podspec I think

Yess, good one, thanks 🙏

erenbesel (approved):

descorp (approved):

Issue #1354, opened 2023-09-01
Support library evolution


ffittschen:
**Describe the bug**
As we are using Adyen alongside an internal binary framework, we need to compile it with library evolution enabled (`BUILD_LIBRARY_FOR_DISTRIBUTION=YES`), so that binary compatibility between our binary framework and its Adyen dependencies is guaranteed.

Starting with [this PR](https://github.com/Adyen/adyen-ios/pull/895), Adyen uses the underscored attribute `@_spi` to mark declarations as System Programming Interface (SPI). These `@_spi` annotations cause build errors when compiling Adyen with the `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` build setting to enable library evolution.

**To Reproduce**
1. Check out this repository
2. Run the following command

```bash
xcodebuild -project Adyen.xcodeproj -scheme Adyen -destination 'generic/platform=iOS' -archivePath "archives/Adyen-iphoneos.xcarchive" SKIP_INSTALL=NO BUILD_LIBRARY_FOR_DISTRIBUTION=YES clean archive
```

**Expected behavior**
I'd expect no build errors when using the `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` build setting.

**Screenshots**

**Environment**
 - Device: n/a
 - iOS Version: n/a
 - SDK Version: n/a
 - Adyen API Version: n/a
 - Package management system: n/a
 - Xcode version: n/a
 - Mac OS type: n/a

**Additional context**
This problem was first brought up in this issue: https://github.com/Adyen/adyen-ios/issues/1130

descorp:
Hey @ffittschen 

Thank you for your feedback!
We have noticed this behavior lately but didn't have any research on this topic yet.
We will keep you posted.

descorp:
Hey @ffittschen 

We would like to have a better understanding of your case.
The "internal binary framework", who uses it?

descorp:
Prospect PR #1413

goergisn:
Hi @ffittschen,

We merged a fix for the described issue to develop yesterday 🤝 
Feel free to check it out for confirmation for now.

Best,
Alex

goergisn:
Hi @ffittschen,

we just released [5.5.0](https://github.com/Adyen/adyen-ios/releases/tag/5.5.0) that contains support for library evolution.

Please let us know if you are running into any issues.
And feel free to re-open the issue, if so. 👍 

Best,
Alex

[...]

PR #1834, opened 2024-09-17
Fix build library for distribution

goergisn:
# Background
When trying to ship the Checkout SDK as a binary using `BUILD_LIBRARY_FOR_DISTRIBUTION=YES` it fails because of some `@_spi` usage. See: https://github.com/Adyen/adyen-ios/issues/1354
It turned out that it was related to some protocols having other protocol requirements to `Component`.
Restructuring the dependency/requirement hierarchy should fix this issue without having to switch to the `package` access level.

## Summary
- Restructuring some protocol dependencies to make compiling with ```xcodebuild clean build -scheme "SCHEME" -destination "generic/platform=iOS" -derivedDataPath "_build" -sdk `xcrun --sdk iphonesimulator --show-sdk-path` BUILD_LIBRARY_FOR_DISTRIBUTION=YES``` succeed

### Release notes

<new>
- Full support for [library evolution](https://www.swift.org/blog/library-evolution/)
</new>


Scripts/validate-library-evolution.swift
```diff
@@ -0,0 +1,165 @@
...
+    
+    return """
+// swift-tools-version: 5.8
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "\(scheme)",
+    platforms: [.iOS(.v13)],
```
atmamont: Should this be 12 as we still support it?

goergisn: Fair - the binary is compiled for iOS 12 but of course it makes sense to use it here as well 👌

atmamont (dismissed):

nauaros (dismissed):

atmamont (approved):

nauaros (approved):

.github/workflows/test_library_evolution.yml
```diff
@@ -28,9 +28,12 @@ jobs:
     - name: "xcodebuild clean archive BUILD_LIBRARY_FOR_DISTRIBUTION=YES"
       run: |
         xcodebuild -version
-        xcodebuild clean archive -project "${project}" -scheme "${scheme}" ${params} -destination "${destination}" | xcpretty --utf --color && exit ${PIPESTATUS[0]}
+        xcodebuild clean archive -project "${project}" -scheme "${scheme}" -destination 'generic/platform=iOS Simulator' -archivePath "${archivePath}" -skipPackagePluginValidation SKIP_INSTALL=NO BUILD_LIBRARY_FOR_DISTRIBUTION=YES
+        ./Scripts/validate-library-evolution.swift "${archivePath}" "${validationProjectName}"
+        cd ${validationProjectName}
+        xcodebuild clean build -scheme "${validationProjectName}" -destination 'generic/platform=iOS Simulator' | xcpretty --utf --color && exit ${PIPESTATUS[0]}
```
descorp, line 34: Minor suggestion:
We can split those in 2 or 3 separate steps - Xcode build logs usually huge, so when if fail - we know where to look

descorp (approved):

erenbesel (approved):

PR #1836, opened 2024-09-19
Add accessibility identifier for "close" button to enable DropIn Card component e2e testing 

atmamont:
In order to make e2e tests more robust and future-proof, this adds missing accessibility id for the close button in the DropIn card component.

image

### DIFF
--- a/Adyen/Core/ToolBar/ModalToolbar.swift
+++ b/Adyen/Core/ToolBar/ModalToolbar.swift
@@ -85,6 +85,9 @@ public class ModalToolbar: UIView, AnyNavigationBar {
             addSubview(stackView)
             stackView.adyen.anchor(inside: self)
         }
+        
+        self.accessibilityIdentifier = "adyen.ModalToolbar"
+        cancelButton.accessibilityIdentifier = ViewIdentifierBuilder.build(scopeInstance: self, postfix: "closeButton")
 
         setupStyle()
     }


atmamont:
@Robert-SD JFYI

Adyen/Core/ToolBar/ModalToolbar.swift
```diff
@@ -85,6 +85,9 @@ public class ModalToolbar: UIView, AnyNavigationBar {
             addSubview(stackView)
             stackView.adyen.anchor(inside: self)
         }
+        
+        self.accessibilityIdentifier = "adyen.ModalToolbar"
```
goergisn, line 89: Is this needed? Why not use the ViewIdentifierBuilder here as well?
Maybe we can adapt the identifier builder to not require a postfix...

goergisn (approved):

Adyen/Core/ToolBar/ModalToolbar.swift
atmamont: I followed the codebase style and found this as an example. ViewBuilder inside uses `accessibiltyIdentifier` for the `scope` which in our case is `ModalToolbar` itself so it should have a defined accessibility identifier.

**QRCodeView.swift**
```
    internal init(model: Model) {
        self.model = model
        super.init(frame: .zero)
        
        backgroundColor = model.style.backgroundColor
        accessibilityIdentifier = "adyen.QRCode"
        
        setupUI()
    }
```

goergisn: Ahhh I see 👌

PR #1840, opened 2024-09-26
Fix crash: duplicate items when rendering payment methods list

atmamont:
# Summary
<fixed>
- On iOS 18, if the shopper has two identical stored payment methods, Drop-in/Component no longer crashes when the shopper selects a payment method that's not the default stored one.
</fixed>

# Details 
iOS18 release introduced new crash which didn't happen before.

When rendering payments list, in rare occasions we could have multiple payment methods having same title and description.
Providing ListItem that is used to render the list uses title as part of it's `Hashable` and `Equatable` conformance, resulting in having two identical entries.
This list typically has few sections: stored payment methods and generic payment methods. While any stored payment method always has an unique identifier coming from the backend, the rest of the sections do not.

That's why this fix does not use any backend-provided identifiers and relies purely on `UUID` that is created upon `ListItem` instantiation and doesn't get exposed anywhere else.
It can be improved by providing a parameter in the init so it could be injected from the outside.

## Note:
`ListItem` has `identifier` property that can be mistreated easily and abused. We use it so set an identifier that is later used as `accessibilityIdentifier` for UI elements. This is misleading as for such a case when we have two identical stored payment methods this could be a potential accessibility problem that will make second duped element on the list inaccessible.

## Screenshot 
image

<youtrack>COIOS-797</youtrack>

### DIFF
--- a/Tests/UnitTests/Core/ListItemTests.swift
+++ b/Tests/UnitTests/Core/ListItemTests.swift
@@ -8,7 +8,7 @@ import XCTest
 @testable @_spi(AdyenInternal) import Adyen
 
 final class ListItemTests: XCTestCase {
-    // Checks COIOS-797 fix
+    // Checks COIOS-797: NSInternalInconsistencyException 1008: Item identifiers are not unique
     func test_listItem_isAlwaysUnique() {
         let item1 = ListItem(title: "Test title 11", icon: .dummy(), trailingText: "Text")
         let item2 = ListItem(title: "Test title 11", icon: .dummy(), trailingText: "Text")
--- a/Tests/UnitTests/Core/ListItemTests.swift
+++ b/Tests/UnitTests/Core/ListItemTests.swift
@@ -1,0 +1,25 @@
+//
+// Copyright (c) 2024 Adyen N.V.
+//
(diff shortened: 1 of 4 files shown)


Tests/UnitTests/Core/ListItemTests.swift
```diff
@@ -0,0 +1,25 @@
...
+// Copyright (c) 2024 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import XCTest
+@testable @_spi(AdyenInternal) import Adyen
+
+final class ListItemTests: XCTestCase {
+    // Checks COIOS-797 fix
```
goergisn: I have a feeling this might not be descriptive enough - maybe add a one-liner?

erenbesel (dismissed):

goergisn (dismissed):

nauaros (dismissed):

Tests/UnitTests/Core/ListItemTests.swift
atmamont: Done!

goergisn (approved):

PR #1842, opened 2024-10-01
Fix crash: duplicate items when rendering payment methods list

atmamont:
# Summary
<fixed>
- On iOS 18, if the shopper has two identical stored payment methods, Drop-in/Component no longer crashes when the shopper selects a payment method that's not the default stored one.
</fixed>

Replicates https://github.com/Adyen/adyen-ios/pull/1840 for v4

<ticket>COIOS-800</ticket>

### DIFF
--- a/Adyen/UI/List/ListItem.swift
+++ b/Adyen/UI/List/ListItem.swift
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2021 Adyen N.V.
+// Copyright (c) 2018 Adyen N.V.
 //
 // This file is open source and available under the MIT license. See the LICENSE file for more info.
 //
@@ -69,6 +69,9 @@ public class ListItem: FormItem {
         builder.build(with: self)
     }
     
+    // MARK: - Private
+        
+    private let listIdentifier: UUID = .init()
 }
 
 // MARK: - Hashable & Equatable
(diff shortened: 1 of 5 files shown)


nauaros (dismissed):

erenbesel (dismissed):

erenbesel (approved):

Adyen/UI/List/ListItem.swift
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2021 Adyen N.V.
+// Copyright (c) 2018 Adyen N.V.
```
goergisn, line 2: Traveling back in time? :P

Demo/Configuration.swift
```diff
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2024 Adyen N.V.
+// Copyright (c) 2017 Adyen N.V.
```
goergisn, line 2: 🤔

Adyen/UI/List/ListItem.swift
atmamont: v4 `swiftformat` apparently uses file creation date for this job

goergisn (approved):

PR #1843, opened 2024-10-01
Always make the initial analytics call

erenbesel:
We now always make the initial analytics call to just fetch the checkout attempt id.

<ticket>
COIOS-782 
</ticket>

### DIFF
--- a/Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
+++ b/Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
@@ -32,7 +32,6 @@ internal final class AnalyticsProvider: AnalyticsProviderProtocol {
         static let infoLimit = 50
         static let logLimit = 5
         static let errorLimit = 5
-        static let attemptIdFetchFailed = "fetch-checkoutAttemptId-failed"
     }
 
     // MARK: - Properties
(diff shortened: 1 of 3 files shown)


descorp (dismissed):

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
```diff
@@ -45,6 +46,10 @@ internal final class AnalyticsProvider: AnalyticsProviderProtocol {
     
     private var batchTimer: Timer?
     private let batchInterval: TimeInterval
+    
+    private var isValidAttemptId: Bool {
```
atmamont: Doesn't it start turning into a whole logic around this id?
Maybe an enum could be a simple logic wrapper:
```
enum CheckoutAttempt {
    case id(String)
    case doNotTrack

    var isValid: Bool
}
```

Tests/UnitTests/Analytics/AnalyticsProviderTests.swift
```diff
@@ -59,10 +59,38 @@ class AnalyticsProviderTests: XCTestCase {
...
 
         // When
         sut.sendInitialAnalytics(with: .components(type: .affirm), additionalFields: nil)
-        XCTAssertEqual(sut.checkoutAttemptId, "do-not-track")
+        wait(until: sut, at: \.checkoutAttemptId, is: expectedCheckoutAttemptId)
+    }
+    
+    func testFetchCheckoutAttemptIdFailedResult() throws {
+        // Given
+        var analyticsConfiguration = AnalyticsConfiguration()
```
atmamont: This configuration and apiClient creation code is quite repetitive, let's extract it to an internal helper function?

atmamont (dismissed):

atmamont (approved):

goergisn (approved):

PR #1865, opened 2024-10-15
Make UISwitch form element accessible

atmamont:
# Summary

This change aims to address [this issue](https://github.com/Adyen/adyen-react-native/issues/532) enabling UISwitch elements to be accessible. They do have their own accessibility identifiers defined.

# Release notes

<new>
- Switch elements are now accessible for external testing with tools like Appium.
</new>

# Ticket

<ticket>
COIOS-808
</ticket>

### DIFF
--- a/Adyen/UI/Form/Items/Toggle/FormToggleItemView.swift
+++ b/Adyen/UI/Form/Items/Toggle/FormToggleItemView.swift
@@ -36,7 +36,7 @@ public final class FormToggleItemView: FormItemView<FormToggleItem> {
         switchControl.translatesAutoresizingMaskIntoConstraints = false
         switchControl.isOn = item.value
         switchControl.onTintColor = item.style.tintColor
-        switchControl.isAccessibilityElement = false
+        switchControl.isAccessibilityElement = true
         switchControl.setContentHuggingPriority(.required, for: .horizontal)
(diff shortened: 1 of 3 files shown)


Adyen/UI/Form/Items/Toggle/FormToggleItemView.swift
```diff
@@ -36,7 +36,7 @@ public final class FormToggleItemView: FormItemView<FormToggleItem> {
         switchControl.translatesAutoresizingMaskIntoConstraints = false
         switchControl.isOn = item.value
         switchControl.onTintColor = item.style.tintColor
-        switchControl.isAccessibilityElement = false
```
goergisn, line 39: ~This most probably is messing with the actual accessibility as it's set to false on purpose if you check out the init of the `FormToggleItemView`~

Actually it does not change anything so all good 👌

goergisn (approved):

atmamont:
Closing as this approach doesn't work.
Appium is not able to traverse the view hierarchy down if a container has `isAccessibleElement` set to true. We use this for the toggle to improve accessibility and it works perfectly fine with tools like VoiceOver. 
The goal is to keep the level of accessibility we provide now and make it testable for tools like Appium at the same time.
I will spend a bit more time on this and create a new PR later.

PR #1866, opened 2024-10-16
Add an event for closing the dropin

erenbesel:
# Summary
Sends a new log event for when the close button of dropin is tapped.

Also updates the timer logic to start after the first analytics call rather than instantly starting and doing nothing.

# Ticket

<ticket>
COIOS-663
</ticket>


atmamont (changes requested):
Let's cover timer change logic with tests?

goergisn (approved):

nauaros (approved):

PR #1873, opened 2024-10-18
Enable `FormItemToggleView` external testability

atmamont:
# Summary

This change aims to improve SDK testability for tools like Appium. In the case VoiceOver is not active, it exposes `UISwitch` as a separate accessibility element instead of a composite [UIlLabel, UISwitch] container that hides internal components from Accessibility.  

# Demo

| Appium Inspector | Accessibility Inspector |
|---|---|
| image | image |

# Release notes

<changed> 
For testing during development, the `UISwitch` testability is now enabled for external tools like Appium.
</changed>

# Ticket

<ticket>
COIOS-808
</ticket>


erenbesel (approved):

Adyen/UI/Form/Items/Toggle/FormToggleItemView.swift
```diff
@@ -36,7 +36,7 @@ public final class FormToggleItemView: FormItemView<FormToggleItem> {
         switchControl.translatesAutoresizingMaskIntoConstraints = false
         switchControl.isOn = item.value
         switchControl.onTintColor = item.style.tintColor
-        switchControl.isAccessibilityElement = false
```
goergisn, line 39: Shouldn't the current default values stay that way?

atmamont: My thinking was that we make it testable by default and then VoiceOver notification triggers logic change. Do you see any potential hiccups here?

goergisn: Makes sense 👌

goergisn (approved):

PR #1875, opened 2024-10-21
COIOS-811: Expiry date sanitization fix

goergisn:
# Summary
- Fixed a bug where the expiry date field ended up in an invalid state when trying to add additional characters to the textfield

### What steps will reproduce the problem?

1. Enter an expiry date 03 / 30 (It jumps to the next field CVC)
2. Go back into the expiry date field and add another character e.g. 5

### What is the expected result?
- The character is ignored (sanitized out)
### What happens instead?
- It does not show it in the UI but the value of the item is now 03305 and errors out but there is no visual indicator why

# Release notes

<fixed>
In the **Expiry date** field, when the shopper tries to enter additional characters to the field after having already entered characters in it, the field no longer ends up in an invalid state.
</fixed>

# Ticket

<ticket>
COIOS-811
</ticket>


atmamont (dismissed):
Is it possible to add new unit test for this fix?

goergisn:
> Is it possible to add new unit test for this fix?

Mhh can you come up with an additional one to the one I added in the "CardExpiryDateFormatterTests"?

atmamont:
> > Is it possible to add new unit test for this fix?
> 
> Mhh can you come up with an additional one to the one I added in the "CardExpiryDateFormatterTests"?

I guess I should not look into the code today. Sorry Alex I haven't noticed tests were updated! 👍

nauaros (dismissed):

erenbesel (dismissed):

nauaros (dismissed):

atmamont (dismissed):

atmamont (approved):

nauaros (approved):

PR #1876, opened 2024-10-23
COIOS-774: Pay by Bank - Confirmation Sheet

goergisn:
# Summary

- Adds a confirmation sheet in front of SFSafariViewController and triggers the instant payment method only after triggering the submit button

# Demo

[Screencast](https://github.com/user-attachments/assets/3805b25c-6dd9-4e9e-89a0-2e1bc79471e9)

# Ticket

<ticket>
COIOS-774
</ticket>


erenbesel (approved):

AdyenComponents/PayByBank/US/PayByBankUSComponent+ConfirmationViewController.swift
```diff
@@ -0,0 +1,188 @@
...
+            contentStack.alignment = .center
+            contentStack.setCustomSpacing(10, after: subtitleLabel)
+            
+            let contentButtonStack = UIStackView(arrangedSubviews: [
+                contentStack,
+                submitButton
+            ])
+            contentButtonStack.spacing = 32
+            contentButtonStack.axis = .vertical
+            contentButtonStack.alignment = .fill
```
atmamont, line 95: It's default value, isn't it?

AdyenComponents/PayByBank/US/PayByBankUSComponent+ConfirmationViewController.swift
```diff
@@ -0,0 +1,188 @@
...
+            
+            view.addSubview(contentButtonStack)
+            contentButtonStack.adyen.anchor(inside: view.layoutMarginsGuide, with: .init(top: 16, left: 0, bottom: 8, right: 0))
```
atmamont: Ideally this should go to some style constants in the future to avoid these magic numbers and make style changes easier.

AdyenComponents/PayByBank/US/PayByBankUSComponent+ConfirmationViewController.swift
```diff
@@ -0,0 +1,188 @@
...
+        public required init?(coder: NSCoder) {
+            fatalError("init(coder:) has not been implemented")
+        }
+        
+        @objc private func submitTapped() {
+            submitButton.showsActivityIndicator = true
+            model.continueHandler()
+        }
+        
+        override public var preferredContentSize: CGSize {
```
atmamont, line 123: What happens if we don't provide this manually?

AdyenComponents/PayByBank/US/PayByBankUSComponent+ConfirmationViewController.swift
```diff
@@ -0,0 +1,188 @@
...
+        private var imageLoadingTask: AdyenCancellable? {
+            willSet { imageLoadingTask?.cancel() }
+        }
+        
+        internal init(
+            headerImageUrl: URL,
+            style: PayByBankUSComponent.Style,
+            continueHandler: @escaping () -> Void
+        ) {
+            let message: [String] = [
```
atmamont: Should this be localized?

AdyenComponents/PayByBank/US/PayByBankUSComponent+ConfirmationViewController.swift
```diff
@@ -0,0 +1,188 @@
...
+                "By connecting your bank account you are authorizing debits",
+                "to your account for any amount owed for use of our services",
+                "and/or purchase of our products, until this authorization is revoked."
+            ]
+            
+            self.headerImageUrl = headerImageUrl
+            self.title = "Pay by Bank"
```
atmamont: Same here and further down

Adyen/Core/Payment Methods/PayByBankUSPaymentMethod.swift
```diff
@@ -0,0 +1,26 @@
+//
+// Copyright (c) 2024 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+/// PayByBank US payment.
```
nauaros: Can you add method at the end?

Adyen/Core/Payment Methods/StoredPayByBankUSPaymentMethod.swift
```diff
@@ -6,8 +6,8 @@
 
 import Foundation
 
-/// Stored PayByBank payment.
-public struct StoredPayByBankPlaidPaymentMethod: StoredPaymentMethod {
+/// Stored PayByBank US payment.
```
nauaros: Same here

AdyenDropIn/Utilities/ComponentManager+PaymentComponentBuilder.swift
```diff
@@ -419,7 +433,7 @@ extension ComponentManager: PaymentComponentBuilder {
         )
     }
 
-    private func createBoletoComponent(_ paymentMethod: BoletoPaymentMethod) -> BoletoComponent {
+    func createBoletoComponent(_ paymentMethod: BoletoPaymentMethod) -> BoletoComponent {
```
atmamont, line 436: We have a linter rule to prevent us from NOT having a scope defined, right?

atmamont (approved):
Easy to read and review, thanks 👍

atmamont:
⚠️ This is a fully breaking change

Adyen/Core/Payment Methods/StoredPayByBankUSPaymentMethod.swift
```diff
@@ -6,8 +6,8 @@
...
+public struct StoredPayByBankUSPaymentMethod: StoredPaymentMethod {
```
atmamont, line 10: ⚠️ This is a fully breaking change, can we keep struct and maybe create a typealias?

AdyenComponents/PayByBank/US/PayByBankUSComponent+ConfirmationViewController.swift
goergisn: Default value is .center

Adyen/Core/Payment Methods/StoredPayByBankUSPaymentMethod.swift
atmamont: I found `StoredPayByBankPlaidPaymentMethod` was added in the base branch. Let's rethink which branch to target when we run api diff? Options:
- develop (then we might have all long-living branch changes)
- latest release tag (then every time we have all development changes combined)

Or we can switch off api diff for all PRs not targeting develop and deal with all changes when we merge long-living branch into develop. The only concern I have is that usually it's a common sentiment to merge as fast as possible because "everything was reviewed", everyone has seen the code multiple times and naturally you just want to merge it asap.

AdyenComponents/PayByBank/US/PayByBankUSComponent+ConfirmationViewController.swift
atmamont: Oh, interesting! I just checked it [here](https://developer.apple.com/documentation/uikit/uistackview/1616243-alignment)
image

AdyenComponents/PayByBank/US/PayByBankUSComponent+ConfirmationViewController.swift
goergisn: Good catch 😬

Adyen/Core/Payment Methods/StoredPayByBankUSPaymentMethod.swift
goergisn: Yeah that's a tricky one - also thought about it.
I want to see the changes to the target branch but also agree that it would be nice to see the full change set to develop/release.
We could run it twice if the target branch is not develop and we're not in a release branch - so it creates 2 comments (one to the target branch and one to develop) - 🤔

AdyenDropIn/Utilities/ComponentManager+PaymentComponentBuilder.swift
goergisn: It does not catch here for some reason - it's in a private extension

atmamont: Maybe we can turn swiftlint warnings into errors so every time one doesn't need to even focus on this during code review.

AdyenComponents/PayByBank/US/PayByBankUSComponent+ConfirmationViewController.swift
goergisn: it will not get the height it needs

PR #1877, opened 2024-10-24
COIOS-774: Add logos as option for trailing view in ListItem

goergisn:
# Summary
- ListItem now supports an additional case `Logos` as trailing view

# Demo

(This view is not part of the PR but showcases how it would look like)
Screenshot 2024-10-23 at 16 23 54

# Ticket

<ticket>
COIOS-774
</ticket>

### DIFF
--- a/Adyen/Core/Models/DisplayInformation.swift
+++ b/Adyen/Core/Models/DisplayInformation.swift
@@ -11,7 +11,7 @@ public struct DisplayInformation: Equatable {
 
     public enum TrailingInfoType: Equatable {
         case text(String)
-        case logos(named: [String])
+        case logos(named: [String], trailingText: String?)
         
         internal var accessibilityLabel: String? {
             switch self {
--- a/Adyen/Core/Models/DisplayInformation.swift
+++ b/Adyen/Core/Models/DisplayInformation.swift
@@ -9,6 +9,18 @@ import Foundation
 /// Describes a payment method display information.
 public struct DisplayInformation: Equatable {
 
+    public enum TrailingInfoType: Equatable {
+        case text(String)
+        case view(UIView)
+        
+        var accessibilityLabel: String? {
+            switch self {
+            case let .text(text): return text
+            case .view: return nil
+            }
+        }
+    }
+    
     /// The title for the payment method, adapted for displaying in a list.
     /// In the case of stored payment methods, this will include information identifying the stored payment method.
     /// For example, this could be the last 4 digits of the card number, or the used email address.
@@ -23,9 +35,9 @@ public struct DisplayInformation: Equatable {
     @_spi(AdyenInternal)
     public let logoName: String
 
-    /// The trailing disclosure text.
+    /// The trailing info element
     @_spi(AdyenInternal)
-    public let disclosureText: String?
+    public let trailingInfo: TrailingInfoType?
 
     /// The footnote if any.
     @_spi(AdyenInternal)
@@ -35,7 +47,7 @@ public struct DisplayInformation: Equatable {
     @_spi(AdyenInternal)
     public let accessibilityLabel: String?
 
-    /// Initializes a`DisplayInformation`.
+    /// Initializes a `DisplayInformation`
     ///
     /// - Parameter title: The title.
     /// - Parameter subtitle: The subtitle.
@@ -44,6 +56,7 @@ public struct DisplayInformation: Equatable {
     /// - Parameter footnoteText: The footnote text if any.
     /// - Parameter accessibilityLabel: An optional custom `accessibilityLabel` to use.
     /// Set this if the title / subtitle might not be sufficient enough to provide a good accessibility
+    @available(*, deprecated, renamed: "init(title:subtitle:logoName:trailingInfo:footnoteText:accessibilityLabel:)")
     public init(
         title: String,
         subtitle: String?,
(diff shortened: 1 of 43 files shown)


Adyen/Core/Models/DisplayInformation.swift
```diff
@@ -44,18 +56,37 @@ public struct DisplayInformation: Equatable {
     /// - Parameter footnoteText: The footnote text if any.
     /// - Parameter accessibilityLabel: An optional custom `accessibilityLabel` to use.
     /// Set this if the title / subtitle might not be sufficient enough to provide a good accessibility
+    @available(*, deprecated, renamed: "init(title:subtitle:logoName:trailingInfo:footnoteText:accessibilityLabel:)")
```
erenbesel: Not sure about this deprecation. What happens if they use the old deprecated method and set something for the disclosure text property which won't exist anymore?

goergisn: the `disclosureText` was `@_spi(AdyenInternal)` so there is no other publicly noticable change than the deprecation

erenbesel: The proerty yes, but I meant what happens if it's set on init, expecting to see some disclosure text but nothing will be visible?

Adyen/UI/Views/SupportedPaymentMethodsView.swift
```diff
@@ -0,0 +1,88 @@
...
+            ])
+            imageView.load(url: url, using: imageLoader)
+            return imageView
+        }
+        
+        let label = UILabel()
+        label.text = trailingText
+        label.isHidden = (trailingText ?? "").isEmpty
+        label.setContentHuggingPriority(.required, for: .horizontal)
+        label.font = .preferredFont(forTextStyle: .callout)
```
descorp, line 80: Should this come from styling configuration?

Adyen/UI/Views/SupportedPaymentMethodsView.swift
```diff
@@ -0,0 +1,88 @@
...
+        label.textColor = UIColor.Adyen.componentSecondaryLabel
```
descorp: Should this come from styling configuration?

Adyen/Core/Models/DisplayInformation.swift
```diff
@@ -44,18 +56,37 @@ public struct DisplayInformation: Equatable {
...
+            footnoteText: footnoteText,
+            accessibilityLabel: accessibilityLabel
+        )
+    }
+    
+    public init(
+        title: String,
+        subtitle: String?,
+        logoName: String,
+        trailingInfo: TrailingInfoType?,
```
nauaros, line 83: What do you think of having `nil` as the default value?

goergisn: I wanted to but then it clashes with the other init as it has the same signature if you remove the default values and the compiler cannot decide which ones to use

Adyen/Core/Models/DisplayInformation.swift
goergisn: It will be `trailingInfo: disclosureText.map { .text($0) },`

Tests/IntegrationTests/UIKit/Views/SupportedPaymentMethodLogosViewTests.swift
```diff
@@ -0,0 +1,41 @@
...
+        let trailingText = "TRAILING_TEXT"
+        let supportedLogosView = SupportedPaymentMethodLogosView(
+            imageUrls: [URL(string: "https://adyen.com")!],
+            trailingText: trailingText
+        )
+        XCTAssertEqual(supportedLogosView.subviews.count, 0)
+        XCTAssertNil(supportedLogosView.content)
+        
+        let viewController = ADYViewController(view: supportedLogosView)
+        setupRootViewController(viewController) // Providing a window
```
atmamont: Could you thy another approach that I used to convert UI tests to integration tests actually to remove relying on mounting the whole view hierarchy into a real window? This will speed up the test significantly.

atmamont (dismissed):

erenbesel (dismissed):

atmamont (dismissed):

atmamont (approved):

Adyen/UI/List/ListItemView.swift
```diff
@@ -90,6 +89,37 @@ public final class ListItemView: UIView, AnyFormItemView {
...
+            trailingView.isHidden = string.isEmpty
+        case let .logos(urls, trailingText):
+            let trailingLogosView = SupportedPaymentMethodLogosView(
+                imageUrls: urls,
+                trailingText: trailingText
+            )
+            trailingLogosView.accessibilityIdentifier = item?.identifier.map {
+                ViewIdentifierBuilder.build(scopeInstance: $0, postfix: "trailingLogosView")
+            }
+            trailingView = trailingLogosView
```
erenbesel, line 114: trailingView may be hidden from a previous case I think. We can either set it visible at the top of the function or set it here (and maybe even add tests to make sure?)

goergisn: I will write tests for it once it gets used by the PayByBank payment method - right now the new view is not in use yet 👌

erenbesel (approved):

PR #1879, opened 2024-10-25
COIOS-802: Identify native redirect flow (v5)

nauaros:
## Summary
This development enables us to differentiate between regular and native redirects. To handle cases where native redirects fail (indicated by `nativeRedirectData` being `nil`), it's essential to track the originating flow type of each redirect.

## Motivation
Native redirect flows can occasionally fail if `nativeRedirectData` is `nil` within the action object. Currently, we handle this by discarding the native redirect (checking if `nativeRedirectData` is nil), and defaulting to a "direct issuer flow" using a `/details` call.

To avoid the additional steps of the "direct issuer flow," we can address this issue on the backend. By identifying the native redirect flow, we can still retrieve the native redirect result directly.

# Release notes

<new> 

- Added support for distinguishing between regular and native redirects to improve handling of failed native redirect flows.

</new>

# Ticket

<ticket>
COIOS-802
</ticket>


AdyenActions/Actions/RedirectAction.swift
```diff
@@ -7,26 +7,21 @@
...
+
     /// The URL to which to redirect the user.
     public let url: URL
-    
+
     /// The server-generated payment data that should be submitted to the `/payments/details` endpoint.
     public let paymentData: String?
-    
-    /// Native redirect data.
-    public let nativeRedirectData: String?
```
erenbesel, line 19: This is a breaking change

AdyenActions/Actions/RedirectAction.swift
```diff
@@ -7,26 +7,21 @@
 import Foundation
 
 /// Describes an action in which the user is redirected to a URL.
-public struct RedirectAction: Decodable {
-    
+public class RedirectAction: Decodable {
```
erenbesel: Any reason to switch to class?

AdyenActions/Actions/RedirectAction.swift
```diff
@@ -9,24 +9,68 @@ import Foundation
 /// Describes an action in which the user is redirected to a URL.
 public struct RedirectAction: Decodable {
     
+    /// Defines the type of redirect flow utilized by the `RedirectAction` object.
+    public enum RedirectType: Decodable {
```
goergisn:
```suggestion
    public enum RedirectType: String, Decodable {
```

AdyenActions/Actions/RedirectAction.swift
```diff
@@ -9,24 +9,68 @@ import Foundation
...
+        case redirect
+        case nativeRedirect
+
+        public init(from decoder: Decoder) throws {
+            let container = try decoder.singleValueContainer()
+            let type = try? container.decode(String.self)
+            switch type {
```
goergisn: self = RedirectType(rawValue: type) ?? .redirect

goergisn (dismissed):

AdyenActions/Actions/RedirectAction.swift
```diff
@@ -9,24 +9,64 @@ import Foundation
...
     /// The URL to which to redirect the user.
     public let url: URL
-    
+
     /// The server-generated payment data that should be submitted to the `/payments/details` endpoint.
     public let paymentData: String?
-    
+
+    /// Redirect type.
+    public let type: RedirectType
```
erenbesel: Do we need to expose the property itself? We could keep it internal at the start

nauaros: Made it private

goergisn (dismissed):

AdyenActions/Actions/RedirectAction.swift
```diff
@@ -9,24 +9,63 @@ import Foundation
 /// Describes an action in which the user is redirected to a URL.
 public struct RedirectAction: Decodable {
     
+    /// Defines the type of redirect flow utilized by the `RedirectAction` object.
+    public enum RedirectType: String, Decodable {
+        // swiftlint:disable redundant_string_enum_value
+        case redirect = "redirect"
+        case nativeRedirect = "nativeRedirect"
```
descorp, line 16: Optional:

```suggestion
        case redirect
        case nativeRedirect
```

AdyenActions/Actions/RedirectAction.swift
```diff
@@ -25,17 +40,37 @@ public struct RedirectAction: Decodable {
...
-    ///   - nativeRedirectData: Native redirect data.
+    ///   - type: The redirect flow  used by the action. Defaults to `redirect`.
+    ///   - nativeRedirectData: Native redirect data. Defaults to `nil`.
     ///   - paymentMethodType: The type of the payment method.
     public init(
         url: URL,
         paymentData: String?,
+        type: RedirectType = .redirect,
         nativeRedirectData: String? = nil,
         paymentMethodType: String? = nil
```
nauaros, line 50: @erenbesel can you review the changes here? This includes your latest addition of `paymentMethodType`.

AdyenActions/Actions/RedirectAction.swift
```diff
@@ -25,17 +40,39 @@ public struct RedirectAction: Decodable {
...
+        self.type = type
         self.nativeRedirectData = nativeRedirectData
         self.paymentMethodType = paymentMethodType
     }
+
+    public init(from decoder: any Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.url = try container.decode(URL.self, forKey: .url)
+        self.paymentData = try container.decodeIfPresent(String.self, forKey: .paymentData)
+        self.type = try container.decode(RedirectType.self, forKey: .type)
```
goergisn: are we sure the RedirectType will always be returned?

otherwise we could do:

```suggestion
        self.type = (try? container.decode(RedirectType.self, forKey: .type)) ?? FALLBACK_VALUE
```

goergisn (dismissed):

AdyenActions/Actions/RedirectAction.swift
```diff
@@ -9,12 +9,27 @@ import Foundation
 /// Describes an action in which the user is redirected to a URL.
 public struct RedirectAction: Decodable {
     
+    /// Defines the type of redirect flow utilized by the `RedirectAction` object.
+    public enum RedirectType: String, Decodable {
+        // swiftlint:disable redundant_string_enum_value
```
erenbesel: Why do we need to add the string values when they are matching?

atmamont: just a guess - to not break anything if someone decides to rename enum case in the future?

AdyenActions/Actions/RedirectAction.swift
atmamont: Is there any other way to check its optionality? https://docs.adyen.com/api-explorer/Checkout/71/post/payments#responses-200-action-CheckoutRedirectAction-type

Tests/IntegrationTests/Actions Tests/Redirect/RedirectComponentTests.swift
```diff
@@ -292,7 +292,12 @@ class RedirectComponentTests: XCTestCase {
         }
         delegate.onDidFail = { _, _ in XCTFail("Should not call onDidFail") }
         
-        let action = RedirectAction(url: URL(string: "https://google.com")!, paymentData: nil, nativeRedirectData: "test_nativeRedirectData")
+        let action = RedirectAction(
+            url: URL(string: "https://google.com")!,
+            paymentData: nil,
+            type: .nativeRedirect,
+            nativeRedirectData: "test_nativeRedirectData"
```
atmamont, line 299: [Documentation](https://docs.adyen.com/api-explorer/Checkout/71/post/payments#responses-200-action-CheckoutNativeRedirectAction-nativeRedirectData) says it's just a string with no further explanation. Is this an URL or anything else? Should we have tests for this string?

atmamont (dismissed):

AdyenActions/Actions/RedirectAction.swift
nauaros: I double checked with the backend team. This property is always coming.

AdyenActions/Actions/RedirectAction.swift
nauaros: I like adding them to make it more explicit these are the values expected. But indeed, there's no need for it. I removed them.

goergisn (approved):

neelSharma12 (approved):

erenbesel (approved):

PR #1884, opened 2024-10-28
Splitting responsibilities of analytics provider

erenbesel:
# Summary

Splits analytics provider responsibilities into two by moving the conditional follow up events on its own class `EventAnalyticsProvider` with the timer/batching logic.
This object is only created by the context if the configuration is turned on. 
The original analytics provider is now only responsible for the initial call, is not aware of any of the events provider's lifetime and only passes the calls to it.

# Ticket

<ticket>
COIOS-815
</ticket>


Adyen/Core/AdyenContext/AdyenContext.swift
```diff
@@ -67,13 +67,23 @@ public final class AdyenContext: PaymentAware {
             )
         else { return nil }
         
-        let eventDataSource = AnalyticsEventDataSource()
-        let syncEventDataSource = ThreadSafeAnalyticsEventDataSource(dataSource: eventDataSource)
+        var eventAnalyticsProvider: AnyEventAnalyticsProvider?
+        
+        
+        if analyticsConfiguration.isEnabled {
```
atmamont, line 72: Nice!

atmamont (dismissed):

goergisn (dismissed):

Tests/UnitTests/Analytics/EventAnalyticsProviderTests.swift
```diff
@@ -0,0 +1,179 @@
...
+        }
+        
+        let sut = createSUT(apiClient: apiClient)
+
+        sut.sendEventsIfNeeded()
+        
+        wait(for: [expectation], timeout: 0.1)
+    }
+
+    func testShouldNotSendEventWhenAttemptIdButNoEvents() {
```
atmamont: I am missing some verbs here :)
```suggestion
func testShouldNotSendEventWhenNoEventsAreStored()
```

The name also suggested the attempt id was set but I see no specific `checkoutAttemptId` setting in the preparation stage so I removed this from the test name.

Tests/UnitTests/Analytics/EventAnalyticsProviderTests.swift
```diff
@@ -0,0 +1,179 @@
...
+    
+    func testEventShouldSendWhenAttemptIdAndEventsExist() {
+        let apiClient = APIClientMock()
+        let sut = createSUTWithSuccessMock(apiClient: apiClient)
+        
+        let infoEvent = AnalyticsEventInfo(component: "card", type: .rendered)
+        sut.add(info: infoEvent)
+        
+        let shouldSendExpectation = expectation(description: "send event is called")
+        let expectedId = infoEvent.id
```
atmamont, line 63: Is this `expectedCheckoutAttemptId` or a local event id?

Tests/UnitTests/Core/AdyenContextTests.swift
```diff
@@ -54,6 +54,31 @@ class AdyenContextTests: XCTestCase {
         let context = AdyenContext(apiContext: apiContext, payment: Dummy.payment)
         XCTAssertNil(context.analyticsProvider)
     }
+    
+    func testBothAnalyticsProviderShouldBeCreated() {
```
atmamont, line 58: This "both" triggers a set of questions to anyone seeing this for the first time. Why do we have two analytic providers? How do I know which one to use?

Do you think we can improve this even more by extracting initial analytics call to a separate entity? Then even the variable name would give a lot more context (like `initialDataAnalyticsProvider`)

atmamont (dismissed):
Nice job Eren! 🥂 
Some comments regarding tests readability.

goergisn (dismissed):
Looks good!

Tests/UnitTests/Analytics/EventAnalyticsProviderTests.swift
erenbesel: Good catch! They are set on the createSUT function

Tests/UnitTests/Analytics/EventAnalyticsProviderTests.swift
erenbesel: It's the id of that info event, no relation to the attempt id

Tests/UnitTests/Core/AdyenContextTests.swift
erenbesel: Good point on the first part. For the 2nd part, could be another next step but not entirely sure how to keep the components agnostic. Right now all the component knows is having an `AnyAnalyticsProvider`

Tests/UnitTests/Analytics/EventAnalyticsProviderTests.swift
atmamont: 🤔 

Then I have to admit it's not clear when do you set a `checkoutAttemptId` itself as multiple test names suggest it should be either set of nil.

erenbesel: Hmm, since attempt id is needed for all these tests, it's set on the helper function.
What part is confusing?

Tests/UnitTests/Core/AdyenContextTests.swift
atmamont: I see the problem, right!

I would play around with completely extracting that single call into a separate entity and calling it once. It would allow to declutter **main** analytics provider (that could be something like `AnalyticsCoordinator`, this is a long shot, I just try to show my thinking). 

By doing so you can hide `UniqueAssetAPIClient` and `InitialAnalyticsRequest` inside that entity hence declutter main analytics provider.
Apart from that I guess we could try to get rid of `func sendInitialAnalytics()` interface method and just send this analytics upon the new entity instantiation, I assume we have all necessary data to create the initial request and we don't actually need to wait for any additional input or network response.

I suggest not making any more changes to this PR, this is quite good and I just try to explore more ideas here for the future!

Tests/UnitTests/Analytics/EventAnalyticsProviderTests.swift
atmamont: I guess then all cases when you assume that no `checkoutAttemptId` is set are a bit obscure. I see there are two helpers: `createSUT()` and `createSUTWithSuccessMock()`

Let's add attempt id to the interface so it is super clear when you read the test?

For example:
func createSUT(checkoutAttemptId: String?) -> EventAnalyticsProvider {
}

Tests/UnitTests/Analytics/EventAnalyticsProviderTests.swift
atmamont: I replied in the comment above.

atmamont (approved):

goergisn (approved):

PR #1885, opened 2024-10-29
Card number text field fix

goergisn:
# Summary
- The cursor position now stays in place even when the formatting type changes

## Tests
- Enter card number (character by character) that's longer than the pan length -> should call `shouldCallNotifyDelegateOfMaxLength` **once**
- Paste card number that's exactly the the pan length -> should call `shouldCallNotifyDelegateOfMaxLength` **once**
- Deleting card number that's longer than the pan length (character by character) -> should call `shouldCallNotifyDelegateOfMaxLength` **once**
- Reaching `shouldCallNotifyDelegateOfMaxLength` should trigger the `FormViewController` to move `firstResponder` to the next field

## Background
- There is an observation on `formattedValue` which caused the text of the textField to just be replaced with that value
- This caused the cursor to jump to the end of the textfield

## Fix
- Whenever the formattedValue changes we just notify about max length on the card number field

## Also
- Reverted max length check to what it was before (Not super nice but no change in unproblematic behavior)

# Ticket

<ticket>
COIOS-000
</ticket>

### DIFF
--- a/Adyen/UI/Form/Items/Text/FormTextItemView.swift
+++ b/Adyen/UI/Form/Items/Text/FormTextItemView.swift
@@ -46,8 +46,7 @@ open class FormTextItemView<ItemType: FormTextItem>: FormValidatableValueItemVie
 
         bind(item.$placeholder, to: textField, at: \.placeholder)
         observe(item.$formattedValue) { [weak self] newValue in
-            self?.textField.text = newValue
-            self?.updateValidationStatus()
+            self?.handleFormattedValueDidChange(newValue)
         }
         
         updateValidationStatus()
@@ -159,6 +158,11 @@ open class FormTextItemView<ItemType: FormTextItem>: FormValidatableValueItemVie
     }
     
     // MARK: - Private
+    
+    open func handleFormattedValueDidChange(_ newValue: String) {
+        textField.text = newValue
+        updateValidationStatus()
+    }
     
     @_spi(AdyenInternal)
     @objc open func textDidChange(textField: UITextField) {
@@ -272,7 +276,7 @@ open class FormTextItemView<ItemType: FormTextItem>: FormValidatableValueItemVie
     
     public func notifyDelegateOfMaxLengthIfNeeded() {
         let maximumLength = item.validator?.maximumLength(for: item.value) ?? .max
-        if item.value.count >= maximumLength {
+        if item.value.count == maximumLength {
             delegate?.didReachMaximumLength(in: self)
         }
     }
--- a/Tests/IntegrationTests/Card Tests/Items/CardNumberItem/FormCardNumberItemViewTests.swift
+++ b/Tests/IntegrationTests/Card Tests/Items/CardNumberItem/FormCardNumberItemViewTests.swift
@@ -124,28 +124,42 @@ class FormCardNumberItemViewTests: XCTestCase {
     
     func test_notifyDelegateOfMaxLengthIfNeeded() {
         
+        let panLength = 5
+        
         let cardNumberValidator = CardNumberValidator(
             isLuhnCheckEnabled: true,
             isEnteredBrandSupported: true,
-            panLength: 5
+            panLength: panLength
         )
 
         let sut = setupSut(validator: cardNumberValidator)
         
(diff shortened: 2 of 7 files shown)


Adyen/UI/Form/Items/Text/FormTextItemView.swift
```diff
@@ -272,7 +276,7 @@ open class FormTextItemView<ItemType: FormTextItem>: FormValidatableValueItemVie
     
     public func notifyDelegateOfMaxLengthIfNeeded() {
         let maximumLength = item.validator?.maximumLength(for: item.value) ?? .max
-        if item.value.count >= maximumLength {
+        if item.value.count == maximumLength {
```
descorp, line 279: No chance that `item.value` will be longer? Or sanitisation will clean exceeding?

nauaros (dismissed):

Adyen/UI/Form/Items/Text/FormTextItemView.swift
atmamont: Yeah I recall previous change had the truncation logic but the overall complexity for this feature seems high.

Tests/IntegrationTests/Card Tests/Items/CardNumberItem/FormCardNumberItemViewTests.swift
```diff
@@ -130,8 +119,50 @@ class FormCardNumberItemViewTests: XCTestCase {
...
+        
+        let expectation = expectation(description: "Handle did reach maximum length was called once")
+        
+        let delegate = FormTextItemViewDelegateMock<FormCardNumberItem, FormCardNumberItemView>()
+        delegate.handleDidReachMaximumLength = { itemView in
+            expectation.fulfill()
+        }
+        sut.delegate = delegate
+        
+        "5555341244".forEach { character in
```
atmamont: Let's also add a case for adding the whole string from the pasteboard?

Adyen/UI/Form/Items/Text/FormTextItemView.swift
goergisn: This is just reverting it to how it was before - what it does it not great but I did not want to alter the behavior.
(Ideally it would truncate it and only allow a value up to the maximumLength - but that was never implemented)

nauaros (dismissed):

Tests/IntegrationTests/Card Tests/Items/CardNumberItem/FormCardNumberItemViewTests.swift
```diff
@@ -124,28 +124,42 @@ class FormCardNumberItemViewTests: XCTestCase {
...
         "5555341244".forEach { character in
             sut.textField.text = sut.item.value + String(character)
             sut.textDidChange(textField: sut.textField)
         }
         
-        wait(for: [expectation], timeout: 0.1)
+        let pasteSetup = setupDelegateExpectation()
+        
+        // Testing setting the text with the exact panLength
+        sut.textField.text = (0..<panLength).reduce("") { $0 + "\($1)" }
```
atmamont: Should we also try pasting a string longer than allowed length?

nauaros (dismissed):

Tests/IntegrationTests/Card Tests/Items/CardNumberItem/FormCardNumberItemViewTests.swift
```diff
@@ -179,4 +241,31 @@ private extension FormCardNumberItemViewTests {
         item.formatter = formatter
         return FormCardNumberItemView(item: item)
     }
+    
+    func setupSingleCallDidReachMaximumLengthExpectation(
```
atmamont:
```suggestion
// MARK: - Helpers

func makeSingleCallDidReachMaximumLengthExpectation(
```

atmamont (dismissed):

nauaros (dismissed):

atmamont (dismissed):
:shipit:

atmamont (approved):

nauaros (approved):

PR #1901, opened 2024-11-18
Send a new field to determine the analytics level in the initial call.

erenbesel:
# Summary
Send a new field called `level` on the inital analytics call to determine the level. As of now it's only `initial` or `all` based on the existing configuration flag.

# Ticket

<ticket>
COIOS-817
</ticket>


Adyen/Analytics/Models/AdyenAnalytics.swift
```diff
@@ -74,3 +74,20 @@ public struct AdditionalAnalyticsFields {
         self.sessionId = sessionId
     }
 }
+
+/// Describes the levels that determine which analytics calls are made.
+internal enum AnalyticsLevel: String, Encodable {
+    
+    /// Indicates all analytics are enabled.
+    case all
```
goergisn, line 82: Some ideas for naming:
- full
- complete
- extensive

goergisn (dismissed):

goergisn (approved):

nauaros (approved):

PR #1903, opened 2024-11-20
COIOS-806: Support all Boleto Bancario tx-variants

nauaros:
# Summary

<new>

Add support for the following Boleto tx-variants:

- Boleto Bancario Itaú (`boletoBancarioItau`)
- PrimeiroPay Boleto (`primeiroPayBoleto`)
- Boleto Bancario (`boletoBancario`)

</new>

# Ticket

<ticket>
COIOS-806
</ticket>


Adyen/Core/Payment Methods/Abstract/PaymentMethodType.swift
```diff
@@ -37,7 +37,10 @@ public enum PaymentMethodType: RawRepresentable, Hashable, Codable {
     case econtextStores
     case econtextATM
     case econtextOnline
-    case boleto
+    case boletoBancario
```
descorp: Technically, this is a breaking change.
Maybe keep the old one as-is, just mark as deprecated?

descorp (dismissed):

Adyen/Core/Payment Methods/Abstract/PaymentMethodType.swift
nauaros: As discussed, we keep `boleto` but mark it as deprecated

Adyen/Core/Payment Methods/Abstract/PaymentMethodType.swift
```diff
@@ -37,7 +37,17 @@ public enum PaymentMethodType: RawRepresentable, Hashable, Codable {
     case econtextStores
     case econtextATM
     case econtextOnline
+    case boletoBancario
+    @available(
+        *,
+        deprecated,
+        renamed: "boletoBancarioSantander",
```
Robert-SD, line 44: Let's guide merchants regarding the right tx variant. In case of Santander, let's suggest to use  boletoBancarioSantander otherwise to the default boletoBancario.

Robert-SD (dismissed):

goergisn (approved):

Adyen/Core/Payment Methods/Abstract/PaymentMethodType.swift
```diff
@@ -37,7 +37,17 @@ public enum PaymentMethodType: RawRepresentable, Hashable, Codable {
...
+        message: "Use 'boletoBancarioSantander' instead. Consider 'boletoBancario' as the default payment method for Boleto Bancario."
```
erenbesel, line 45: The message text is a bit confusing starting from the `Consider...` part. As in, if boletoBancario is default why is it renamed type of confusion.

erenbesel (approved):
Approved with a tiny comment on deprecation message

Adyen/Core/Payment Methods/Abstract/PaymentMethodType.swift
nauaros: There are two issues here:
1. What was named as `boleto` before, in reality represents the tx-variant "boletoBancarioSantander". To prevent confusion and mismatch we decided to deprecate and rename into `boletoBancarioSantander`.
2. Since we were only offering `boletoBancarioSantander`, merchants had one option only so that's always their default tx-variant. However, from the business side, merchants are always encouraged to use `boletoBancario` since its success rate is higher. (this information came from @Robert-SD who's in touch with IMs working with Boleto)

PR #1913, opened 2024-12-04
Allow setting `maxIssuerNumber` for Twint via Drop-In configuration

goergisn:
# Summary
- The `maxIssuerNumber` for Twint can now be set when using Drop-In
- Statically linking the `TwintSDK` to prevent name clashing when using another library that includes the `TwintSDK` as well (e.g. Datatrans)

# Release notes

<new>
- When using Drop-In: The `maxIssuerNumber` for Twint can now be set
</new>

# Ticket

<ticket>
COIOS-000
</ticket>


atmamont (dismissed):

Package.swift
```diff
@@ -182,7 +182,7 @@ let package = Package(
         ),
         .binaryTarget(
             name: "TwintSDK",
-            path: "XCFramework/Dynamic/TwintSDK.xcframework"
+            path: "XCFramework/Static/TwintSDK.xcframework"
```
atmamont: Let's not rush and extract this to a separate change? Is this the official SPM way to change the packaging type by adjusting the _path_?

goergisn: We're (already) shipping both frameworks - now it points to the static instead of the dynamic one

goergisn: I reverted the change so we can merge it

atmamont (approved):

erenbesel (approved):

nauaros (approved):

PR #1918, opened 2024-12-12
Error events for 3ds2/redirect/encryption

erenbesel:
# Summary

Several error events for:

- [x] ThreeDS2 flow (including the fingerPrintSubmit API error)
- [x] Redirect
- [x] Encryption

# Ticket

<ticket>
COIOS-841
</ticket>


Adyen/Analytics/AnalyticsProvider/EventAnalyticsProvider.swift
```diff
@@ -16,7 +16,14 @@ internal final class EventAnalyticsProvider: AnyEventAnalyticsProvider {
         static let errorLimit = 5
     }
     
-    internal var checkoutAttemptId: String?
+    internal var checkoutAttemptId: String? {
+        didSet {
+            if checkoutAttemptId != nil {
+                startNextTimer()
```
atmamont, line 22: This is a bit bigger than the scope of this. Any reason to change this?
I quickly looked through tests changes and didn't find any new/changed tests for this part of logic.

AdyenCard/Components/Card/CardComponentExtensions.swift
```diff
@@ -58,9 +58,19 @@ extension CardComponent {
 
             submit(data: data)
         } catch {
+            sendEncryptionErrorEvent()
```
goergisn, line 61: are we not interested in the actual error message?

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -462,9 +462,19 @@ extension GiftCardComponent {
                 storePaymentMethod: false
             ))
         } catch {
+            sendEncryptionErrorEvent()
```
goergisn, line 465: Same here - is the error description irrelevant?

goergisn (dismissed):
Looks good - just double checking

AdyenActions/Components/3DS2/Action handlers/3DS2 Without Delegated Authentication/ThreeDS2CoreActionHandler.swift
```diff
@@ -283,22 +296,20 @@ internal class ThreeDS2CoreActionHandler: AnyThreeDS2CoreActionHandler {
...
     }
     
-    private func sendChallengeEvent(_ subtype: AnalyticsEventLog.LogSubType) {
-        let logEvent = AnalyticsEventLog(
-            component: Constants.challengeEvent,
-            type: .threeDS2,
-            subType: subtype
-        )
-        context.analyticsProvider?.add(log: logEvent)
+    private func sendErrorEvent(_ code: AnalyticsConstants.ErrorCode, for component: String, message: String? = nil) {
```
atmamont, line 308: There is some inconsistency between method name and it's parameters.
First one with name omitted is code and from the outside it suggests and event (sendError**Event**).
I would suggest to make it more consistent to not accumulate ambiguity.

It also does two things at a time: 
- creates an instance of `AnalyticsEventError` which is not a direct responsibility of `ThreeDS2CoreActionHandler`
- asks to send it 

I see we now have `AnalyticsEventError` instantiation across multiple files, do you think we can extract it so in the end we only send events using a predefined enum(s) from specific handlers such as `ThreeDS2CoreActionHandler`?

AdyenCard/Components/Stored Card/StoredCardAlertManager.swift
```diff
@@ -115,10 +115,20 @@ internal final class StoredCardAlertManager: NSObject, UITextFieldDelegate, Adye
...
         }
     }
     
+    private func sendEncryptionErrorEvent() {
+        var errorEvent = AnalyticsEventError(
+            component: paymentMethod.type.rawValue,
+            type: .internal
+        )
+        errorEvent.code = AnalyticsConstants.ErrorCode.encryptionError.stringValue
+        context.analyticsProvider?.add(error: errorEvent)
```
atmamont, line 129: Double dot syntax is always a trigger. It is something that would make refactor in the future really hard (extracting analytics out of context for example). We can conform context to analytics handling protocol (we can create one if necessary) and just proxy all `add()` calls for now.

Adyen/Analytics/Models/AnalyticsConstants.swift
```diff
@@ -12,6 +12,37 @@ public enum AnalyticsConstants {
     /// A constant to pass into the payment data object in the case where fetching the checkout attempt Id fails.
     public static let fetchCheckoutAttemptIdFailed = "fetch-checkoutAttemptId-failed"
     
+    /// Struct to hold error codes as type-safe static variables.
+    public struct ErrorCode {
+        
+        public static let redirectFailed = ErrorCode(600)
```
atmamont, line 18: There is a bunch of `code.stringValue` across this PR, is there a reason to make it Int instead of having an enum or just using a String raw value?

atmamont (changes requested):

Adyen/Analytics/Models/AnalyticsConstants.swift
erenbesel: It is a string on the backend unfortunately. I used static constants with a raw value Int, and a converted `stringValue`

Adyen/Analytics/AnalyticsProvider/EventAnalyticsProvider.swift
erenbesel: This was actually how it was before the refactor. With the split I seem to have misplaced it on the new class.
We don't have any timer tests yet indeed that would catch this indeed, should add them separately.

Adyen/Analytics/Models/AnalyticsConstants.swift
atmamont: My point was more like why wouldn't we use string as rawValue? Having these as static let doesn't help with keeping them unique as it would be if using enum. This is how we can remove all data type conversions

Adyen/Analytics/AnalyticsProvider/EventAnalyticsProvider.swift
atmamont: Yes this is good point, would help to catch such accidental misplacements.

AdyenCard/Components/Card/CardComponentExtensions.swift
erenbesel: The error code is the important part.

Adyen/Analytics/Models/AnalyticsConstants.swift
erenbesel: Do you mean using enums? With enums we would be creating an instance every single time. Since these are meant as constants, I made them static. Minor performance gain maybe but isn't it helpful?

erenbesel: Ah will remove the `stringValue` part and make the `rawValue` string if you are good with the static part

atmamont: Yes, static is a valid approach, sure. Just using string would make the code simpler.

atmamont (dismissed):

AdyenActions/Actions/RedirectAction.swift
```diff
@@ -18,15 +18,23 @@ public struct RedirectAction: Decodable {
...
     public let nativeRedirectData: String?
     
+    internal let paymentMethodType: String?
+    
     /// Initializes a redirect action.
     ///
     /// - Parameters:
     ///   - url: The URL to which to redirect the user.
     ///   - paymentData: The server-generated payment data that should be submitted to the `/payments/details` endpoint.
     ///   - nativeRedirectData: Native redirect data.
```
goergisn, line 28: The comment is missing the `paymentMethodType` info

```suggestion
    ///   - nativeRedirectData: Native redirect data.
    ///   - paymentMethodType: The type of the payment method.
```

atmamont (dismissed):

goergisn (dismissed):

goergisn (approved):

Issue #1093, opened 2022-12-02
[Feature] Ability to change highlight color when pressing listitem


xanderdeseyn:
![Simulator Screen Shot - iPhone 14 Pro - 2022-12-02 at 11 44 02](https://user-images.githubusercontent.com/1395284/205275370-01ee46f8-0a01-4237-8fbe-4666ae834c44.png)

The default does not fit for a dark theme.

descorp:
Hey @xanderdeseyn 

Sorry for a long reply 😅
Are you making your own custom "dark mode" for this?

Default selection in dark mode will looks like this for v5.2.0:

![Screenshot 2023-04-19 at 13 59 11 (2)](https://user-images.githubusercontent.com/2648655/233068467-ee27e955-59b8-4671-8ebb-5fc443cd32a2.png)

goergisn:
Closing ticket due to inactivity.

You could force drop-in into dark mode using

```
let viewController = dropIn.viewController
if #available(iOS 13.0, *) {
    viewController.overrideUserInterfaceStyle = .dark
}
present(viewController: viewController, completion: nil)
```

PR #1919, opened 2024-12-13
Add highlight styling to list style

jjvarghese:
# Summary

Added the ability to customize highlighted (tapped state) background color of List Items. Previously, it was locked to the default OS value.

# What problem you’re solving

Unable to change the highlighted background color of list items.

# Your approach to fixing the problem

Added a style property to `ListItemStyle` - as an optional value, because the default value is OS protected and can only be approximated. This also means it won't result in any changes to existing implementations.

# Any tests you wrote

As a highlighted state that only occurs on tapping, I did not see a way to integrate with snapshot tests. Happy to add some if someone can point me in the right direction.

# Release notes
<changed>
You can now highlight a list item by changing its background color when the shopper selects it. To do this, use the new optional `ListItemStyle.highlightedBackgroundColor` property.
</changed>


goergisn:
Hi @jjvarghese,
Thank you for your contribution :)

Do you mind explaining what your intended usecase is?
If you want to have a dark highlight you might want to try [applying this suggestion](https://github.com/Adyen/adyen-ios/issues/1093#issuecomment-2397203784).

jjvarghese:
> Hi @jjvarghese, Thank you for your contribution :)
> 
> Do you mind explaining what your intended usecase is? If you want to have a dark highlight you might want to try [applying this suggestion](https://github.com/Adyen/adyen-ios/issues/1093#issuecomment-2397203784).

Hi @goergisn, thanks for the suggestion. My use case is applying a specific theme from a designer to our Adyen Drop In component, including a specific colour value for the list item highlighted state. While the automatic colours in dark mode would indeed look better than the light ones for my use case, being able to specify the exact colour to use is what I am currently after.

atmamont:
Hi Joshua,
Thanks for the contribution!

Do you mind dropping me an email? I have some follow-up questions that might help us to understand the requirements better. We aim to build for the best experience from your side and it would be incredibly helpful to learn a bit more.
My email is [redacted]

atmamont (approved):

goergisn (approved):

erenbesel (approved):

PR #1920, opened 2024-12-13
Fix card component stale state on dropin component reuse

atmamont:
# Summary

Every time certain components are reused within same Dropin session, it has stale parts like:
- error messages for gift card component
- brand icons for card component

This fix adds:
- missing `reset()` call for `FormErrorItemView` (https://github.com/Adyen/adyen-ios/pull/1920/commits/3bf58e0de5d8c3520297607df90f4f0de9c64c7c)
- missing nullification for detected brands logos urls for `FormCardNumberItemView` (https://github.com/Adyen/adyen-ios/pull/1920/commits/d24f7d7df82dfbb6d249bbd71a51ea012e11be61)

It is recommended to review by commits provided.

# Release notes

<fixed> 
For Drop-in, some components that are used in the same Drop-in session no longer have stale parts such as the following:   
- For gift cards, error messages.  
- For cards, brand logos.  
</fixed>

# Ticket

<ticket>
COIOS-840
</ticket>


Adyen/UI/Form/Items/Error/FormErrorItemView.swift
```diff
@@ -80,4 +80,10 @@ internal final class FormErrorItemView: FormItemView<FormErrorItem> {
         return view
     }()
 
+    override internal func reset() {
```
erenbesel, line 83: NIce find!

AdyenCard/Form/DualBrandView.swift
```diff
@@ -90,6 +90,9 @@ extension FormCardNumberItemView {
             guard let firstLogo = logos.first else { return }
             let secondLogo = logos.adyen[safeIndex: 1]
             
+            primaryLogoUrl = firstLogo.url
```
erenbesel, line 93: Tried to look commit by commit but I see these 2 lines were already in the code, why is it shown as newly added?

atmamont: This is my bad, I think I commited these two lines when was extracting this to a separate file 🙈

atmamont: It was like this before 
image

atmamont: Good catch, I missed the right method 🙈 
Updated and force pushed, check [this commit ](https://github.com/Adyen/adyen-ios/pull/1920/commits/d24f7d7df82dfbb6d249bbd71a51ea012e11be61)

erenbesel (approved):

nauaros (dismissed):

erenbesel (dismissed):
Looks good! Maybe we can add some tests? seems we don't have any

atmamont:
> Looks good! Maybe we can add some tests? seems we don't have any

👌 
In progress, DualBandView has zero coverage now so I'll add more apart from those that check this specific fix.

atmamont:
@goergisn @erenbesel Unit tests added, can I have another look?

Tests/IntegrationTests/Card Tests/DualBrandViewTests.swift
```diff
@@ -0,0 +1,162 @@
...
+    }
+    
+    override func tearDown() {
+        sut = nil
+        imageLoader = nil
+        brandSelectionCount = nil
+        super.tearDown()
+    }
+    
+    func testUpdateCurrentLogos_WhenResettingLoadedImages_ShouldResetToPlaceholder() {
```
atmamont, line 36: Let me know if this guy looks too big and it is not easy to follow. If so, I'll break it to 3-4 separate tests.

Tests/IntegrationTests/Card Tests/DualBrandViewTests.swift
```diff
@@ -0,0 +1,162 @@
...
+        borderColor: nil,
+        borderWidth: 0.0,
+        cornerRadius: 0.0,
+        clipsToBounds: true,
+        contentMode: .scaleAspectFit
+    )
+
+}
+
+private class MockImageLoader: ImageLoading {
```
erenbesel: nice!

erenbesel (dismissed):

goergisn (dismissed):

Tests/IntegrationTests/Card Tests/DualBrandViewTests.swift
goergisn: we already have an `ImageLoaderMock` - maybe that can be adapted instead if needed?

atmamont: Good catch, indeed there is no need to create duplicates

erenbesel (approved):

goergisn (approved):

PR #1924, opened 2024-12-18
Replace 3rd party workflow actions with Github API to access labels and PR body

atmamont:
# Summary

## Problem
External PRs from forks do not have access to Github token so none of external actions could be run successfully.

## Options
1. Expand access level by using different even (`pull_request` to `pull_request_target`). [Documentation](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target) says labels and comments are perfect cases for such expansion but thinking security first I decided do explore other options.
2. ~~Using Github API and removing 3rd party actions requiring Github token. Using Github javascript-like API to post comments.~~

## Solution

Option 1 was chosen as I didn't manage to get option 2 working, it failed for the "post comment" section.

# Ticket

<ticket>
COIOS-000
</ticket>

# Manual tests
- [x] Fails if no labels are set
- [x] Fails if any of release-related labels are set with no matching XML in PR body
- [x] Comments if job fails
- [x] Removes comment if job suceeds
- [x] Does not interfere with other Github Bot comments
- [x] All of the above triggered from a fork ([an example PR](https://github.com/Adyen/adyen-ios/pull/1928))

# Test release notes

<changed> This is a changed text wrapped in XML tag </changed>


goergisn (dismissed):

nauaros (approved):

erenbesel (approved):

.github/workflows/validate_pr_labels_and_release_notes.yml
```diff
@@ -1,9 +1,13 @@
...
 on:
+  pull_request_target:
+    types: [opened, synchronize, reopened, labeled, unlabeled, edited]
   pull_request:
     types: [opened, synchronize, reopened, labeled, unlabeled, edited]
-
+    
+permissions:
+  pull-requests: write
+  issues: write
```
goergisn, line 10: is this needed to access to access `github.rest.issues`?

.github/workflows/validate_pr_labels_and_release_notes.yml
```diff
@@ -14,6 +18,7 @@ jobs:
           fetch-depth: 0
 
       - name: Get the list of allowed pull request labels
+        id: get-allowed_labels
```
goergisn, line 21: Just to make it look nicer

```suggestion
        id: get-allowed-labels
```

.github/workflows/validate_pr_labels_and_release_notes.yml
```diff
@@ -56,15 +85,85 @@ jobs:
...
+          if [ "$SHOULD_FAIL" = true ]; then
+            echo "$ERROR_MESSAGE"
+            echo "error_message=${ERROR_MESSAGE}" >> $GITHUB_OUTPUT
             exit 1
           fi
+          
         env:
           PR_BODY: ${{github.event.pull_request.body}}
           PR_LABELS: ${{steps.pr-labels.outputs.labels}}
           ALLOWED_LABELS: ${{ steps.get-allowed_labels.outputs.LABELS }}
```
goergisn, line 104:
```suggestion
          ALLOWED_LABELS: ${{ steps.get-allowed-labels.outputs.LABELS }}
```

.github/workflows/validate_pr_labels_and_release_notes.yml
```diff
@@ -56,15 +85,85 @@ jobs:
               cmd="awk '/<$LABEL>.*<\/$LABEL>/'"
               FEATURE_RELEASE_NOTE="$(echo $PR_BODY | eval $cmd)"
               if [[ -z "$FEATURE_RELEASE_NOTE" ]]; then
-                 echo -e "${RED}Pull requests labeled with '$LABEL' must have the release note in the pull request body in the following format '<$LABEL> {THE RELEASE NOTE}  </$LABEL>'${NC}"
-                 SHOULD_FAIL_JOB="true"
+                 ERROR_MESSAGE="❌ Pull requests labeled with '${LABEL}' must have the release note wrapped in the same xml tag."
```
goergisn, line 88: Shall we keep the info about which label to use?
It might help unfamiliar contributors

.github/workflows/validate_pr_labels_and_release_notes.yml
```diff
@@ -56,15 +85,85 @@ jobs:
...
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number
+            });
+            
+            const botComment = comments.find(comment => 
+              comment.user.login === 'github-actions[bot]' &&
+              comment.body.includes(commentIdentifier)
+            );
+
```
goergisn, line 125:
```suggestion
```

goergisn (approved):

.github/workflows/validate_pr_labels_and_release_notes.yml
atmamont: I totally failed to post a comment with properly escaped backticks and xml tags so I just gave up. I don't mind having extra help here :)

.github/workflows/validate_pr_labels_and_release_notes.yml
atmamont: Yes, according to the [github docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token), I find it misleading but they mean everything under "issues"

PR #1931, opened 2024-12-31
Cashapp/Twint error events

erenbesel:
# Summary

Third party related error events (Twint/Cashapp)

# Ticket

<ticket>
COIOS-841
</ticket>


nauaros (dismissed):

AdyenActions/Components/SDK/TwintSDKActionComponent.swift
```diff
@@ -212,10 +218,14 @@ import Foundation
             presentationDelegate.present(component: presentableComponent)
         }
 
-        private func handleShowError(_ error: String) {
+        private func handleShowError(_ errorMessage: String, componentName: String) {
```
goergisn, line 221: This might read a bit nicer

```suggestion
        private func handleShow(errorMessage: String, for componentName: String) {
```

AdyenActions/Components/SDK/TwintSDKActionComponent.swift
```diff
@@ -235,6 +245,17 @@ import Foundation
         private func cleanup() {
             pollingComponent?.didCancel()
         }
+        
+        private func sendThirdPartyErrorEvent(with message: String?, componentName: String) {
```
goergisn, line 249:
```suggestion
        private func sendThirdPartyErrorEvent(with message: String?, for componentName: String) {
```

AdyenCashAppPay/CashAppPayComponent.swift
```diff
@@ -225,24 +231,37 @@ extension CashAppPayComponent: CashAppPayObserver {
...
+            fail(with: error, message: error.localizedDescription)
         default:
             break
         }
     }
 
-    private func fail(with error: Swift.Error) {
+    private func fail(with error: Swift.Error, message: String? = nil) {
         stopLoading()
+        sendThirdPartyErrorEvent(with: message)
```
goergisn, line 251: is this a backend decision? Otherwise we could try to prevent optionals
```suggestion
        sendThirdPartyErrorEvent(with: message ?? error.localizedDescription)
```

goergisn (dismissed):

goergisn:
⚠️  https://github.com/Adyen/adyen-ios/pull/1931#issuecomment-2566421490

AdyenCashAppPay/CashAppPayComponent.swift
erenbesel: we only send error description if it's manually created by us to make sure it doesn't contain sensitive data

nauaros (approved):

goergisn (approved):

AdyenActions/Components/SDK/TwintSDKActionComponent.swift
atmamont, line 249: Do we need to add extra method to send events for each domain?
I see we have an enum that has clear domain split `AnalyticsConstants.ErrorCode.thirdPartyError` so could we use a generic `sendError(...)` or `sendEvent(...)` method if we have so?

AdyenCashAppPay/CashAppPayComponent.swift
```diff
@@ -23,6 +23,12 @@ public final class CashAppPayComponent: PaymentComponent,
         static let storeDetailsItem = "storeDetailsItem"
         static let cashAppButtonItem = "cashAppButtonItem"
     }
+    
+    private enum ErrorMessage {
+        static let unexpectedError = "CashApp unexpected error"
```
atmamont, line 28: Do we have an agreement with data team to send human readable values? Is it the same for Android and Web?

AdyenCashAppPay/CashAppPayComponent.swift
```diff
@@ -225,24 +231,37 @@ extension CashAppPayComponent: CashAppPayObserver {
...
     }
 
-    private func fail(with error: Swift.Error) {
+    private func fail(with error: Swift.Error, message: String? = nil) {
         stopLoading()
+        sendThirdPartyErrorEvent(with: message)
         delegate?.didFail(with: error, from: self)
     }
+    
+    private func sendThirdPartyErrorEvent(with message: String?) {
```
atmamont, line 255: So this method repeats again. What do you think of having an additional generic method in `analyticsProvider`, something like 
```swift
analyticsProvider.addError(
    paymentMethod: paymentMethod.type.rawValue,
    errorType: .thirdParty,
    message: message)
```

erenbesel: I am not a fan of repeating this either but maybe we can replace the `add(error)` function completely then as having 2 different ones also doesn't seem useful.

AdyenCashAppPay/CashAppPayComponent.swift
erenbesel: The message field is only for context, not to be used as a filtering property so each platform is free to send any logical, non sensitive text.

AdyenCashAppPay/CashAppPayComponent.swift
atmamont: Agreed. You have this on your fingertips, I'm just trying to provide feedback but please make your own decision on how to simplify this.

AdyenCashAppPay/CashAppPayComponent.swift
atmamont: Why do we have such parameter then? What's the benefit from a product or data point?

erenbesel: Should have also added that message field is only sent when the same error code event is sent from multiple places. Just an extra help when we filter by for example thirdParty errors for component X, we will see how many unique cases there are.

neelSharma12 (approved):

PR #1938, opened 2025-01-06
Update LICENSE

descorp:
# Summary

<ticket>
COIOS-000
</ticket>

* Align company legal name on LICENSE file with other SDKs

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License
 
-Copyright (c) Adyen N.V. (https://www.adyen.com)
+Copyright (c) Adyen (https://www.adyen.com)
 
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,6 +1,6 @@
 The MIT License
 
-Copyright (c) Adyen B.V. (https://www.adyen.com)
+Copyright (c) Adyen (https://www.adyen.com)
```
atmamont, line 3: Aren't we Adyen NV?

descorp: We are, but all other SDKs use just "Adyen"

atmamont (approved):

goergisn (approved):

PR #1948, opened 2025-01-15
Fix broken CI workflows (v4)

nauaros:
## Summary

- Updated workflow runners to `macos-14-xlarge`.
- Upgraded device destinations in all affected jobs.
- Modified `Coder` to output using `sortedKeys`.
- Defined `encodeWithSortedKeys` in `JSONEncoder`.
- Fixed unit tests broken by updates in Swift 6 and Xcode 16.

## Issue

GitHub Actions deprecated the `macos-12` runners, causing workflows relying on them to fail. To restore functionality, we've updated to newer runners. This update also necessitated upgrading the selected Xcode version and device destinations in the jobs. Additionally, some unit tests were broken due to changes in the latest Xcode version, which have now been fixed.

## Ticket

<ticket>
COIOS-853
</ticket>


.github/workflows/build.yml
```diff
@@ -10,20 +10,45 @@ on:
 jobs:
 
   tests:
-    runs-on: macos-12
+    runs-on: macos-14-xlarge
```
atmamont, line 13: Do we need `xlarge` machine? We could use `macos-latest` and ideally forget about this type of updates.

.github/workflows/build.yml
```diff
@@ -10,20 +10,45 @@ on:
...
+        
+        - version: '16.4'
+          runtime: 'iOS-16-4'
+          device: 'iPhone 14'
+          displayname: 'iPhone-14'
+          os: 'macos-14-xlarge'
+          xcode_version: '16.1'
+          needs_custom_sim: false # Takes the shipped simulator that comes with Xcode 16
+          
+        # - destination: 'name=iPhone 11 Pro'
```
atmamont: commented code

.github/workflows/pr_scan.yml
```diff
@@ -23,7 +26,7 @@ jobs:
     - name: Select latest Xcode
       uses: maxim-lobanov/setup-xcode@v1
       with:
-        xcode-version: latest-stable
+        xcode-version: '16.1'
```
atmamont, line 29: Why do we need to lock it now?

.github/workflows/test_cocoapods_integration.yml
```diff
@@ -12,24 +12,27 @@ on:
...
+    - uses: actions/checkout@v4
     - uses: n1hility/cancel-previous-runs@v3
       with:
         token: ${{ secrets.GITHUB_TOKEN }}
 
     - name: Select latest Xcode
       uses: maxim-lobanov/setup-xcode@v1
       with:
-        xcode-version: latest-stable
+        xcode-version: '16.1'
```
atmamont, line 25: This looks repetitive, I guess it could be extracted to the top level as env var
```
env:
  xcode_version: '16.1'
```

and later used as 
```
with:
    xcode-version: ${{env.xcode_version}}
````

File: Scripts/test-carthage-integration.sh [hunk 5]
```diff
@@ -125,7 +125,7 @@ cp "../Demo/Configuration.swift" Source/Configuration.swift
 xcodegen generate
 
 echo_header "Run Tests"
-xcodebuild build test -project $PROJECT_NAME.xcodeproj -scheme App -destination "name=iPhone 13" CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO | xcpretty && exit ${PIPESTATUS[0]}
+xcodebuild build test -project $PROJECT_NAME.xcodeproj -scheme App -destination "name=iPhone 16" CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO | xcpretty && exit ${PIPESTATUS[0]}
```
atmamont, line 128: I guess this is a good candidate for env var extraction as we will be updating this once in a while

Scripts/test-CocoaPods-integration.sh
```diff
@@ -41,41 +41,43 @@ rm -rf $PROJECT_NAME
...
-            dependencies: [\"TempProject\"]),
-    ]
-)
-
-" > Package.swift
-
-swift package update
-
-swift package generate-xcodeproj
+echo "
```
atmamont, line 44: This file change is substantial and is not very clear, can you provide brief explanation what's happening in here?

.github/workflows/pr_scan.yml
goergisn: We want to sync it with our local machines (like with v5) so the tests are reproducible

.github/workflows/build.yml
nauaros: removed

atmamont (dismissed):
Good job Naufal!

.github/workflows/test-carthage-integration.yml
```diff
@@ -12,27 +12,28 @@ on:
 jobs:
 
   carthage:
-    runs-on: macos-12
+    runs-on: macos-latest
```
goergisn: The combination of macos-latest + a hardcoded xcode version might break in the future - In general I like the idea of using the latest macos but it might result in random workflow failures - maybe better to choose a proven macos version 🤔

.github/workflows/test-carthage-integration.yml
```diff
@@ -12,27 +12,28 @@ on:
...
       with:
-        xcode-version: latest-stable
+        xcode-version: '16.1'
 
     - name: Resolve dependencies
       run: |
         brew update
         brew install xcodegen
+        brew install carthage
         brew upgrade carthage
```
goergisn, line 32: is upgrade carthage needed when you just installed it?

Tests/AdyenTests/Card Tests/3DS2 Component/ThreeDS2ClassicActionHandlerTests.swift
```diff
@@ -269,4 +274,23 @@ class ThreeDS2ClassicActionHandlerTests: XCTestCase {
...
+
+    func compareJSONStrings(_ jsonString1: String, _ jsonString2: String) -> Bool {
+        guard let data1 = jsonString1.data(using: .utf8),
+              let data2 = jsonString2.data(using: .utf8) else {
+            print("Invalid data.")
+            return false
+        }
+
+        do {
+            let dict1 = try JSONSerialization.jsonObject(with: data1, options: [.allowFragments]) as? [String: Any]
```
goergisn: try to unwrap it to catch the case where both json strings cannot be casted to [String: Any]
```suggestion
            let dict1 = try XCTUnwrap(JSONSerialization.jsonObject(with: data1, options: [.allowFragments]) as? [String: Any])
```

Tests/AdyenTests/Card Tests/3DS2 Component/ThreeDS2ClassicActionHandlerTests.swift
```diff
@@ -269,4 +274,23 @@ class ThreeDS2ClassicActionHandlerTests: XCTestCase {
...
+            let dict2 = try JSONSerialization.jsonObject(with: data2, options: [.allowFragments]) as? [String: Any]
+
+            return NSDictionary(dictionary: dict1 ?? [:]).isEqual(to: dict2 ?? [:])
```
goergisn:
```suggestion
            return NSDictionary(dictionary: dict1).isEqual(to: dict2)
```

.github/workflows/test-carthage-integration.yml
nauaros: It is a good practice. At least, from the literature online

Tests/AdyenTests/Card Tests/3DS2 Component/ThreeDS2ClassicActionHandlerTests.swift
nauaros: Removed this method. By using sortedKeys in the encoders, its not needed anymore.

atmamont (approved):

goergisn (approved):

PR #1950, opened 2025-01-15
switching to Void? for AssociatedKeys

goergisn:
# Summary
- Switching to `Void?` for `AssociatedKey`s as described [here](https://github.com/atrick/swift-evolution/blob/diagnose-implicit-raw-bitwise/proposals/nnnn-implicit-raw-bitwise-conversion.md#workarounds-for-common-cases)
- Removing superfluous `swiftlint:disable`

# Ticket

<ticket>
COIOS-000
</ticket>

### DIFF
--- a/Adyen/Core/Components/Base/Component.swift
+++ b/Adyen/Core/Components/Base/Component.swift
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2022 Adyen N.V.
+// Copyright (c) 2019 Adyen N.V.
 //
 // This file is open source and available under the MIT license. See the LICENSE file for more info.
 //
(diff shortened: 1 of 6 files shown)


nauaros (approved):

descorp (approved):

Adyen/Core/Components/Base/Component.swift
```diff
@@ -78,9 +78,9 @@ public extension Component {
 }
 
 private enum AssociatedKeys {
-    internal static var isDropIn = "isDropInObject"
+    internal static var isDropIn: Void?
```
atmamont, line 81: This solution's readability could be better. Shall we either add a comment as in the link you mentioned?
```swift
private enum AssociatedKeys {
    internal static var isDropIn: Void? // "isDropin"
}
```

neelSharma12 (approved):

PR #1958, opened 2025-01-20
Fix Warnings

goergisn:
# Summary
- Checking if the compiler knows about the backported apple pay enum case
- Updating Project settings
- Adding [@retroactive](https://forums.swift.org/t/supressing-retroactive-conformance-warning-while-supporting-both-swift-5-10-and-6-0/73784) for extensions of foundational types

## Background regarding #if compiler(>=6)

Xcode started complaining:
```
Extension declares a conformance of imported type 'Array' to imported protocol 'AdyenCompatible'; 
this will not behave correctly if the owners of 'Swift' introduce this conformance in the future
```

## Issue

When adding the @retroactive tag it silences the warning but it will start failing if you build with a compiler that doesn’t know about this label. Which is the case on our CI because we need older Xcodes (with older toolchain) to get access to older Simulators to be able to test older iOS versions.

## Solution
You can check for the swift or compiler version that is used to compile your project via a compiler directive.

e.g.
```
#if swift(>=6) // Checks for the minimum swift version
#if compiler(>=6) // Checks for the minimum compiler version
```

So wrapping the code in an `#if compiler(>=6)` block and providing the old code in the `#else` block silences the warnings AND allows you to build the code with older toolchains

# Ticket

<ticket>
COIOS-000
</ticket>


atmamont (dismissed):

atmamont (approved):

nauaros (approved):

goergisn:
Closing due to quite some changes for very little benefit

PR #1999, opened 2025-01-31
Release note workflow escaping

goergisn:
# Summary
- Using escaping for release notes instead of a token
- Improving output

<ticket>COIOS-000</ticket>


atmamont (approved):
We need to make it a rule to leave comments for complex actions syntax.
Great stuff!

neelSharma12 (approved):

PR #2003, opened 2025-02-03
OCR: Card image parser

nauaros:
# Summary

This PR includes the main business layer to perform the card image scanning.

The `CardImageParser` class contains logic to process an image (`CIImage`) following these steps:
1. The scanned image is transformed by applying a series of `CIFilter`s. These are the following:
    - Image noise reduction (`CINoiseReduction`)
    - Image color adjustment (`CIColorControls`)
    - Image sharpness enhancement (`CISharpenLuminance`)
2. Using `Vision`, a text recognition request is created, with an accurate level of recognition.
3. In order to perform text recognition concurrently, GCD is used to analyize text fetch from the image to identify an PAN number and expire date.
4. The card parsing completes when both pan number and expire date are recognized.

# Ticket

<ticket>
COIOS-825
COIOS-826
</ticket>


AdyenCardScanner/Sources/CardImageParser.swift
```diff
@@ -0,0 +1,233 @@
...
+        colorControlsFilter?.setValue(1.2, forKey: kCIInputSaturationKey)
+        outputImage = colorControlsFilter?.outputImage
+
+        // Apply sharpness enhancement filter.
+        let sharpenFilter = CIFilter(name: "CISharpenLuminance")
+        sharpenFilter?.setValue(outputImage, forKey: kCIInputImageKey)
+        sharpenFilter?.setValue(0.5, forKey: kCIInputSharpnessKey)
+        outputImage = sharpenFilter?.outputImage
+
+//        outputImage = convertToGrayscale(image: outputImage!)
```
nauaros: This is intended commented code.

goergisn: Maybe leave a comment on why? Same with the "do not remove" block below?

AdyenCardScanner/Sources/CardImageParser.swift
```diff
@@ -0,0 +1,233 @@
...
+        dispatchGroup.enter()
+        DispatchQueue.global().async {
+            if self.expireDate == nil {
+                self.expireDate = self.extractExpireDate(from: results)
+            }
+            dispatchGroup.leave()
+        }
+
+        dispatchGroup.notify(queue: .main) {
+            guard let cardNumber = self.cardNumber, let expireDate = self.expireDate else { return }
```
erenbesel: The completion block is ignored on this case, leaving the caller unaware. Is this intended or otherwise we may need to call an empty/error state completion.

AdyenCardScanner/Sources/CardImageParser.swift
```diff
@@ -0,0 +1,233 @@
...
+
+        return colorMatrixFilter?.outputImage
+    }
+
+    // ==================================================
+
+    private func extractCardNumber(from textObservations: [VNRecognizedTextObservation]) -> String? {
+        for observation in textObservations {
+            let candidates = observation.topCandidates(Constants.topCandidates)
+            guard let topCandidate = candidates.first,
```
erenbesel: How about moving this logic to a new function to get rid of `continue`? May also help this function look simple with like 2-3 lines.

AdyenCardScanner/Sources/CardImageParser.swift
```diff
@@ -0,0 +1,233 @@
...
+
+            return sanitizedCardNumber
+        }
+
+        return nil
+    }
+
+    private func extractExpireDate(from textObservations: [VNRecognizedTextObservation]) -> Date? {
+        for observation in textObservations {
+            let candidates = observation.topCandidates(Constants.topCandidates)
```
erenbesel: Same as above with `continue`

erenbesel (commented):
Nice start! Small questions

AdyenCardScanner/AdyenCardScanner.docc/AdyenCardScanner.md
descorp, line 1: Is this a draft? Maybe remove it from PR if you are not planing to work on documentation?

AdyenCardScanner/Sources/CardImageParser.swift
```diff
@@ -0,0 +1,233 @@
...
+        colorControlsFilter?.setValue(0.2, forKey: kCIInputBrightnessKey)
+        colorControlsFilter?.setValue(1.5, forKey: kCIInputContrastKey)
+        colorControlsFilter?.setValue(1.2, forKey: kCIInputSaturationKey)
+        outputImage = colorControlsFilter?.outputImage
+
+        // Apply sharpness enhancement filter.
+        let sharpenFilter = CIFilter(name: "CISharpenLuminance")
+        sharpenFilter?.setValue(outputImage, forKey: kCIInputImageKey)
+        sharpenFilter?.setValue(0.5, forKey: kCIInputSharpnessKey)
+        outputImage = sharpenFilter?.outputImage
```
descorp: I am suggesting move all CIImage related logic into separate file as a set of extensions.
Would be nice to use fluent interface here instead of long method with comments:

```suggestion
        var outputImage = image
            .applyNoiseReductionFilter()
            .applyColorControlsFilter()
            .applySharpenFilter()
```

```swift

extension CIImage {
    func applyNoiseReductionFilter() { 
        let noiseReductionFilter = CIFilter(name: "CINoiseReduction")
        noiseReductionFilter?.setValue(self, forKey: kCIInputImageKey)
        noiseReductionFilter?.setValue(0.02, forKey: "inputNoiseLevel")
        noiseReductionFilter?.setValue(0.4, forKey: "inputSharpness")
        return noiseReductionFilter?.outputImage
    }
        
    func applyColorControlsFilter() { 
        let colorControlsFilter = CIFilter(name: "CIColorControls")
        colorControlsFilter?.setValue(self, forKey: kCIInputImageKey)
        colorControlsFilter?.setValue(0.2, forKey: kCIInputBrightnessKey)
        colorControlsFilter?.setValue(1.5, forKey: kCIInputContrastKey)
        colorControlsFilter?.setValue(1.2, forKey: kCIInputSaturationKey)
        return colorControlsFilter?.outputImage
    }   
     
    func applySharpenFilter() {
        let sharpenFilter = CIFilter(name: "CISharpenLuminance")
        sharpenFilter?.setValue(self, forKey: kCIInputImageKey)
        sharpenFilter?.setValue(0.5, forKey: kCIInputSharpnessKey)
        return sharpenFilter?.outputImage
    }
}
```

AdyenCardScanner/Sources/CardImageParser.swift
```diff
@@ -0,0 +1,233 @@
...
+
+    private let dateFormatter: DateFormatter
+
+    private var cardNumber: String?
+    private var expireDate: Date?
+
+    // MARK: - Initializers
+
+    init() {
+        self.dateFormatter = DateFormatter()
```
goergisn: what about making the dateFormatter a lazy var and put the setup code locally?

AdyenCardScanner/Sources/CardImageParser.swift
```diff
@@ -0,0 +1,233 @@
...
+        self.dateFormatter.locale = Locale.current
+        self.dateFormatter.timeZone = TimeZone(secondsFromGMT: 0)
+    }
+
+    // MARK: - CardImageParsing
+
+    func parse(
```
goergisn, line 42: what about using swift concurrency here?

AdyenCardScanner/Sources/CardImageParser.swift
goergisn: +1 for extensions :P

AdyenCardScanner/Sources/CardImageParser.swift
```diff
@@ -0,0 +1,233 @@
...
+        colorControlsFilter?.setValue(outputImage, forKey: kCIInputImageKey)
+        colorControlsFilter?.setValue(0.2, forKey: kCIInputBrightnessKey)
+        colorControlsFilter?.setValue(1.5, forKey: kCIInputContrastKey)
+        colorControlsFilter?.setValue(1.2, forKey: kCIInputSaturationKey)
+        outputImage = colorControlsFilter?.outputImage
+
+        // Apply sharpness enhancement filter.
+        let sharpenFilter = CIFilter(name: "CISharpenLuminance")
+        sharpenFilter?.setValue(outputImage, forKey: kCIInputImageKey)
+        sharpenFilter?.setValue(0.5, forKey: kCIInputSharpnessKey)
```
descorp, line 110: BTW, use use `kCIInputSharpnessKey` here and on line "96":
I assume only "CINoiseReduction" might be enough 🤔

descorp: [CISharpenLuminance](https://developer.apple.com/documentation/coreimage/cisharpenluminance) also have required parameter "radius".

AdyenCardScanner/Sources/CardImageParser.swift
```diff
@@ -0,0 +1,233 @@
...
+
+    private func extractCardNumber(from textObservations: [VNRecognizedTextObservation]) -> String? {
+        for observation in textObservations {
+            let candidates = observation.topCandidates(Constants.topCandidates)
+            guard let topCandidate = candidates.first,
+                  topCandidate.confidence > Constants.cardNumberConfidence else { continue }
+
+            debugPrint("candidates >> \(topCandidate.string)")
+
+            let sanitizedCardNumber = topCandidate.string.replacingOccurrences(of: " ", with: "")
```
goergisn: what about having a function that sanitizes the string? (it could remove all spaces + non-numbers)

AdyenCardScanner/Sources/CardImageParser.swift
```diff
@@ -0,0 +1,233 @@
...
+                sum += digit
+            }
+        }
+
+        return sum % 10 == 0
+    }
+
+    private func date(from dateString: String) -> Date? {
+        // First, try the short ("MM/YY") format
+        dateFormatter.dateFormat = CardExpireDateFormat.short.rawValue
```
goergisn: these functions are called from a background thread right?
maybe better to have 2 DateFormatters instead of updating a shared one

AdyenCardScanner/Sources/CardImageParser.swift
```diff
@@ -0,0 +1,233 @@
...
+import Vision
+
+protocol CardImageParsing {
+    func parse(
+        image: CIImage,
+        completion: @escaping (CreditCard) -> Void
+    )
+}
+
+class CardImageParser: CardImageParsing {
```
descorp, line 18: BTW, Apple significantly improved CoreImages with `import CoreImage.CIFilterBuiltins`

Improvement:

**BEFORE:**

```swift
import CoreImage

    let colorControlsFilter = CIFilter(name: "CIColorControls")
    colorControlsFilter?.setValue(image, forKey: kCIInputImageKey)
    colorControlsFilter?.setValue(0.0, forKey: kCIInputSaturationKey) // No color
    colorControlsFilter?.setValue(1.0, forKey: kCIInputBrightnessKey) // Full brightness
    colorControlsFilter?.setValue(1.0, forKey: kCIInputContrastKey) // Full contrast
```

**AFTER:**

```swift
import CoreImage
import CoreImage.CIFilterBuiltins

    let colorControlsFilter = CIFilter.colorControls()
    colorControlsFilter.inputImage = inputImage
    colorControlsFilter.saturation = 0 // No color
    colorControlsFilter.brightness = 1 // Full brightness
    colorControlsFilter.contrast = 1 // Full contrast
```

AdyenCardScannerTests/AdyenCardScannerTests.swift
```diff
@@ -0,0 +1,16 @@
...
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+@testable import AdyenCardScanner
+import Testing
+
+struct AdyenCardScannerTests {
+
+    @Test func example() async throws {
+        // Write your test here and use APIs like `#expect(...)` to check expected conditions.
```
descorp, line 13: Would be nice to have some tests with locally stored "card" image 🤔

AdyenCardScanner/Sources/CardImageParser.swift
nauaros: Yes, this is intended behavior. It's hard to see it like this, but this method is continuously called from the camera stream. In the case of no complete match, then we just return without calling the completion block.

AdyenCardScanner/Sources/Extensions/String+Extensions.swift
```diff
@@ -0,0 +1,13 @@
+//
+// Copyright (c) 2025 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+extension String {
```
descorp: Better move it to `AdyenScope` to avoid code collisions

AdyenCardScanner/Sources/CardImageParser.swift
nauaros: Do you suggest filtering the candidates first, and then traversing the list?

AdyenCardScanner/Sources/CardImageParser.swift
```diff
@@ -0,0 +1,233 @@
...
+            if index % 2 == 1 {
+                let doubled = digit * 2
+                sum += (doubled > 9) ? (doubled - 9) : doubled
+            } else {
+                sum += digit
+            }
+        }
+
+        return sum % 10 == 0
+    }
```
descorp, line 141: Can we create `String` extension and/or reuse Luhn check from `CardNumberValidator` instead of copy-pasting the code?

AdyenCardScanner/AdyenCardScanner.docc/AdyenCardScanner.md
nauaros: Yes, it was automatically added by Xcode. I didn't see it. Let me remove it

AdyenCardScanner/Sources/CardImageParser.swift
```diff
@@ -0,0 +1,233 @@
...
+    private func extractExpireDate(from textObservations: [VNRecognizedTextObservation]) -> Date? {
+        for observation in textObservations {
+            let candidates = observation.topCandidates(Constants.topCandidates)
+            guard let topCandidate = candidates.first,
+                  topCandidate.confidence > Constants.expireDateConfidence else { continue }
+
+            let expireDateMatch = extractMatch(from: topCandidate.string, using: Constants.expireDateRegex)
+            guard let expireDateMatch else { continue }
+
+            debugPrint("VALID DATE: \(expireDateMatch)")
```
descorp: Better use `AdyenLogging`

AdyenCardScanner/Sources/CardImageParser.swift
nauaros: Nice! I will refactor it

AdyenCardScanner/Sources/CardImageParser.swift
goergisn: Or a custom date formatter / parser to move this logic out of this class

AdyenCardScanner/Sources/CardImageParser.swift
```diff
@@ -0,0 +1,233 @@
...
+            return date(from: expireDateMatch)
+        }
+
+        return nil
+    }
+
+    private func isValidLuhn(_ number: String) -> Bool {
```
goergisn, line 125: We have that function on the CardNumberValidator right? Can we somehow inject that instead of duplicating code?

AdyenCardScanner/Sources/CardImageParser.swift
nauaros: Moved into an extension.

AdyenCardScanner/Sources/Extensions/String+Extensions.swift
nauaros: I want to prevent any dependency to any of our modules. This is not public, I guess there can be no name collisions.

AdyenCardScanner/Sources/CardImageParser.swift
nauaros: This debug prints are gonna dissappear. I want to prevent any dependency with other modules.

AdyenCardScanner/Sources/CardImageParser.swift
nauaros: Done

AdyenCardScanner/Sources/Extensions/String+Extensions.swift
descorp: > I want to prevent any dependency to any of our modules

Nobel goal, but I guess single "Adyen Core" dependency is fine - this library is not intended to be used outside of Adyen Checkout, right?

AdyenCardScanner/Sources/CardImageParser.swift
nauaros: Done

AdyenCardScanner/Sources/CardImageParser.swift
```diff
@@ -211,26 +211,26 @@ class CardImageParser: CardImageParsing {
 private extension CIImage {
 
     func applyNoiseReductionFilter() -> CIImage? {
```
descorp, line 162: No need for `CIImage?`

AdyenCardScanner/Sources/CardImageParser.swift
```diff
@@ -0,0 +1,236 @@
...
+
+    private func convertToGrayscale(image: CIImage) -> CIImage? {
+        let colorControlsFilter = CIFilter(name: "CIColorControls")
+        colorControlsFilter?.setValue(image, forKey: kCIInputImageKey)
+        colorControlsFilter?.setValue(0.0, forKey: kCIInputSaturationKey) // No color
+        colorControlsFilter?.setValue(1.0, forKey: kCIInputBrightnessKey) // Full brightness
+        colorControlsFilter?.setValue(1.0, forKey: kCIInputContrastKey) // Full contrast
+        return colorControlsFilter?.outputImage
+    }
+
```
descorp: You can also move `convertToGrayscale` and `applyThreshold` to extension as fluent interface methods.

BTW: Don't forget to apply `CoreImage.CIFilterBuiltins` to them and remove optional `CIImage?`

AdyenCardScannerTests/AdyenCardScannerTests.swift
nauaros: Yes, tests will be added in following PRs

AdyenCardScanner/Sources/CardImageParser.swift
erenbesel: A function inside the for loop so we call 
```
for observation in textObservations {
    let candidates = observation.topCandidates(Constants.topCandidates)
    findCardNumber(from candidates)
}
```
or move candidate line into the function too etc

AdyenCardScanner/Sources/CardImageParser.swift
nauaros: The output image is always an optional image. Applying filters can fail.

AdyenCardScanner/Sources/CardImageParser.swift
nauaros: Refactor the method. No for loop is present

AdyenCardScanner/Sources/CardImageParser.swift
nauaros: Refactor the method. No for loop is present

AdyenCardScanner/Sources/CardImageParser.swift
nauaros: For now I will let it like this since this is only being used here in the parser. 
Injecting the CardNumberValidador for a luhn check fuction that doesn't change creates more hassle than the benefits (knowing it will never change). Also, no need to have a dependency with `AdyenCard`. `CardNumberValidator` does more than just a luhn check.
We don't validate the card number in the card scanner, that will be done on the card component.

AdyenCardScanner/Sources/CardImageParser.swift
```diff
@@ -90,84 +90,36 @@ class CardImageParser: CardImageParsing {
...
-        return nil
+        if let cachedCardNumber { return cachedCardNumber }
+
+        let cardNumberMatch = textObservations
+            .compactMap { $0.topCandidates(Constants.topCandidates).first }
+            .filter { $0.confidence > Constants.cardNumberConfidence }
+            .map { $0.string.replacingOccurrences(of: " ", with: "") }
+            .filter { $0.isOnlyNumbers }
+            .filter { $0.count >= 13 && $0.count <= 19 }
+            .filter { isValidLuhn($0) }
```
descorp: Maybe convert this into a single property?

```swift
let cardNumberMatch = textObservations
    .compactMap { $0.topCandidates(Constants.topCandidates).first }
    .filter { $0.confidence > Constants.cardNumberConfidence }
    .filter { $0.string.isCardNumber }
    .first
    .string
...

extension String {
    internal var isCardNumber: Bool {
        let trimmed = $0.string.replacingOccurrences(of: " ", with: "")
        return trimmed.isOnlyNumbers && (13...19).contains(trimmed.count) &&  trimmed.isValidLuhn
    }
    private var isOnlyNumbers: Bool { ... }
    private var isValidLuhn: Bool { ... }
}
```

AdyenCardScanner/Sources/CardImageParser.swift
nauaros: I don't want to remove non numbers. The idea is to find a string that's only numbers and spaces or numbers, and filter out any candidate that has any non-numeric character.

AdyenCardScanner/Sources/CardImageParser.swift
nauaros: Yes, that's a good catch

AdyenCardScanner/Sources/CardImageParser.swift
nauaros: I can do, good suggestion

AdyenCardScanner/Sources/CardImageParser.swift
nauaros: Created an `ExpireDateFormatter` that uses two different `DateFormatter` instances. We avoid any data races this way.

AdyenCardScanner/Sources/CardImageParser.swift
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    private func extractCardNumber(from textObservations: [VNRecognizedTextObservation]) -> String? {
+        if let cachedCardNumber { return cachedCardNumber }
+
+        let cardNumberMatch = textObservations
+            .compactMap { $0.topCandidates(Constants.topCandidates).first }
+            .filter { $0.confidence > Constants.cardNumberConfidence }
+            .map { $0.string.replacingOccurrences(of: " ", with: "") }
+            .filter { $0.isCardNumber }
```
goergisn:
```suggestion
            .firstWhere { $0.isCardNumber && luhncheck... }
```

AdyenCardScanner/Sources/Formatters/ExpireDateFormatter.swift
```diff
@@ -0,0 +1,51 @@
...
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+protocol ExpireDateFormatting {
+    func date(from string: String) -> Date?
+}
+
+class ExpireDateFormatter: ExpireDateFormatting {
```
goergisn: I believe expireDate does not exist.

it's either expiryDate or expirationDate

AdyenCardScanner/Sources/CardImageParser.swift
nauaros: Done, I forgot we have `firstWhere`

AdyenCardScanner/Sources/Formatters/ExpireDateFormatter.swift
nauaros: Yes, you're right. I read it so many times out there 😂

AdyenCardScanner/Sources/Formatters/ExpirationDateFormatter.swift
```diff
@@ -0,0 +1,51 @@
...
+
+import Foundation
+
+protocol ExpirationDateFormatting {
+    func date(from string: String) -> Date?
+}
+
+class ExpirationDateFormatter: ExpirationDateFormatting {
+
+    enum CardExpireDateFormat {
```
goergisn: 😶‍🌫️

goergisn (approved):

AdyenCardScanner/Sources/Formatters/ExpirationDateFormatter.swift
nauaros: upss

AdyenCardScanner/Sources/CardImageParser.swift
```diff
@@ -0,0 +1,185 @@
...
+        completion: @escaping (CreditCard) -> Void
+    ) {
+        guard let transformedImage = transform(image: image) else { return }
+
+        let recognizeTextRequest = VNRecognizeTextRequest()
+        recognizeTextRequest.recognitionLevel = .accurate
+        recognizeTextRequest.usesLanguageCorrection = false
+
+        let imageRequestHandler = VNImageRequestHandler(ciImage: transformedImage, options: [:])
+        try? imageRequestHandler.perform([recognizeTextRequest])
```
descorp, line 53: Maybe better put it in try/catch instead or make `parse(image: completion)` throws?

AdyenCardScanner/Sources/CardImageParser.swift
```diff
@@ -0,0 +1,185 @@
...
+                return
+            }
+            let card = CreditCard(number: cardNumber, expirationDate: expirationDate)
+            completion(card)
+        }
+    }
+
+    // MARK: - Private
+
+    private func transform(image: CIImage) -> CIImage? {
```
descorp, line 89: No need for `CIImage?`

AdyenCardScanner/Sources/CardImageParser.swift
descorp: `CardNumberValidator` can be moved to `AdyenEncryption` module, since it is useful for both CardComponent and for custom card integration.

PR #2004, opened 2025-02-03
Base UI for payTo

neelSharma12:
# Summary

<new>
Base UI including following:

1. PayTo title text
2. PayTo flow type selection title text
3. Segmented control item including 1. PayID 2.BSB
4. Continue button

</new>

# Demo

Simulator Screenshot - iPhone 16 - 2025-02-03 at 09 36 59

# Ticket

<ticket>
COIOS-864
</ticket>

### DIFF
--- a/AdyenComponents/PayTo/PayToComponent.swift
+++ b/AdyenComponents/PayTo/PayToComponent.swift
@@ -12,8 +12,8 @@ public final class PayToComponent: PaymentComponent,
                                    PresentableComponent {
 
     private enum ViewIdentifier {
-        static let flowSelectionTitleLabelItem = "flowSelectionTitleLabelItem"
-        static let flowSelectionItem = "flowSelectionSegmentedControlItem"
+        static let flowSelectionTitleLabelItem = "flowSelectionTitleLabel"
+        static let flowSelectionItem = "flowSelectionSegmentedControl"
         static let continueButtonItem = "continueButton"
     }
 
@@ -61,6 +61,7 @@ public final class PayToComponent: PaymentComponent,
 
     /// The payment flow selection title  label item.
     internal lazy var flowSelectionTitleLabelItem: FormLabelItem = {
+        // TODO: Add translation
         let item = FormLabelItem(
             text: "How would you like to use Payto?",
             style: configuration.style.footnoteLabel
--- a/AdyenComponents/PayTo/PayToComponent.swift
+++ b/AdyenComponents/PayTo/PayToComponent.swift
@@ -11,6 +11,12 @@ import UIKit
 public final class PayToComponent: PaymentComponent,
                                    PresentableComponent {
 
+    private enum ViewIdentifier {
+        static let flowSelectionTitleLabelItem = "flowSelectionTitleLabelItem"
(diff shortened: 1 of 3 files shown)


AdyenComponents/PayTo/PayToComponent.swift
```diff
@@ -46,4 +58,75 @@ public final class PayToComponent: PaymentComponent,
         self.context = context
         self.configuration = configuration
     }
+
+    /// The payment flow selection title  label item.
+    internal lazy var flowSelectionTitleLabelItem: FormLabelItem = {
+        let item = FormLabelItem(
+            text: "How would you like to use Payto?",
```
erenbesel, line 66: Let's add TODOs so we don't forget localization

erenbesel (approved):
Approved but tiny comment :)

AdyenComponents/PayTo/PayToComponent.swift
goergisn: tbh not sure if we should merge this before we have the translation for it.
Otherwise the purpose of having a clean base branch is a bit defeated.

From my point of view the idea behind a base branch is that it's properly reviewed and ideally would not have to be re-reviewed before merging. By approving and merging TODOs and unfinished code we would still have to re-review the full base branch PR. 🤔

goergisn: Same counts for failing tests/workflows

neelSharma12: Ah!! this we should discuss further :)

atmamont: I have a suggestion:
- we do not block development as waiting for localization can take some time for new project (when we build the train and the tracks at the same time) 
- we add TODO
- we fail every PR to `develop` that has TODOs (swiftlint rule)
- eventually we do not have long living TODOs in the code, they go to YouTrack instead

AdyenComponents/PayTo/PayToComponent.swift
```diff
@@ -11,6 +11,12 @@ import UIKit
 public final class PayToComponent: PaymentComponent,
                                    PresentableComponent {
 
+    private enum ViewIdentifier {
+        static let flowSelectionTitleLabelItem = "flowSelectionTitleLabelItem"
+        static let flowSelectionItem = "flowSelectionSegmentedControlItem"
+        static let continueButtonItem = "continueButton"
```
atmamont, line 17: It's a bit inconsistent, let's remove Item from prior two or ad "item" to the button identifier? If its main purpose it accessibility ID, I suggest removeing "Item" from all string identifiers as it doesn't bring any value on accessibility layer.

atmamont:
+1 to have pipeline always green

AdyenComponents/PayTo/PayToComponent.swift
neelSharma12: Added TODO for string translation.

AdyenComponents/PayTo/PayToComponent.swift
neelSharma12: Removed "item" from string identifier ce0a1a60db9fc6ca39b38da911d6df89429afec1

PR #2012, opened 2025-02-12
PayID tab UI elements

neelSharma12:
# Summary

<new>
- Added Identifier picker UI.
- Added First name input field.
- Added last name input field.
</new>

# Ticket

<ticket>
COIOS-865
</ticket>

Simulator Screenshot - iPhone 16 - 2025-02-17 at 13 18 37


AdyenComponents/PayTo/PayToComponent.swift
```diff
@@ -15,6 +15,29 @@ public final class PayToComponent: PaymentComponent,
...
+    private enum AccountIdentifier: CustomStringConvertible, CaseIterable {
+        case phone
+        case email
+        case abn
+        case organizationID
+
+        public var description: String {
+            switch self {
+            case .phone:
+                return "Phone"
```
erenbesel, line 34: Todo for translation for these words as well. (except ABN i think?)

Adyen/UI/Form/Items/Value Pickers/Identifier Picker/FormAccountIdentifierPickerItem.swift
```diff
@@ -0,0 +1,48 @@
+//
+// Copyright (c) 2025 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+@_spi(AdyenInternal)
+public protocol AccountIdentifier: CustomStringConvertible, Equatable {
```
erenbesel: This protocol seems redundant. Also there is an enum below with the same name

AdyenComponents/PayTo/PayToComponent.swift
```diff
@@ -101,6 +125,58 @@ public final class PayToComponent: PaymentComponent,
...
+            scopeInstance: self,
+            postfix: ViewIdentifier.lastNameInputItem
+        )
+        return item
+    }()
+
+    /// The identifier picker item.
+    internal lazy var identifierPickerItem: FormIdentifierPickerItem = {
+        let selectableValues = AccountIdentifier.allCases.map { account in
+            BasePickerElement(
```
erenbesel: I see both `FormIdentifierPickerElement` and `BasePickerElement` and it's confusing. Shouldn't there be only one of them? In this case, the new one `FormIdentifierPickerElement`

AdyenComponents/PayTo/PayToComponent.swift
```diff
@@ -101,6 +125,58 @@ public final class PayToComponent: PaymentComponent,
...
+                identifier: account.description,
+                element: FormIdentifierPickerElement(
+                    identifier: account.description,
```
erenbesel: account.description used for both identifier and title values

AdyenComponents/PayTo/PayToComponent.swift
neelSharma12: yes, just saw the thread on channel. I added this questions in my to-ask list but now I know the answer :) I gonna add TODO here as well.

Adyen/UI/Form/Items/Value Pickers/Identifier Picker/FormAccountIdentifierPickerItem.swift
neelSharma12: Removed it

AdyenComponents/PayTo/PayToComponent.swift
neelSharma12: Modified the code :)

AdyenComponents/PayTo/PayToComponent.swift
neelSharma12: Ah!! Changed it :)

AdyenComponents/PayTo/PayToComponent.swift
```diff
@@ -9,12 +9,50 @@ import UIKit
...
+                return "Email"
+            case .abn:
+                return "ABN"
+            case .organizationID:
+                return "Organization ID"
+            }
+        }
+
+        public var identifier: String {
+            switch self {
```
atmamont: Can be `self.rawValue`?

nauaros (approved):

AdyenComponents/PayTo/PayToComponent.swift
```diff
@@ -101,6 +140,52 @@ public final class PayToComponent: PaymentComponent,
...
+    }()
+
+    /// The identifier picker item.
+    internal lazy var identifierPickerItem: FormStringPickerItem = {
+        let selectableValues = AccountIdentifiers.allCases.map { accountIdentifier in
+            FormStringPickerElement(identifier: accountIdentifier.identifier, title: accountIdentifier.description)
+        }
+
+        let item = FormStringPickerItem(
+            preselectedIdentifier: selectableValues[0],
```
atmamont: Could be a crash

Tests/IntegrationTests/Components Tests/PayTo/PayToComponentTests.swift
```diff
@@ -77,4 +77,52 @@ class PayToComponentTests: XCTestCase {
...
     }
 
+    func test_identifierPicker_exists() throws {
+        // Given
+        let sut = try PayToComponent(
+            paymentMethod: AdyenCoder.decode(payto),
+            context: Dummy.context
+        )
+
+        setupRootViewController(sut.viewController)
```
atmamont: `sut.viewController.loadIfNeeded()`

AdyenComponents/PayTo/PayToComponent.swift
neelSharma12: This is added in case we need different identifier from rawValue for api call.

atmamont: makes sense!

AdyenComponents/PayTo/PayToComponent.swift
```diff
@@ -109,17 +194,36 @@ public final class PayToComponent: PaymentComponent,
...
-            formViewController.append(continueButton)
+            formViewController.append(continueButtonItem)
         }
 
         return formViewController
     }()
+
+    // MARK: - Private
+
+    private func contentView(formVC: FormViewController) {
```
goergisn: What are these used for? The function name could be a bit more descriptive

AdyenComponents/PayTo/PayToComponent.swift
erenbesel: In this case rawValue can and should be identifier. What else is there to make this enum's elements unique? Nice catch!

Adyen/UI/Form/Items/Value Pickers/Identifier Picker/FormStringPickerItem.swift
```diff
@@ -0,0 +1,49 @@
...
+        self.title = title
+    }
+}
+
+/// A identifier picker form item
+@_spi(AdyenInternal)
+public final class FormStringPickerItem: BaseFormPickerItem<FormStringPickerElement> {
+
+    public init(
+        preselectedIdentifier: FormStringPickerElement,
```
erenbesel: parameter names still have identifier. Can rename to `preselectedValue` or `preselectedStringValue`

AdyenComponents/PayTo/PayToComponent.swift
neelSharma12: if the selectableValues is nil?

Adyen/UI/Form/Items/Value Pickers/Identifier Picker/FormStringPickerItem.swift
neelSharma12: Renamed them :)

AdyenComponents/PayTo/PayToComponent.swift
neelSharma12: It is for constructing UI but I can rename it :)

AdyenComponents/PayTo/PayToComponent.swift
goergisn: If the list is are empty accessing the first element will crash - ideally use `selectableValues.first` and try to unwrap the optional

AdyenComponents/PayTo/PayToComponent.swift
```diff
@@ -109,17 +194,36 @@ public final class PayToComponent: PaymentComponent,
...
-            formViewController.append(continueButton)
+            formViewController.append(continueButtonItem)
         }
 
         return formViewController
     }()
+
+    // MARK: - Private
+
+    private func payIDContentView(formVC: FormViewController) {
```
goergisn: The function still does not really say what it does - based on the name I'd assume it returns a contentView but it actually appends item to the form.

Maybe instead call it `func appendNameInputItems(to formVC: FormViewController)` or something similar to not have to go through the function implementation to figure out what it does.

Also if the function is only used once - it might be even ok to move the calls from the function body to the call site. So all the appending to the form happens in one place.

AdyenComponents/PayTo/PayToComponent.swift
erenbesel: or when you create the array check for isEmpty but directly checking the first would be the clearest way.

neelSharma12: I have added an assertion :)

AdyenComponents/PayTo/PayToComponent.swift
neelSharma12: I have added this function to separate static and dynamic UI content (will be added in later PRs). This function will call methods for static and dynamic UI.

AdyenComponents/PayTo/PayToComponent.swift
neelSharma12: Added rawValue as identifier.

Tests/IntegrationTests/Components Tests/PayTo/PayToComponentTests.swift
neelSharma12: Changed it.

AdyenComponents/PayTo/PayToComponent.swift
neelSharma12: I have renamed the method to `appendItemsTo(formVC: FormViewController)`

PR #2013, opened 2025-02-14
PayTo: Phone number item

erenbesel:
# Summary

Adds the phone number item with single extension (thus no selection).

# Ticket

<ticket>
COIOS-866
</ticket>


Adyen/UI/Form/Items/Phone Number/Phone Extension Picker/FormPhoneExtensionPickerItem.swift
```diff
@@ -19,6 +19,10 @@ extension PhoneExtension: FormPickable {
 @_spi(AdyenInternal)
 public final class FormPhoneExtensionPickerItem: FormPickerItem<PhoneExtension> {
     
+    internal var allowsSelection: Bool {
```
atmamont, line 22: This is outside of this PR scope, why is this called PhoneExtension? This is country code prefix, right?

Tests/IntegrationTests/Components Tests/PayTo/PayToComponentTests.swift
```diff
@@ -60,6 +60,22 @@ class PayToComponentTests: XCTestCase {
...
     }
+    
+    func test_phoneNumberItem_exists() throws {
+        // Given
+        let sut = try PayToComponent(
+            paymentMethod: AdyenCoder.decode(payto),
+            context: Dummy.context
+        )
+
+        setupRootViewController(sut.viewController)
```
atmamont: Please replace it with `viewController.loadIfNeeded()` so this can be an integration not UI test.

Adyen/UI/Form/Items/Phone Number/Phone Extension Picker/FormPhoneExtensionPickerItem.swift
neelSharma12: I think we are using the phone picker to get only one country Code here which is +61 and that's why we are removing the selection on it.

AdyenComponents/PayTo/PayToComponent.swift
```diff
@@ -9,7 +9,7 @@ import UIKit
 
 /// A component that provides PayTo flows for PayTo component.
 public final class PayToComponent: PaymentComponent,
-                                   PresentableComponent {
+    PresentableComponent {
```
neelSharma12, line 12: This is not indented properly I think.

Adyen/UI/Form/Items/Phone Number/Phone Extension Picker/FormPhoneExtensionPickerItem.swift
goergisn: I agree but I believe PhoneExtension is public - so changing it is breaking - we might want to rename it and create a typealias to keep the API stable - but yeah a bit out of scope here 👌

atmamont: Oh, please disregard this specific line I commented on. 

It's about generic naming of this `FormPhoneExtensionPickerItem`, to me it looks more like a `FormPhoneCountryCodePickerItem`

AdyenComponents/PayTo/PayToComponent.swift
```diff
@@ -86,6 +91,18 @@ public final class PayToComponent: PaymentComponent,
...
     }()
+    
+    internal lazy var phoneNumberItem: FormPhoneNumberItem = {
+        let item = FormPhoneNumberItem(
+            phoneNumber: nil,
+            selectableValues: payToExtensions,
+            style: configuration.style.textField,
+            localizationParameters: configuration.localizationParameters,
+            presenter: .init(self)
+        )
```
neelSharma12, line 103: Is title missing here?

Adyen/UI/Form/Items/Phone Number/Phone Extension Picker/FormPhoneExtensionPickerItemView.swift
```diff
@@ -68,7 +68,13 @@ public final class FormPhoneExtensionPickerItemView: FormItemView<FormPhoneExten
     }
     
     internal func setupView() {
-        let stackView = UIStackView(arrangedSubviews: [countryCodeLabel, chevronView, valueLabel])
+        var arrangedSubviews: [UIView] = []
```
goergisn:
```suggestion
        let arrangedSubviews: [UIView]
```

Better make it a let for compile-time safety

Adyen/UI/Form/Items/Phone Number/Phone Extension Picker/FormPhoneExtensionPickerItem.swift
erenbesel: I'm confused, what is out of scope and what is breaking? I don't see any breaking changes

atmamont: Naming is the issue, the name of the class is `FormPhoneExtensionPickerItem`. 
We cannot rename it to `FormPhoneCountryCodePickerItem` because it's `public` now.

And overall this discussion is outside of the scope of current PR, I raised this as `FormPhoneExtensionPickerItem` name doesn't provide correct context.

AdyenComponents/PayTo/PayToComponent.swift
erenbesel: It's set inside the item init

goergisn (approved):

PR #2024, opened 2025-02-21
Payto: segment UI element and UI logic

erenbesel:
# Summary

PayID/BSB segment and UI logic to display/hide related elements.

Simulator Screenshot - iPhone 16 Pro - 2025-02-21 at 13 28 40

# Ticket

<ticket>
COIOS-871
</ticket>

### DIFF
--- a/AdyenComponents/PayTo/PayToComponent.swift
+++ b/AdyenComponents/PayTo/PayToComponent.swift
@@ -342,7 +342,7 @@ public final class PayToComponent: PaymentComponent,
         addTopItems(to: formViewController)
         formViewController.append(FormSpacerItem(numberOfSpaces: 2))
         
-        addDynamicitems(to: formViewController)
+        addDynamicItems(to: formViewController)
         addBottomItems(to: formViewController)
 
         // continue button last
@@ -423,7 +423,7 @@ private extension PayToComponent {
         add(topItems, to: formViewController, spacing: 1)
     }
     
-    func addDynamicitems(to formViewController: FormViewController) {
+    func addDynamicItems(to formViewController: FormViewController) {
         add(
             payIdDynamicItems,
             to: formViewController,
--- a/AdyenComponents/PayTo/PayToComponent.swift
+++ b/AdyenComponents/PayTo/PayToComponent.swift
@@ -11,6 +11,13 @@ import UIKit
 public final class PayToComponent: PaymentComponent,
     PresentableComponent {
 
+    /// The flow types for PayTo component.
+    public enum PayToFlowType: Int {
+        // TODO: Add comments
+        case payID = 0
+        case BSB = 1
+    }
+
     private enum ViewIdentifier {
         static let flowSelectionTitleLabelItem = "flowSelectionTitleLabel"
         static let flowSelectionItem = "flowSelectionSegmentedControl"
@@ -22,6 +29,9 @@ public final class PayToComponent: PaymentComponent,
         static let emailInputItem = "emailTextfield"
         static let abnInputItem = "abnTextfield"
         static let organizationIDInputItem = "organizationIDTextfield"
+        static let accountNumberInputItem = "accountNumberTextfield"
+        static let bankStateBranchInputItem = "bankStateBranchTextfield"
+        static let paymentInstructionTitleLabelItem = "paymentInstructionTitle"
     }
 
     private enum AccountIdentifiers: String, CustomStringConvertible, CaseIterable {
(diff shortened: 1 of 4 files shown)


goergisn (dismissed):

goergisn (approved):

nauaros (approved):

AdyenComponents/PayTo/PayToComponent.swift
```diff
@@ -63,17 +77,47 @@ public final class PayToComponent: PaymentComponent,
...
+    private lazy var selectedPaymentOption: PaymentOption = .payID(selectedIdentifier) {
+        didSet {
+            UIView.performWithoutAnimation {
+                updateInterface()
+            }
+            
+        }
+    }
+    
+    private lazy var payIdDynamicItems: [FormItem] = [
```
atmamont, line 106: In what sense are they dynamic? Do they change later?

AdyenComponents/PayTo/PayToComponent.swift
```diff
@@ -63,17 +77,47 @@ public final class PayToComponent: PaymentComponent,
...
+    private var selectedIdentifier: PayIDIdentifier = .phone
+
+    /// Represents the selected payTo flow for the payment component.
+    /// Determines the specific payTo transaction process to follow.
+    private lazy var selectedPaymentOption: PaymentOption = .payID(selectedIdentifier) {
+        didSet {
+            UIView.performWithoutAnimation {
+                updateInterface()
+            }
+
```
atmamont, line 102: Empty line

AdyenComponents/PayTo/PayToComponent.swift
```diff
@@ -181,12 +227,15 @@ public final class PayToComponent: PaymentComponent,
     }()
 
     /// The identifier picker item.
-    internal lazy var identifierPickerItem: FormStringPickerItem = {
-        let selectableValues = AccountIdentifiers.allCases.map { accountIdentifier in
-            FormStringPickerElement(identifier: accountIdentifier.rawValue, title: accountIdentifier.description)
+    internal lazy var identifierPickerItem: FormContainerItem<FormStringPickerItem> = {
+        let selectableValues = PayIDIdentifier.allCases.map { payIDIdentifier in
```
atmamont, line 231: This triggers some questions:
- is it fine to keep enum order, do we need to change it in the future?
- this is not testable, do we need to test it?

AdyenComponents/PayTo/PayToComponent.swift
```diff
@@ -199,7 +248,8 @@ public final class PayToComponent: PaymentComponent,
             scopeInstance: self,
             postfix: ViewIdentifier.identifierPickerItem
         )
-        return item
+        // we return a container item to prevent the picker from becoming firstResponder
```
atmamont, line 251: Shouldn't this be decided in a view controller? This is clearly UIKit logic leaking to a business logic layer. I tried to check `FormContainerItemView` if it overrides `canBecomeFirstResponder` and I didn't find anything, can you guide me so I get the idea?

AdyenComponents/PayTo/PayToComponent.swift
```diff
@@ -313,10 +387,115 @@ extension PayToComponent: ViewControllerPresenter {
...
+        }
+    }
+    
+    func updateInterface() {
+        switch selectedPaymentOption {
+        case let .payID(identifier):
+            resetPayIdItemsVisibility()
+            
+            switch identifier {
+            case .phone:
```
atmamont, line 474: Just an idea. `payIdDynamicItems` could be a dictionary so this can look like
payIdDynamicItems[identifier].isHidden.wrappedValue = false
This will also simplify helper methods like `resetPayIdItemsVisibility`

AdyenComponents/PayTo/PayToComponent.swift
erenbesel: Ideally yes. But if we don't add this the picker becomes first responder since we make the first eligible item first responder in FormViewController.
Or we try to add a property to this picker element to not trigger first responder.
Or we don't add it at all and have the picker auto open at start.

AdyenComponents/PayTo/PayToComponent.swift
erenbesel: These are hardcoded values that is reportedly not to change.

AdyenComponents/PayTo/PayToComponent.swift
erenbesel: Tried it but amount of code or complexity stayed the same. I'd argue it was not better even. 
`items[.phone]?.isHidden.wrappedValue = false` is worse than `phoneItem.isHidden.wrappedValue = false`

atmamont: Just for the sake of conversation :) 
It could be extracted to a helper method like `func hide(_ item: FormItem) {}`

AdyenComponents/PayTo/PayToComponent.swift
atmamont: Well this sounds like a behavior that should be overriden in a child class, meaning we shouldn't create workarounds in a child class just because of a specific behavior in a base class

PR #2029, opened 2025-02-26
AdyenCardScanner - Public API

nauaros:
# Summary

This PR introduces the public API to be consumed in order to make use of the card scanner SDK.
The SDK's entry point is the function `createCardScanner` in `CardScanner`.

**Usage:**
_Analytics_: Expected events to be sent, detailed in the code block below.

```swift

guard let cardScanningAvailable = CardScanner.isAvailable else {
  // Event 1: Button rendered / card scanning availability
  return
}

let cardScannerViewController = CardScanner.createCardScanner { result in
    switch result {
    case .success(let newCard):
        // Event 3: Successful card scanning.
        // Event 4: Complete / partial scan.
        // Handle success scenario.
    case .failure(let error):
        // Event 5: Failed card scanning.
        // Handle error scenario.
    }
}

// Event 2: Card scanner presented.
// Event 6: Card scanner cancelled.
present(viewController: cardScannerViewController)
```

## Next Steps
- [ ] Add unit tests for `CardImageParser`.
    - _The idea is to have a series of card images against we can write automated tests._
- [ ] Add unit tests for the MVVM module.

# Ticket

<ticket>
COIOS-826
</ticket>


AdyenCardScanner/Sources/CardScanner.swift
```diff
@@ -0,0 +1,36 @@
...
+    ) -> UIViewController?
+
+    static var isAvailable: Bool { get }
+}
+
+public enum CardScanner: CardScanning {
+
+    // MARK: - Properties
+
+    private static let cardScannerAssembler = CardScannerAssembler()
```
erenbesel, line 22: I know it's not injected but is it supposed to be `CardScannerAssembling`?

AdyenCardScanner/Sources/CardScanner.swift
```diff
@@ -0,0 +1,36 @@
...
+
+    // MARK: - CardScanning
+
+    public static func createCardScanner(
+        completion: @escaping (Result<CreditCard, CardScannerError>) -> Void
+    ) -> UIViewController? {
+        let cardScannerViewController = cardScannerAssembler.resolveCardScannerViewController(completion: completion)
```
erenbesel, line 29: I see `CardScannerAssembling `s function and  `CardScanning ` `createCardScanner(..` have the same signatures. (completion and return values). Do we need 2 different ones then? 
The second layer seems redundant as it's practically the same thing.
We can do the resolving elements inside this block maybe?

AdyenCardScanner/Sources/CardScanner.swift
nauaros: Yeah, it is but since it cannot be injected, I just leave it like this. You see any issues with it like this?
I could declare it as` private static let cardScannerAssembler: CardScannerAssembling = CardScannerAssembler()`

AdyenCardScanner/Sources/CardScanner.swift
nauaros: We don't necessarily need the two layers. Everything can be done directly in the CardScanning one. I just prefer to have anything internal within the assembler layer and use this one in the only public class of the SDK.

atmamont:
Could you add your vision on analytics to the sample usage section?

atmamont (approved):

goergisn (approved):

PR #2030, opened 2025-02-26
Payto items provider

erenbesel:
# Summary

Reduce complexity of the component by splitting items logic into a provider and moving the enums to a its own file.

# Ticket

<ticket>
COIOS-000
</ticket>

### DIFF
--- a/AdyenComponents/PayTo/PayToComponent.swift
+++ b/AdyenComponents/PayTo/PayToComponent.swift
@@ -11,55 +11,6 @@ import UIKit
 public final class PayToComponent: PaymentComponent,
     PresentableComponent, AdyenObserver {
 
-    private enum ViewIdentifier {
-        static let flowSelectionTitleItem = "flowSelectionTitleLabel"
-        static let flowSelectionItem = "flowSelectionSegmentedControl"
-        static let phoneNumberItem = "phoneNumberItem"
-        static let continueButtonItem = "continueButton"
-        static let identifierPickerItem = "identifierPicker"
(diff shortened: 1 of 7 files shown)


nauaros (dismissed):
It looks good like this with the items provider.

descorp (dismissed):

nauaros (approved):

AdyenComponents/PayTo/PayToComponent.swift
```diff
@@ -139,198 +100,73 @@ public final class PayToComponent: PaymentComponent,
...
-        let item = FormLabelItem(
-            text: localizedString(LocalizationKey(key: "How would you like to use Payto?"), configuration.localizationParameters),
-            style: configuration.style.textField.text
-        )
-        item.identifier = ViewIdentifierBuilder.build(
-            scopeInstance: self,
-            postfix: ViewIdentifier.flowSelectionTitleItem
-        )
-        return item
+        itemsProvider.flowSelectionTitleItem
```
goergisn, line 103: the `PayToItemsProvider.flowSelectionTitleItem` is already a lazy var - can we make the `PayToComponent.flowSelectionTitleItem` a computed var? 

or the item provider provided version return without keeping track of it?

goergisn (approved):

atmamont (approved):

AdyenComponents/PayTo/PayToComponent.swift
erenbesel: Yeah that was the plan good catch, will fix on next PR since this one is approved

PR #2034, opened 2025-03-03
Payto: validation logic for form items

erenbesel:
# Summary

Validators for form items with tests.

# Ticket

<ticket>
COIOS-872
</ticket>


AdyenComponents/PayTo/PayToComponent.swift
```diff
@@ -100,70 +100,70 @@ public final class PayToComponent: PaymentComponent,
...
     /// The  account holder firstname text input item.
     internal lazy var firstNameInputItem: FormTextInputItem = {
-        itemsProvider.firstNameInputItem
+        itemsProvider.firstNameInputItem()
     }()
 
     /// The  account holder lastname text input item.
     internal lazy var lastNameInputItem: FormTextInputItem = {
-        itemsProvider.lastNameInputItem
+        itemsProvider.lastNameInputItem()
```
goergisn: Maybe to make it more explicit you could name it something like

`itemsProvider.provideLastNameInputItem()`
`itemsProvider.buildLastNameInputItem()`
`itemsProvider.assembleLastNameInputItem()`

Basically just a verb that indicates that it's generating an item instead of accessing one 🤷‍♂️

goergisn (dismissed):

AdyenComponents/PayTo/PayToComponent.swift
atmamont: Agreed, otherwise we hide it inside and it might be improtant to know we get a new instance every time we access it.
Another name option is `makeLastNameInputItem()`

AdyenComponents/PayTo/PayToItemsProvider.swift
```diff
@@ -44,20 +44,41 @@ internal class PayToItemsProvider: PayToItemsProviding {
...
     
     private enum ValidationRegex {
-        static let phone = #"^\+[0-9]{1,3}-[1-9]{1,1}[0-9]{1,29}$"#
+        static let phone = #"^[1-9]{1,1}[0-9]{1,29}$"#
         static let abn = #"^((\d{9})|(\d{11}))$"#
         static let organizationId = #"^[!-@\[-~][ -@\[-~]{0,254}[!-@\[-~]$"#
-        static let bsb = #"^\d{6}"#
-        static let accountNumber = #"^\[ -~]{1,28}$"#
+        static let bsb = #"^\d{6}$"#
+        static let accountNumber = #"^[ -~]{1,28}$"#
```
atmamont, line 51: We don't receive this from backend, right?

AdyenComponents/PayTo/PayToComponent.swift
erenbesel: Alright, in that case, the keyword we seem to use most seems to be `create`. We good with that one?

Tests/IntegrationTests/Components Tests/PayTo/PayToComponentTests.swift
```diff
@@ -20,21 +30,11 @@ class PayToComponentTests: XCTestCase {
     }
 
     func test_paymentMethodType_isPayto() throws {
-        let sut = try PayToComponent(
```
atmamont, line 23: I see a risk of tests not being atomic anymore that some time when the logic becomes convoluted and amount of tests increase we will have side effects applied to sut based on tests execution order. I suggest creating `sut` every time by using `makeSUT()` method that incapsulates all possible variations via parameters injection.

Tests/IntegrationTests/Components Tests/PayTo/PayToComponentTests.swift
```diff
@@ -173,13 +172,33 @@ class PayToComponentTests: XCTestCase {
...
+        XCTAssertFalse(sut.organizationIdInputItem.isValid())
+        sut.organizationIdInputItem.value = "A valid sentence?"
+        XCTAssertFalse(sut.organizationIdInputItem.isValid())
+        sut.organizationIdInputItem.value = "A!V"
+        XCTAssertFalse(sut.organizationIdInputItem.isValid())
+        
+        // valid
+        sut.organizationIdInputItem.value = "123123"
+        XCTAssertTrue(sut.organizationIdInputItem.isValid())
+        sut.organizationIdInputItem.value = "!@"
```
atmamont: This and following values are confusing to say the least!

Tests/IntegrationTests/Components Tests/PayTo/PayToComponentTests.swift
```diff
@@ -189,13 +208,25 @@ class PayToComponentTests: XCTestCase {
         XCTAssertNotNil(organizationIDInputItem, "organizationID input field should exist")
     }
 
-    func test_accountNumber_textfield_exists() throws {
+    func test_accountNumber_textfield() throws {
+        // invalid
+        XCTAssertFalse(sut.accountNumberInputItem.isValid())
```
atmamont, line 229: This should be preceded with explicitly setting value to empty string or nil

Tests/IntegrationTests/Components Tests/PayTo/PayToComponentTests.swift
```diff
@@ -205,13 +236,30 @@ class PayToComponentTests: XCTestCase {
         XCTAssertNotNil(accountNumberInputItem, "Bank account number input field should exist")
     }
 
-    func test_bank_state_number_textfield_exists() throws {
+    func test_bank_state_number_textfield() throws {
+        
+        // invalud
```
atmamont: small typo

AdyenComponents/PayTo/PayToItemsProvider.swift
erenbesel: Unfortunately not

Tests/IntegrationTests/Components Tests/PayTo/PayToComponentTests.swift
erenbesel: updated, they were auto generated values

AdyenComponents/PayTo/PayToComponent.swift
atmamont: Sure, if it is consistent with such cases in the existing codebase

AdyenComponents/PayTo/PayToItemsProvider.swift
atmamont: Can we challenge this and suggest fetching these rules from server?

goergisn (dismissed):

AdyenComponents/PayTo/PayToItemsProvider.swift
erenbesel: I asked at the very beginning and it is not possible. Also they deem these to be highly unlikely to change which adds another reason not to spend effort on such solutions.

atmamont (approved):

goergisn (approved):

PR #2045, opened 2025-03-10
Card scanning - integrate scanning library

atmamont:
# Summary

This change finally makes use of `AdyenCardScanner`

# Demo

Not available for privacy reasons (card data)

# Release notes

<ticket>
COIOS-827
</ticket>

# TODO
- [ ] discuss error handling (errors mapping and UX on the app side) after initial phase of internal testing


AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -9,36 +9,34 @@ import UIKit
 #if canImport(AdyenEncryption)
     import AdyenEncryption
 #endif
+#if canImport(AdyenCardScanner)
+    import AdyenCardScanner
+
+    private let isCardScannerAvailable = true
```
goergisn: Is there a scenario where the AdyenCardScanner is bundled but e.g. the device does not support card scanning? (Thinking about Mac Catalyst, Some iPod touch models maybe?)
Maybe also good to do the iOS version check here (computed property) as well.

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -409,7 +407,33 @@ extension CardViewController: CardViewControllerProtocol {
...
+            )
+            presentViewController(scannerNavigationController, animated: true)
+        }
+    }
+    
+    private func processScannedCard(_ card: AdyenCardScanner.CreditCard) {
+        items.numberContainerItem.numberItem.value = card.number ?? ""
+        
+        if let expirationDate = card.expirationDate {
+            items.expiryDateItem.setExpiryDate(expirationDate)
```
goergisn: Any reason not to use the nil coalescing operator here? (Like you did with the numberItem.value)

AdyenCard/Form/FormCardNumberItemView.swift
```diff
@@ -41,6 +41,7 @@ internal final class FormCardNumberItemView: FormTextItemView<FormCardNumberItem
     }
     
     override public func handleFormattedValueDidChange(_ newValue: String) {
+        textField.text = newValue
```
goergisn, line 44: Why is this needed here?
How did it work before?

AdyenCardScanner/Sources/CardScannerAssembler.swift
```diff
@@ -33,7 +33,10 @@ class CardScannerAssembler: CardScannerAssembling {
     func resolveCardScannerViewController(
         completion: @escaping (Result<CreditCard, CardScannerError>) -> Void
     ) -> UIViewController? {
-        guard let captureDevice else { return nil }
+        guard let captureDevice else {
```
goergisn, line 36: Maybe good to expose this as a boolean to not having to try to open a viewController to get that "information"

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -168,42 +166,42 @@ internal class CardViewController: FormViewController {
...
         guard configuration.socialSecurityNumberMode != .hide else { return nil }
         return items.socialSecurityNumberItem.nonEmptyValue
     }
-
+    
     internal var storePayment: Bool? {
         configuration.showsStorePaymentMethodField ? items.storeDetailsItem.value : nil
     }
-
+
```
erenbesel, line 191: Why do these empty changes happen? Can cause us to miss an actual line change in between

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -409,7 +407,33 @@ extension CardViewController: CardViewControllerProtocol {
...
+            scannerNavigationController.setViewControllers(
+                [scannerViewController],
+                animated: false
+            )
+            presentViewController(scannerNavigationController, animated: true)
+        }
+    }
+    
+    private func processScannedCard(_ card: AdyenCardScanner.CreditCard) {
+        items.numberContainerItem.numberItem.value = card.number ?? ""
```
erenbesel: Let's add the same function logic as setting the expiry date? Even better to add it to `FormCardNumberContainerItem` as a convenience and pass the value down to its CardNumberItem.

AdyenCardScanner/Sources/CardScannerAssembler.swift
```diff
@@ -33,7 +33,10 @@ class CardScannerAssembler: CardScannerAssembling {
...
+            print("Card scanner: cannot resolve view controller, missing capture device")
```
erenbesel, line 38: TODO to remove this when error handling is implemented.

AdyenCard/Components/Card/CardViewController.swift
atmamont: There is another combined check [here](https://github.com/Adyen/adyen-ios/pull/2045/files#diff-e2b149ce65acfbc811da99a674f2533ac7679d427f89c4864cddde18065c5bf6R39).
Would it help renaming this to `isCardScannerLibraryAvailable`?

AdyenCard/Components/Card/CardViewController.swift
atmamont: Passing an optional date into `setExpiryDate()`? Let me know if I follow this right

AdyenCard/Form/FormCardNumberItemView.swift
atmamont: It didn't work before, however we did not have any programmatic .value updates before

AdyenCardScanner/Sources/CardScannerAssembler.swift
atmamont: @nauaros what do you think, could it become a part of API like `public var isCaptureAvailable: Bool`?

AdyenCard/Components/Card/CardViewController.swift
atmamont: I am super keen to figure out why does my swiftformat performs these changes. I'm on the latest Internal repo

AdyenCard/Components/Card/CardViewController.swift
atmamont: Yes we can go for it. The way we expose the whole chain of logic elements and they are always ordered in the way that depends on UI is not great overall.

AdyenCard/Components/Card/CardViewController.swift
atmamont: Currently `AdyenCardScaning` can return `nil` for expiry date, this is why this check belongs here in `CardViewController` as it captures the rule "We do not nullify expiry date if i was entered before as we do not always expect to have it".

@nauaros Can we make card number non-nil in the `AdyenCardScanner` API so it captures framework behavior via API?

atmamont: Great comment Alex, I removed `else` part :)

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -409,7 +407,30 @@ extension CardViewController: CardViewControllerProtocol {
 
 extension CardViewController {
     private func openCardScanner() {
-        // TODO: Use CardScanning module
+        let scannerNavigationController = UINavigationController()
+        if let scannerViewController = AdyenCardScanner.CardScanner.createCardScanner(completion: { [weak self] result in
```
erenbesel: Seems there is nothing after the `if let` so maybe `guard let` to reduce indentation?

AdyenCard/Components/Card/CardViewController.swift
goergisn: Makes it more clear what is actually meant with this flag, yes 👌

AdyenCardScanner/Sources/CardScannerAssembler.swift
atmamont: I found this flag in the CardScanner but still not clear what it would return if camera permissions are still not requested. Will get back to this later when camera permissions are handled in the library.

AdyenCard/Components/Card/CardViewController.swift
atmamont: Done

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -16,30 +16,37 @@ internal protocol CardViewControllerProtocol {
...
-    
     private let initialCountryCode: String
-    
     private let scope: String
-    
     private let cardLogos: [FormCardLogosItem.CardTypeLogo]
+    private lazy var cardScannerController: CardScannerControlling = {
+        let controller = CardScannerController(presenter: self)
+        controller.onScanComplete = { [weak self] result in
+            switch result {
```
goergisn: maybe better to extract this into a function

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -16,30 +16,37 @@ internal protocol CardViewControllerProtocol {
...
+                self?.processScannedCard(number: number, expiryDate: date)
+            case let .failure(error):
+                // TODO: Error handling
+                print("Scanner error: \(error)")
+            }
+        }
+        return controller
+    }()
+    
+    private var isScannerAvailable: Bool { cardScannerController.isScannerAvailable }
```
goergisn: Which scanner?
```suggestion
    private var isCardScannerAvailable: Bool { cardScannerController.isScannerAvailable }
```

goergisn (approved):

AdyenCard/Components/Card/CardViewController.swift
atmamont: Was on my mind and forgot in the end, good one.

AdyenCard/Components/Card/CardViewController.swift
atmamont: Makes sense

neelSharma12 (approved):

PR #2063, opened 2025-03-17
Style card scanning screen

atmamont:
# Summary

This change adds:
- navigation bar title, reusing same string used for "Scan your card" button
- Cancel button (system default, no localization needed)

# Demo

| Light mode | Dark mode |
|--------|--------|
| light | dark |

<ticket>
COIOS-827
</ticket>


AdyenCard/Components/Card/CardScannerController.swift
```diff
@@ -62,14 +66,36 @@ internal protocol CardScannerControlling {
                 .failure(CardScannerError.scanningError)
             }
         }
+
+        private func makeNavigationController() -> UINavigationController {
+            guard #available(iOS 13.0, *) else { return UINavigationController() }
+
+            let appearance = UINavigationBarAppearance()
+            appearance.configureWithDefaultBackground()
```
erenbesel, line 108: A bit rusty on this, will this appearance change not affect all other navigation bars using appearance?

AdyenCard/Components/Card/CardScannerController.swift
```diff
@@ -62,14 +66,36 @@ internal protocol CardScannerControlling {
...
+        }
     }
 
 #else // canImport(AdyenCardScanner)
 
     internal final class CardScannerController: CardScannerControlling {
         internal var isScannerAvailable: Bool { false }
         internal var onScanComplete: ((Result<CardModel, any Error>) -> Void)?
-        internal func openCardScanner() {}
+        internal func openCardScanner(title: String?) {}
```
erenbesel: I'd suggest `openCardScanner(titled: )` or `openCardScanner(with title)`

nauaros: I prefer any of these options.

goergisn: Other thought - this could just be a property on the `CardScannerController` 🤷

atmamont: Good points! I quickly checked with [API guidelines](https://www.swift.org/documentation/api-design-guidelines/):
- `titled` is not aligned with the common API design approach we use across the codebase
- `with title` converts into openCardScanner(with: "Scan Your Card") which has poor clarity
So I'd go with a simple variable. The sacrifice is that one can forget to set it, but at least it follows the `UIViewController` approach which is commonly known

AdyenCard/Components/Card/CardScannerController.swift
atmamont: `UINavigationBar.appearance()` is for global changes, here I am setting this for a specific navigation bar instance.

nauaros (approved):

erenbesel (approved):

neelSharma12 (approved):

PR #2067, opened 2025-03-19
DSP-1867: create a provider for the ThreeDSService

robertdalmeida:
# Summary
Created a type `ThreeDSServiceProvider` which handles creating the ThreeDSService.
This type would consume the configuration and create a service implementation.

<ticket>
DSP-1867
</ticket>

### DIFF
--- a/AdyenActions/Components/3DS2/3DS2 SDK Adapters/ThreeDSServiceProvider.swift
+++ b/AdyenActions/Components/3DS2/3DS2 SDK Adapters/ThreeDSServiceProvider.swift
@@ -1,0 +1,72 @@
+//
+// Copyright (c) 2025 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+@_spi(AdyenInternal) import Adyen
+
+internal protocol ThreeDSConfigurable {
+    var configuration: ThreeDSFeatureChecker? { get set }
+}
+
+internal protocol ThreeDSFeatureChecker {
+    func isFeatureEnabled(_ name: String) -> Bool
+}
+
+/// This type creates a ThreeDSServiceable implementation and consumes ThreeDSConfigurable.
+internal final class ThreeDSServiceProvider: ThreeDSServiceable, ThreeDSConfigurable {
+    
+    /// An optional configuration that can be provided to the service provider.
+    internal var configuration: ThreeDSFeatureChecker?
+    
+    private var service: ThreeDSServiceable?
+    private let serviceCreationHandler: (ThreeDSFeatureChecker?) -> ThreeDSServiceable
+
+    internal init(
+        serviceCreationHandler: @escaping (ThreeDSFeatureChecker?) -> ThreeDSServiceable = createService
+    ) {
+        self.serviceCreationHandler = serviceCreationHandler
+    }
+    
+    internal func performFingerprint(
+        parameters: FingerprintServiceParameters,
+        completionHandler: @escaping (Result<any AnyAuthenticationRequestParameters, ThreeDSServiceFingerprintError>) -> Void
+    ) {
+        self.resetTransaction()
+        let service = serviceCreationHandler(configuration)
+        self.service = service
+        service.performFingerprint(
+            parameters: parameters,
+            completionHandler: completionHandler
+        )
+    }
+    
+    internal func performChallenge(
+        with parameters: ChallengeParameters,
+        completionHandler: @escaping (Result<any AnyChallengeResult, ThreeDSServiceChallengeError>) -> Void
+    ) {
+        // The service needs to be already created during the fingerprint step.
+        // There is no way a challenge action should happen without a fingerprint action.
+        guard let service = self.service else {
+            completionHandler(.failure(.transactionNotInitialized))
+            return
+        }
+        service.performChallenge(with: parameters) { result in
+            self.resetTransaction()
+            completionHandler(result)
+        }
+    }
+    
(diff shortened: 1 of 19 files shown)


AdyenActions/Components/3DS2/3DS2 SDK Adapters/DefaultThreeDSServiceProvider.swift
```diff
@@ -0,0 +1,72 @@
...
+    
+    internal func resetTransaction() {
+        service?.resetTransaction()
+        service = nil
+    }
+    
+    internal static func creator(
+        configuration: ThreeDSFeatureConfig?
+    ) -> ThreeDSServiceable {
+        ThreeDSServiceLegacy()
```
atmamont: Later we would branch here to switch between legacy and new SDK, right?

AdyenActions/Components/3DS2/3DS2 SDK Adapters/DefaultThreeDSServiceProvider.swift
```diff
@@ -0,0 +1,72 @@
...
+internal final class DefaultThreeDSServiceProvider: ThreeDSServiceable, ThreeDSConfigurable {
+    
+    /// An optional configuration that can be provided to the service provider.
+    internal var configuration: ThreeDSFeatureConfig?
+    
+    private var service: ThreeDSServiceable?
+    private let serviceHandler: (ThreeDSFeatureConfig?) -> ThreeDSServiceable
+
+    internal init(
+        serviceHandler: @escaping (ThreeDSFeatureConfig?) -> ThreeDSServiceable = creator
```
atmamont: I jumped onto a long existential thought train on this line

AdyenActions/Components/3DS2/3DS2 SDK Adapters/DefaultThreeDSServiceProvider.swift
robertdalmeida: Exactly.

AdyenActions/Components/3DS2/3DS2 SDK Adapters/DefaultThreeDSServiceProvider.swift
robertdalmeida: Changing this to `serviceCreator`, thanks for the tip.

AdyenActions/Components/3DS2/3DS2 SDK Adapters/DefaultThreeDSServiceProvider.swift
erenbesel: Then why is it also injectable if this part is the decider of which service to provide? A bit cloudy to me
On the naming what about  just `defaultService`?

robertdalmeida: It is injectable to make this type testable.
For the name right now, i've renamed it to `defaultServiceCreator` hope that is good enough?

erenbesel: To be more clear, it's great that it's injectable. But if it is, and also by looking at the responsibility of this class, seems like it shouldn't be the one deciding which service to create but rather it should come from another part that decides/creates the service and injects it into here.
The function `defaultServiceCreator` implies it doesn't have any deciding logic in it but only creates one default/base service.

robertdalmeida: @erenbesel Actually, i did expect this type to decide the service & provide the service(decorator), otherwise this type doesn't have any other value.  :)
I do see that it can be read as "default--ServiceCreator" or "defaultService--Creator" which may lead to confusion. 🤯
Do you think if i renamed it to `defaultService` it would be better? or what would you suggest? I'm ok with anything here as long as it is easy to understand for anyone.

erenbesel: It is confusing a bit indeed. Why do we need to have the `default` keyword? default infers a base/fallback sort of provider when other options are not available. How about just `ThreeDSServiceProvider`?

robertdalmeida: Discussed it with Eren - 
DefaultThreeDSServiceProvider -> ThreeDSServiceProvider.
defaultServiceCreator -> CreateService

AdyenActions/Components/3DS2/3DS2 SDK Adapters/ThreeDSServiceProvider.swift
```diff
@@ -0,0 +1,72 @@
...
+//
+
+import Foundation
+@_spi(AdyenInternal) import Adyen
+
+internal protocol ThreeDSConfigurable {
+    var configuration: ThreeDSFeatureConfig? { get set }
+}
+
+internal protocol ThreeDSFeatureConfig {
```
erenbesel: ThreeDSFeatureConfiguration maybe? to not have shortened words

robertdalmeida: changed it to. - ThreeDSFeatureChecker. Thanks

erenbesel (approved):

goergisn (approved):

atmamont (approved):

PR #2068, opened 2025-03-20
Change card details output type

nauaros:
# Summary

- Make `CreditCard` internal.
- Use card details tuple in completion.

# Ticket

<ticket>
COIOS-000
</ticket>

### DIFF
--- a/AdyenCardScanner/Sources/CardScanner.swift
+++ b/AdyenCardScanner/Sources/CardScanner.swift
@@ -7,9 +7,11 @@
 import Foundation
 import UIKit
 
+public typealias CardDetails = (number: String?, expirationDate: Date?)
+
 protocol CardScanning {
     static func createCardScanner(
-        completion: @escaping (Result<CreditCard, CardScannerError>) -> Void
+        completion: @escaping (Result<CardDetails, CardScannerError>) -> Void
     ) -> UIViewController?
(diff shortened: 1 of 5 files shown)


atmamont (approved):
Perfect!

AdyenCardScanner/Sources/CardScanner.swift
```diff
@@ -7,9 +7,11 @@
 import Foundation
 import UIKit
 
+public typealias CardDetails = (number: String?, expirationDate: Date?)
```
erenbesel: We have a `CardDetails` struct, this may collide with it

atmamont: It's a framework, and its usage would be strictly limited to one file `CardScannerController.swift` and I will just use namespace prefix for that matter. Do you think the rest of the existing occurrences would conflict with this one that is not imported in those files?

nauaros: I can rename it into `CardData` or `CardScanDetails` instead

nauaros: Renamed to `CardScanDetails`. This way we avoid any possibility of collision.

erenbesel (approved):

PR #2069, opened 2025-03-24
Tests | CardImageParser

nauaros:
# Summary

Unit tests for:
- [x] CardImageParser
- [x] ExpirationDateFormatter

# Ticket

<ticket>
COIOS-000
</ticket>


AdyenCardScannerTests/CardImageParserTests.swift
```diff
@@ -0,0 +1,153 @@
...
+        sut.parse(image: testCreditCard.image) { receivedCreditCard in
+            expectation.fulfill()
+
+            let expectedCreditCard = testCreditCard.creditCard
+            XCTAssertEqual(expectedCreditCard.number, receivedCreditCard.number)
+            XCTAssertEqual(expectedCreditCard.expirationDate, receivedCreditCard.expirationDate)
+        }
+
+        // Then
+        waitForExpectations(timeout: 10.0)
```
neelSharma12: We have added `wait` api in  `XCTestCase extension`. Should we use that one instead of using `waitForExpectations` ?

neelSharma12 (approved):

AdyenCardScannerTests/CardImageParserTests.swift
atmamont: We do not expect 10 seconds delay in real scenario, let's reduce it to something close to reality with a safe buffer for CI variability? Something Like 0.1-0.5.
Personally I would also never use generic "wait for all expectations" as it reduces readability, it is safer for the future to use a specific expectations array we want to wait for.
Example:
`wait(for: [expectation], timeout: 0.1)`
Makes a difference when you wait for multiple expectations and you can tell all of them from this one line.

AdyenCardScannerTests/CardImageParserTests.swift
```diff
@@ -0,0 +1,153 @@
...
+        )
+        return TestCreditCard(
+            image: originalImage,
+            creditCard: creditCard
+        )
+    }
+
+
+    private func dateFrom(_ string: String) -> Date? {
+        let dateFormatter = ExpirationDateFormatter()
```
atmamont: Could be a part of test class along with `sut` so it is not instantiated every time (formatters are expensive).

AdyenCardScannerTests/ExpirationDateFormatterTests.swift
```diff
@@ -0,0 +1,71 @@
...
+@testable import AdyenCardScanner
+
+final class ExpirationDateFormatterTests: XCTestCase {
+
+    var sut: ExpirationDateFormatter!
+    var dateFormatter: DateFormatter!
+
+    override func setUpWithError() throws {
+        try super.setUpWithError()
+        dateFormatter = DateFormatter()
```
atmamont: Could be instantiated once per test suite.

erenbesel (approved):

AdyenCardScannerTests/Mocks/ExpirationDateFormattingMock.swift
```diff
@@ -0,0 +1,29 @@
+//
+// Copyright (c) 2025 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+@testable import AdyenCardScanner
+import Foundation
+
+class ExpirationDateFormattingMock: ExpirationDateFormatting {
```
atmamont, line 10: Is it a good time to think of adding sourcery for mock generation?

AdyenCardScannerTests/CardImageParserTests.swift
nauaros: Good point, I change to the suggested example. We keep it more realistic this way.

AdyenCardScannerTests/CardImageParserTests.swift
nauaros: Done

AdyenCardScannerTests/ExpirationDateFormatterTests.swift
nauaros: Done

AdyenCardScannerTests/Mocks/ExpirationDateFormattingMock.swift
nauaros: Yeah, I thought of it like we did for MyStore. Since we don't have that many mocks, I decided to skip it for now. But yeah, it can be a good addition.

atmamont (approved):

PR #2075, opened 2025-03-26
Configuration builder base

erenbesel:
# Summary

Foundation work start on v6. Configuration builder skeleton.

# Ticket

<ticket>
COIOS-888
</ticket>


Adyen/CheckoutConfiguration/ConfigurationBuilder.swift
```diff
@@ -0,0 +1,29 @@
+//
+// Copyright (c) 2025 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+@resultBuilder
+public struct ConfigurationBuilder {
```
goergisn:
```suggestion
public struct CheckoutConfigurationBuilder {
```

goergisn: It's a public struct and `ConfigurationBuilder` is quite generic - so maybe `CheckoutConfigurationBuilder` is a safer name :)

goergisn (commented):
Looks great already!

Adyen/CheckoutConfiguration/CheckoutConfiguration.swift
```diff
@@ -0,0 +1,60 @@
...
+    ///   - environment: The environment to retrieve internal resources from.
+    ///   - amount: Payment amount.
+    ///   - clientKey: The client key that corresponds to the web service user you will use for initiating the payment.
+    ///   - content: Configuration builder to provide the desired configuration instances.
+    ///   See https://docs.adyen.com/user-management/client-side-authentication for more information.
+    /// - Throws: `ClientKeyError.invalidClientKey` if the client key is invalid.
+    public init(
+        environment: Environment,
+        amount: Amount,
+        clientKey: String, @ConfigurationBuilder content: () -> CheckoutConfigurable
```
goergisn:
```suggestion
        clientKey: String, 
        @ConfigurationBuilder content: () -> CheckoutConfigurable
```

Adyen/CheckoutConfiguration/ConfigurationBuilder.swift
erenbesel: It was CheckoutConfigurationBuilder but thought to myself, why not ConfigurationBuilder, how common could that be

Adyen/CheckoutConfiguration/CheckoutConfiguration.swift
erenbesel: Swiftlint's rule seems to be buggy with this. If I move it down it complains that it's not aligned

PR #2081, opened 2025-04-01
DSP-2075: Add option to force a redirect

robertdalmeida:
# Summary

Add test app option to force redirect flow. 

# Demo

Simulator Screenshot - iPhone 16 Pro - 2025-04-01 at 14 46 26# Ticket

<ticket>
DSP-2075
</ticket>

### DIFF
--- a/Demo/Common/Configuration/ConfigurationView.swift
+++ b/Demo/Common/Configuration/ConfigurationView.swift
@@ -16,6 +16,7 @@ internal struct ConfigurationView: View {
         case region = "Region"
         case payment = "Payment"
         case components = "Components"
+        case authentication = "Authentication"
         case dropIn = "DropIn"
     }
     
@@ -51,6 +52,7 @@ internal struct ConfigurationView: View {
                 wrapInSection(view: paymentSection, section: .payment)
                 wrapInSection(view: dropInSection, section: .dropIn)
                 wrapInSection(view: componentsSection, section: .components)
+                wrapInSection(view: authenticationSection, section: .authentication)
             }.navigationBarTitle("Configuration", displayMode: .inline)
                 .navigationBarItems(
                     leading: Button("Default", action: viewModel.defaultTapped),
@@ -148,6 +150,12 @@ internal struct ConfigurationView: View {
             cardComponentSection
             applePaySection
             analyticsSection
+        }
+    }
+    
+    private var authenticationSection: some View {
+        Toggle(isOn: $viewModel.allowForceCardRedirectAction) {
+            Text("Force Card RedirectAction")
         }
     }
 
--- a/Demo/Common/Configuration/ConfigurationViewModel.swift
+++ b/Demo/Common/Configuration/ConfigurationViewModel.swift
@@ -26,6 +26,10 @@ internal final class ConfigurationViewModel: ObservableObject {
     @Published internal var allowDisablingStoredPaymentMethods: Bool = false
     @Published internal var allowsSkippingPaymentList: Bool = false
     @Published internal var allowPreselectedPaymentView: Bool = true
+    
+    /// Controls the setting to configure the option to force redirect action to authenticate using a Card instead of native.
+    @Published internal var allowForceCardRedirectAction: Bool = false
+    
     @Published internal var applePayMerchantIdentifier: String = ""
     @Published internal var allowOnboarding: Bool = false
     @Published internal var analyticsIsEnabled: Bool = true
@@ -61,6 +65,7 @@ internal final class ConfigurationViewModel: ObservableObject {
         self.allowDisablingStoredPaymentMethods = configuration.dropInSettings.allowDisablingStoredPaymentMethods
         self.allowsSkippingPaymentList = configuration.dropInSettings.allowsSkippingPaymentList
         self.allowPreselectedPaymentView = configuration.dropInSettings.allowPreselectedPaymentView
+        self.allowForceCardRedirectAction = configuration.threeDSConfigurationSettings.allowForceCardRedirectAction
         self.applePayMerchantIdentifier = configuration.applePaySettings.merchantIdentifier
         self.allowOnboarding = configuration.applePaySettings.allowOnboarding
         self.analyticsIsEnabled = configuration.analyticsSettings.isEnabled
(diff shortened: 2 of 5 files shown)


Demo/Common/Configuration/ConfigurationView.swift
```diff
@@ -131,6 +131,9 @@ internal struct ConfigurationView: View {
                 placeholder: "Amount",
                 keyboardType: .numberPad
             )
+            Toggle(isOn: $viewModel.allowForcedRedirect) {
```
robertdalmeida: I'm not sure if this should sit here.  As this has to do with the payments request i kept it along with the `Payment` section. 
Please advice.

goergisn: Maybe put it into a separate 3DS2 section? Or add some info about this toggle being related to 3DS2 - otherwise we could mistake it for being related to Native Redirects

robertdalmeida: Thanks for the quick feedback, i've moved it to another section called "Authentication" and renamed the option to "Force card RedirectAction". How does this sound?

Demo/Common/Configuration/ConfigurationViewModel.swift
```diff
@@ -26,6 +26,7 @@ internal final class ConfigurationViewModel: ObservableObject {
     @Published internal var allowDisablingStoredPaymentMethods: Bool = false
     @Published internal var allowsSkippingPaymentList: Bool = false
     @Published internal var allowPreselectedPaymentView: Bool = true
+    @Published internal var allowForcedRedirect: Bool = false
```
atmamont: This name lacks proper context the same way as tableview comment from @goergisn, let's make clear it belongs to authentication/3ds

Demo/Common/Networking/PaymentsRequest.swift
```diff
@@ -68,6 +68,39 @@ internal struct PaymentsRequest: APIRequest {
         try container.encodeIfPresent(data.checkoutAttemptId, forKey: .checkoutAttemptId)
     }
     
+    private struct AdditionalData: Encodable {
+        let allow3DS2: Bool?
+        let executeThreeD: Bool
+        
+        enum Configuration {
```
atmamont, line 75: We have similar naming for app to app redirects, including "native". I suggest giving more context to the `AdditionalData` namespace

Demo/Common/Networking/SessionRequest.swift
```diff
@@ -65,6 +64,13 @@ internal struct SessionRequest: APIRequest {
         if ConfigurationConstants.current.dropInSettings.allowDisablingStoredPaymentMethods {
             try container.encode(true, forKey: .showRemovePaymentMethodButton)
         }
+        
+        if ConfigurationConstants.current.threeDSConfigurationSettings.allowForcedRedirect {
```
atmamont: This reads great!

Demo/Common/Networking/PaymentsRequest.swift
```diff
@@ -68,6 +68,39 @@ internal struct PaymentsRequest: APIRequest {
...
+            case native
```
atmamont: We have similar naming for app to app redirects, including "native". I suggest giving more context to the `AdditionalData` namespace

Demo/Common/Networking/SessionRequest.swift
```diff
@@ -65,6 +64,13 @@ internal struct SessionRequest: APIRequest {
...
+            try container.encode(
+                AuthenticationData(configuration: .forceRedirect),
```
erenbesel: I see this and paymentsRequest are slightly different. Can we not use the same for both?

erenbesel (dismissed):

Demo/Common/Configuration/ConfigurationViewModel.swift
robertdalmeida: As this is a view model property, shall i rename this to?
```
@Published internal var allowForceCardRedirectAction: Bool = false
```

Demo/Common/Networking/PaymentsRequest.swift
robertdalmeida: Will this give better context to the configuration type?

```
    enum Configuration {
        case nativeThreeDSActionForAuthentication
        case cardRedirectActionForAuthentication
    }

```

Demo/Common/Networking/SessionRequest.swift
robertdalmeida: The request for sessions to trigger a redirect is different. 
I've now kept the configuration enum values to make the connection between them. Would this work? 

        enum AuthenticationConfiguration {
            case nativeThreeDSAction
            case cardRedirectAction
        }

Demo/Common/Networking/PaymentsRequest.swift
atmamont: Or it could be adding context to the enum: `enum AuthenticationConfiguration` and then using full enum like `AuthenticationConfiguration.threeDsNative`

robertdalmeida: That is better indeed. how does it look now?

atmamont (dismissed):

Demo/Common/Networking/PaymentsRequest.swift
atmamont: Yes it reads much easier now and has zero ambiguity!

Demo/Common/Networking/SessionRequest.swift
```diff
@@ -65,6 +64,13 @@ internal struct SessionRequest: APIRequest {
         if ConfigurationConstants.current.dropInSettings.allowDisablingStoredPaymentMethods {
             try container.encode(true, forKey: .showRemovePaymentMethodButton)
         }
+        
+        let configuration: AuthenticationData.AuthenticationConfiguration = ConfigurationConstants.current.threeDSConfigurationSettings.allowForceCardRedirectAction ?
+            .cardRedirectAction : .nativeThreeDSAction
+        try container.encodeIfPresent(
```
erenbesel, line 71: Looks cleaner indeed. But can we unify the encoding usage? One has its own function `encodeAdditionalData...` and here it's seemingly the similar encoding without that function. Let's choose one way

Demo/Common/Networking/PaymentsRequest.swift
```diff
@@ -59,7 +59,7 @@ internal struct PaymentsRequest: APIRequest {
         try container.encode(currentConfiguration.countryCode, forKey: .countryCode)
         try container.encode(ConfigurationConstants.returnUrl.absoluteString, forKey: .returnUrl)
         try container.encode(ConfigurationConstants.shopperReference, forKey: .shopperReference)
-        try container.encode(ConfigurationConstants.additionalData, forKey: .additionalData)
+        try encodeAdditionalData(to: &container, currentConfiguration: currentConfiguration)
```
erenbesel: `encodeAdditionalData(in container...` ?

robertdalmeida: Removed this method.

Demo/Common/Networking/SessionRequest.swift
robertdalmeida: Sure, removed the method encodeAdditionalData.

erenbesel (approved):

nauaros (approved):

atmamont (approved):

PR #2082, opened 2025-04-02
v6 adyenCheckout base

erenbesel:
# Summary

Base for the new AdyenCheckout object. Contains code regarding:
- How super component can build components via ComponentBuilder
- How CheckoutConfiguration stores internal configurations, passes them to relevant components
- New protocol to help with the above CheckoutComponentConfiguration
- BLIK configuration following the new interface (not done)

# Ticket

<ticket>
COIOS-884
</ticket>


AdyenCheckout/AdyenCheckout.swift
```diff
@@ -9,11 +9,39 @@
 @_spi(AdyenInternal) import AdyenSession
 @_spi(AdyenInternal) import AdyenNetworking
 
-public class AdyenCheckout {
+internal protocol AdyenCheckoutProtocol {
+    static func setup(
+        with sessionId: String,
+        sessionData: String,
+        configuration: CheckoutConfiguration,
+        completion: @escaping (Result<AdyenCheckout, Error>) -> Void
```
descorp: Should we also have `async` analog(s) ?

descorp (approved):

neelSharma12 (approved):

AdyenCheckout/AdyenCheckout.swift
atmamont: We had a quick conversation on that. Rewriting everything internally is not an option right now, so at some point we will have a layer for just converting async into completions. 
While it's nice to support this feature, we could add it later in minor v6 updates as an addition to the current API.

AdyenCheckout/AdyenCheckout.swift
```diff
@@ -9,11 +9,39 @@
...
+    )
+    
+    static func setupSession(
+        with configuration: CheckoutConfiguration,
+        order: PartialPaymentOrder?,
+        completion: @escaping (Result<AdyenSession, Error>) -> Void
+    )
+    
+    static func fetchCheckoutAttemptId(
```
atmamont: I'm in doubt if this belongs here. Let's apply interface segregation and introduce `CheckoutAttemptIdProviding`?

Adyen/CheckoutConfiguration/CheckoutComponentType.swift
```diff
@@ -0,0 +1,39 @@
+//
+// Copyright (c) 2025 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+public enum CheckoutComponentType: Hashable {
+    case payment(PaymentMethodType)
```
atmamont, line 10: Is there a world where these two have same abstract public API and we hide the complexity inside?

atmamont (approved):
Got some questions but maybe we will go through them face to face.
Shaping out nicely!

PR #2091, opened 2025-04-08
Refactor card scanner completion

atmamont:
# Summary

This PR:
- fixes tests target by adding analytics handler where it is needed (no specific analytics coverage though in the current scope)
- wraps card scanner provider into addiitional layer using composition. New layer responsibility is to dispatch card scanner completion once. Unit tests added.

This is a short-term solution, this problem belongs more to the `AdyenCardScanner` module.

# Todo
- [x] fix broken integrations for non-included AdyenCardScanner case
- [x] unit tests
- [x] manual testing with logs check

# Ticket

<ticket>
COIOS-886
</ticket>

### DIFF
--- a/AdyenCard/Components/Card/CardScannerController.swift
+++ b/AdyenCard/Components/Card/CardScannerController.swift
@@ -41,8 +41,28 @@ internal protocol CardScannerControlling: CardScannerAvailability {
         }
     }
 
-    private struct CardScannerProviderWrapper: CardScannerProviding {
-        func createCardScanner(completion: @escaping (Result<CardScanDetails, Error>) -> Void) -> UIViewController? {
+    internal class CardScannerProviderDispatchOnce: CardScannerProviding {
+        private let scannerProvider: CardScannerProviding
+
+        internal init(scannerProvider: CardScannerProviding) {
+            self.scannerProvider = scannerProvider
+        }
+
+        private var isDispatched = false
+
+        private var completion: ((Result<CardScanDetails, any Error>) -> Void)?
+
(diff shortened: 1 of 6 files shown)


AdyenCard/Components/Card/CardScannerController.swift
```diff
@@ -20,13 +20,6 @@ internal protocol CardScannerProviding {
 }
 
 internal protocol CardScannerControlling: CardScannerAvailability {
-
-    init(
```
atmamont, line 24: @nauaros 
Having this defined on the interface level typically creates unnecessary load when it comes down to juggling around with various entities conforming to a specific protocol. It also exposes some implementation details and dependencies on the interface level.

neelSharma12 (approved):

erenbesel (approved):

AdyenCard/Components/Card/CardScannerController.swift
```diff
@@ -41,8 +34,26 @@ internal protocol CardScannerControlling: CardScannerAvailability {
...
+
+        internal init(scannerProvider: CardScannerProviding) {
+            self.scannerProvider = scannerProvider
+        }
+
+        private var isDispatched = false
+
+        internal func createCardScanner(completion: @escaping (Result<CardScanDetails, any Error>) -> Void) -> UIViewController? {
+            self.scannerProvider.createCardScanner { result in
+                guard !self.isDispatched else { return }
```
nauaros, line 48: Thanks! It's a provisional solution that I will address from the SDK side.

PR #2095, opened 2025-04-14
COIOS-252: AdyenCardScanner

atmamont:
# Ticket

<ticket>
COIOS-252
</ticket>


Demo/UIKit/Info.plist
```diff
@@ -128,13 +126,17 @@
...
+	<string>12.0.0</string>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
 	<key>NSAppTransportSecurity</key>
 	<dict>
 		<key>NSAllowsLocalNetworking</key>
 		<true/>
 	</dict>
+	<key>NSCameraUsageDescription</key>
+	<string>Scan credit card details</string>
```
atmamont, line 139: This should be in the docs, and this seems to be a breaking change actually. My guess that a simple upgrade to the next SDK version will cause app crash on "scan your card" button tap.  I need to investigate and make a decision if necessary.

@erenbesel @neelSharma12 @nauaros

atmamont: It is a bot of overthinking from my side. It would not crash unless one imports AdyenCardScanner.

PR #2097, opened 2025-04-16
Expose `AdyenCardScanner` for dependency managers

atmamont:
# Summary

This is the part of exploration on how to make `canImport` compile time check work in the app that uses Cocoapods (MyStore).
The only way I found to produce a separate framework file is to expose dependency as a separate pod.
This change:
- adds podspec for AdyenCardScanner 
- makes its code fully compatible with iOS12, otherwise no merchant could build an iOS12 project with this framework
- replacing compile time canImport checks with some availability checks (typealias CardScannerController = DummyCardScannerController) as a result of approach change
- exposing `AdyenCardScanner` as SPM library

# Ticket

<ticket>
COIOS-894
</ticket>

### DIFF
--- a/Adyen.podspec
+++ b/Adyen.podspec
@@ -60,14 +60,14 @@ Pod::Spec.new do |s|
     }
   end
 
-  s.subspec 'AdyenCardScanner' do |plugin|
-    plugin.source_files = 'AdyenCardScanner/**/*.swift'
-    plugin.framework_name = 'AdyenCardScanner'
-    plugin.pod_target_xcconfig = { 
-      'PRODUCT_MODULE_NAME' => 'AdyenCardScanner',
-      'DEFINES_MODULE' => 'YES'
-    }  
-  end
+  # s.subspec 'AdyenCardScanner' do |plugin|
+  #   plugin.source_files = 'AdyenCardScanner/**/*.swift'
+  #   plugin.framework_name = 'AdyenCardScanner'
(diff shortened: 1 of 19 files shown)


Adyen.podspec
```diff
@@ -15,7 +15,7 @@ Pod::Spec.new do |s|
   s.swift_version = '5.7'
   s.frameworks = 'Foundation'
   s.default_subspecs = 'Core', 'Components', 'Actions', 'Card', 'Encryption', 'DropIn', 'Session'
-  s.pod_target_xcconfig = {'SWIFT_SUPPRESS_WARNINGS' => 'YES', 'BUILD_LIBRARY_FOR_DISTRIBUTION' => 'YES' }
+  s.pod_target_xcconfig = {'SWIFT_SUPPRESS_WARNINGS' => 'YES' }
```
neelSharma12, line 18: Why are we removing `BUILD_LIBRARY_FOR_DISTRIBUTION' => 'YES` ?

atmamont, line 18: It was added during my experimentations with cocoapods in the target branch without a PR. This line is just reverting things to as they were before. Sorry for confusion!

neelSharma12: Yes, I figured.I already deleted the comment but I think it was not updated :)

neelSharma12 (approved):

erenbesel (approved):

PR #2098, opened 2025-04-16
Use new scanner entity every scan attempt

atmamont:
# Summary

After introducing new callback controller that dispatches scanner completion only once, an issue appeared caused by the dropin retaining all components. This fix modifies `CardScannerProviderDispatchOnce` behavior in such a way that it resets its state after each `createCardScanner` request and becomes reusable.

# Note

I had to disable podspec linting because it doesn't work with local podspecs and always fetches them from the repo.
Whenever (and if) we publish `AdyenCardScanner.podspec` to the main repo, we can re-enable it.

# Ticket

<ticket>
COIOS-826
</ticket>

### DIFF
--- a/Adyen.podspec
+++ b/Adyen.podspec
@@ -60,7 +60,6 @@ Pod::Spec.new do |s|
   end
 
   s.subspec 'CardScanner' do |plugin|
-    plugin.dependency 'Adyen/Card'
     plugin.dependency 'AdyenCardScanner'
   end
 
--- a/Adyen.podspec
+++ b/Adyen.podspec
@@ -61,7 +61,7 @@ Pod::Spec.new do |s|
 
   s.subspec 'CardScanner' do |plugin|
     plugin.dependency 'Adyen/Card'
-    plugin.dependency 'AdyenCardScanner', :git => 'https://github.com/Adyen/adyen-ios.git', :branch => 'chore/ocr-cocoapods'
+    plugin.dependency 'AdyenCardScanner'
   end
 
   s.subspec 'Components' do |plugin|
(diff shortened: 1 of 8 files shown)


erenbesel (dismissed):
Approved but what is the api diff warning? 🤔

atmamont:
> Non SPM Bundle Extension

It was an unnecessary record in the Package.swift for the newly exposed module. It relates to the base `ocr-cocoapods` branch but I will fix it here.

Adyen.podspec
```diff
@@ -59,14 +59,10 @@ Pod::Spec.new do |s|
...
-  # s.subspec 'AdyenCardScanner' do |plugin|
-  #   plugin.source_files = 'AdyenCardScanner/**/*.swift'
-  #   plugin.framework_name = 'AdyenCardScanner'
-  #   plugin.pod_target_xcconfig = { 
-  #     'PRODUCT_MODULE_NAME' => 'AdyenCardScanner',
-  #     'DEFINES_MODULE' => 'YES'
-  #   }  
-  # end
+  s.subspec 'CardScanner' do |plugin|
+    plugin.dependency 'Adyen/Card'
```
erenbesel: Wouldn't this create cyclic dependency? Card module will be importing this module right?

atmamont: This is just a safety check but I see your point. Card scanning doesn't make sense without the Card, right? On the other hand right now Card is one of default pods.
Should I remove it?

erenbesel: Yeah, I think so.

erenbesel (dismissed):

erenbesel (dismissed):

erenbesel (dismissed):

neelSharma12 (dismissed):

PR #2103, opened 2025-04-23
Introducing package access modifier

erenbesel:
# Summary

Introducing `package` modifier for the newly added v6 related interface, in hopes to replace SPI.

# Ticket

<ticket>
COIOS-898
</ticket>


AdyenCheckout/AdyenCheckout.swift
```diff
@@ -100,7 +99,7 @@ public class AdyenCheckout: AdyenCheckoutProtocol {
     }
     
     public func createComponent(with paymentMethod: any PaymentMethod) -> AdyenCheckoutComponent? {
-        CheckoutComponentBuilder.build(for: paymentMethod, configuration: configuration)
+        try? CheckoutComponentBuilder.build(for: paymentMethod, configuration: configuration)
```
descorp, line 109: Maybe we should be bold and make this method `throws`.  It will be very helpful for Apple Pay and SDK-based PMs

AdyenCheckout/CheckoutComponentBuilder.swift
```diff
@@ -9,30 +9,23 @@
...
-            
-        }
-        
-        
-        
+        } else if let atomePaymentMethod = paymentMethod as? AtomePaymentMethod {}
         
         // TODO: create real checkout errors
-        throw fatalError()
+        fatalError()
```
descorp, line 31: What happened here?

nauaros (approved):

Adyen.podspec
```diff
@@ -12,10 +12,14 @@ Pod::Spec.new do |s|
   s.source = { :git => 'https://github.com/Adyen/adyen-ios.git', :tag => "#{s.version}" }
   s.platform = :ios
   s.ios.deployment_target = '12.0'
-  s.swift_version = '5.7'
+  s.swift_version = '5.9'
   s.frameworks = 'Foundation'
-  s.default_subspecs = 'Core', 'Components', 'Actions', 'Card', 'Encryption', 'DropIn', 'Session'
-  s.pod_target_xcconfig = {'SWIFT_SUPPRESS_WARNINGS' => 'YES' }
+  s.default_subspecs = 'Core', 'Components', 'Actions', 'Card', 'Encryption', 'DropIn', 'Session', 'AdyenCheckout'
```
atmamont, line 17: `AdyenCheckout` has all of them as dependencies, can we remove everything but `AdyenCheckout`?

Package.swift
```diff
@@ -180,6 +184,19 @@ let package = Package(
             path: "AdyenTwint",
             exclude: ["Info.plist"]
         ),
+        .target(
+            name: "AdyenCheckout",
+            dependencies: [
+                .target(name: "Adyen"),
```
atmamont, line 190: Would we remove these later from the top level libraries?

atmamont (approved):

Adyen.podspec
erenbesel: It may not work with cocoapods as DropIn was the same and it also expliclitly states all the dependencies.

Package.swift
erenbesel: Yes most modules won't be exposed as separate to outside, only usable between our modules

PR #2104, opened 2025-04-25
Support co-branded cards new UI

neelSharma12:
# Summary

# Release notes

<new>
-  For cards: when the shopper enters the details for a co-badged card in the payment form, they can now select the card brand they want to pay with.
</new>

# Ticket

<ticket>
COSDK-148
</ticket>


AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
```diff
@@ -161,6 +161,41 @@ extension CardViewController {
...
+                let selectableItem = SelectableFormItem(
+                    title: brand.type.name,
+                    imageUrl: numberContainerItem.numberItem.cardTypeLogos.filter { $0.type == brand.type }.first?.url,
+                    isSelected: brand.type.rawValue == brands.first?.type.rawValue ? true : false,
+                    style: .init(title: configuration.style.textField.title),
+                    identifier: brand.type.rawValue
+                )
+                selectableItem.selectionHandler = { [weak self, weak selectableItem] in
+                    guard let self else { return }
+                    handleSelection(identifier: selectableItem?.identifier, brand: brand)
```
goergisn: Does it make sense to call the function without having a selectableItem.identifier?

AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
```diff
@@ -241,7 +276,23 @@ extension CardViewController {
             )
             return item
         }()
-        
+
+        internal func handleSelection(identifier: String?, brand: CardBrand? = nil) {
```
goergisn: Should you be able to "un-select" the brand? (Asking because of optional identifier)

AdyenCard/Form/FormCoBadgedCardItemStyle.swift
```diff
@@ -0,0 +1,54 @@
...
+        self.subtitle = subtitle
+    }
+}
+
+extension FormCoBadgedCardItemStyle: Equatable {
+
+    internal static func == (lhs: FormCoBadgedCardItemStyle, rhs: FormCoBadgedCardItemStyle) -> Bool {
+        lhs.title == rhs.title &&
+            lhs.subtitle == rhs.subtitle &&
+            lhs.backgroundColor.cgColor == rhs.backgroundColor.cgColor
```
goergisn: I believe if you made `UIColor: Equatable` you would not have to implement this custom conformance as `TextStyle` is already `Equatable` right?

Adyen/Core/Payment Methods/OnlineBankingPaymentMethod.swift
```diff
@@ -19,6 +19,16 @@ public struct Issuer: Codable, CustomStringConvertible, Equatable {
         name
     }
 
+    /// Initializes a new instance of `Issuer`.
+    ///
+    /// - Parameters:
+    ///   - identifier: The identifier.
+    ///   - name: The name.
+    public init(identifier: String, name: String) {
```
goergisn: Does this have to be public? Or can we use `@_spi`

AdyenCard/Components/Card/CardComponent.swift
```diff
@@ -222,6 +222,14 @@ public class CardComponent: PresentableComponent,
             type: data.type
         )
         infoEvent.target = data.target
+        infoEvent.brand = data.brands?.first?.type.rawValue
+        infoEvent.configData = data.type == .selected ? nil : CoBadgedAnalyticsConfiguration(
```
goergisn: that's quite a long statement - maybe you can split it up into parts (maybe an if statement)

Adyen/Core/Payment Methods/OnlineBankingPaymentMethod.swift
neelSharma12: This was not needed any more.

AdyenCard/Form/FormCoBadgedCardItemStyle.swift
neelSharma12: yes, this was not required. I have removed it.

AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
neelSharma12: I have removed the identifier from this method.

AdyenCard/Components/Card/CardComponent.swift
neelSharma12: I have changed it to if else statement. thanks for the suggestion :)

AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
neelSharma12: I have removed the identifier from this method.

Adyen.xcodeproj/project.pbxproj
```diff
@@ -41,6 +41,7 @@
 		004BA67B29D427F300A8C69B /* DropInExample.swift in Sources */ = {isa = PBXBuildFile; fileRef = 004BA67A29D427F300A8C69B /* DropInExample.swift */; };
 		004BA67F29D45A2E00A8C69B /* InitialDataFlowProtocol.swift in Sources */ = {isa = PBXBuildFile; fileRef = 004BA67E29D45A2E00A8C69B /* InitialDataFlowProtocol.swift */; };
 		004BA68129D4639900A8C69B /* PresenterExampleProtocol.swift in Sources */ = {isa = PBXBuildFile; fileRef = 004BA68029D4639900A8C69B /* PresenterExampleProtocol.swift */; };
+		0055CDA72DD215DF0019F1D9 /* CoBadgedAnalyticsConfiguration.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0055CDA62DD215CC0019F1D9 /* CoBadgedAnalyticsConfiguration.swift */; };
```
atmamont: CoBadgedCardAnalyticsConfiguration

Adyen.xcodeproj/project.pbxproj
```diff
@@ -1546,6 +1550,7 @@
 		004BA67A29D427F300A8C69B /* DropInExample.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DropInExample.swift; sourceTree = "<group>"; };
 		004BA67E29D45A2E00A8C69B /* InitialDataFlowProtocol.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = InitialDataFlowProtocol.swift; sourceTree = "<group>"; };
 		004BA68029D4639900A8C69B /* PresenterExampleProtocol.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PresenterExampleProtocol.swift; sourceTree = "<group>"; };
+		0055CDA62DD215CC0019F1D9 /* CoBadgedAnalyticsConfiguration.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = CoBadgedAnalyticsConfiguration.swift; sourceTree = "<group>"; };
```
atmamont: CoBadgedCardAnalyticsConfiguration

Adyen/Analytics/Models/AdyenAnalytics.swift
```diff
@@ -32,7 +32,7 @@ public enum AnalyticsEventTarget: String, Encodable {
     case expiryDate = "expiry_date"
     case securityCode = "security_code"
     case holderName = "holder_name"
-    case dualBrand
```
atmamont, line 35: This enum is public and this is a breaking change. Should we keep the old one with the deprecation annotation instead?

AdyenCard/Components/Card/CardComponent.swift
```diff
@@ -222,6 +222,12 @@ public class CardComponent: PresentableComponent,
             type: data.type
         )
         infoEvent.target = data.target
+        infoEvent.brand = data.brands?.first?.type.rawValue
+        if data.type == .selected {
```
erenbesel: If type is anything but selected, we are adding the config data? Like `unfocus`, `validationError` for example. Am I seeing it wrong? :D
Maybe there is no tests for this check, if not let's also add it.

Adyen/Analytics/Models/AnalyticsEventInfo.swift
```diff
@@ -40,6 +40,7 @@ public struct AnalyticsEventInfo: AnalyticsEvent {
         case validationError = "ValidationError"
         case rendered = "Rendered"
         case input = "Input"
+        case displayed = "Displayed"
```
atmamont, line 43: What's the difference between rendered and displayed?

Adyen/Assets/Generated/LocalizationKey.swift
```diff
@@ -88,6 +88,10 @@ public struct LocalizationKey {
     public static let cardNumberItemUnknownBrand = LocalizationKey(key: "adyen.card.numberItem.unknownBrand")
     /// Scan your card
     public static let cardScanYourCardButton = LocalizationKey(key: "adyen.card.scanYourCardButton")
+    /// Card Brand
+    public static let creditCardDualBrandTitle = LocalizationKey(key: "adyen.creditCard.dualBrand.title")
```
atmamont, line 92: This feature introduces semantic split for same feature:
- dual brand
- co badged card

I suggest we pick one and stick to one. Otherwise it becomes harder to connect all dots and build a mindmap later when making any changes.

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -198,11 +198,26 @@ internal class CardViewController: FormViewController {
...
         var brands: [CardBrand] = []
+
         // no dual branding if response is from regex (fallback)
         if binInfo.isCreatedLocally, let firstBrand = binInfo.brands?.first {
             brands = [firstBrand]
+            items.coBadgedCardItem.isHidden.wrappedValue = true
+            items.coBadgedCardItem.selectableFormItems = []
         } else {
             brands = binInfo.brands ?? []
+            if brands.count == 2, brands.allSatisfy(\.isSupported) {
```
erenbesel: This if else block logic seems like it can be moved into the `coBadgedCardItem` itself. Similar to cardNumberItem handling it dual brand logic instead of here.

AdyenCard/Form/FormCoBadgedCardItemView.swift
```diff
@@ -0,0 +1,113 @@
...
+        stackView.isLayoutMarginsRelativeArrangement = true
+        stackView.setContentHuggingPriority(.required, for: .vertical)
+        stackView.axis = .vertical
+        stackView.alignment = .leading
+        stackView.distribution = .fillProportionally
+
+        return stackView
+    }()
+
+    private func removeBrandsListFromView() {
```
nauaros, line 88: Suggestion to improve readability and keep it more swifty:

```
private func removeBrandsListFromView() {
    brandsListView.forEach { view in
        contentStackView.removeArrangedSubview(view)
        view.removeFromSuperview()
    }
}
```

AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
```diff
@@ -161,6 +162,47 @@ extension CardViewController {
...
+                    .creditCardDualBrandDescription,
+                    configuration.localizationParameters
+                ),
+                selectableFormItems: [],
+                style: .init(
+                    title: configuration.style.sectionHeader,
+                    subtitle: configuration.style.footnoteLabel
+                )
+            )
+            item.style.subtitle.textAlignment = .left
```
erenbesel: Have you tested this with RTL languages? I feel like we shouldn't need to specify this manually

AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
```diff
@@ -161,6 +162,47 @@ extension CardViewController {
...
+                postfix: "coBadgedCardItem"
+            )
+            return item
+        }()
+
+        internal func selectableFormItems(from brands: [CardBrand]) -> [SelectableFormItem] {
+            brands.map { brand in
+                let selectableItem = SelectableFormItem(
+                    title: brand.type.name,
+                    imageUrl: numberContainerItem.numberItem.cardTypeLogos.filter { $0.type == brand.type }.first?.url,
```
nauaros: Moving this outside the initializer to their own constant declaration would improve readability:

```swift
let imageUrl = numberContainerItem.numberItem.cardTypeLogos.filter { $0.type == brand.type }.first?.url
```

```swift
let isSelected = brand.type.rawValue == brands.first?.type.rawValue ? true : false
```

AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
```diff
@@ -241,7 +283,30 @@ extension CardViewController {
             )
             return item
         }()
-        
+
+        internal func handleSelection(selectedBrand: CardBrand) {
+            coBadgedCardItem.selectableFormItems.forEach { $0.isSelected = false }
```
erenbesel: This also seems like logic belonging to coBadgedCardItem. By passing the new value it can handle the selection/unselection logic under its own area.

AdyenCard/Components/Card/CardComponent.swift
atmamont: This logic is obscure. Let's either use additional variables/methods named in a way that makes it easy to understand?

AdyenCard/Form/FormCoBadgedCardItemView.swift
```diff
@@ -0,0 +1,113 @@
...
+        accessibilityIdentifier = item.identifier
+        accessibilityLabel = item.accessibilityLabel
+        isAccessibilityElement = true
+
+        preservesSuperviewLayoutMargins = true
+
+        observe(item.$selectableFormItems) { [weak self] selectableItem in
+            selectableItem.forEach { [weak self] _ in
+                guard let self else { return }
+                removeBrandsListFromView()
```
erenbesel: Just a question, not suggestion. Do we need to remove and re-add every time or can we just update the views and only remove/add when the number is 0 or non-zero etc?

nauaros (dismissed):

AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
```diff
@@ -161,6 +162,47 @@ extension CardViewController {
...
+            )
+            return item
+        }()
+
+        internal func selectableFormItems(from brands: [CardBrand]) -> [SelectableFormItem] {
+            brands.map { brand in
+                let selectableItem = SelectableFormItem(
+                    title: brand.type.name,
+                    imageUrl: numberContainerItem.numberItem.cardTypeLogos.filter { $0.type == brand.type }.first?.url,
+                    isSelected: brand.type.rawValue == brands.first?.type.rawValue ? true : false,
```
atmamont: Just highlighting it here - provider ideally should just create entity and not decide on what is selected. It is business logic leaking to assembly layer. It's not a blocker for merge but it requires test coverage if possible. 

The use-case I read here: "as a user, whenever I enter co-branded card number, I want the first brand on the list to be auto-selected"

AdyenCard/Form/FormCoBadgedCardItemView.swift
neelSharma12: We are removing and adding only when the change in the card number is detected which results in fetching new binLookup.

AdyenCard/Components/Card/CoBadgedAnalyticsConfiguration.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+@_spi(AdyenInternal) import Adyen
+
+internal struct CoBadgedAnalyticsConfiguration: AnalyticsStringDictionaryConvertible {
+
+    private let dualBrands: String?
```
atmamont: Public property name `brands` is different from private `dualBrands` creating some confusion, let's align?

atmamont (changes requested):
Nice job Neelam!
The PR size is a bit on a large side so I ran out of my capacity to review tests.
If I read properly, we get rid of the previous icon interaction and keep it to only new method?

AdyenCard/Components/Card/CardComponent.swift
neelSharma12: I have changed the condition of the check and also, I have already tests for this :)  31b1b1a51dc4739130ce268fcc4085578f617395

Adyen/Analytics/Models/AdyenAnalytics.swift
erenbesel: It is SPI, no issue.

atmamont: Yes, thanks, this explains why diff tool didn't highlight this change.

AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
goergisn: The section header is centered that's why the manual setting - I guess `.natural` would be the one to choose here

AdyenCard/Components/Card/CardViewController.swift
nauaros: Agree with Eren here

neelSharma12:
> Nice job Neelam! The PR size is a bit on a large side so I ran out of my capacity to review tests. If I read properly, we get rid of the previous icon interaction and keep it to only new method?

Thank you! Yes, we have removed the previous item interaction and keeping it to new method only.

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -101,6 +101,10 @@ internal class CardViewController: FormViewController {
         setupView()
         setupViewRelations()
         observeNumberItem()
+        observeCoBadgedCardItem()
+        items.coBadgedCardItem.onCardBrandSelection = { [weak self] cardBrand in
+            self?.handleSelection(selectedBrand: cardBrand)
```
erenbesel:
```suggestion
            self?.handleBrandSelection(_ selectedBrand: cardBrand)
```

AdyenCard/Form/FormCoBadgedCardItem.swift
```diff
@@ -0,0 +1,125 @@
...
+            guard let defaultSelectedBrand = brands.first else {
+                return
+            }
+            selectableFormItems = selectableFormItems(
+                from: brands,
+                cardLogos: cardLogos,
+                defaultSelectedBrand: defaultSelectedBrand
+            )
+            isHidden.wrappedValue = false
+            cardItemsDisplayed = brands
```
erenbesel, line 110: should this be set to nil at the start?

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -198,17 +202,28 @@ internal class CardViewController: FormViewController {
 
     internal func update(binInfo: BinLookupResponse) {
         var brands: [CardBrand] = []
+
         // no dual branding if response is from regex (fallback)
         if binInfo.isCreatedLocally, let firstBrand = binInfo.brands?.first {
             brands = [firstBrand]
+            items.coBadgedCardItem.isHidden.wrappedValue = true
+            items.coBadgedCardItem.selectableFormItems = []
```
erenbesel: No need to expose or interact with `selectableFormItems` any more. We can set it to empty via the same function below with empty array as the value. or add a new function like `resetItems()`

AdyenCard/Components/Card/CoBadgedCardAnalyticsConfiguration.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+
+import Foundation
+@_spi(AdyenInternal) import Adyen
+
+internal struct CoBadgedCardAnalyticsConfiguration: AnalyticsStringDictionaryConvertible {
+
+    private let dualBrands: String?
+
+    internal init(brands: String?) {
```
erenbesel: let's use the same name on both sides. dualBrands

AdyenCard/Form/FormCoBadgedCardItem.swift
```diff
@@ -0,0 +1,125 @@
...
+            )
+            selectableItem.selectionHandler = { [weak self] in
+                guard let self else { return }
+                onCardBrandSelection?(brand)
+            }
+            return selectableItem
+        }
+    }
+
+    internal func updateSelectableFormItems(_ brands: [CardBrand], cardLogos: [FormCardLogosItem.CardTypeLogo]) {
```
erenbesel: No need to mention selectableFormitems name in explicitly, we can name it to `updateItems(...`
(and `resetItems()` if you prefer a different function to set the array to empty)

erenbesel (commented):
Looks great! a few small comments

AdyenCard/Form/FormCoBadgedCardItem.swift
```diff
@@ -0,0 +1,130 @@
...
+            selectableFormItems = selectableFormItems(
+                from: brands,
+                cardLogos: cardLogos,
+                defaultSelectedBrand: defaultSelectedBrand
+            )
+            isHidden.wrappedValue = false
+            cardItemsDisplayed = brands
+        } else {
+            isHidden.wrappedValue = true
+        }
```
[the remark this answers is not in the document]
atmamont: There is no known plans to support more than two brands for this feature.

atmamont (dismissed):

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -278,11 +278,13 @@ extension CardViewController {
 
     private func observeCoBadgedCardItem() {
         observe(items.coBadgedCardItem.$cardItemsDisplayed) { [weak self] brands in
-            self?.items.triggerInfoEvent(
-                of: .displayed,
-                target: .dualBrandButton,
-                brands: brands
-            )
+            if brands?.isEmpty == false {
```
atmamont: Is this a perfect case for `guard` check?

atmamont (dismissed):

atmamont (approved):

erenbesel (approved):

PR #2105, opened 2025-04-29
COIOS-897: Assert NSCameraUsageDescription

nauaros:
# Summary

`NSCameraUsageDescription` validation in card scanner creation.

- Debug: It will cause a crash in runtime
- Release: It won't have any effect on release mode

# Ticket

<ticket>
COIOS-897
</ticket>


AdyenCardScanner/Sources/CardScanner.swift
```diff
@@ -42,4 +43,24 @@ public enum CardScanner: CardScanning {
...
+        guard let cameraUsageDescription = Bundle.main.object(
+            forInfoDictionaryKey: "NSCameraUsageDescription"
+        ) as? String, !cameraUsageDescription.isEmpty else {
+            #if DEBUG
+                fatalError(
+                    "Error: NSCameraUsageDescription is missing from Info.plist. Please add a description for camera usage as required by the AdyenCardScanner SDK."
+                )
+            #else
+                debugPrint(
+                    "⚠️ Warning: NSCameraUsageDescription is missing from Info.plist. Camera functionality in AdyenCardScanner SDK might be limited or fail."
```
atmamont: Should this be more explicit and mention release build crash?

atmamont (dismissed):

AdyenCardScanner/Sources/CardScanner.swift
nauaros: I can add that as a reason

erenbesel: Think assert could be simpler, fails on debug and nothing on release

nauaros: Actually, Eren is right. `assertionFailure` achieves the same behavior. I changed it.

erenbesel (dismissed):
Top! You can update the description as well

AdyenCardScanner/Sources/CardScanner.swift
```diff
@@ -42,4 +43,16 @@ public enum CardScanner: CardScanning {
     public static var isAvailable: Bool {
         cardScannerAssembler.captureDevice != nil
     }
+
+    // MARK: - Private
+
+    // swiftlint:disable line_length
+    private static func assertCameraUsageDescription() {
+        guard let cameraUsageDescription = Bundle.main.object(
```
goergisn: Any particular reason of using a guard here? I believe an if would be easier to read.

if let cameraUsageDescription ..., !cameraUsageDescription.isEmpty { return }
assertionFailure(...)

goergisn: What happens when you don't have the description setup and run it in release mode on the phone? It does not crash right? But does it do anything? Otherwise you might want to return `nil` to not even allow showing the camera screen

AdyenCardScanner/Sources/CardScanner.swift
```diff
@@ -42,4 +43,16 @@ public enum CardScanner: CardScanning {
     public static var isAvailable: Bool {
         cardScannerAssembler.captureDevice != nil
     }
+
+    // MARK: - Private
+
+    // swiftlint:disable line_length
```
goergisn, line 49: Maybe you can refactor it a tiny bit to not having to silence the linter

AdyenCardScanner/Sources/CardScanner.swift
nauaros: It won't crash on release mode. At the same time, we don't want to prevent them from using the card scanning feature. This is just a validation to make sure they add the `NSCameraUsageDescription` to their plist file. If they don't, the App Store will take care of reminding them.

AdyenCardScanner/Sources/CardScanner.swift
nauaros: The error text is long. Cannot do much about it. Any suggestion?

AdyenCardScanner/Sources/CardScanner.swift
goergisn: Ahh I see - should be fine then

goergisn (dismissed):

erenbesel (approved):

atmamont (approved):

PR #2109, opened 2025-05-01
v6 component callbacks

erenbesel:
# Summary

Experimenting with callbacks on components on how to pass them from the main configuration.
- Maybe components will trigger the event which then shold be handled by the generic component?

# Ticket

<ticket>
COIOS-158
</ticket>


Adyen/CheckoutConfiguration/CheckoutBaseCallbacks.swift
```diff
@@ -27,3 +27,13 @@ package protocol CheckoutBaseCallbacks {
     
     var onComplete: CheckoutSuccessHandler? { get set }
 }
+
+package extension CheckoutBaseCallbacks {
+    
+    mutating func setBaseCallbacks(from configuration: CheckoutConfiguration) {
+        self.onSubmit = configuration.onSubmit
```
atmamont: I imagine this grows as we move forward. wdyt of an idea to store all callbacks in a dictionary with a key enumerating all supported callbacks? Then accessing and updating would be one line and there would be no `if let callback = self.callback else` anymore

AdyenCheckout/CheckoutComponentBuilder.swift
```diff
@@ -57,6 +57,19 @@ extension BLIKPaymentMethod {
...
+            payment: nil,
+            amount: configuration.amount,
+            analyticsConfiguration: configuration.analyticsConfiguration
+        )
+        var component = BLIKComponent(
+            paymentMethod: self,
+            context: context,
+            configuration: blikConfiguration
+        )
+        component.setBaseCallbacks(from: configuration)
```
atmamont: It is not clear why do we this after we instantiated the component from the configuration, from what I see it could be an implementation details without exposing this here.

AdyenComponents/BLIK/BLIKComponent.swift
```diff
@@ -9,7 +9,15 @@ import Foundation
 import UIKit
 
 /// A component that provides a form for BLIK payments.
-public final class BLIKComponent: PaymentComponent, PresentableComponent, PaymentAware, LoadingComponent {
+public final class BLIKComponent: PaymentComponent, PresentableComponent, PaymentAware, LoadingComponent, CheckoutBaseCallbacks {
+    
+    package var onSubmit: Adyen.SubmitHandler?
```
atmamont, line 14: I would prefer to not see this on top, can we extract this to an extension? I would also play around with the default implementation in the protocol itself if possible to just remove this code from all components and override it if the behavior is different for a specific component.

atmamont (commented):
I left comments as I went through the code, some of them are open questions to just not forget to discuss rather than nitpicks.
Great stuff, Eren!

PR #2118, opened 2025-05-13
v6: Super component foundation

erenbesel:
# Summary [Required]
- Base foundation for the new super component which. will handle all the component logic internally.

## Ticket [Optional]
<ticket>
COSDK-158
</ticket>


Adyen/CheckoutConfiguration/CheckoutConfiguration.swift
```diff
@@ -47,18 +47,13 @@ public struct CheckoutConfiguration {
         self.amount = amount
         self.analyticsConfiguration = analyticsConfiguration
         
-        var configDictionary: [CheckoutComponentType: ConfigurationWrapper] = [:]
+        var configDictionary: [CheckoutComponentType: CheckoutComponentConfiguration] = [:]
```
nauaros, line 50: It would help to have a more descriptive name. Such as `componentConfigurations`. Just an idea

AdyenCheckout/AdyenCheckout.swift
```diff
@@ -106,12 +106,11 @@ public final class AdyenCheckout: AdyenCheckoutProtocol {
     }
     
     public func createComponent(with paymentMethod: any PaymentMethod) -> AdyenCheckoutComponent? {
-        try? CheckoutComponentBuilder.build(for: paymentMethod, configuration: configuration)
-//        AdyenCheckoutComponent(paymentMethod: paymentMethod, configuration: configuration)
+        AdyenCheckoutComponent(paymentMethod: paymentMethod, configuration: configuration, session: session)
```
nauaros, line 109: Some formatting here to make it multi-line

Adyen/Core/AdyenContext/AdyenContext.swift
```diff
@@ -18,8 +20,9 @@ public final class AdyenContext: PaymentAware {
     /// The payment information.
     public private(set) var payment: Payment?
     
-    @_spi(AdyenInternal)
-    public let analyticsProvider: AnyAnalyticsProvider?
+    package let analyticsProvider: AnyAnalyticsProvider?
+    
+    package let amount: Amount
```
descorp, line 25: `amount` can be public, right? We will get rid of `Payment` eventually any way

AdyenCheckout/AdyenCheckoutComponent.swift
```diff
@@ -0,0 +1,50 @@
...
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+@_spi(AdyenInternal) import Adyen
+#if canImport(AdyenSession)
+    @_spi(AdyenInternal) import AdyenSession
+#endif
+#if canImport(AdyenDropIn)
+    @_spi(AdyenInternal) import AdyenDropIn
```
descorp, line 12: Side thought: Now `AdyenCheckout` is an "umbrella" module that holds references to all other modules. 

This means DropIn no longer need to have `ComponentManager` and it can become a fancy UI tool with thin business login 🤔

AdyenCheckout/AdyenCheckoutComponent.swift
```diff
@@ -0,0 +1,50 @@
...
+#endif
+
+// TODO: add description
+public final class AdyenCheckoutComponent {
+    
+    private var paymentComponent: PaymentComponent?
+    
+    private var actionComponent: ActionComponent?
+    
+    private var actionHandlingComponent: ActionHandlingComponent?
```
descorp, line 28: What is difference between `actionComponent` and `actionHandlingComponent` in this context ?

AdyenCheckout/AdyenCheckoutComponent.swift
```diff
@@ -0,0 +1,50 @@
...
+    
+    private var configuration: CheckoutConfiguration
+    
+    package init(
+        paymentMethod: PaymentMethod,
+        configuration: CheckoutConfiguration,
+        session: AdyenSession? = nil
```
descorp, line 35: Perhaps we can create a more generic "AnySession" in `AdyenCore` (that would represent an advanced mode); this way, the `AdyenCheckout` module would not need to directly depend on the `AdyenSession` module.

Especially because both `session` and `configuration` contains `Context` with `amount`, but only one on `session` actually matter.

AdyenCheckout/CheckoutComponentBuilder.swift
```diff
@@ -15,14 +15,12 @@
...
 
-public protocol AdyenCheckoutComponent: PresentableComponent {}
-
 internal enum CheckoutComponentBuilder {
     
     internal static func build(
         for paymentMethod: PaymentMethod,
         configuration: CheckoutConfiguration
-    ) throws -> AdyenCheckoutComponent {
+    ) -> PaymentComponent {
```
descorp, line 23: We need to keep `throws` for sake of ApplePay and other SDK-based payment components that might not be generated directly.

AdyenCheckout/CheckoutComponentBuilder.swift
```diff
@@ -34,29 +32,43 @@ internal enum CheckoutComponentBuilder {
...
 extension CheckoutConfiguration {
-    func configuration(for paymentMethod: PaymentMethod) -> CheckoutComponentConfiguration? {
-        configurations[.payment(paymentMethod.type)]?.configuration
+    func componentConfiguration(for paymentMethod: PaymentMethod) -> CheckoutComponentConfiguration? {
+        configurations[.payment(paymentMethod.type)]
     }
+    
+//    func componentConfiguration(for action: Action) -> CheckoutComponentConfiguration? {
+//        configurations[.action(...)]
+//    }
```
descorp, line 47: Commented code

Adyen/Core/AdyenContext/AdyenContext.swift
erenbesel: Let's start with restricted access and can switch to public if needed

Adyen/CheckoutConfiguration/CheckoutConfiguration.swift
erenbesel: It's a local var assigned to actual `configurations` property at the end. IMO it's fine

AdyenCheckout/AdyenCheckoutComponent.swift
erenbesel: Yes indeed! Beauty of this is we can change the internal stuff later

AdyenCheckout/AdyenCheckoutComponent.swift
erenbesel: ActionComponent is the component created via action to support action component creation via super component (which is not final but we want to)
Action handling component is our AdyenActionComponent which is for handling action object automatically.

AdyenCheckout/AdyenCheckoutComponent.swift
erenbesel: We will change the internal code of session for sure. It won't have some of the stuff now that session itself won't be accessible

AdyenCheckout/CheckoutComponentBuilder.swift
erenbesel: It is removed for now for simplicity. Whenever needed, we will add it back.

AdyenCheckout/CheckoutComponentBuilder.swift
erenbesel: Yes :)

atmamont (approved):

AdyenCheckout/AdyenCheckoutComponent.swift
descorp: So  `actionHandlingComponent` is absolute, since `AdyenCheckoutComponent` will do what `AdyenActionComponent` does right now, right ?

descorp (approved):

AdyenCheckout/AdyenCheckoutComponent.swift
erenbesel: Well yes and no. For super component to be able to do it, we still need the functionality. Why add all the logic in this class when there is already a class that does it and add pass that responsibility to that component?

PR #2119, opened 2025-05-14
v6: super component callbacks

erenbesel:
# Summary [Required]
The exposed new way of callbacks structure is through closures such as onSubmit(), onAdditionalDetails(). Instead of changing this on component level, we came up with the idea to have the new AdyenCheckoutComponent be the bridge between the old apporach and the new by acting as the delegate of components while using the closures for outside access.

## Ticket [Optional]
<ticket>
COSDK-467
</ticket>


Adyen/CheckoutConfiguration/CheckoutBaseCallbacks.swift
```diff
@@ -27,3 +27,49 @@ package protocol CheckoutBaseCallbacks {
...
+    
+    private enum CodingKeys: CodingKey {
+        case resultCode
+        case action
+        case order
+    }
+}
+
+/// Data model that contains information regarding the status of a payment.
+public struct CheckoutResult {
```
neelSharma12: Can we make a `Result` struct and keep the resultCode and error in that? Also, I feel in future it will be easier for us to add additional fields like messages if needed. Do you think it would be beneficial for us to  convert this `CheckoutResult` into enum before making it a part of new `Result` struct?

Adyen/CheckoutConfiguration/CheckoutBaseCallbacks.swift
```diff
@@ -27,3 +27,49 @@ package protocol CheckoutBaseCallbacks {
...
+    
+    // TODO: string or enum?
+    public let resultCode: String
```
atmamont: It's a string so let's make it enum, this prevents mistypes when comparing and makes it easier to update in the future.

Adyen/CheckoutConfiguration/CheckoutBaseCallbacks.swift
```diff
@@ -27,3 +27,49 @@ package protocol CheckoutBaseCallbacks {
...
+    
+    package init(resultCode: String) {
+        self.resultCode = resultCode
+    }
+}
+
+/// Wrapper type to contain checkout related errors.
+// TODO: if not needed, can remove?
+public struct CheckoutError: Error {
```
atmamont: Does it make sense to merge both into `Result<CheckoutResult, CheckoutError>`?

Adyen/CheckoutConfiguration/CheckoutBaseCallbacks.swift
```diff
@@ -27,3 +27,49 @@ package protocol CheckoutBaseCallbacks {
...
+    
+    private let error: Error
```
atmamont: I didn't find a finite error list in api explorer, do we have it?

Adyen/CheckoutConfiguration/CheckoutConfiguration.swift
```diff
@@ -6,6 +6,7 @@
 
 import Foundation
 
+// TODO: add description
```
atmamont: A container structure to pass specific payment methods configurations that override default behavior?

AdyenCheckout/AdyenCheckoutComponent.swift
```diff
@@ -48,3 +40,33 @@ public final class AdyenCheckoutComponent {
         self.actionComponent = CheckoutComponentBuilder.build(for: action, configuration: configuration)
     }
 }
+
+extension AdyenCheckoutComponent: PaymentComponentDelegate {
+    
+    public func didSubmit(_ data: PaymentComponentData, from component: any PaymentComponent) {
+        configuration.onSubmit?(data) { [weak self] response in
```
atmamont, line 47: Why is onSubmit optional? This is a part of base callbacks array and it should be always present, right?

atmamont (approved):

Adyen/CheckoutConfiguration/CheckoutBaseCallbacks.swift
erenbesel: result code exists regardless of error. it's not a case of one or the other so we can't create a `Result` with these two. 
also cc: @neelSharma12 relevant to your comment

Adyen/ResponseModels/CheckoutPaymentsResponse.swift
```diff
@@ -0,0 +1,53 @@
...
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+/// Data model to contain relevant parts of `/payments` and `/payment/details` call responses.
+public struct CheckoutPaymentsResponse: Decodable, Sendable {
+    
+    // TODO: should it be string or enum?
+    public let resultCode: String
```
nauaros: I would work on an `enum` basis. Find it easier and less error prone

Adyen/ResponseModels/CheckoutPaymentsResponse.swift
```diff
@@ -0,0 +1,53 @@
...
+    
+    package init(resultCode: String) {
+        self.resultCode = resultCode
+    }
+}
+
+/// Wrapper type to contain checkout related errors.
+// TODO: if not needed, can remove?
+public struct CheckoutError: Error {
```
nauaros, line 46: I think normally BE returns an error description right? An APIErrorResponse with a title if I am not wrong

Adyen/UI/Form/FormViewController+ViewProtocol.swift
```diff
@@ -9,7 +9,7 @@ import Foundation
 @_spi(AdyenInternal)
 public protocol FormViewProtocol {
     
-    func add<T: FormItem>(item: T?)
+    func add(item: (some FormItem)?)
```
nauaros, line 12: Sweeeet

nauaros (approved):

Adyen/ResponseModels/CheckoutPaymentsResponse.swift
goergisn: As a dev I prefer enums - but keeping in mind public breaking changes - maybe having an extra case `.other(String)`

AdyenDropIn/Utilities/ComponentManager+PaymentComponentBuilder.swift
```diff
@@ -438,16 +438,15 @@ private extension ComponentManager {
...
-//        return BLIKComponent(
-//            paymentMethod: paymentMethod,
-//            context: context,
-//            configuration: config
-//        )
+        let config = BLIKComponentConfiguration(
+            style: configuration.style.formComponent,
+            localizationParameters: configuration.localizationParameters
+        )
+        return BLIKComponent(
```
goergisn, line 445: Is it a failable init? Otherwise you might want to make the return value of the function non-optional

PR #2120, opened 2025-05-15
CardScanner: Focus on CVC field after successful scan

nauaros:
## Summary [Required]

Adds logic in the card component to move first responder to the next empty field after a successful scan.
On the `FormViewController`, there's a new method `focusNextInputField` that moves the first responder to the next `FormTextInputItem` with a `nil` value.

## Demo

https://github.com/user-attachments/assets/9398fce7-7204-4028-b223-d60e20ef0dbb

## Ticket [Optional]
<ticket>
COSDK-468
</ticket>

## Checklist [Required]

- [x] Tested changes locally

### DIFF
--- a/Adyen/Core/APIClient/APIContext.swift
+++ b/Adyen/Core/APIClient/APIContext.swift
@@ -26,6 +26,7 @@ public struct APIContext: AnyAPIContext {
     public let headers: [String: String] = ["Content-Type": "application/json"]
 
     /// Environment to retrieve internal resources from.
+    /// - Note: Always use the provided `Environment` object to ensure a correct environment.
     public let environment: AnyAPIEnvironment
     
     /// The client key that corresponds to the web service user you will use for initiating the payment.
@@ -37,6 +38,7 @@ public struct APIContext: AnyAPIContext {
     ///   - environment: The environment to retrieve internal resources from.
     ///   - clientKey: The client key that corresponds to the web service user you will use for initiating the payment.
     /// - Throws: `ClientKeyError.invalidClientKey` if the client key is invalid.
+    /// - Note: Always use the provided `Environment` object to ensure a correct environment.
     public init(environment: AnyAPIEnvironment, clientKey: String) throws {
         guard ClientKeyValidator().isValid(clientKey) else {
             throw ClientKeyError.invalidClientKey
--- a/Adyen/Core/APIClient/APIContext.swift
+++ b/Adyen/Core/APIClient/APIContext.swift
@@ -26,7 +26,7 @@ public struct APIContext: AnyAPIContext {
     public let headers: [String: String] = ["Content-Type": "application/json"]
 
     /// Environment to retrieve internal resources from.
-    /// - Note: Always use the provided `Environment` object to ensure a correct environment.
+    /// - Note: Always use the provided `Environment` type to ensure a correct environment.
     public let environment: AnyAPIEnvironment
     
     /// The client key that corresponds to the web service user you will use for initiating the payment.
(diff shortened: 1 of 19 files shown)


AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -423,15 +423,33 @@ extension CardViewController {
     private func handleCardScanningResult(_ result: Result<CardScannerCardDetails, Error>) {
         switch result {
         case let .success((number, expiryDate)):
-            items.numberContainerItem.setCardNumber(number ?? "")
+            if let number {
```
atmamont, line 426: This is behavior change, previous code **always** updates the text field, new code does it only when the number is  not nil

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -423,15 +423,33 @@ extension CardViewController {
...
+            focusNextInputField()
         case .failure:
-            // TODO: Implement error handling
-            break
+            cardScannerController.dismiss(nil)
         }
     }
+
+    private func focusNextInputField() {
+        let textInputItems: [FormTextInputItem] = [
```
atmamont: This is fragile as we redefine the order of fields the second time. Can we use our existing `CardViewController` form items and filter out them? Then if we have any redesign in the future we wouldn't need to remember updating this block.

atmamont (dismissed):

Adyen/Core/APIClient/APIContext.swift
```diff
@@ -37,6 +38,7 @@ public struct APIContext: AnyAPIContext {
     ///   - environment: The environment to retrieve internal resources from.
     ///   - clientKey: The client key that corresponds to the web service user you will use for initiating the payment.
     /// - Throws: `ClientKeyError.invalidClientKey` if the client key is invalid.
+    /// - Note: Always use the provided `Environment` object to ensure a correct environment.
```
erenbesel:
```suggestion
    /// - Note: Always use the provided `Environment` type to ensure a correct environment.
```

Adyen/Core/APIClient/APIContext.swift
```diff
@@ -26,6 +26,7 @@ public struct APIContext: AnyAPIContext {
     public let headers: [String: String] = ["Content-Type": "application/json"]
 
     /// Environment to retrieve internal resources from.
+    /// - Note: Always use the provided `Environment` object to ensure a correct environment.
```
erenbesel:
```suggestion
    /// - Note: Always use the provided `Environment` type to ensure a correct environment.
```

AdyenCard/Form/FormCardExpiryDateItem.swift
```diff
@@ -74,7 +74,7 @@ internal final class FormCardExpiryDateItem: FormTextItem {
 }
 
 extension FormItemViewBuilder {
-    internal func build(with item: FormCardExpiryDateItem) -> FormItemView<FormCardExpiryDateItem> {
-        FormTextItemView<FormCardExpiryDateItem>(item: item)
+    internal func build(with item: FormCardExpiryDateItem) -> FormItemView<FormTextInputItem> {
```
erenbesel, line 77:
```suggestion
    internal func build(with item: FormCardExpiryDateItem) -> FormItemView< FormCardExpiryDateItem > {
```

atmamont: Do we have such codestyle? Doesn't look aligned with Apple suggested style

nauaros: As discussed in person, it has to do with the way Swift handles type inheritance with generic parameters. In this case, even if `FormCardExpiryDateItem` is a child of `FormTextInput`, Swift forces you to keep the parent class.|

erenbesel (approved):

atmamont (approved):

PR #2141, opened 2025-05-26
Fix brand name for Cobadged cards

neelSharma12:
# Summary

- Update the card brand name value with ```localeBrand``` property and default `brand.type.rawValue`.

# Release notes

<fixed> 

For cards: if the shopper's card is [co-badged](https://docs-admin.is.adyen.com/online-payments/co-badged-cards-compliance#guidelines-for-stored-co-branded-cards), the UI where the shopper selects the brand now shows the `localeBrand` value instead of `brand.type.name`.

For example, the UI now shows **Bancontact card** instead of **bcmc**.

The default for `localeBrand` is `brand.type.rawValue`. 

</fixed>

# Ticket

<ticket>
COIOS-000
</ticket>


AdyenCard/Form/FormCoBadgedCardItem.swift
```diff
@@ -80,7 +80,7 @@ internal final class FormCoBadgedCardItem: FormItem {
             let isSelected = brand.type.rawValue == defaultSelectedBrand.type.rawValue ? true : false
 
             let selectableItem = SelectableFormItem(
-                title: brand.type.name,
+                title: brand.localeBrand ?? brand.type.rawValue,
```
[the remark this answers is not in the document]
neelSharma12: This is not as per the technical requirement.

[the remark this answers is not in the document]
goergisn: If that's the case I'd suggest adding a comment why you use the rawValue here - In a year you might revisit this code and wonder why .name was not used instead of .rawValue

Unfortunately there is no link from the code to this PR

goergisn (dismissed):

nauaros (dismissed):

Tests/IntegrationTests/Card Tests/CardComponentTests.swift
```diff
@@ -1422,6 +1422,27 @@ class CardComponentTests: XCTestCase {
...
+    func testCoBadgedCardsNameOnSelectionUI() throws {
+        // Given
+
+        let sut = CardComponent(
+            paymentMethod: method,
+            context: context,
+            configuration: CardComponent.Configuration()
+        )
+
+        setupRootViewController(sut.viewController)
```
erenbesel: With new tests we try to avoid `setupRootViewController` and instead use loadViewIfNeeded()

Tests/IntegrationTests/Card Tests/CardComponentTests.swift
```diff
@@ -1422,6 +1422,27 @@ class CardComponentTests: XCTestCase {
...
+
+        let newResponse = BinLookupResponse(brands: [CardBrand(type: .bcmc), CardBrand(type: .maestro)], issuingCountryCode: "BE", isCreatedLocally: false)
+        sut.cardViewController.update(binInfo: newResponse)
+
+        wait(for: .aMoment)
+
+        // Then
+        XCTAssertFalse(sut.cardViewController.items.coBadgedCardItem.isHidden.wrappedValue)
+        XCTAssertEqual(sut.cardViewController.items.coBadgedCardItem.selectableFormItems[0].title, newResponse.brands?[0].localeBrand ?? newResponse.brands?[0].type.rawValue)
```
erenbesel: This line does not create a reliable test result. 
- If the first part of `??` is not nil, it returns true
- If it is nil, it goes to the second part, it returns true.

It succeeds on both cases, there is no real outcome.

We should test 2 different cases where the new property is nil and non nil, separately with respective results.

erenbesel (commented):
Comments on testing

Tests/IntegrationTests/Card Tests/CardComponentTests.swift
atmamont: +1

AdyenCard/Form/FormCoBadgedCardItem.swift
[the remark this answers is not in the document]
neelSharma12: Added a comment.

Tests/IntegrationTests/Card Tests/CardComponentTests.swift
neelSharma12: Thanks for pointing this.

Tests/IntegrationTests/Card Tests/CardComponentTests.swift
neelSharma12: Indeed :) Updated the test.

neelSharma12:
> Comments on testing

thanks :) Addressed them.

nauaros (approved):

atmamont (approved):

PR #2143, opened 2025-05-27
COSDK-464: Introduce MVVM architecture in DropInComponent

nauaros:
# Summary

This change introduces MVVM architecture to the `DropInComponent`. Now, each one of the screens in the DropInComponent is its own MVVM module. The idea is to move away from the `Component` pattern that we currently have, and keep this one only for strict payment components.

The DropInComponent is full screen now, there's no usage of `WrapperViewController` or `ModalViewController` to achieve partial screens. Furthermore, the custom presentation is gone too.

# Demo

https://github.com/user-attachments/assets/ca0d2591-427a-4c4b-93db-2dec6842d806

## Ticket
<ticket>
COSDK-464
</ticket>

## Checklist [Required]

- [ ] Tested changes locally


Adyen/UI/Form/FormViewController.swift
```diff
@@ -9,7 +9,7 @@ import UIKit
 /// Displays a form for the user to enter details.
 @objc(ADYFormViewController)
 @_spi(AdyenInternal)
-open class FormViewController: UIViewController, AdyenObserver, PreferredContentSizeConsumer {
+open class FormViewController: UIViewController, AdyenObserver {
```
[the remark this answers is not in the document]
nauaros: I am aware of it. This will be addressed in later PRs.

AdyenDropIn/DropInComponent.swift
```diff
@@ -31,6 +31,17 @@ public final class DropInComponent: NSObject,
     ActionHandlingComponent,
     LoadingComponent {
 
+    // MARK: - Properties
+
+    private var preselectedPaymentMethodView: UIViewController?
```
atmamont, line 36: I'd go with `preselectedPaymentMethodViewController` to not create any possible confusion even in the local scope (it's private). Reading the code would be easier later.

AdyenDropIn/DropInComponent.swift
```diff
@@ -31,6 +31,17 @@ public final class DropInComponent: NSObject,
...
+    private var paymentMethodListView: UIViewController?
+    private var componentView: UIViewController?
+
+    private lazy var componentManager: ComponentManager = {
+        let componentManager = createComponentManager(nil)
```
atmamont, line 41: If parameter could be nil, it is better to keep it's name
`createComponentManager(order: nil)`

AdyenDropIn/DropInComponent.swift
```diff
@@ -229,43 +225,95 @@ public final class DropInComponent: NSObject,
         return handler
     }()
 
-    internal func paymentMethodListComponent(onCancel: (() -> Void)?) -> PaymentMethodListComponent {
-        let paymentComponents = componentManager.sections
-        let component = PaymentMethodListComponent(
+    // ================= ROOT VIEW CONTROLLER ===============
+
+    internal lazy var rootViewController: UIViewController = {
```
atmamont, line 230: This is quite hard to understand and it contains a lot of logic. Would be nice to simplify and move decisions out of here in the future PRs.

AdyenDropIn/DropInComponent.swift
```diff
@@ -229,43 +225,95 @@ public final class DropInComponent: NSObject,
...
+    }()
+
+    // ================= COMPONENT MODULE - ASSEMBLER ===============
+
+    internal func resolveComponentView(
+        from component: PresentableComponent
+    ) -> UIViewController {
+        let viewModel = ComponentViewModel(
+            component: component,
+            isRoot: false,
```
atmamont, line 256: A lot to unpack, I doubt if I should write such comment or not. I'd rather highlight this, but I guess it's just not feasible to do everything in one PR.

`isRoot` looks like a workaround, could be extracted to the current assembly layer that will make a decision and could maybe produce RootComponentViewModel instead?

AdyenDropIn/Presentation/DropInNavigationController.swift
```diff
@@ -7,120 +7,40 @@
...
     internal let style: NavigationStyle
     
     internal lazy var keyboardObserver = KeyboardObserver()
-    
-    internal init(rootComponent: PresentableComponent, style: NavigationStyle, cancelHandler: @escaping CancelHandler) {
+
+    // MARK: - Initializers
+
+    internal init(
+        rootViewController: UIViewController,
```
atmamont, line 25: Do we need this `rootViewController` in init?
`DropInNavigationController` is still just a `UINavigationViewController` and one can just reassign `viewControllers` array later.

atmamont (commented):
Left some comments, feel free to disregard completely as I struggle a bit with the big scope.

AdyenDropIn/DropInComponent.swift
nauaros: Yeah, completely agree with you. This is something to be changed, its on my plans to reduce the complexity of that decision tree.

AdyenDropIn/DropInComponent.swift
nauaros: Yes, it is a workaround that I continue from the old implementation just to keep the DropIn functional. It will also go away once the DropIn has its own module. Thanks for highlighting it. With this changes, we can see the number of workarounds that that to be done due to the poor architecture

AdyenDropIn/Presentation/DropInNavigationController.swift
```diff
@@ -7,120 +7,40 @@
...
+        style: NavigationStyle,
+        cancelHandler: @escaping CancelHandler
+    ) {
         self.style = style
         self.cancelHandler = cancelHandler
```
descorp, line 30:
```suggestion
```

No need for `style` and `cancelHandler` - those two no longer in use.

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewModel.swift
```diff
@@ -0,0 +1,106 @@
...
+            style: style,
+            listItemStyle: listItemStyle
+        )
+        self.preselectedPaymentMethodComponent.localizationParameters = localizationParameters
+        self.preselectedPaymentMethodComponent.delegate = self
+    }
+
+    // MARK: - PreselectedPaymentMethodViewModelProtocol
+
+    internal var paymentMethodView: UIViewController {
```
descorp, line 50:
```suggestion
    internal var paymentMethodViewController: UIViewController {
```

Full name would be more readable.

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewModel.swift
```diff
@@ -0,0 +1,106 @@
...
+        router?.showAllPaymentMethods()
+    }
+
+    internal func didProceed(with component: any Adyen.PaymentComponent) {
+        print("Proceed with component: \(component)")
+        router?.proceed(with: component)
+    }
+}
+
+internal class PreselectedPaymentMethodViewController: UIViewController {
```
descorp, line 66: The only purpose of `PreselectedPaymentMethodViewController` and `PaymentMethodListViewController` is to host dedicated ViewController 

Maybe we can use a helper instead?

```swift

protocol ViewModelProtocol {
    var paymentMethodView: UIViewController { get }
}

extension ViewModelProtocol {
   func setupPaymentMethodView() -> UIViewController {
        let paymentMethodView = viewModel.paymentMethodView
        var newViewController = UIViewController()
        paymentMethodView.willMove(toParent: newViewController)
        newViewController.addChild(paymentMethodView)
        newViewController.addSubview(paymentMethodView.view)
        paymentMethodView.didMove(toParent: newViewController)
        paymentMethodView.view.adyen.anchor(inside: newViewController.view)
        return newViewController
    }
}
```

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewModel.swift
```diff
@@ -0,0 +1,123 @@
...
+
+internal protocol PaymentMethodListViewModelProtocol {
+    var paymentMethodListView: UIViewController { get }
+}
+
+internal class PaymentMethodListViewModel: PaymentMethodListViewModelProtocol, PaymentMethodListComponentDelegate {
+
+    // MARK: - Properties
+
+    private weak var router: PaymentMethodListRouterProtocol?
```
descorp, line 28: This also feels like there could be a generic base class.

AdyenDropIn/Presentation/DropInNavigationController.swift
nauaros: Thanks Vlad! Yes, the DropInNavigationController will go away in a future PR

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewModel.swift
nauaros: The idea is to have an MVVM module for that screen in the Drop-in. Since there are gonna be changes on the preselected payment view, once we are ready to develop them, we can get rid all together of the PreselectedPaymentMethodComponent and develop a new UI. I am just preparing the grounds for it to happen.

Same thing for the payment method list.

It's just we cannot develop everything at the same. The idea is to swipe it out when the moment comes

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewModel.swift
nauaros: What do you mean by more generic base class?

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewModel.swift
nauaros: In next PR I am creating this is gone, so you won't see this property any more. Just something temporary

AdyenDropIn/Presentation/DropInNavigationController.swift
nauaros: Nope, actually this whole class will dissappear. So I won't change it for now, no need. It will be gone

AdyenDropIn/DropInComponent.swift
nauaros: Agree! I will change this in the next PR I am creating

AdyenDropIn/DropInComponent.swift
nauaros: This is just temporary, it will be gone in the next PR

erenbesel (approved):

atmamont (approved):

PR #2149, opened 2025-05-30
COSDK-483 Conform Card component to European Accessibility Act

atmamont:
# Summary

This change is an improvement to card component accessibility aiming to provide screen reader with more context on two inputs: card expiry date and security code.

# Demo (updated)

**Expiry date accessibility label**
image

**Security code accessibility label**
image

### Use appropriate tags for different types of changes:
<fixed>
- For cards, enhanced the accessibility labels so that the screen reader provides more context on the expiry date and security code fields. 
</fixed>

## Testing Instructions

1. Launch demo app in Simulator
2. Launch Xcode accessibility inspector
3. Walk through card component UI elements
4. Check if screen reader provides expected format for expiry date input (should have date format)
5. Check if screen reader provides number of digits for security code input (should have number of digits)

## Ticket
<ticket>
COSDK-483
</ticket>

## Checklist [Required]

- [x] Tested changes locally
- [x] Update localizations for all languages


AdyenCard/Form/FormCardExpiryDateItemView.swift
```diff
@@ -0,0 +1,30 @@
...
+/// Copyright (c) 2021 Adyen N.V.
+///
+/// This file is open source and available under the MIT license. See the LICENSE file for more info.
+///
+
+@_spi(AdyenInternal) import Adyen
+import UIKit
+
+/// A view representing a form card expiry date item.
+internal final class FormCardExpiryDateItemView: FormTextItemView<FormCardExpiryDateItem> {
```
goergisn: What makes the `FormCardExpiryDateItemView` special to not use `FormTextInputItemView` instead?

Adyen/Utilities/Observable/AdyenObservable.swift
```diff
@@ -18,7 +18,19 @@ public final class AdyenObservable<ValueType: Equatable>: EventPublisher {
     }
     
     // MARK: - Value
-    
+
+    /// The value being observed.
```
goergisn, line 24: IMO this comment could move to the class documentation to be more visible

AdyenCard/Form/FormCardExpiryDateItemView.swift
atmamont: Business logic that constructs a11y label and belongs to this specific view.
Do you think it could be extracted to a builder method of the item itself?

Adyen/Utilities/Observable/AdyenObservable.swift
atmamont: Done, I've simplified it and moved to the class scope

AdyenCard/Form/FormCardExpiryDateItemView.swift
```diff
@@ -0,0 +1,30 @@
...
+        }
+        observe(item.$title) { [weak self] _ in
+            self?.setupAccessibility()
+        }
+    }
+
+    private func setupAccessibility() {
+        let title = item.title ?? ""
+        let placeholder = item.placeholder ?? "MM/YY"
+        textField.accessibilityLabel = title.isEmpty ? placeholder : "\(title), \(placeholder)"
```
erenbesel: This logic can be calculated on the item (viewModel) and passed here as `item.accesibilityLabel`, wdyt?

atmamont: Oh now I see this is a pandora box really. 
We have few `accessibilityLabel` parameters in:
- `SelectableFormItem`
- `ListItem`

And we have `identifier` set in lots of Items, defined in the "FormItem" base class.

I see your point, but at the same time adding another `accessibilityLabel` to `FormTextInputItem` doesn't feel alright while this should belong to a base Form class, what do you think?
I'm open to changes but I'd rather make a change that lasts and doesn't require more refactors in the future. 
Should we add `accessibilityLabel` to the `FormItem` the same way we have `identifier`?

erenbesel (approved):

AdyenCard/Form/FormCardExpiryDateItemView.swift
atmamont: We agreed to try using FormTextItemView.accessibilityLabelView to provide context from outside (no new view needed)

nauaros (approved):

Tests/IntegrationTests/Card Tests/Views/FormCardExpiryDateItemViewTests.swift
```diff
@@ -0,0 +1,73 @@
...
+}
+
+extension FormCardExpiryDateItemViewTests {
+    private func makeSUT(item: FormCardExpiryDateItem) -> FormTextInputItemView {
+        let view = FormTextInputItemView(item: item)
+        view.observe(item.$title) { _ in
+            view.accessibilityLabelView?.accessibilityLabel = item.accessibilityValue
+        }
+        view.accessibilityLabelView?.accessibilityLabel = item.accessibilityValue
+        return view
```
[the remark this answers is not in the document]
atmamont: Tried that.

This approach is first that comes to mind but unfortunately it doesn't work unless there is a dedicated view type produced. Here we don't have a dedicated view type as we produce generic `FormTextInputItemView `, therefore base `build()` function is called from `FormItemViewBuilder.swift` line 39 and no observation logic is applied.

Good catch though.

neelSharma12 (approved):

PR #2152, opened 2025-06-04
Card component - Add a11y label localization for expiry date format

atmamont:
# Summary

This is a stacked diff to the main feature branch.
This change contains:
- removes newly added view (it breaks tests because of `ViewIdentifierBuilder` relies on view hierarchy and ids are hardcoded currently in tests
- new localization addition for expiry date format (aligning with Android)
- `accessibilityLabelView` property moved up the view hierarchy 

Real strings will be added to the main PR when Smartling is ready later on the main PR.


AdyenCard/Form/FormCardExpiryDateItem.swift
```diff
@@ -74,7 +74,23 @@ internal final class FormCardExpiryDateItem: FormTextInputItem {
...
+        }
+
+        let view = FormTextInputItemView(item: item)
+        view.observe(item.$placeholder) { _ in
+            updateAccessibilityClosure(view.accessibilityLabelView)
+        }
+        view.observe(item.$title) { _ in
+            updateAccessibilityClosure(view.accessibilityLabelView)
+        }
+        updateAccessibilityClosure(view.accessibilityLabelView)
```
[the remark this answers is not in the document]
atmamont: `accessibilityLabelView` points to `textField` in `FormTextItemView`.

AdyenCard/Form/FormCardExpiryDateItemView.swift
```diff
@@ -9,22 +9,19 @@ import UIKit
...
             self?.setupAccessibility()
         }
+        setupAccessibility()
     }
 
     private func setupAccessibility() {
         let title = item.title ?? ""
-        let placeholder = item.placeholder ?? "MM/YY"
-        textField.accessibilityLabel = title.isEmpty ? placeholder : "\(title), \(placeholder)"
+        let format = localizedString(.cardExpiryItemAccessibilityLabel, item.localizationParameters)
```
erenbesel: So my nitpick was about this, on the `FormCardExpiryDateItem`, similar to how expiryMonth and year values are done:
```
internal var accessibilityValue: String {
    let format = localizedString(.cardExpiryItemAccessibilityLabel, localizationParameters)
    return title.isEmpty ? format : "\(title), \(format)"
}
```
And in this class we can just call `textField.accessiblityLabel = item.accessibilityValue`
Easier to test since it's on the item

Adyen/Assets/Generated/LocalizationKey.swift
```diff
@@ -70,6 +70,8 @@ public struct LocalizationKey {
     public static let cardExpiryItemPlaceholder = LocalizationKey(key: "adyen.card.expiryItem.placeholder")
     /// Invalid expiry date
     public static let cardExpiryItemInvalid = LocalizationKey(key: "adyen.card.expiryItem.invalid")
+    /// Expiry date accessibility label for VoiceOver
```
erenbesel: Description of each key is normally the english translation for that key. Why is this different?

AdyenCard/Form/FormCardExpiryDateItemView.swift
atmamont: Good point, it simplifies everything in a good way.
Updated.

Adyen/Assets/Generated/LocalizationKey.swift
atmamont: If we don't set accessibilityLabel, typically title is used which is translated and localized. Here we add a suffix to the title so we have to add localization, otherwise it would be a mix from localized "Expiry date" and then non-localized format string.

Adyen/UI/Form/Items/FormItemView.swift
```diff
@@ -13,6 +13,12 @@ open class FormItemView<ItemType: FormItem>: UIView, AnyFormItemView, AdyenObser
     /// The item represented by the view.
     public let item: ItemType
     
+    /// The view to set the `accessibilityLabel` on when invalid
+    internal var accessibilityLabelView: UIView? {
```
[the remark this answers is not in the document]
atmamont: Fixed.

Adyen/Assets/Generated/LocalizationKey.swift
goergisn:
```suggestion
    /// Month, 2 digits, Year, 2 digits
```

erenbesel (dismissed):

Adyen/Assets/Generated/LocalizationKey.swift
atmamont: Misunderstood the question, let's sync on this later as a team

neelSharma12 (approved):

goergisn (approved):

nauaros (approved):

PR #2153, opened 2025-06-05
v6: AdyenCheckout to be delegate for components and session

erenbesel:
# Summary [Required]
- Moves bridging logic of payment/action components delegation from AdyenCheckoutComponent to AdyenCheckout.
- Also integrates session into the AdyenCheckout, where it will act as the session delegate and expose final callbacks.
- Includes some refactoring on session, replacing result code responses with the new one and removing now unused code.
- Includes refactoring on AdyenCheckout to split some responsibility into different layers. Such as `AdyenCheckoutProvider` which handles the static creation.

## Ticket [Optional]
<ticket>
COSDK-491
</ticket>


AdyenCheckout/AdyenCheckoutProvider.swift
```diff
@@ -0,0 +1,118 @@
...
+        // fetch and store checkout attempt id
+        group.enter()
+        fetchCheckoutAttemptId(with: configuration) { result in
+            switch result {
+            case let .success(attemptId):
+                checkoutAttemptIdInstance = attemptId
+            case .failure:
+                // TODO: decide what to do for failed attempt
+                break
+            }
```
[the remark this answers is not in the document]
atmamont: While this should not be an error, is worth including adding telemetry for this case?

atmamont (approved):
I love that so much code was removed!
At this point I would like to see (not in this PR of course):
- tests added for new objects. While we can't have all of them green, we could establish a solid foundation
- v6 migration guide. We could start shaping out it's skeleton

AdyenCheckout/AdyenCheckoutProtocol.swift
```diff
@@ -9,40 +9,42 @@
...
 
-internal protocol AdyenCheckoutProtocol: AdyenSessionProviding, CheckoutAttemptIdProviding {
-    static func setup(
+internal protocol AdyenCheckoutProtocol {
+    
+    func createComponent(with paymentMethod: PaymentMethod) -> AdyenCheckoutComponent?
+    
+    func createComponent(with action: Action) -> AdyenCheckoutComponent?
+    
+    func createDropIn() -> DropInComponent?
```
nauaros, line 18: For this we'll need to align once I finish the changes in DropIn. Can you put a TODO here so we check together?
Points I am thinking about are:
- Configuration
- Localization
- Styling

AdyenCheckout/AdyenCheckoutProtocol.swift
```diff
@@ -9,40 +9,42 @@
...
+    func setup(
         with sessionId: String,
         sessionData: String,
         configuration: CheckoutConfiguration,
         presentationDelegate: PresentationDelegate?,
         completion: @escaping (Result<AdyenCheckout, Error>) -> Void
     )
     
-    static func setup(
+    func setup(
```
nauaros, line 30: nice!

AdyenCheckout/AdyenCheckoutProvider.swift
```diff
@@ -0,0 +1,118 @@
...
+        let group = DispatchGroup()
+        
+        // create and store session and payment methods
+        group.enter()
+        setupSession(with: configuration) { result in
+            switch result {
+            case let .success(session):
+                sessionInstance = session
+            case .failure:
+                // TODO: add error
```
nauaros, line 39: You think this should be a fatal error? Just asking

nauaros (approved):

AdyenCheckout/AdyenCheckoutProvider.swift
erenbesel: That's something we will need to decide for any errors to be asserts/fatals/throws etc.

PR #2170, opened 2025-06-20
DropInComponent modularization

nauaros:
# Summary

This PR refactors the `DropInComponent` to a more modular and testable architecture using a MVVM+R pattern. It introduces separate modules for different parts of the Drop-in UI (Root, PreselectedPaymentMethod, PaymentMethodList, and Component) with dedicated assemblers, routers, view controllers, and view models.

**Key changes:**
- Introduces a clear separation of concerns with modular components.
- Adds a centralized `DropInRootRouter` to manage navigation flow.
- Cleans up the DropInComponent by delegating responsibilities to new modules.


AdyenDropIn/Modules/Component/ComponentContainerViewModel.swift
```diff
@@ -0,0 +1,77 @@
...
+
+import Adyen
+import Foundation
+import UIKit
+#if canImport(AdyenActions)
+    @_spi(AdyenInternal) import AdyenActions
+#endif
+
+internal protocol ComponentContainerViewModelProtocol {
+    var view: UIViewController { get }
```
goergisn: I guess I understand the idea behind the naming but it might still be confusing to access a `view` and get a `UIViewController`

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListRouter.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    // MARK: - PaymentMethodListRouterProtocol
+
+    internal var rootViewController: UIViewController {
+        navigationController
+    }
+
+    internal func start() {
+        guard let view else {
+            fatalError("Router's view was not set.")
```
goergisn, line 46: Might be a good case for a throwing function?

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewController.swift
```diff
@@ -0,0 +1,62 @@
...
+        paymentMethodListView.willMove(toParent: self)
+        addChild(paymentMethodListView)
+        view.addSubview(paymentMethodListView.view)
+        paymentMethodListView.didMove(toParent: self)
+        paymentMethodListView.view.adyen.anchor(inside: view)
+    }
+
+    private func setupNavigationItem() {
+        navigationItem.title = "2"
+        let cancelButton = UIBarButtonItem(
```
goergisn, line 54: Not sure if we have translations for this. Maybe interesting to look at: https://developer.apple.com/documentation/uikit/uibarbuttonitem/systemitem/cancel

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewModel.swift
```diff
@@ -4,27 +4,21 @@
...
-internal protocol PaymentMethodListRouterProtocol: AnyObject {
-    func didLoad()
-    func present(_ component: PaymentComponent)
-    func delete(
-        storedPaymentMethod: StoredPaymentMethod,
-        completion: @escaping (Bool) -> Void
-    )
-}
-
+@objc
```
goergisn, line 11: Why the @objc?

AdyenDropIn/Modules/Component/ComponentContainerViewModel.swift
nauaros: Yes, its a good point. I can rename it to `componentView`

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewModel.swift
nauaros: The cancel method is being used from a `UIBarButtonItem`

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewController.swift
nauaros: Yeah... it is only available from iOS 14. I will localize these titles in future PRs

AdyenDropIn/DropInComponent.swift
```diff
@@ -125,7 +130,10 @@
 
     // MARK: - Presentable Component Protocol
 
-    public var viewController: UIViewController { navigationController }
+    public var viewController: UIViewController {
+        dropInRootRouter.start()
+        return dropInRootRouter.rootViewController
+    }
```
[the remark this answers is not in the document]
erenbesel: This seems like a valid concern. Can you reply here as to why it is fine to resolve the comment @nauaros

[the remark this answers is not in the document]
neelSharma12: @erenbesel Do you think start() method name makes complete sense here?

AdyenDropIn/Modules/Component/ComponentContainerViewModel.swift
neelSharma12: with `componentView` also it is quite misleading  I think.

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListRouter.swift
```diff
@@ -0,0 +1,62 @@
...
+import Adyen
+import Foundation
+import UIKit
+
+internal protocol PaymentMethodListRouterProtocol: AnyObject {
+    var rootViewController: UIViewController { get }
+    func start()
+    var delegate: PaymentMethodListRouterDelegate? { get set }
+    func cancel(completion: (() -> Void)?)
+    func didSelect(_ component: PresentableComponent)
```
neelSharma12, line 16: We can improve the readability here by first declaring the properties and then the methods rather than using mixed approach.

AdyenDropIn/Modules/Component/ComponentContainerViewModel.swift
nauaros: What do you suggest @neelSharma12 ?

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListRouter.swift
nauaros: Yes, will change it. Thanks

AdyenDropIn/Modules/Component/ComponentContainerViewModel.swift
neelSharma12: It should reflect that it is a viewcontroller instance. According to me `componentViewController`  is a good candidate. What do you think of this name?

nauaros: Renamed to `componentViewController`

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListRouter.swift
nauaros: Reordered for better readability

AdyenDropIn/DropInComponent.swift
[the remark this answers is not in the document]
nauaros: You're right @erenbesel . Initially I thought that `viewController` is only accessed one but it is true that the possibility of merchants accessing it multiple times is always there. I moved the start to the `DropInComponent` initializer. It's safer this way.

neelSharma12 (approved):

erenbesel (approved):

PR #2172, opened 2025-07-03
Fix to show cobadged brands UI only if certain brands are present in the binLookup response

neelSharma12:
# Summary

The new cobadged brands should only be shown if the /binLookup response contains cartebancaire, bcmc, or dankort.
This PR performs the necessary check

# Demo 

Simulator Screenshot - iPhone 12 - 2025-07-03 at 15 54 50

Simulator Screenshot - iPhone 12 - 2025-07-03 at 15 57 18

<fixed>
- For cards: in the payment form, the UI to [select a card brand for a co-badged card](https://docs.adyen.com/online-payments/co-badged-cards-compliance/) now appears only if the the following brands are available: 
- Cartes Bancaires (`cartebancaire`) 
- Bancontact card (`bcmc`) 
- Dankort (`dankort`) 
</fixed>

## Ticket [Optional]
<ticket>
COSDK-531
</ticket>

## Checklist
Manually tested with dual brands that are present in the "inclusion" list and those that aren't

### DIFF
--- a/AdyenCard/Components/Card/CardViewController.swift
+++ b/AdyenCard/Components/Card/CardViewController.swift
@@ -230,7 +230,7 @@ internal class CardViewController: FormViewController {
             items.coBadgedCardItem.resetItems()
         } else {
             brands = binInfo.brands ?? []
-            items.coBadgedCardItem.updateItems(brands, cardLogos: cardLogos)
+            showCoBadgedCardsUI(for: brands)
         }
         issuingCountryCode = binInfo.issuingCountryCode
         items.numberContainerItem.update(brands: brands)
@@ -243,6 +243,16 @@ internal class CardViewController: FormViewController {
         items.numberContainerItem.numberItem.selectBrand(cardBrand: selectedBrand)
 
         items.triggerInfoEvent(of: .selected, target: .dualBrandButton, brands: [selectedBrand])
+    }
+
+    private func showCoBadgedCardsUI(for brands: [CardBrand]) {
+        let allowedCoBadgedCardTypes = ["cartebancaire", "bcmc", "dankort"]
+        let coBadgedBrands = brands.filter { brand in
+            allowedCoBadgedCardTypes.contains(brand.type.rawValue)
+        }
+        if coBadgedBrands.count > 0 {
+            items.coBadgedCardItem.updateItems(brands, cardLogos: cardLogos)
+        }
     }
 }
 
--- a/AdyenCard/Components/Card/CardViewController.swift
+++ b/AdyenCard/Components/Card/CardViewController.swift
@@ -26,6 +26,7 @@ internal class CardViewController: FormViewController {
     private let initialCountryCode: String
     private let scope: String
     private let cardLogos: [FormCardLogosItem.CardTypeLogo]
+    private let allowedCoBadgedCardTypes = Set(["cartebancaire", "bcmc", "dankort"])
     private let cardScannerAnalyticsHandler: CardScannerAnalyticsHandler
     private lazy var cardScannerController: CardScannerControlling = {
         var controller: CardScannerControlling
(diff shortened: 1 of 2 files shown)


AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -244,6 +244,16 @@ internal class CardViewController: FormViewController {
 
         items.triggerInfoEvent(of: .selected, target: .dualBrandButton, brands: [selectedBrand])
     }
+
+    private func showCoBadgedCardsUI(for brands: [CardBrand]) {
+        let allowedCoBadgedCardTypes = ["cartebancaire", "bcmc", "dankort"]
```
erenbesel: Let's do this with the CardType enum not rawvalue.  Can also define this as a constant either on cardComponent to be injected here or can define at top of this class. `let allowedCoBadgedCardTypes: [CardType] = [.bcmc, .....]`

Lastly let's confirm these on our integration tests

descorp:
```suggestion
        let allowedCoBadgedCardTypes = Set(["cartebancaire", "bcmc", "dankort"])
```

descorp: Maybe move this list as a constant to top of the class ?
Or ideally move this check to `BinLookupResponse` because AFAIU in the future it will be part of API response

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -244,6 +244,16 @@ internal class CardViewController: FormViewController {
...
+        let coBadgedBrands = brands.filter { brand in
+            allowedCoBadgedCardTypes.contains(brand.type.rawValue)
+        }
+        if coBadgedBrands.count > 0 {
+            items.coBadgedCardItem.updateItems(brands, cardLogos: cardLogos)
+        }
```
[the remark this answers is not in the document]
descorp: I don't think this is necececy :)

descorp:
Can we also cover this with tests?

AdyenCard/Components/Card/CardViewController.swift
neelSharma12: d7ceb3e088d21ef6b59d88b38104cb37b460b283

neelSharma12:
> Can we also cover this with tests?

Modified/Added test d7ceb3e088d21ef6b59d88b38104cb37b460b283

AdyenCard/Components/Card/CardViewController.swift
neelSharma12: d7ceb3e088d21ef6b59d88b38104cb37b460b283

neelSharma12: Thanks for this suggestion :) de173ad7f48cb72548e6293fd0f38937592d1781

AdyenCard/Components/Card/CardViewController.swift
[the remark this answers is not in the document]
neelSharma12: yes, you are absolutely right :) d7ceb3e088d21ef6b59d88b38104cb37b460b283

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -244,6 +245,15 @@ internal class CardViewController: FormViewController {
 
         items.triggerInfoEvent(of: .selected, target: .dualBrandButton, brands: [selectedBrand])
     }
+
+    internal func showCoBadgedCardsUI(for brands: [CardBrand]) {
+        let coBadgedBrands = brands.filter { brand in
+            return allowedCoBadgedCardTypes.contains(brand.type)
+        }
+        if !coBadgedBrands.isEmpty { // Check if there are any co-badged brands
+            items.coBadgedCardItem.updateItems(brands, cardLogos: cardLogos)
```
erenbesel, line 254: Is passing the brands here intended? Shouldn't we pass the newly filtered array?
How is the test passing as well? :D

erenbesel: Could you add more test cases where the filtered array is not 2 brands anymore so we see what happens to the UI?

neelSharma12: The newly filtered array is just to check if the brands belongs to EU region or not. If we get any of the brand from `bcmc`, `dankort` or `cartebancair` then we need to show the UI. I have added for EU and AU.

neelSharma12: As discussed internally,  it will always be a local scheme (bcmc, eftos_australia) which will be paired up with a bigger scheme (visa, mc) so we do not need to anything for such scenarios.

erenbesel (approved):

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -244,6 +245,15 @@ internal class CardViewController: FormViewController {
 
         items.triggerInfoEvent(of: .selected, target: .dualBrandButton, brands: [selectedBrand])
     }
+
+    internal func showCoBadgedCardsUI(for brands: [CardBrand]) {
+        let coBadgedBrands = brands.filter { brand in
+            return allowedCoBadgedCardTypes.contains(brand.type)
+        }
+        if !coBadgedBrands.isEmpty { // Check if there are any co-badged brands
```
nauaros, line 253: The comment is not really needed. Code speaks by itself

nauaros (approved):

PR #2191, opened 2025-07-22
Refactor: Component Container Architecture for Adyen Drop-in

nauaros:
# Summary

This pull request implements a significant architectural refactoring within the Adyen Drop-in module. The primary goal is to introduce a new 'Component Container' pattern, centralizing the presentation and event handling logic for individual payment components. This change streamlines the internal flow of control and data, enhancing modularity and maintainability by clearly defining responsibilities across the Drop-in's various sub-modules.

### Highlights

* **New Component Container Architecture**: Introduced `ComponentContainerRouter` and `ComponentContainerViewModel` to encapsulate the presentation and event handling of individual payment components, promoting modularity and a more structured event flow.
* **Drop-in Module Renaming & Restructuring**: Performed a significant refactoring of the Drop-in module's internal components, including renaming many 'Root' prefixed classes (e.g., `DropInRootViewModel` to `DropInViewModel`) for clearer responsibilities and simplifying the overall structure.
* **Centralized Event Delegation**: Refactored the event flow so that `ComponentContainerViewModel` now handles various component delegates (Payment, Action, Presentation), centralizing event submission and cancellation logic, and forwarding events through the new router hierarchy to the main `DropInComponent`.
* **Enhanced Dependency Injection**: Updated all relevant assemblers (`ComponentContainerAssembler`, `PaymentMethodListAssembler`, `PreselectedPaymentMethodAssembler`, `DropInAssembler`) to consistently pass `context`, `configuration`, and component-specific delegates down the dependency chain, ensuring proper setup and configuration.
* **Formalized Initial Flow Logic**: Introduced a new `DropInRoot` enum to explicitly define and manage the different starting points of the Drop-in flow (preselected payment, single component, or payment method list), improving clarity and control over the initial user experience.

## Ticket
<ticket>
COSDK-557
</ticket>


AdyenDropIn/Modules/Component/ComponentContainerViewModel.swift
```diff
@@ -4,39 +4,60 @@
...
         self.component = component
-        self.isRoot = isRoot
-        self.cancelHandler = cancelHandler
+        self.context = context
+        self.configuration = configuration
+        self.cardComponentDelegate = cardComponentDelegate
+        self.partialPaymentDelegate = partialPaymentDelegate
+        self.actionComponent = AdyenActionComponent(context: context)
 
         setupComponent()
```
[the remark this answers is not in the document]
nauaros: Fixed!

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListRouter.swift
```diff
@@ -49,14 +51,50 @@ internal class PaymentMethodListRouter: PaymentMethodListRouterProtocol {
...
+
+        componentContainerRouter.start()
+
+        let componentContainerViewController = componentContainerRouter.rootViewController
+
+        if component.requiresModalPresentation {
+            view?.navigationController?.pushViewController(componentContainerViewController, animated: true)
+        } else {
+            view?.present(componentContainerViewController, animated: true)
+        }
```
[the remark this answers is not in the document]
nauaros: Yes! That's right, that flag needs to be changed in every component as it still carries the logic of the previous implementation where the `pushViewController` had a modal animation style.

AdyenCard/Form/FormCardNumberItemView.swift
```diff
@@ -44,15 +44,13 @@ internal final class FormCardNumberItemView: FormTextItemView<FormCardNumberItem
         textField.text = newValue
         updateValidationStatus()
     }
-    
-    @_spi(AdyenInternal)
```
erenbesel, line 48: Removing these didn't cause compile errors? 🤔

AdyenDropIn/Modules/Component/ComponentContainerRouter.swift
```diff
@@ -0,0 +1,70 @@
...
+
+    // MARK: - Properties
+
+    internal weak var delegate: ComponentContainerRouterDelegate?
+    private let view: UIViewController
+    private var viewModel: ComponentContainerViewModelProtocol
+
+    // MARK: - Initializers
+
+    internal init(
```
erenbesel: Wou can also pass the delegate in the init if there is no scenario where delegate should not be set.

AdyenDropIn/Modules/Component/ComponentContainerViewModel.swift
```diff
@@ -4,39 +4,60 @@
...
+    private let context: AdyenContext
+    private let configuration: DropInComponent.Configuration
+    private weak var cardComponentDelegate: CardComponentDelegate?
+    private weak var partialPaymentDelegate: PartialPaymentDelegate?
+
+    private let actionComponent: AdyenActionComponent
 
     // MARK: - Initializers
 
     internal init(
```
erenbesel, line 44: Same here, can pass delegate too if we want it to be set always.

AdyenDropIn/Modules/Component/ComponentContainerViewModel.swift
```diff
@@ -61,17 +94,79 @@ internal class ComponentContainerViewModel: ComponentContainerViewModelProtocol
...
+        from component: any PaymentComponent
     ) {
-        // TODO: -
-        print("SUBMIT")
+        let checkoutAttemptId = component.context.analyticsProvider?.checkoutAttemptId
+        let updatedData = data.replacing(
+            checkoutAttemptId: checkoutAttemptId
+        )
+
+        guard updatedData.browserInfo == nil else {
```
erenbesel, line 110: I just realized something about having 2 submit functions in payment component and dropin. Adding a comment here to discuss offline more.

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListAssembler.swift
```diff
@@ -18,30 +18,41 @@ internal struct PaymentMethodListAssembler: PaymentMethodListAssemblerProtocol {
...
+            partialPaymentDelegate: partialPaymentDelegate
+        )
         let router = PaymentMethodListRouter(componentContainerAssembler: componentContainerAssembler)
         let viewModel = PaymentMethodListViewModel(
-            router: router,
             context: context,
             componentManager: componentManager,
             configuration: configuration
         )
+        viewModel.delegate = router
```
erenbesel: We seem to be always setting delegate indeed. Moving it to init would prevent the caller from forgetting this line.

Demo/Common/Networking/APIRequest.swift
```diff
@@ -14,19 +14,19 @@ internal protocol APIRequest: Request where ErrorResponseType == APIError {}
 internal struct APIError: ErrorResponse, LocalizedError {
     
     /// The status.
-    public let status: Int?
```
erenbesel: Swiftformat remnant?

AdyenDropIn/Modules/Root/DropInRouter.swift
```diff
@@ -9,52 +9,51 @@ import AdyenNetworking
...
-    private enum DropInRoot {
-        case preselected(_ paymentComponent: PaymentComponent)
-        case component(_ paymentComponent: PresentableComponent)
-        case paymentMethodList
-    }
+internal class DropInRouter: DropInRouterProtocol {
 
     // MARK: - Properties
 
+    internal weak var delegate: DropInRouterDelegate?
```
erenbesel, line 33: Same for here to pass it on init (only if it should be always be set, for others too).

Tests/IntegrationTests/Components Tests/PaymentComponent/PaymentComponentSubject.swift
```diff
@@ -19,7 +19,7 @@ class PaymentComponentSubject: PaymentComponent {
 
     // MARK: - Initializers
 
-    public init(
```
erenbesel: if these compile, doesn't matter.

Tests/UnitTests/Core/PaymentMethodTests.swift
```diff
@@ -845,7 +845,7 @@ class PaymentMethodTests: XCTestCase {
         testCoding(paymentMethod)
     }
     
-    public func expectedStoredCardPaymentMethodDisplayInfo(method: StoredCardPaymentMethod, localizationParameters: LocalizationParameters?) -> DisplayInformation {
```
erenbesel: if these compile, doesn't matter.

Tests/UnitTests/Core/PaymentMethodTests.swift
```diff
@@ -1038,7 +1038,7 @@ class PaymentMethodTests: XCTestCase {
         testCoding(paymentMethod)
     }
     
-    public func expectedBancontactCardDisplayInfo(
```
erenbesel: if these compile, doesn't matter.

erenbesel (commented):
Super nice changes!

AdyenDropIn/Modules/Component/ComponentViewModel.swift
```diff
@@ -75,7 +75,7 @@ internal final class ComponentViewController: UIViewController {
         setupNavigationItem()
     }
 
-    override public func viewDidDisappear(_ animated: Bool) {
+    override func public viewDidDisappear(_ animated: Bool) {
         super.viewDidDisappear(animated)
         componentView.resignFirstResponder()
         viewModel.didCancel()
```
atmamont: Does this affect the multi-screen flow? Imagine user goes back and forth between screens, triggering cancelation

AdyenDropIn/DropInComponent.swift
```diff
@@ -80,15 +80,19 @@ public final class DropInComponent: NSObject,
             .retryAPIClient(with: scheduler)
             .retryOnErrorAPIClient()
 
-        let dropInRootAssembler = DropInRootAssembler(
+        let dropInRootAssembler = DropInAssembler(
```
atmamont: let's rename `dropInRootAssembler` to `dropInAssembler`?

AdyenDropIn/Modules/Component/ComponentContainerAssembler.swift
```diff
@@ -9,20 +9,51 @@ import Foundation
...
+    private let configuration: DropInComponent.Configuration
+    private let cardComponentDelegate: CardComponentDelegate?
+    private let partialPaymentDelegate: PartialPaymentDelegate?
+
+    // MARK: - Initializers
+
+    internal init(
+        context: AdyenContext,
+        configuration: DropInComponent.Configuration,
+        cardComponentDelegate: CardComponentDelegate?,
```
atmamont, line 35: I didn't find any usages for this being optional.
I'd rather ditch struct here and go with classes composition or simple inheritance to use different instances depending on the need:
- `ComponentContainerAssembler`
- `CardComponentContainerAssembler`

Maybe I'm missing a bigger picture, but from this PR it looks like `cardComponentDelegate` is something very specific that leaks into an abstract interface. Probably just removing this scpecific delegate from the interface and making it optional would allow to hide the assigning logic in the scope when it matters and makes sense.

Same applies to `PartialPaymentDelegate` type

AdyenDropIn/Modules/Component/ComponentContainerRouter.swift
```diff
@@ -0,0 +1,70 @@
...
+
+internal protocol ComponentContainerRouterDelegate: AnyObject {
+    func didSubmit(_ data: PaymentComponentData, from component: any PaymentComponent)
+    func didFail(with error: any Error)
+    func didCancel(component: any PaymentComponent)
+}
+
+internal protocol ComponentContainerRouterProtocol {
+    var delegate: ComponentContainerRouterDelegate? { get set }
+    var rootViewController: UIViewController { get }
```
atmamont, line 18: Does it need to have `root` in the name?

AdyenDropIn/Modules/Component/ComponentContainerViewModel.swift
```diff
@@ -45,14 +66,26 @@ internal class ComponentContainerViewModel: ComponentContainerViewModelProtocol
...
 
     private func setupComponent() {
         (component as? PaymentComponent)?.delegate = self
+        (component as? CardComponent)?.cardComponentDelegate = cardComponentDelegate
+        (component as? PartialPaymentComponent)?.partialPaymentDelegate = partialPaymentDelegate
+        (component as? PartialPaymentComponent)?.readyToSubmitComponentDelegate = self
+    }
+
+    private func setupActionComponent() {
+        actionComponent.delegate = self
```
atmamont, line 87: Why is component creation and setup are split? I'd make it either lazy var and create and setup on demand or call `setupActionComponent` right away after its creation. Otherwise current approach could open up a possibility to have at some point component created but not configured.

AdyenDropIn/Modules/Component/ComponentContainerViewModel.swift
```diff
@@ -61,17 +94,79 @@ internal class ComponentContainerViewModel: ComponentContainerViewModelProtocol
...
+            checkoutAttemptId: checkoutAttemptId
+        )
+
+        guard updatedData.browserInfo == nil else {
+            delegate?.didSubmit(updatedData, from: component)
+            return
+        }
+        updatedData.dataByAddingBrowserInfo { [weak self] in
+            guard let self else { return }
+            delegate?.didSubmit($0, from: component)
```
atmamont, line 116: This looks a bit hard to read, let's streamline the BrowserInfo data addition logic so we have only one delegate?.didSubmit call instead of two

AdyenDropIn/Modules/Component/ComponentContainerViewModel.swift
```diff
@@ -61,17 +94,79 @@ internal class ComponentContainerViewModel: ComponentContainerViewModelProtocol
...
+            handleCancellation()
+        } else {
+            delegate?.didFail(with: error)
+        }
+    }
+
+    // MARK: - Private
+
+    private func handleCancellation() {
+        component.cancelIfNeeded()
```
atmamont: can we simplify and call it `cancel`? on the API level it adds unnecessary context

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListRouter.swift
```diff
@@ -49,14 +51,51 @@ internal class PaymentMethodListRouter: PaymentMethodListRouterProtocol {
...
+        )
+        self.componentContainerRouter = componentContainerRouter
+
+        componentContainerRouter.start()
+
+        let componentContainerViewController = componentContainerRouter.rootViewController
+
+        // TODO: - Invert `requiresModalPresentation` logic or remove it fully.
+        if component.requiresModalPresentation {
+            view?.navigationController?.pushViewController(componentContainerViewController, animated: true)
```
atmamont, line 79: I think it adds unnecessary UIKit internals exposure to an abstract router. In this case this could be extracted to a `presenter.push()` that encapsulates UIKit presentation logic and maybe later could be hot-swapped with another presenter. It would also make it testable.

atmamont (approved):
Massive improvements, Naufal 👏 
I left some comments that relate mostly to readability and ambiguity.

I'd also be happy to see new routing/presenting/delegates logic captured vis test coverage.

AdyenCard/Form/FormCardNumberItemView.swift
nauaros: Good catch. No idea where the change came from. Reverting it. It wasnt intentional

Demo/Common/Networking/APIRequest.swift
nauaros: Indeed. Reverted

atmamont:
Updated branch as diff tool output was not readable

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListAssembler.swift
nauaros: Nice one, change it that way.

AdyenDropIn/Modules/Root/DropInRouter.swift
nauaros: In this case the `DropInComponent` acts as the `delegate`. So it can only be set once the `DropInComponent` is initialized. Look at this:

```swift
    public init(
        paymentMethods: PaymentMethods,
        context: AdyenContext,
        configuration: Configuration = .init(),
        title: String? = nil
    ) {
        self.title = title ?? Bundle.main.displayName
        self.context = context
        self.paymentMethods = paymentMethods

        let scheduler = SimpleScheduler(maximumCount: 3)
        self.apiClient = APIClient(apiContext: context.apiContext)
            .retryAPIClient(with: scheduler)
            .retryOnErrorAPIClient()

        let dropInRootAssembler = DropInAssembler(
            paymentMethods: paymentMethods,
            context: context,
            configuration: configuration,
            cardComponentDelegate: cardComponentDelegate,
            partialPaymentDelegate: partialPaymentDelegate
        )
        self.dropInRouter = dropInRootAssembler.resolveDropInRootRouter()
        super.init()

        self.dropInRouter.delegate = self
        self.dropInRouter.start()
    }
```

AdyenDropIn/Modules/Component/ComponentContainerViewModel.swift
nauaros: Change the way of assembling here too

AdyenDropIn/Modules/Component/ComponentContainerRouter.swift
nauaros: Done!

AdyenDropIn/DropInComponent.swift
nauaros: good catch ;)

AdyenDropIn/Modules/Component/ComponentContainerRouter.swift
nauaros: Not necessarily. Just wanna emphasize this is the start view of the router. It can named `viewController`

AdyenDropIn/Modules/Component/ComponentContainerViewModel.swift
nauaros: Good point about creating it on demand, I haven't thought of it. I moved it to a computed property.

```swift
    private var actionComponent: AdyenActionComponent {
        let actionComponent = AdyenActionComponent(context: context)
        actionComponent.delegate = self
        actionComponent.presentationDelegate = self
        actionComponent.configuration.style = configuration.style.actionComponent
        actionComponent.configuration.localizationParameters = configuration.localizationParameters
        actionComponent.configuration.threeDS = configuration.actionComponent.threeDS
        actionComponent.configuration.twint = configuration.actionComponent.twint
        return actionComponent
    }
```

AdyenDropIn/Modules/Component/ComponentContainerViewModel.swift
nauaros: Cannot really find a nice way of rewriting it with a single delegate call. What do you think of this?

```swift
func didSubmit(
    _ data: PaymentComponentData,
    from component: any PaymentComponent
) {
    let checkoutAttemptId = component.context.analyticsProvider?.checkoutAttemptId
    let enrichedData = data.replacing(checkoutAttemptId: checkoutAttemptId)

    let submit: (PaymentComponentData) -> Void = { [weak self] submission in
        delegate?.didSubmit(submission, from: component)
    }

    if let _ = enrichedData.browserInfo {
        submit(enrichedData)
    } else {
        enrichedData.dataByAddingBrowserInfo(submit)
    }
}
```

AdyenDropIn/Modules/Component/ComponentContainerViewModel.swift
nauaros: Move this logic to a `cancel` method that just calling `handleCancellation `.

AdyenDropIn/Modules/Component/ComponentViewModel.swift
nauaros: This is only on a DropIn level. The behavior is the following:
1. Shopper opens card component.
2. Shopper start payment flow after tapping payment button.
3. Shopper goes back to the payment list while payment is in progress.
4. When view disappears, the component is cancelled and its delegate is notified of the cancellation.

Tests/UnitTests/Core/PaymentMethodTests.swift
nauaros: Done by the linter. Reverted

Tests/UnitTests/Core/PaymentMethodTests.swift
nauaros: Done by the linter. Reverted

Tests/IntegrationTests/Components Tests/PaymentComponent/PaymentComponentSubject.swift
nauaros: Done by the linter. Reverted

AdyenDropIn/Modules/Component/ComponentContainerViewModel.swift
atmamont: Replied in PM

atmamont: If I would be to refactor this, I'd start with this approach:
- extract multiple responsibilities from this method (enrich data with `checkoutAttemptId`, enrich data with `browserInfo`). It is too much of business logic going on in `didSubmit`.
- I would try composition instead, injecting a sort of `dataEnricher` entity that would handle all responsibilities. It could be done using two different entities (single responsibility principle) or the same one, composed on an assembly layer, one for checkoutAttemptId and the second one for browserInfo. I am not sure it would be easy to come up with a single interface for both of them though. If not, then two separate entities could be a solution.

This way we will get:
- `didSubmit` flow streamlined, easy to read
- new interfaces/entities easy to plug and reuse for any other place that requires data enriching

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListRouter.swift
nauaros: I want to revisit this, not finished yet. Good idea the presenter, I will consider it.

erenbesel (approved):

PR #2193, opened 2025-07-24
Use .xcconfig file to store secrets

atmamont:
# Summary

## Ticket 
<ticket>
COSDK-274
</ticket>

## Checklist 

- [x] Tested changes locally
- [x] Tested changes on CI by releasing Demo app to Testflight


erenbesel (approved):
Nice!

neelSharma12 (approved):

goergisn (approved):

.gitignore
```diff
@@ -44,3 +44,6 @@ TempProject/
 .build
 .vscode
 .tmp
+
+# Configuration files with secrets
+Demo/Secrets.test.xcconfig
```
goergisn: `test` because they are the secrets for the test environment?

atmamont: Yes

Demo/UIKit/Info.plist
```diff
@@ -2,6 +2,16 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>ADYEN_CLIENT_KEY</key>
```
robertdalmeida, line 5: My only concern here is that the test app would have these values exposed for the testflight builds right?

robertdalmeida: compared to previously that it would be in code so not easily accessible.

atmamont: Yes, indeed there is no way to prevent attacker that has ipa file from accessing those values. The goal of this change is to prevent us from occasional key leaks.

Adyen.xcodeproj/project.pbxproj
```diff
@@ -6985,11 +6997,13 @@
 			files = (
 				F9EDB78923955ABF00CFB3C9 /* AdyenUIHost.strings in Resources */,
 				E9909FB422B7C816008DE50C /* PaymentMethods.json in Resources */,
+				B6244E0B2E3281B800F42780 /* Secrets.xcconfig in Resources */,
```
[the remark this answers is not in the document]
atmamont: https://github.com/Adyen/adyen-ios/pull/2193/commits/ce9a07d8d204fed7446365d4846498ee6cbe0878

nauaros (approved):

erenbesel (approved):

PR #2202, opened 2025-08-04
blik component v6 example on advanced flow

erenbesel:
# Summary [Required]
AdyenCheckout and CheckoutConfiguration should be ready to be initialized.
So it's a good time to start the demo sample code for Blik component and see if there is anything missing to prevent us from a deliverable BLIK component.

- [x] Create BLIKComponentExample file on the session or advanced flow (both need to be done eventually)
   - [ ] Session flow
   - [x] Advanced flow
- [x] Instantiate CheckoutConfig with BLIK configuration
- [x] Supply onSubmit and onAdditionalDetails if advanced flow (can also be used on session flow if they are passed)
- [x] Instantiate AdyenCheckout and pass the configuration
- [x] Present the component
- [ ] Error handling

## Ticket [Optional]
<ticket>
COSDK-576
</ticket>

## Checklist [Required]

- [x] Tested changes locally

### DIFF
--- a/Adyen/Analytics/Models/AdyenAnalytics.swift
+++ b/Adyen/Analytics/Models/AdyenAnalytics.swift
@@ -50,7 +50,7 @@ public struct AnalyticsConfiguration {
     // MARK: - Properties
 
     /// A Boolean value that determines whether analytics is enabled.
-    public var isEnabled = true
+    public let isEnabled: Bool
     
     @_spi(AdyenInternal)
     public var context: AnalyticsContext = .init()
@@ -58,7 +58,9 @@ public struct AnalyticsConfiguration {
     // MARK: - Initializers
     
     /// Initializes a new instance of `AnalyticsConfiguration`
-    public init() { /* Empty implementation */ }
+    public init(isEnabled: Bool = true) {
+        self.isEnabled = isEnabled
+    }
 }
 
 /// Additional fields to be provided with an ``InitialAnalyticsRequest``
--- a/AdyenCheckout/AdyenCheckout+DelegateCallbacks.swift
+++ b/AdyenCheckout/AdyenCheckout+DelegateCallbacks.swift
@@ -43,6 +43,7 @@ extension AdyenCheckout: PaymentComponentDelegate {
         if let action = paymentsResponse.action {
             actionHandlingComponent.handle(action)
         } else {
+            // TODO: check for error cases here
             finish(with: CheckoutResult(resultCode: paymentsResponse.resultCode))
         }
     }
--- a/AdyenCheckout/AdyenCheckoutComponent.swift
+++ b/AdyenCheckout/AdyenCheckoutComponent.swift
@@ -9,6 +9,7 @@
 @_spi(AdyenInternal) import AdyenDropIn
 @_spi(AdyenInternal) import AdyenComponents
 @_spi(AdyenInternal) import AdyenActions
+import UIKit
 
 package typealias CheckoutComponentDelegate = (PaymentComponentDelegate & ActionComponentDelegate)
 
--- a/AdyenCheckout/AdyenCheckoutComponent.swift
+++ b/AdyenCheckout/AdyenCheckoutComponent.swift
@@ -23,6 +23,13 @@ public final class AdyenCheckoutComponent {
     
     internal weak var delegate: CheckoutComponentDelegate?
     
+    public var viewController: UIViewController? {
+        guard let presentableComponent = paymentComponent as? PresentableComponent else {
+            return nil
+        }
+        return presentableComponent.viewController
+    }
+    
     package init(
         paymentMethod: PaymentMethod,
         configuration: CheckoutConfiguration,
--- a/Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
+++ b/Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
@@ -84,14 +84,12 @@ internal final class BLIKComponentAdvancedFlowExample: InitialDataAdvancedFlowPr
                 self.component = component
                 self.presenter?.present(viewController: viewController(for: component!), completion: nil)
             case let .failure(error):
-                print("failed to create adyencheckout \(error)")
+                print("failed to create adyen checkout \(error)")
             }
         }
         
-        // add all the config options to settings? like showsubmitbutton
+        // add all the config options to settings? like show submit button
     }
-    
-//    private func callPaymentsAsync(with data: PaymentComponentData) async -> CheckoutPaymentsResponse {}
     
     private func callPayments(with data: PaymentComponentData, completion: PaymentsResponseHandler?) {
         let request = PaymentsRequest(data: data)
--- a/Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
+++ b/Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
@@ -56,15 +56,15 @@ internal final class BLIKComponentAdvancedFlowExample: InitialDataAdvancedFlowPr
             BLIKComponentConfiguration()
                 .showsSubmitButton(false)
         }
-        .onSubmit { data, handler in
-            self.callPayments(with: data, completion: handler)
+        .onSubmit { [weak self] data, handler in
+            self?.callPayments(with: data, completion: handler)
         }
-        .onAdditionalDetails { data, handler in
-            self.callDetails(with: data, completion: handler)
+        .onAdditionalDetails { [weak self] data, handler in
+            self?.callDetails(with: data, completion: handler)
         }
-        .onComplete { result in
-            self.presenter?.dismiss {
-                self.presenter?.presentAlert(withTitle: "Result Code", message: result.resultCode.rawValue)
+        .onComplete { [weak self] result in
+            self?.presenter?.dismiss {
+                self?.presenter?.presentAlert(withTitle: "Result Code", message: result.resultCode.rawValue)
             }
         }
         
--- a/Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
+++ b/Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
@@ -1,0 +1,147 @@
+//
+// Copyright (c) 2025 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Adyen
+import AdyenActions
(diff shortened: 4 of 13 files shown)


AdyenCheckout/AdyenCheckout+DelegateCallbacks.swift
```diff
@@ -43,6 +43,7 @@ extension AdyenCheckout: PaymentComponentDelegate {
         if let action = paymentsResponse.action {
             actionHandlingComponent.handle(action)
         } else {
+            // TODO: check for error cases here
```
nauaros, line 46: How do you plan on handling errors here? Delegate or callback to the merchant

Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
```diff
@@ -0,0 +1,145 @@
...
+        ) { [weak self] result in
+            guard let self else { return }
+            
+            self.presenter?.hideLoadingIndicator()
+            
+            switch result {
+            case let .success(checkout):
+                self.adyenCheckout = checkout
+                let component = checkout.createComponent(with: blikPaymentMethod)
+                self.component = component
```
nauaros: What's the reason to keep a reference to the component? Trying to think from a merchant perspective as the callbacks have already been defined

AdyenCheckout/AdyenCheckout+DelegateCallbacks.swift
erenbesel: Callback to merchant most likely

Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
erenbesel: Component's still have delegates and they are weak, they won't get triggered if it's not stored.

Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
```diff
@@ -0,0 +1,147 @@
...
+    
+    private func viewController(for component: AdyenCheckoutComponent) -> UIViewController {
+        let navigation = UINavigationController(rootViewController: component.viewController!)
+        component.viewController?.navigationItem.leftBarButtonItem = .init(
+            barButtonSystemItem: .cancel,
+            target: self,
+            action: #selector(cancelPressed)
+        )
+        return navigation
+    }
```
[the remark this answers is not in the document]
atmamont: I'd suggest to avoid merging force-unwraps (fine for tests though)

Adyen/Analytics/Models/AdyenAnalytics.swift
```diff
@@ -50,15 +50,17 @@ public struct AnalyticsConfiguration {
...
+    public let isEnabled: Bool
     
     @_spi(AdyenInternal)
     public var context: AnalyticsContext = .init()
 
     // MARK: - Initializers
     
     /// Initializes a new instance of `AnalyticsConfiguration`
-    public init() { /* Empty implementation */ }
+    public init(isEnabled: Bool = true) {
```
atmamont, line 61: This reads as "configuration is enabled". While it might be the case, it also might be this flag affect the whole analytics subsystem to be enabled/disabled. In this case I would suggest renaming it to "isAnalyticsEnabled"

AdyenCheckout/AdyenCheckoutComponent.swift
```diff
@@ -9,6 +9,7 @@
 @_spi(AdyenInternal) import AdyenDropIn
 @_spi(AdyenInternal) import AdyenComponents
 @_spi(AdyenInternal) import AdyenActions
+import UIKit
```
atmamont, line 12: In a perfect world this component should know nothing about UIKit or WatchKit or SwiftUI, and this viewController is instantiated in another layer.

Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
```diff
@@ -0,0 +1,145 @@
...
+import AdyenActions
+import AdyenCheckout
+import AdyenComponents
+
+internal final class BLIKComponentAdvancedFlowExample: InitialDataAdvancedFlowProtocol {
+
+    internal weak var presenter: PresenterExampleProtocol?
+    
+    private var adyenCheckout: AdyenCheckout?
+    private var component: AdyenCheckoutComponent?
```
atmamont: For the sake of consistency on the merchant side I'd suggest renaming this one to `adyenComponent` to match `adyenCheckout`

Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
```diff
@@ -0,0 +1,145 @@
...
+    
+    internal lazy var apiClient = ApiClientHelper.generateApiClient()
+    
+    // comes from demo app protocol, unused on new structure
+    internal lazy var context: AdyenContext = generateContext()
```
atmamont, line 22: Is it still used or can be removed?

Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
```diff
@@ -0,0 +1,145 @@
...
+    
+    internal init() {}
+    
+    internal func start() {
+        presenter?.showLoadingIndicator()
+        requestPaymentMethods(order: nil) { [weak self] result in
```
atmamont, line 28: The `order` still doesn't sit well with me as because of Uplift order will mean something different in a year or so. I don't have a better name now :)

Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
```diff
@@ -0,0 +1,145 @@
...
+            guard let self else { return }
+            
+            do {
+                switch result {
+                case let .success(paymentMethods):
+                    try self.presentBlik(from: paymentMethods)
```
atmamont, line 34: This level of indentation is not the best but it's also not bad either. 👍 
Readability is good!

Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
```diff
@@ -0,0 +1,145 @@
...
+                case let .failure(error):
+                    throw error
+                }
+            } catch {
+                self.presenter?.hideLoadingIndicator()
```
atmamont, line 39: Shall we move it to `self.handle(error)` to make it even cleaner? Then you will get rid of self?. inside the new method

Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
```diff
@@ -0,0 +1,145 @@
...
+    
+    private func presentBlik(from paymentMethods: PaymentMethods) throws {
+        guard let blikPaymentMethod = paymentMethods.paymentMethod(ofType: BLIKPaymentMethod.self) else { throw IntegrationError.paymentMethodNotAvailable(paymentMethod: BLIKPaymentMethod.self) }
+        
+        let configuration = try CheckoutConfiguration(
+            environment: ConfigurationConstants.componentsEnvironment,
+            amount: ConfigurationConstants.current.amount,
+            clientKey: ConfigurationConstants.clientKey,
+            analyticsConfiguration: .init(
+                isEnabled: ConfigurationConstants.current.analyticsSettings.isEnabled
```
atmamont, line 53: Confirming my assumption from above - `isEnabled` is a bit ambiguous now

atmamont:
Great stuff, Eren! I brought some open topics for discussion, but nothing is a blocker

Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
```diff
@@ -0,0 +1,147 @@
...
+            guard let self else { return }
+            
+            self.presenter?.hideLoadingIndicator()
+            
+            switch result {
+            case let .success(checkout):
+                self.adyenCheckout = checkout
+                let component = checkout.createComponent(with: blikPaymentMethod)
+                self.component = component
+                self.presenter?.present(viewController: viewController(for: component!), completion: nil)
```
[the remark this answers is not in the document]
neelSharma12: I know it is a demo app but should we remove force-unwrapping here?

Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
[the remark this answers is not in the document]
neelSharma12: Should we remove force-unwrapping here?

Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
erenbesel: The protocol is used on the other examples so can't remove yet.

AdyenCheckout/AdyenCheckoutComponent.swift
erenbesel: But if we want to expose the viewController property, shouldn't the import be there? We are not creating the viewcontroller here, just exposing the inner component's vc and yet the import was required by SPM

Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
erenbesel: It hides the loading indicator logic if we put it on a separate function and I wanted to make it visible where it was shown/hidden on the same function block.

AdyenCheckout/AdyenCheckoutComponent.swift
atmamont: Yes, this is because you explicitly use `UIViewController` type that for instance doesn't exist in WatchKit, making this file impossible to use on Apple Watch (I know we don't have to, but this is a good practice to think about it to properly split the layers)

Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
erenbesel: I'm not sure on this one, to me `isEnabled` is clear enough since its a property of the configuration that represents analytics and init directly uses the property.

neelSharma12 (approved):

atmamont (approved):

PR #2218, opened 2025-08-13
Switch to async/await for adyencheckout/session

erenbesel:
# Summary [Required]
Replaces the old closure approach to AdyenCheckout and AdyenSession with async/await
Also updates the Advanced Blik example.

## Ticket [Optional]
<ticket>
COSDK-590
</ticket>

## Checklist [Required]

- [ ] Tested changes locally


AdyenSession/AdyenSession.swift
```diff
@@ -72,77 +72,62 @@ public final class AdyenSession: AdyenSessionProtocol {
...
-                )
-                completion(.success(sessionState))
-            case let .failure(error):
-                completion(.failure(error))
+        
+        let response = try await withCheckedThrowingContinuation { continuation in
+            baseAPIClient.perform(request) { result in
+                continuation.resume(with: result)
             }
         }
```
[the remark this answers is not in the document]
erenbesel: This is not a problem as apiclient is now envoloped under async/await block which swift concurrency will keep alive.

AdyenSession/AdyenSession+PaymentComponentDelegate.swift
```diff
@@ -155,26 +155,25 @@ extension AdyenSession {
...
-        }
-    }
-    
-    private func updateState(with result: Result<State, Error>, component: Component) {
-        switch result {
-        case let .success(state):
-            self.state = state
-        case let .failure(error):
-            finish(with: error, component: component)
+        Task {
```
atmamont, line 168: The `handle` method itself has quite a generic name and it has a lot going on inside. We can think of breaking it down and give smaller methods specific names following the SRP, this would make code easier to read and understand.

AdyenSession/AdyenSession.swift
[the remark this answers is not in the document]
atmamont: I'd be careful here and add a unit test to check the api client retention.

Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
```diff
@@ -28,21 +28,24 @@ internal final class BLIKComponentAdvancedFlowExample: InitialDataAdvancedFlowPr
...
+            Task {
+                do {
+                    switch result {
+                    case let .success(paymentMethods):
+                        try await self.presentBlik(from: paymentMethods)
+                    case let .failure(error):
+                        throw error
+                    }
+                } catch {
+                    self.presenter?.hideLoadingIndicator()
```
atmamont: In other places we tried to keep loading indicator show/hide logic in the same scope, and here I don't see hiding indicator logic for success branch. 
I can assume you want to keep the indicator while calling AdyenCheckout.setup [here](https://github.com/Adyen/adyen-ios/pull/2218/files#diff-34ab34a188622e19e80b9fbe62e3674b7f18516a21216951adb81b8acbb9f310R73), but still the indicator logic feels spread too much here.

Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
```diff
@@ -68,30 +70,18 @@ internal final class BLIKComponentAdvancedFlowExample: InitialDataAdvancedFlowPr
...
-                self.presenter?.present(viewController: viewController(for: component), completion: nil)
-            case let .failure(error):
-                print("failed to create adyen checkout \(error)")
-            }
+        let checkout = try await AdyenCheckout.setup(with: paymentMethods, configuration: configuration, presentationDelegate: self)
+        
+        presenter?.hideLoadingIndicator()
+        
+        guard let component = checkout.createComponent(with: blikPaymentMethod) else {
+            print("component not found")
```
atmamont: Let's capitalize the error message and add context like this?
`print("Component not found for payment method: \(blikPaymentMethod)")`

atmamont (approved):
Bye bye closures! Solid stuff, Eren. I left some comments, up to your consideration 👍

AdyenSession/AdyenSession.swift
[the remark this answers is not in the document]
erenbesel: The tests already cover this as the apiclient is a local variable and tests succeed. I've also done extended tests to in a separate playground to make sure.

atmamont (approved):

Demo/Common/IntegrationExamples/Components/BLIKComponentAdvancedFlowExample.swift
erenbesel: Yeah, I mentioned it before it wasn't easy to find a spot. I will look into this on the next example

PR #2221, opened 2025-08-15
COSDK-599: Fix Github Actions

nauaros:
# Summary [Required]
 - Update Github Action Xcode version from 16.1 to 16.4
 - Change destination version from 18.1 to 18.6
 - Change runner version to macos-15 in Compatibility tests [Xcode 26]

## Ticket
<ticket>
COSDK-599
</ticket>

## Checklist

- [x] Tested changes locally


.github/workflows/detect_api_changes.yml
```diff
@@ -29,7 +29,7 @@ jobs:
     - name: Select latest Xcode
       uses: maxim-lobanov/setup-xcode@v1
       with:
-        xcode-version: '16.1'
+        xcode-version: '16.2'
```
erenbesel, line 32: This one is 16.2 while the rest 16.4

nauaros: Xcode 16.4 is only available in macos-15, but this job does not run properly on macos-15, something is wrong with the api diff project (related with SwiftLint). As 16.1 is not available anymore, switching to 16.2 is the best option for now while keeping macos-14

atmamont, line 32: Should we keep Xcode version the same across all jobs? This one is different from the rest

atmamont (approved):

.github/workflows/detect_api_changes.yml
nauaros: Eren asked the same question: https://github.com/Adyen/adyen-ios/pull/2221#discussion_r2281920068

neelSharma12 (approved):

PR #2239, opened 2025-08-28
Added new styling for toggle

neelSharma12:
# Summary [Required]

This PR contains changes below changes-

- New styling for toggle


AdyenUI/Styles/AdyenToggleStyle.swift
```diff
@@ -0,0 +1,72 @@
...
+    public func backgroundColor(_ backgroundColor: UIColor) -> AdyenToggleStyle {
+        var newStyle = self
+        newStyle.backgroundColor = backgroundColor
+        return newStyle
+    }
+    
+    /// Returns a new AdyenToggleStyle with the specified cornerRadius.
+    /// - Parameter cornerRadius: The cornerRadius to set.
+    /// - Returns: A new `AdyenToggleStyle` instance.
+    public func backgroundColor(_ cornerRadius: CornerRounding) -> AdyenToggleStyle {
```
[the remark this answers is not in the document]
atmamont: 😮

Tests/UnitTests/AdyenTheme/AdyenThemeTests.swift
```diff
@@ -32,5 +32,25 @@ final class AdyenThemeTests: XCTestCase {
...
+        
+        newToggleStyle.tintColor = .black
+
+        // When
+        let updatedTheme = theme.toggle(newToggleStyle)
+
+        // Then
+        XCTAssertEqual(updatedTheme.toggleStyle.title.font, AdyenFonts.default.bodyEmphasized)
+        XCTAssertEqual(updatedTheme.toggleStyle.title.color, .red)
+        XCTAssertEqual(updatedTheme.toggleStyle.tintColor, .black)
```
atmamont, line 52: Let's add tests for every property that toggle style has to offer

neelSharma12: Added tests for all the properties :)

AdyenUI/Styles/AdyenToggleStyle.swift
```diff
@@ -0,0 +1,72 @@
...
+/// Contains the styling customization options for a switch item in a form.
+public struct AdyenToggleStyle {
+    
+    /// The title style.
+    internal var title = AdyenLabelStyle()
+    
+    /// The color of `onTintColor` of switch.
+    public var tintColor: UIColor?
+
+    public var backgroundColor: UIColor = .clear
```
nauaros: This ones lacks documentation

AdyenUI/Styles/AdyenToggleStyle.swift
```diff
@@ -0,0 +1,72 @@
...
+
+import UIKit
+
+/// Contains the styling customization options for a switch item in a form.
+public struct AdyenToggleStyle {
+    
+    /// The title style.
+    internal var title = AdyenLabelStyle()
+    
+    /// The color of `onTintColor` of switch.
```
nauaros, line 15: Is it possible to rephrase this line? What is `onTintColor` exactly?

atmamont: As a developer using SDK, I would assume it is UISwitch ON color. I also had to style the OFF state background, so we can possibly expect such requests in the future.

atmamont (approved):

nauaros (approved):

AdyenUI/Styles/AdyenToggleStyle.swift
```diff
@@ -0,0 +1,73 @@
+//
+// Copyright (c) 2025 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import UIKit
+
+/// Contains the styling customization options for a switch item in a form.
+public struct AdyenToggleStyle {
```
erenbesel: This shouldn't be public right? This struct will inherit the values from the main theme but not be exposed outside
All of the public variables, functions should be `package`

neelSharma12: Thanks for pointing this out. I have changed it to package.

AdyenUI/Styles/AdyenToggleStyle.swift
```diff
@@ -0,0 +1,73 @@
...
+
+    // The background color of the switch.
+    var backgroundColor: UIColor = .clear
+
+    var cornerRadius: CornerRounding = .fixed(AdyenUIConstants.defaultCornerRadius)
+    
+    /// Initializes the form switch item style.
+    ///
+    /// - Parameter title: The title style.
+    init(title: AdyenLabelStyle) {
```
erenbesel: missing access modifier keyword, means it will be internal. Do we want this internal or package?

AdyenUI/Styles/AdyenToggleStyle.swift
```diff
@@ -0,0 +1,73 @@
...
+    init() {}
+}
+
+// This extension adds the method chaining to the AdyenToggleStyle struct.
+extension AdyenToggleStyle {
+
+    /// Returns a new AdyenToggleStyle with the specified title.
+    /// - Parameter title: The title to set.
+    /// - Returns: A new `AdyenToggleStyle` instance.
+    func title(_ title: AdyenLabelStyle) -> AdyenToggleStyle {
```
erenbesel, line 40: Making it package reminded me, do we need these functions? These functions are for the configuration object to be chainable. But if it's not public, what is the purpose of these? Where are they used? Maybe the example code update should also be in this PR, not separate so we catch these things.

AdyenUI/AdyenTheme.swift
```diff
@@ -15,14 +15,17 @@ public struct AdyenTheme {
     // Available styles
     public var buttonStyle = AdyenButtonStyles()
     public var labelStyle = AdyenLabelStyle()
+    package var toggleStyle = AdyenToggleStyle()
 
-    // Initialize with a default ButtonStyle and LabelStyle if none is provided
-    public init(
+    // Initialize with a default ButtonStyle, LabelStyle and ToggleStyle if none is provided
+    package init(
```
erenbesel, line 21: Should this be package? Then I guess this is because we won't give the `.theme(_ : AdyenTheme)` option?

AdyenUI/Styles/AdyenToggleStyle.swift
neelSharma12: Will address this in upcoming PR.

AdyenUI/AdyenTheme.swift
neelSharma12: Will address this in upcoming PR.

AdyenUI/Styles/AdyenToggleStyle.swift
neelSharma12: Added package access specifier.

PR #2253, opened 2025-09-15
v6: New styling for textField

neelSharma12:
# Summary [Required]
This PR contains following changes 

-  Added new textfield style.
- Added tests 

## Ticket [Optional]
<ticket>
COSDK-635
</ticket>


AdyenUI/Styles/AdyenTextFieldStyle.swift
```diff
@@ -0,0 +1,150 @@
...
+    /// A default instance of AdyenTextFieldStyle.
+    public static let `default` = AdyenTextFieldStyle()
+
+    /// Initializes a new AdyenTextFieldStyle with default values.
+    public init(
+        backgroundColor: UIColor = AdyenColorScheme.default.background,
+        textColor: UIColor = AdyenColorScheme.default.text,
+        activeColor: UIColor = AdyenColorScheme.default.highlight,
+        errorColor: UIColor = AdyenColorScheme.default.destructive,
+        cornerRadius: CornerRounding = CornerRounding.fixed(12.0),
```
nauaros: We have a global constant for the corner radius? I was wondering if we can use it here instead of the 12.0

AdyenUI/Styles/AdyenTextFieldStyle.swift
```diff
@@ -0,0 +1,150 @@
...
+        self.borderColor = borderColor
+        self.borderWidth = borderWidth
+    }
+
+    public init() {
+        self.backgroundColor = AdyenColorScheme.default.background
+        self.textColor = AdyenColorScheme.default.text
+        self.activeColor = AdyenColorScheme.default.highlight
+        self.errorColor = AdyenColorScheme.default.destructive
+        self.cornerRadius = CornerRounding.fixed(12.0)
```
nauaros: Same here

nauaros (approved):

AdyenUI/Styles/AdyenTextFieldStyle.swift
neelSharma12: As discussed I will keep this as it follows the designs.

AdyenUI/Styles/AdyenTextFieldStyle.swift
neelSharma12: As discussed I will keep this as it follows the designs.

atmamont: Sorry to reopen this, do we have few corner radiuses across the design system so we can't define a single one?

neelSharma12: It would be ideal if we can apply same corner radius throughout but this we plan to discuss with Arjen soon, till then we are following the designs. We will keep you posted about this.

neelSharma12: I got the answer from design team today about this and we can apply same corner radius to all ui elements. So, I have made this change here.

Tests/UnitTests/AdyenTheme/AdyenThemeTests.swift
```diff
@@ -95,5 +95,174 @@ final class AdyenThemeTests: XCTestCase {
...
+        XCTAssertEqual(updatedTheme.textFieldStyle.cornerRadius, cornerRadius)
+    }
+    
+    func test_textFieldMethod_shouldUpdateBorderWidth() {
+        // Given
+        let theme = AdyenTheme()
+        
+        var newTextFieldStyle = AdyenTextFieldStyle()
+        
+        newTextFieldStyle.borderWidth = 2.0
```
atmamont: Using magic numbers makes code less readable. It's a good practice to use following or similar:

```swift
let expectedBorderWidth: CGFloat = 2.0
newTextFieldStyle.borderWidth = expectedBorderWidth
// the rest
XCTAssertEqual(updatedTheme.textFieldStyle.borderWidth, expectedBorderWidth)
```

The same applies to the rest of the test cases.

Tests/UnitTests/AdyenTheme/AdyenThemeTests.swift
```diff
@@ -95,5 +95,174 @@ final class AdyenThemeTests: XCTestCase {
         // Then
         XCTAssertEqual(updatedTheme.toggleStyle.cornerRadius, cornerRadius)
     }
-
+    
+    func test_textfieldMethod_shouldUpdateBackgroundColor() {
```
atmamont: This file grows, please consider creating separate set of tests placed in their own file for each semantic theme element (toggle, textfield, etc)

Tests/UnitTests/AdyenTheme/AdyenThemeTests.swift
```diff
@@ -95,5 +95,174 @@ final class AdyenThemeTests: XCTestCase {
...
+        let theme = AdyenTheme()
+        
+        var newTextFieldStyle = AdyenTextFieldStyle()
+        newTextFieldStyle.backgroundColor = AdyenColorScheme.default.primary
+        
+        // When
+        let updatedTheme = theme.textfield(newTextFieldStyle)
+        
+        // Then
+        XCTAssertEqual(updatedTheme.textFieldStyle.backgroundColor, AdyenColorScheme.default.primary)
```
atmamont: I didn't find any tests covering default values for a case when merchant doesn't tinker theming at all. Let's add them if we don't have them yet?

AdyenUI/Styles/AdyenTextFieldStyle.swift
```diff
@@ -0,0 +1,150 @@
...
+        self.textColor = AdyenColorScheme.default.text
+        self.activeColor = AdyenColorScheme.default.highlight
+        self.errorColor = AdyenColorScheme.default.destructive
+        self.cornerRadius = CornerRounding.fixed(AdyenUIConstants.defaultCornerRadius)
+        self.borderColor = AdyenColorScheme.default.outline
+        self.borderWidth = AdyenUIConstants.defaultBorderWidth
+    }
+}
+
+// MARK: - Method Chaining for AdyenTextFieldStyle Properties
```
atmamont: This is LLM generated comment, no value and all capitalized, please remove it.

AdyenUI/Styles/AdyenTextFieldStyle.swift
```diff
@@ -0,0 +1,150 @@
...
+
+// This extension adds the method chaining to the AdyenTextFieldStyle struct.
+extension AdyenTextFieldStyle {
+    
+    /// Returns a new style with the specified background color.
+    /// - Parameter backgroundColor: The color to set.
+    /// - Returns: A new `AdyenTextFieldStyle` instance.
+    public func with(backgroundColor: UIColor) -> AdyenTextFieldStyle {
```
atmamont: This naming is not matching neither of previous elements and should be reconsidered.
So far there is a variety of patterns used:
- `AdyenLabelStyle.font(:)`
- `AdyenButtonStyle.primary(:)`
- `AdyenToggleStyle.tintColor(:)`

Adding new `with` pattern is not great, and makes it more complex and fragmented.

It is fine to use LLM, but it requires a lot of attention and thinking through to make good use of it.

AdyenUI/Styles/AdyenTextFieldStyle.swift
```diff
@@ -0,0 +1,150 @@
...
+    public func with(cornerRadius: CornerRounding) -> AdyenTextFieldStyle {
+        var newStyle = self
+        newStyle.cornerRadius = cornerRadius
+        return newStyle
+    }
+    
+    /// Returns a new style with the specified border color.
+    /// - Parameter borderColor: The color to set.
+    /// - Returns: A new `AdyenTextFieldStyle` instance.
+    public func with(borderColor: UIColor) -> AdyenTextFieldStyle {
```
atmamont: I find it a bit confusing from a merchant perspective, it's not clear if this `borderColor` would be always applied including active and non-active states, as we already have a dedicated semantic color for `activeColor`, which can override default `borderColor`.

Tests/UnitTests/AdyenTheme/AdyenThemeTests.swift
```diff
@@ -95,5 +95,174 @@ final class AdyenThemeTests: XCTestCase {
...
+        let theme = AdyenTheme()
+        
+        var newTextFieldStyle = AdyenTextFieldStyle()
+        newTextFieldStyle.activeColor = AdyenColorScheme.default.highlight
+        
+        // When
+        let updatedTheme = theme.textfield(newTextFieldStyle)
+        
+        // Then
+        XCTAssertEqual(updatedTheme.textFieldStyle.activeColor, AdyenColorScheme.default.highlight)
```
atmamont: Are we ever checking if calling `theme.textfield(newTextFieldStyle)` or any other similar updater preserves values that were set before?
For example, we could check if other theme styles are safe after we update textField style.

atmamont (changes requested):

atmamont:
Would it be possible to add a bridge that applies new styling to existing form text input element? It's fine to do it separately if this would significantly increase the PR size.

neelSharma12:
> Would it be possible to add a bridge that applies new styling to existing form text input element? It's fine to do it separately if this would significantly increase the PR size.

I will be doing this in a separate PR.

AdyenUI/AdyenTheme.swift
```diff
@@ -17,15 +17,18 @@ public struct AdyenTheme {
     public var buttonStyles = AdyenButtonStyles()
     public var labelStyle = AdyenLabelStyle()
     package var toggleStyle = AdyenToggleStyle()
+    public var textFieldStyle = AdyenTextFieldStyle()
```
erenbesel: package

AdyenUI/AdyenUIConstants.swift
```diff
@@ -8,5 +8,6 @@ public enum AdyenUIConstants {
 
     public static let defaultCornerRadius = 14.0
     public static let submitButtonHeight = 52.0
+    public static let defaultBorderWidth = 1.5
```
erenbesel: all these values are public, should they be?

AdyenUI/Styles/AdyenTextFieldStyle.swift
```diff
@@ -0,0 +1,148 @@
...
+    package var placeholderText: AdyenLabelStyle?
+
+    /// The color of the background.
+    public var backgroundColor: UIColor
+
+    /// The color of the text.
+    public var textColor: UIColor
+
+    /// The color of the text when the element is active.
+    public var activeColor: UIColor
```
erenbesel: I see half the properties package, the rest public. Which one should it be?

AdyenUI/Styles/AdyenTextFieldStyle.swift
```diff
@@ -0,0 +1,148 @@
...
+    )
+    
+    /// The text field's placeholder text style.
+    package var placeholderText: AdyenLabelStyle?
+
+    /// The color of the background.
+    public var backgroundColor: UIColor
+
+    /// The color of the text.
+    public var textColor: UIColor
```
andriim-m, line 33: I noticed we have two properties - `text` and `textColor`
There is a conflict, it is better to give them better names that avoid conflicts.

AdyenUI/Styles/AdyenTextFieldStyle.swift
neelSharma12: We do not want to expose `text`, `placeholderText` and `title` property to the merchant. We can change this later if needed.

atmamont: We need to get to to samples stage to see how all of this works in combination together to ensure everything is consistent. For example, as a merchant, I wonder how placeholder or title text color (defined in `AdyenLabelStyle`) would match with the `borderColor` that we allow to change. Without passing the accordingly adjusted `AdyenLabelStyle` this could cause color mismatch. 

Which leads to the question - should we expose the `borderColor`, `textColor` and similar elements that can cause misuse and design inconsistency?

Tests/UnitTests/AdyenTheme/AdyenThemeTextFieldTests.swift
```diff
@@ -0,0 +1,198 @@
...
+        // Then
+        XCTAssertEqual(updatedTheme.textFieldStyle.errorColor, AdyenColorScheme.default.textOnDestructive)
+    }
+    
+    func test_textfieldMethod_shouldUpdateBorderColor() {
+        // Given
+        let theme = AdyenTheme()
+        
+        var newTextFieldStyle = AdyenTextFieldStyle()
+        newTextFieldStyle.borderColor = AdyenColorScheme.default.outline
```
atmamont: `AdyenColorScheme.default.outline` is a default color, this test doesn't test anything.

Tests/UnitTests/AdyenTheme/AdyenThemeTextFieldTests.swift
```diff
@@ -0,0 +1,198 @@
...
+        let theme = AdyenTheme()
+        
+        var newTextFieldStyle = AdyenTextFieldStyle()
+        newTextFieldStyle.activeColor = AdyenColorScheme.default.highlight
+        
+        // When
+        let updatedTheme = theme.textfield(newTextFieldStyle)
+        
+        // Then
+        XCTAssertEqual(updatedTheme.textFieldStyle.activeColor, AdyenColorScheme.default.highlight)
```
atmamont: `AdyenColorScheme.default.highlight` is a default color for `activeColor`

atmamont: As one of the ways to exclude occasional matching of default colors and to ensure test is useful, I
d suggest using simple `UIColor` predefined colors that rarely used in designs like `.pink` or `.purple`.
And as my previous comment saids, let's use `let expectedValue = UIColor.pink` to make reading and copy-pasting easier.

Tests/UnitTests/AdyenTheme/AdyenThemeTextFieldTests.swift
```diff
@@ -0,0 +1,198 @@
...
+        
+        // When
+        let updatedTheme = theme.textfield(newTextFieldStyle)
+        
+        // Then
+        XCTAssertEqual(updatedTheme.textFieldStyle.cornerRadius, cornerRadius)
+    }
+    
+    func test_textFieldMethod_shouldUpdateBorderWidth() {
+        let expectedBorderWidth: CGFloat = 2.0
```
atmamont: 👍

atmamont: Just a thought: we could use `AdyenUIConstants.defaultBorderWidth + 1` as a value to avoid re-using the default value by accident.

Tests/UnitTests/AdyenTheme/AdyenThemeTextFieldTests.swift
```diff
@@ -0,0 +1,198 @@
...
+        
+        var newTextFieldStyle = AdyenTextFieldStyle()
+        
+        newTextFieldStyle.placeholderText = newPlaceholderTextStyle
+        
+        // When
+        let updatedTheme = theme.textfield(newTextFieldStyle)
+        
+        // Then
+        XCTAssertEqual(updatedTheme.textFieldStyle.placeholderText?.color, .red)
```
atmamont: It is better to validate against the values you passed. This will make this code easier to maintain and change in the future.
`XCTAssertEqual(updatedTheme.textFieldStyle.placeholderText?.color, newPlaceholderTextStyle.color)`

Tests/UnitTests/AdyenTheme/AdyenThemeLabelTests.swift
```diff
@@ -0,0 +1,225 @@
...
+        let expectedTextAlignment: NSTextAlignment = .left
+        let theme = AdyenTheme()
+        
+        let newLabelStyle = AdyenLabelStyle(textAlignment: expectedTextAlignment)
+        
+        // When
+        let updatedTheme = theme.label(newLabelStyle)
+        
+        // Then
+        XCTAssertEqual(updatedTheme.textFieldStyle.backgroundColor, AdyenColorScheme.default.background)
```
atmamont: I see we now have good coverage but have some repetitions.
Do you think we can extract some repetitive assertions into helper methods?

Something like this:

```swift
func expect(textFieldStyle: AdyenTextFiedStyle, toMatch expectedStyle: AdyentextFiedStyle, file: StaticString = #file, line: UInt = #line) {
    XCTAssertEqual(textFieldStyle.backgroundColor, expectedStyle.backgroundColor)
   // and so on for the rest
}
```

Having these helpers for each style element (toggle, label, textField) would allow to check all at one and reduce LOC

neelSharma12: I have added helper methods for assertions

atmamont (approved):

PR #2262, opened 2025-09-30
COSDK-637: ActionComponent impl. in DropIn V6

nauaros:
## Summary

This PR introduces action handling within the new DropIn architecture.  
The `DropInComponent` has been restructured to adopt a modular, router-driven design, reducing responsibilities within the component itself and improving the overall flow of data and control. The result is a cleaner separation of concerns, simpler navigation, and more predictable handling of payment actions across the DropIn experience.

## Highlights

- **Refined DropInComponent** – `DropInComponent` delegates action handling to `DropInRouter`, removing direct dependency on `AdyenActionComponent` and clarifying its core responsibilities.  
- **Router-Oriented Architecture** – Routers now coordinate control flow across `ComponentContainer`, `PaymentMethodList`, `PreselectedPaymentMethod`, and `Root`. Protocols and initializers have been restructured for consistency and modularity.  
- **Streamlined Navigation** – Replaced the custom `DropInNavigationController` with `UINavigationController`, simplifying the navigation stack and aligning with UIKit conventions.  
- **Consistent Delegate Propagation** – `PaymentComponentDelegate` and `ActionComponentDelegate` methods flow uniformly through the router chain, ensuring predictable behavior for submissions, cancellations, errors, and external app transitions.  
- **UI/UX Enhancements** – Navigation titles are now dynamic, large titles are enabled by default, and cancel actions are properly integrated, resulting in a more modern and consistent user interface.  

# Demo
 

https://github.com/user-attachments/assets/e208723b-82cd-4f64-8a4e-9370c6deb5ef

## Ticket [Optional]
<ticket>
COSDK-637
</ticket>

## Checklist [Required]

- [ ] Tested changes locally


AdyenDropIn/Modules/ComponentContainer/ComponentContainerRouter.swift
```diff
@@ -0,0 +1,61 @@
...
+// Copyright (c) 2025 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Adyen
+import Foundation
+import UIKit
+
+internal protocol ComponentContainerRouterDelegate: AnyObject {
```
erenbesel: I see the `...RouterDelegate` protocol methods appear exactly on the `...RouterProtocol` protocol. In all following protocols it is the same. Do we need this duplication? I couldn't understand why they are needed on the second protocol `ComponentContainerRouterProtocol`.

Also the protocol names are too close we need to find something else :D

Adyen.xcodeproj/project.pbxproj
```diff
@@ -698,13 +698,14 @@
...
 		C9025A4D2DF31E4700E0D191 /* PreselectedPaymentMethodAssembler.swift in Sources */ = {isa = PBXBuildFile; fileRef = C9025A4C2DF31E4700E0D191 /* PreselectedPaymentMethodAssembler.swift */; };
 		C9037F102D563DCD003A5154 /* CardScannerViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = C9037F0F2D563DCD003A5154 /* CardScannerViewModel.swift */; };
 		C9037F122D5A1976003A5154 /* CardScannerOverlayView.swift in Sources */ = {isa = PBXBuildFile; fileRef = C9037F112D5A1976003A5154 /* CardScannerOverlayView.swift */; };
 		C9037F142D5A19BD003A5154 /* ROIView.swift in Sources */ = {isa = PBXBuildFile; fileRef = C9037F132D5A19BD003A5154 /* ROIView.swift */; };
 		C9037F162D5A1AC0003A5154 /* CardScannerViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = C9037F152D5A1AC0003A5154 /* CardScannerViewController.swift */; };
+		C90494542E8FE5D2007C4EB5 /* PaymentComponentRouting.swift in Sources */ = {isa = PBXBuildFile; fileRef = C90494532E8FE5D2007C4EB5 /* PaymentComponentRouting.swift */; };
+		C90494562E8FF2CE007C4EB5 /* LoadControllable.swift in Sources */ = {isa = PBXBuildFile; fileRef = C90494552E8FF2CE007C4EB5 /* LoadControllable.swift */; };
```
atmamont, line 716: This name is great for reflecting the purpose of this interface 👍

Adyen/Core/Core Protocols/Component.swift
```diff
@@ -17,7 +17,7 @@ extension Component {
     /// - Parameter completion: The block to execute after the component finalizes its activity.
     /// Use of this block is recommended for ApplePayComponent. You may specify nil for this parameter.
     public func finalizeIfNeeded(with success: Bool, completion: (() -> Void)?) {
```
atmamont, line 19: I'd eventually rename this as well and fully get rid of the 'ifNeeded'

AdyenDropIn/Modules/ComponentContainer/ComponentContainerAssembler.swift
```diff
@@ -45,19 +45,33 @@ internal struct ComponentContainerAssembler: ComponentContainerAssemblerProtocol
...
             component: component,
             context: context,
-            delegate: router,
             configuration: configuration,
             cardComponentDelegate: cardComponentDelegate,
             partialPaymentDelegate: partialPaymentDelegate
         )
-        let view = ComponentContainerViewController(viewModel: viewModel)
-        router.view = view
+        let viewController = resolveRouterViewController(for: component, viewModel: viewModel)
```
atmamont: The name `resolveRouterViewController` is a bit confusing, the left part `let viewController = ` suggests it returns view controller, but the `router` part then makes no sense. We do not expect a router in this architecture to have its own view controller, do we?

AdyenDropIn/Modules/ComponentContainer/ComponentContainerAssembler.swift
```diff
@@ -45,19 +45,33 @@ internal struct ComponentContainerAssembler: ComponentContainerAssemblerProtocol
...
     }
+    
+    private func resolveRouterViewController(
+        for component: PresentableComponent,
+        viewModel: ComponentContainerViewModelProtocol
+    ) -> UIViewController {
+        if let alertController = component.viewController as? UIAlertController {
+            return alertController
+        } else {
+            return ComponentContainerViewController(viewModel: viewModel)
```
atmamont: From what I see here, `resolveComponentContainerViewController` name could make this method signature explicit and remove the confusion I mentioned above.

AdyenDropIn/Modules/ComponentContainer/ComponentContainerRouter.swift
```diff
@@ -0,0 +1,64 @@
...
+import Foundation
+import UIKit
+
+internal protocol ComponentContainerRouterListener: AnyObject {
+    func didSubmit(_ data: PaymentComponentData, from component: any PaymentComponent)
+    func didFail(with error: any Error, from component: any PaymentComponent)
+    func didCancel(component: any PaymentComponent)
+}
+
+internal protocol ComponentContainerRouting: AnyObject, PaymentComponentRouting {
```
atmamont: While me might not introduce "presenter" as a separate architecture block yet, it can be helpful to reflect the primary responsibility of this protocol in its name - `ComponentContainerPresenting`. 
The current code relationships could stay the same, the router would conform to this protocol, but it will make extracting the presentation logic out of the router so much easier (in the case it becomes complex and goes beyond router responsibilities, like managing navigation controller stack or handling custom transition animations).

AdyenDropIn/Modules/ComponentContainer/ComponentContainerRouter.swift
```diff
@@ -0,0 +1,64 @@
...
+    func present(_ viewController: UIViewController, animated: Bool)
+}
+
+internal class ComponentContainerRouter: Router, ComponentContainerRouting {
+
+    // MARK: - Properties
+
+    internal let rootViewController: UIViewController
```
atmamont: Do you think `containerViewController` could explain this better?

AdyenDropIn/Modules/ComponentContainer/ComponentContainerRouter.swift
```diff
@@ -0,0 +1,64 @@
...
+    ) {
+        self.rootViewController = viewController
+        self.loadable = loadable
+        self.listener = listener
+    }
+
+    // MARK: - ComponentContainerRouting
+    
+    internal func present(_ viewController: UIViewController, animated: Bool) {
+        rootViewController.present(viewController, animated: animated)
```
atmamont: Oh now I see this is indeed not a `container` but `root` that handles presenting. 
What about dismissing logic, it's good practice to always mirror `present` with `dismiss`, otherwise the control flow becomes obscure.

AdyenDropIn/Modules/ComponentContainer/ComponentContainerViewController.swift
```diff
@@ -61,22 +59,8 @@ internal final class ComponentContainerViewController: UIViewController {
...
-            action: #selector(cancelButtonTapped)
-        )
-        navigationItem.rightBarButtonItem = cancelButton
-    }
-
-    @objc private func cancelButtonTapped() {
-        navigationController?.dismiss(animated: true)
+        navigationItem.title = componentView.title
+        navigationItem.largeTitleDisplayMode = .always
+        navigationController?.navigationBar.prefersLargeTitles = true
```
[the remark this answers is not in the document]
atmamont: Please consider we might be moving configuration to server eventually in the future updates.

AdyenDropIn/Modules/ComponentContainer/ComponentContainerViewModel.swift
```diff
@@ -66,11 +52,13 @@ internal class ComponentContainerViewModel: ComponentContainerViewModelProtocol
...
     internal func cancel() {
-        component.cancelIfNeeded()
+        component.cancel()
 
         if let component = (component as? PaymentComponent) {
-            delegate?.didCancel(component: component)
+            router?.cancel(component: component)
         }
+        
+        stopLoading()
```
atmamont: Do we need it? I see we call `component.cancel()` and then canceling the router flow (I assume we also dismiss the UI completely) before we stop loading. Wouldn't the UI including the activity indicator be dismissed by this moment?

atmamont: And one more idea/thought. If we already have the case when view model can't fully control its own state and should delegate "stop loading" to someone else, can we take full control over loading from view model and give it to router instead?
It would be much easier to read, maintain and test the code that has start/stop loading logic in one place.

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListRouter.swift
```diff
@@ -8,94 +8,106 @@ import Adyen
...
-
-        componentContainerRouter.start()
+        self.childRouter = componentContainerRouter
 
         let componentContainerViewController = componentContainerRouter.rootViewController
 
         // TODO: - Invert `requiresModalPresentation` logic or remove it fully.
         if component.requiresModalPresentation {
-            view?.navigationController?.pushViewController(componentContainerViewController, animated: true)
+            viewController.navigationController?.pushViewController(componentContainerViewController, animated: true)
```
atmamont, line 68: This is an example of how router now has too much responsibilities (knowing about viewController structure and how to properly manage navigation controller) apart of its main duty and this could go to a "presenter" in the future.

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListRouter.swift
```diff
@@ -8,94 +8,106 @@ import Adyen
...
+    internal func submit(_ data: PaymentComponentData, from component: any PaymentComponent) {
+        listener?.didSubmit(data, from: component)
+    }
+    
+    internal func fail(with error: any Error, from component: any PaymentComponent) {
+        listener?.didFail(with: error, from: component)
+    }
+    
+    internal func cancel(component: any PaymentComponent) {
+        loadable.stopLoading()
```
atmamont: Interesting to notice this router manages loading state and doesn't delegate it to a view model, unlike the previous case.

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewController.swift
```diff
@@ -50,13 +48,23 @@ internal class PaymentMethodListViewController: UIViewController {
     }
 
     private func setupNavigationItem() {
-        navigationItem.title = "2"
+        navigationItem.title = viewModel.paymentMethodListView.title
+        navigationItem.largeTitleDisplayMode = .always
+        navigationController?.navigationBar.prefersLargeTitles = true
+        
+        setupCancelButton()
```
atmamont, line 55:
```suggested
setupNavigationCancelButton()
```

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewModel.swift
```diff
@@ -92,3 +89,49 @@ extension PaymentMethodListViewModel: PaymentMethodListComponentDelegate {
         // TODO: - Logic to delete stored payment method
     }
 }
+
+// MARK: - PaymentComponentDelegate
+
+extension PaymentMethodListViewModel: PaymentComponentDelegate {
+    
+    internal func didSubmit(
```
atmamont, line 109: Main responsibility of `PaymentMethodListViewModel` is to manage payment methods list, not to do any data enrichment/conversion. This logic doesn't belong here.

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewModel.swift
```diff
@@ -92,3 +89,49 @@ extension PaymentMethodListViewModel: PaymentMethodListComponentDelegate {
...
+        _ data: PaymentComponentData,
+        from component: any PaymentComponent
+    ) {
+        let checkoutAttemptId = component.context.analyticsProvider?.checkoutAttemptId
```
atmamont, line 115: This line performs hard lock on component and context internal structure.
The simplest way to hide implementation details is adding a helper function that does this internally, capturing this relationship once:
`let checkoutAttemptId = component.checkoutAttemptId()`

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListRouter.swift
```diff
@@ -8,94 +8,106 @@ import Adyen
...
-    func didFail(with error: any Error) {
-        delegate?.didFail(with: error)
+    func didFail(with error: any Error, from component: any Adyen.PaymentComponent) {
+        listener?.didFail(with: error, from: component)
     }
     
-    func didCancel(component: any PaymentComponent) {
-        delegate?.didCancel(component: component)
+    func didCancel(component: any Adyen.PaymentComponent) {
+        loadable.stopLoading()
```
atmamont: For didFail we call `stopLoading` from the view model itself, for cancellation we call it from the router. Can we unify the flow and keep them together?

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodRouter.swift
```diff
@@ -8,73 +8,92 @@ import Adyen
...
     internal func proceed(with paymentComponent: any PresentableComponent) {
-        // TODO: - Handle logic with preselected payment method
-//        delegate?.didProceed(with: paymentComponent)
-
         let componentContainerRouter = componentContainerAssembler.resolveComponentContainerRouter(
             for: paymentComponent,
             delegate: self
         )
-        componentContainerRouter.start()
+        self.componentContainerRouter = componentContainerRouter
```
atmamont: I've seen `childRouter` in the code. Should we abstract parent/child router relationships into a protocol? It might be helpful for other developers to follow the new architecture pattern.

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodRouter.swift
```diff
@@ -8,73 +8,92 @@ import Adyen
...
+    internal func submit(_ data: PaymentComponentData, from component: any PaymentComponent) {
+        listener?.didSubmit(data, from: component)
+    }
+    
+    internal func fail(with error: any Error, from component: any PaymentComponent) {
+        listener?.didFail(with: error, from: component)
+    }
+    
+    internal func cancel(component: any PaymentComponent) {
+        rootViewController.dismiss(animated: true)
```
atmamont: Let's extract this to a local `dismiss()` method?

AdyenDropIn/Modules/ComponentContainer/ComponentContainerViewController.swift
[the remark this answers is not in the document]
erenbesel: Nice comment, worth thinking about.

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListRouter.swift
```diff
@@ -8,94 +8,106 @@ import Adyen
...
 }
 
-internal class PaymentMethodListRouter: PaymentMethodListRouterProtocol {
+internal class PaymentMethodListRouter: Router, PaymentMethodListRouting {
 
     // MARK: - Properties
 
-    internal weak var delegate: PaymentMethodListRouterDelegate?
+    private let viewController: UIViewController
+    private let loadable: LoadControllable
```
erenbesel: Is this the component? Property name can be what this object actually is or similar to it. Loadable is too vauge I think.

AdyenDropIn/Modules/Root/DropInRouter.swift
```diff
@@ -9,123 +9,164 @@ import AdyenNetworking
...
-    func start()
+    func stopLoading()
 }
 
-internal protocol DropInRouterProtocol: Router {
-    var delegate: DropInRouterDelegate? { get set }
-    var rootViewController: UIViewController { get }
-    func start()
+extension Router {
+    func stopLoading() { /* Optional implementation */ }
```
atmamont: Is the goal of this extension to remove requirement to implement `stopLoading` for every instance of `Router`?

AdyenDropIn/Modules/Root/DropInRouter.swift
```diff
@@ -9,123 +9,164 @@ import AdyenNetworking
...
 }
 
-internal class DropInRouter: DropInRouterProtocol {
+internal protocol DropInRouting: Router, AnyObject {
+    func handle(action: Action)
+    func present(_ viewController: UIViewController, animated: Bool)
```
atmamont: `present` and `openExternalApplication` are good candidates for interface segregation and could go to a separate interface.

AdyenDropIn/Modules/Root/DropInRouter.swift
```diff
@@ -9,123 +9,164 @@ import AdyenNetworking
...
+    
+    internal func handle(action: Action) {
+        viewModel.handle(action: action)
     }
-
-    internal var rootViewController: UIViewController {
-        navigationController
+    
+    internal func present(_ viewController: UIViewController, animated: Bool) {
+        childRouter?.rootViewController.present(viewController, animated: animated)
```
atmamont: Router should know nothing about internal implementation details of its child router. 

```suggested
childRouter.present(viewController, animated: animated)
```

AdyenDropIn/Modules/Root/DropInRouter.swift
```diff
@@ -9,123 +9,164 @@ import AdyenNetworking
...
+    }
+    
+    internal func openExternalApplication(component: any ActionComponent) {
+        listener?.didOpenExternalApplication(component: component)
```
atmamont: Method name is action, the call name itself is a notification, let's align.

AdyenDropIn/Modules/Root/DropInRouter.swift
```diff
@@ -9,123 +9,164 @@ import AdyenNetworking
...
+    }
+    
+    internal func provide(_ data: ActionComponentData, from component: any ActionComponent) {
+        listener?.didProvide(data, from: component)
```
atmamont: Same semantic conflict here and below.

AdyenDropIn/Modules/Root/DropInRouter.swift
```diff
@@ -9,123 +9,164 @@ import AdyenNetworking
...
+                delegate: self,
                 component: paymentComponent,
-                title: "Preselected PM"
+                title: "DropIn V6"
             )
-            preselectedPaymentMethodRouter.delegate = self
-            self.preselectedPaymentMethodRouter = preselectedPaymentMethodRouter
-            return preselectedPaymentMethodRouter.rootViewController
+            self.childRouter = preselectedPaymentMethodRouter
+            let preselectedPaymentMethodViewController = preselectedPaymentMethodRouter.rootViewController
```
atmamont, line 80: It is better to inject `rootViewController` in line 116 (in the assembly layer) instead of retrieving it from the router, then you'll have ability to test the router<->viewController interactions.

AdyenDropIn/Modules/Root/DropInRouter.swift
```diff
@@ -9,123 +9,164 @@ import AdyenNetworking
...
         case let .component(paymentComponent):
-            // TODO: - Handle standalone component case
-//            let componentView = componentContainerAssembler.resolveContainerView(for: paymentComponent)
-//            return componentView
-            return UIViewController()
+            let componentContainerRouter = componentContainerAssembler.resolveComponentContainerRouter(
+                for: paymentComponent,
+                delegate: self
+            )
+            self.childRouter = componentContainerRouter
```
atmamont, line 89: This unnecessary internal detail implementation detail could be abstracted to a protocol method so it looks like:

```swift
parentRouter.attach(childRouter)
```

AdyenDropIn/Modules/Root/DropInRouter.swift
```diff
@@ -9,123 +9,164 @@ import AdyenNetworking
...
-    func showAllPaymentMethods() {
-        let paymentMethodListRouter = paymentMethodListAssembler.resolvePaymentMethodListRouter()
-        self.paymentMethodListRouter = paymentMethodListRouter
-        paymentMethodListRouter.delegate = self
-        paymentMethodListRouter.start()
+extension DropInRouter: PreselectedPaymentMethodRouterListener {
         
-        navigationController.present(paymentMethodListRouter.rootViewController, animated: true)
+    func didPresentPaymentMethodList() {
+        let paymentMethodListRouter = paymentMethodListAssembler.resolvePaymentMethodListRouter(delegate: self)
```
atmamont: Method name `didPresentPaymentMethodList` suggests this is a notification, actions inside indicate this is a response to call for action. Should this method be named `presentPaymentMethodList` instead?

AdyenDropIn/Modules/Root/DropInRouter.swift
```diff
@@ -9,123 +9,164 @@ import AdyenNetworking
...
+        self.childRouter = paymentMethodListRouter
+        rootViewController.present(paymentMethodListRouter.rootViewController, animated: true)
```
atmamont: Exposing internal implementation details of `paymentMethodListRouter` creates hard dependency and locks its implementation making any future refactors hard.
It's hard to come up on the spot, but think of presenting logic being a responsibility of a method I mentioned before:

`attach(childRouter)`

It could be fine unless we have an edge case when the child router should be attached but not presented, like no-UI module or something like this.

AdyenDropIn/Modules/Root/DropInRouter.swift
```diff
@@ -9,123 +9,164 @@ import AdyenNetworking
...
 
-    func paymentMethodListDidCancel(completion: (() -> Void)?) {
-        navigationController.presentingViewController?.dismiss(animated: true)
-        preselectedPaymentMethodRouter = nil
-        paymentMethodListRouter = nil
+extension DropInRouter: PaymentMethodListRouterListener, ComponentContainerRouterListener {
+        
+    func didDismiss(completion: (() -> Void)?) {
+        // TODO: - Decide wether dismissal this logic belongs to dropIn or merchant's side
+        rootViewController.presentingViewController?.dismiss(animated: true)
```
atmamont: Having double-dot syntax captures implementation details and makes future refactors hard.
At this moment it looks like a candidate to moved into a dedicated Presenter. Otherwise it's hardly possible to write clean tests.

A line below nullifies the `childRouter`, so please think about abstracting this logic to a `detach(childRouter)` interface.

AdyenDropIn/Modules/Root/DropInViewModel.swift
```diff
@@ -54,11 +59,57 @@ internal class DropInViewModel: DropInViewModelProtocol {
...
 }
 
+// MARK: - PresentationDelegate
+
 extension DropInViewModel: PresentationDelegate {
 
-    internal func present(component: any Adyen.PresentableComponent) {
-        // TODO: -
+    internal func present(component: any PresentableComponent) {
+        let componentViewController = component.viewController
```
atmamont: If we have more than one occurrence of such logic, maybe it makes sense to delegate `viewController` retrieval to the router itself?
So it could become 
```swift
router?.present(component, animated: true)
```

AdyenDropIn/DropInComponent.swift
```diff
@@ -33,7 +33,16 @@ public final class DropInComponent: NSObject,
 
     // MARK: - Properties
 
-    private let dropInRouter: DropInRouterProtocol
+    private lazy var router: DropInRouting = {
```
atmamont: Should it this be internal or package to enable mocking? Ideally it should be injected from the assembly layer.

AdyenDropIn/DropInComponent.swift
```diff
@@ -134,8 +131,8 @@ public final class DropInComponent: NSObject,
 
     // MARK: - Presentable Component Protocol
 
-    public lazy var viewController: UIViewController = {
-        dropInRouter.rootViewController
+    public private(set) lazy var viewController: UIViewController = {
+        router.rootViewController
```
atmamont, line 138: What's the role of DropInComponent in our new architecture? 
I see this component doesn't own a view controller, instead proxies the router property. Maybe something to rethink, or at least draw "component - view controller - router - view model" relationships and optimize it.

Tests/SnapshotTests/Components/PayToComponentUITests.swift
```diff
@@ -437,7 +437,7 @@ class PayToComponentUITests: XCTestCase {
             XCTAssertEqual(details.shopperName?.firstName, "test")
             XCTAssertEqual(details.shopperName?.lastName, "lastname")
 
-            sut.stopLoadingIfNeeded()
+            sut.stopLoading()
```
atmamont, line 440: There are no new tests added, which is fine while the new architecture approach is jelly. However, tests written consciously that capture the ideas behind the architecture could be helpful to understand the expectations of its components. If written against expectations and interfaces and not capturing a lot of implementation details, they can still stay robust even if we change internal implementation details.

atmamont (commented):
Great job Naufal, thanks for taking the risk and making such a bold move.
There are plenty comments from my side, please treat them as an invitation for discussion and a feedback on new approach ease of use and understanding.

Nothing of this is a blocker to merge this PR.

However, it would be great if we can improve clarity, readability and add important tests that capture architecture expectations and main components interactions between router, view model, view controller and a component.

This will lay a solid foundation for all future refactors, this is the reason I'd invest more time now and less later.

AdyenDropIn/DropInComponent.swift
nauaros: For mocking, its a good idea to make it `internal private(set)`

AdyenDropIn/DropInComponent.swift
nauaros: The idea of the `DropInComponent` is to act as the input / output between the merchant and the internal DropIn implementation. It's all that merchants know.

Like the rest of the components, it exposes a view controller. This keeps the same pattern we currently follow.
How do you think we could optimize it?

atmamont: Even without changing the architecture, the "component" should know nothing about the router. Minimum we can do here is to pass the view controller as a dependency.
We also have a viewModel which feels like a better candidate for communications with a router

AdyenDropIn/Modules/ComponentContainer/ComponentContainerAssembler.swift
nauaros: The difficulty comes from the fact that alert controllers are displayed from the payment method list. These alert controllers are not meant to be treated like regular standalone components. I need to find a better way of handling this edge case, but for now we can continue with this.

I replaced the naming with your suggested one. Makes it more obvious.

AdyenDropIn/Modules/ComponentContainer/ComponentContainerRouter.swift
nauaros: I want to keep a consistency across routers. Any router conforming to `Router` needs to have a `rootViewController`. We can revisit the naming tho.

AdyenDropIn/Modules/ComponentContainer/ComponentContainerViewModel.swift
nauaros: Yeah, unfortunately only components know they start loading but as we delegate this on merchants or sessions, the stopLoading needs to come from the outside. It also annoys me but all async operations are handled by a third party.

nauaros: Regarding the `stopLoading`, you are raising a good point. Since components are not fully deallocated, we need to make sure any loading is stopped. It will be redundant when we switch to a component on demand creation. That will be when the payment method list is refactored.

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewModel.swift
```diff
@@ -52,11 +50,18 @@ internal class PaymentMethodListViewModel: PaymentMethodListViewModelProtocol {
...
     }
 
     // MARK: - Private
+    
+    private func startLoading(for component: any PaymentComponent) {
+        paymentMethodListComponent.startLoading(for: component)
+    }
+    
+    private func stopLoading() {
+        paymentMethodListComponent.cancel()
```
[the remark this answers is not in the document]
atmamont: This is a good point, keeping semantic contract promise is important to reduce obscurity and to prevent possible misuse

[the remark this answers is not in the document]
nauaros: Yes, this is a good point. I misused the cancel method. It should only be used by merchants. Not internally.

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewModel.swift
```diff
@@ -52,42 +53,72 @@ internal class PreselectedPaymentMethodViewModel: PreselectedPaymentMethodViewMo
...
-    private func startLoading() {
+    
+    private func startLoading(for component: PaymentComponent) {
         preselectedPaymentMethodComponent.startLoading(for: component)
     }
-
+    
     private func stopLoading() {
-        preselectedPaymentMethodComponent.stopLoadingIfNeeded()
+        preselectedPaymentMethodComponent.cancel()
```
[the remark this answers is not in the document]
nauaros: Changed! Good point

AdyenDropIn/Modules/Root/DropInRouter.swift
```diff
@@ -53,79 +41,86 @@ internal class DropInRouter: DropInRouterProtocol {
...
+        let actionViewController = component.viewController
+        let viewControllerToPresent: UIViewController
+        
+        if actionViewController is SFSafariViewController {
+            viewControllerToPresent = actionViewController
+        } else {
+            viewControllerToPresent = ActionWrapperViewController(
+                rootViewController: component.viewController
+            ) { [weak self] in
+                // TODO: - Handle action cancellation - stopLoading in component
```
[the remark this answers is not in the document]
nauaros: Yes! To be addressed in a follow up PR

PR #2267, opened 2025-10-10
Refactor card component styling

atmamont:
# Summary

This change serves a purpose of collecting all styling code for card component screen in a single place, making small improvement along the way: 
- aligning with the current design 
- supporting new visual states
- code cleanup
- preparations to pass `AdyenTheme` as a dependency to `FormItemView` and remove `style` from the `FormItem`
- added unit tests for old and new styles to `NSAttributedString` conversions

# Demo 

| Normal state | Error state |
| -------- | -------- |
|  image | image | 

## Checklist [Required]

- [ ] Tested changes locally


AdyenCard/Form/FormCardSecurityCodeItemView.swift
```diff
@@ -22,8 +22,8 @@ internal final class FormCardSecurityCodeItemView: FormTextItemView<FormCardSecu
             let number = cardsType == CardType.americanExpress ? "4" : "3"
             let localizedPlaceholder = localizedString(.cardCvcItemPlaceholderDigits, item.localizationParameters, number)
 
-            if let textField = self?.textField {
-                textField.apply(placeholderText: localizedPlaceholder, with: item.style.placeholderText)
+            if let textField = self?.textField, let style = self?.style {
```
atmamont, line 25: We would find more overrides like this in the future, it was easy to left it unnoticed if old placeholder color would match the new placeholder color!

Breaking the current styling by moving `FormItem.style` to `FormItemView.style` as agreed would help to spot all such cases hopefully.

AdyenCard/Components/Card/CardViewControllerItemsProvider.swift
```diff
@@ -20,8 +20,8 @@ extension CardViewController {
     }
 
     internal final class ItemsProvider {
-
         private let formStyle: FormComponentStyle
+        private let theme: AdyenTheme
```
atmamont, line 24: This should propagate to the `FormItemView.style` in the future updates.

AdyenUI/Styles/AdyenTextFieldStyle.swift
```diff
@@ -40,33 +46,38 @@ public struct AdyenTextFieldStyle {
...
-        self.title = AdyenLabelStyle.default
+        self.title = AdyenLabelStyle.default.font(AdyenFonts.default.bodyEmphasized)
         self.text = AdyenLabelStyle.default
+        self.placeholder = AdyenLabelStyle.default.color(AdyenColorScheme.default.textSecondary)
         self.backgroundColor = AdyenColorScheme.default.background
+        self.containerColor = AdyenColorScheme.default.container
         self.errorColor = AdyenColorScheme.default.destructive
         self.cornerRadius = CornerRounding.fixed(AdyenUIConstants.defaultCornerRadius)
-        self.borderColor = AdyenColorScheme.default.outline
+        self.borderColor = AdyenColorScheme.default.container
```
atmamont, line 78: This should use `AdyenColorScheme.default.outline` but with proper base color fine-tuning (it should match the container color so there is no visual border for inactive state)

AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
```diff
@@ -38,7 +38,10 @@ open class FormTextItemView<ItemType: FormTextItem>: FormValidatableValueItemVie
     AnyFormTextItemView {
     
     override public var accessibilityLabelView: UIView? { textField }
-    
+
+    // TODO: TO be passed as a dependency by FormViewController.ItemManager
+    package let style: AdyenTextFieldStyle = .init()
```
atmamont, line 43: This is the first step to prepare for future refactor.

AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
```diff
@@ -289,6 +285,49 @@ open class FormTextItemView<ItemType: FormTextItem>: FormValidatableValueItemVie
...
+
+        // Container
+        entryTextStackView.backgroundColor = style.containerColor
+        entryTextStackView.layer.borderWidth = style.borderWidth
+        entryTextStackView.layer.cornerRadius = AdyenUIConstants.defaultCornerRadius
+
+        // Alert
+        alertLabel.textColor = style.errorColor
+
+        switch style.cornerRadius {
```
atmamont, line 315: We might need a conversion helper at some point.

AdyenUI/Styles/AdyenTextFieldStyle.swift
nauaros: It's not a bad idea to keep it like this. It reinforces the idea the borderColor should be exactly as the container background.

AdyenUI/Styles/AdyenTextFieldStyle.swift
```diff
@@ -109,6 +129,15 @@ extension AdyenTextFieldStyle {
         return newStyle
     }
     
+    /// Returns a new style with the specified border color for active state.
+    /// - Parameter borderActiveColor: The color to set.
+    /// - Returns: A new `AdyenTextFieldStyle` instance.
+    public func borderActiveColor(borderActiveColor: UIColor) -> AdyenTextFieldStyle {
```
nauaros: Why is this public? The naming with the signature of the func being the same one as the parameter is somehow confusing.

AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
nauaros: Yeah, great. Should we expect this `style` to be accessed from the outside?
Just wondering if it should internal or private.

AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
```diff
@@ -289,6 +285,49 @@ open class FormTextItemView<ItemType: FormTextItem>: FormValidatableValueItemVie
...
+        titleLabel.textColor = style.title.color
+        titleLabel.textAlignment = style.title.textAlignment
+
+        // Text field
+        textField.font = style.text.font
+        textField.textColor = style.text.color
+        textField.textAlignment = style.text.textAlignment
+
+        // Placeholder
+        textField.apply(placeholderText: textField.placeholder, with: style.placeholder)
```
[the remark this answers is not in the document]
nauaros: Good catch

AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
nauaros: Maybe as an added property to the enum `CornerRounding`?
Should we rethink `CornerRounding`? I dont see any usage of `percent`.

```swift
/// The size of corner rounding.
public enum CornerRounding {
    
    /// No rounding.
    case none
    
    /// The fixed radius of corners in display points.
    case fixed(CGFloat)
    
    /// The radius of corners in percent of length of smallest dimension (width or height).
    /// Value should be in range of 0 to 0,5. Negative and exceeding positive values will be rounded to 0 or 0,5 accordingly.
    case percent(CGFloat)
    
}
```

cc @neelSharma12

AdyenUI/Styles/AdyenTextFieldStyle.swift
atmamont: Yes, I find it more flexible and still we can decide later if we let merchants to have a dedicated outline color that is different from a container background or not.
I'd say we should stay as compact as possible and expand based on demand, while having technical possibilities to support this expansion.

AdyenUI/Styles/AdyenTextFieldStyle.swift
atmamont: Good catch, I didn't pay attention to this builder methods signature. I guess it should become 
```swift
package func borderActiveColor(_ borderActiveColor: UIColor) -> AdyenTextFieldStyle
```

AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
atmamont: At the moment I can't answer this question, I learn the UI layer code org as I go. 
I stick to the "every code change should make code better" principle. 

It's difficult to make weighted decision while being uncertain about the final vision.
Please dissect any change you review asking "if this makes code easier and simple"? And let me know if it doesn't!

AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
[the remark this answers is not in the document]
atmamont: Yes, indeed a good one. I didn't catch it immediately even after Gemini has left this comment.

AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
atmamont: Yes, @neelSharma12 we need your input here, as `percent` is not used anywhere now.

I can make an assumption that merchants can change the pay button height for instance, and then we need to change the corner radius accordingly?

neelSharma12: @nauaros @atmamont I moved this from core module and indeed percent is not being used anywhere. I think it is a good idea to get rid of it for now and add it later when needed. I can also check if there is any better way to apply cornerRounding.

atmamont: Nice, Neelam! Let's simplify it as much as possible to cover existing cases only

neelSharma12 (approved):

AdyenUI/Styles/AdyenTextFieldStyle.swift
erenbesel: It has similar functions with the same naming and also public. If they are not meant for external use, let's make them all package and also add the `_` to the param name

AdyenUI/UI/Form/Items/Text/TextField.swift
```diff
@@ -70,4 +70,11 @@ extension TextField {
         }
     }
 
+    package func apply(placeholderText: String?, with style: AdyenLabelStyle) {
+        if let text = placeholderText {
+            attributedPlaceholder = NSAttributedString(string: text, attributes: style.stringAttributes)
+        } else {
+            placeholder = placeholderText
+        }
+    }
```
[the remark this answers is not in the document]
erenbesel: This one as well?

AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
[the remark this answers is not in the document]
atmamont: Addressed in ff3c254f7ee7c98c63391ebbdd20dfb2c2fcc58f

AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
```diff
@@ -289,6 +285,49 @@ open class FormTextItemView<ItemType: FormTextItem>: FormValidatableValueItemVie
...
+            entryTextStackView.layer.cornerRadius = radius
+        default:
+            break
+        }
+
+        updateBorderStyling()
+    }
+
+    func updateBorderStyling() {
+        entryTextStackView.layer.borderWidth = AdyenUIConstants.defaultBorderWidth
```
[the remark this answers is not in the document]
atmamont: Addressed in 5df6519883fb6480ae3afb717bb93a178bf6c81b

AdyenUI/UI/Form/Items/Text/TextField.swift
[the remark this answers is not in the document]
atmamont: Addressed in 1f4ccc8c4e965b9f4fcef76ca044290d81b4e6ed

AdyenUI/Styles/AdyenTextFieldStyle.swift
atmamont: This is a typical LLM rookie mistake, it produces quite a lot of code and while it works, sometimes you need to really look into it, as such small things can slip through.
Thanks for catching this! Good lesson for all of us.

atmamont: Addressed in d42da8ce70ac67eecb57def10704d29974c678de

erenbesel (approved):

PR #2270, opened 2025-10-10
Tests | PIX Improvements

nauaros:
# Summary
This PR introduces test coverage for the new PIX improvements:
- Added unit tests
- Update integration tests

## Ticket
<ticket>
COSDK-553
</ticket>

## Checklist [Required]

- [x] Tested changes locally


Tests/IntegrationTests/Actions Tests/QRCode/QRCodeViewControllerTests.swift
```diff
@@ -0,0 +1,99 @@
...
+final class QRCodeViewControllerTests: XCTestCase {
+    
+    var viewModel: QRCodeViewModelMock!
+    var sut: QRCodeViewController!
+    
+    override func setUp() {
+        super.setUp()
+        viewModel = QRCodeViewModelMock()
+        sut = QRCodeViewController(viewModel: viewModel, style: qrCodeViewStyleMock)
+        _ = sut.view // Force view to load
```
atmamont: A better way would be to use `loadIfNeeded()`. Using `view` is perfectly fine but it increases reading complexity, also assuming the reader knows all details of the `UIViewController` lifecycle that is triggered by accessing `view` property. The comment will become redundant as a bonus.

Tests/IntegrationTests/Actions Tests/QRCode/QRCodeViewControllerTests.swift
```diff
@@ -0,0 +1,99 @@
...
+
+final class QRCodeViewControllerTests: XCTestCase {
+    
+    var viewModel: QRCodeViewModelMock!
+    var sut: QRCodeViewController!
+    
+    override func setUp() {
+        super.setUp()
+        viewModel = QRCodeViewModelMock()
+        sut = QRCodeViewController(viewModel: viewModel, style: qrCodeViewStyleMock)
```
atmamont: Creating a `sut` in the setUp() has subtle nuances that will make testing less robust. It creates a hard dependency on `sut` configuration and every test assumes it's in the proper initial state.
In the future this test suite might evolve, introduce state, or will rely on different initial states of the `sut`. It might also need to have another way to pass dependencies.

This is why setUp() should only create building blocks but never touch their state, it should be always a part of `Given` statement for a specific test. This improves test suite robustness and makes it easier to read.

Extracting `sut` creation into `private func makeSut() -> QRCodeViewController` would be one way to achieve this, and this can be extended with additional initializer parameters in the future (with default values to not break all tests at once)

Tests/IntegrationTests/Actions Tests/QRCode/QRCodeViewControllerTests.swift
```diff
@@ -0,0 +1,99 @@
...
+    }
+    
+    // MARK: - View Lifecycle
+    
+    func test_viewDidLoad_loadsLogoImage() {
+        XCTAssertTrue(viewModel.loadLogoImageCalled, "viewDidLoad should trigger loadLogoImage on the view model")
+    }
+    
+    func test_viewHierarchy_containsExpectedSubviews() {
+        XCTAssertTrue(sut.view.subviews.contains(where: { $0 is UIScrollView }))
```
atmamont: This is really hard to "read". Missing given/when/then structure heavily increases the effort one needs to make.

Tests/IntegrationTests/Actions Tests/QRCode/QRCodeViewControllerTests.swift
```diff
@@ -0,0 +1,99 @@
...
+    
+    func test_actionButton_displaysCorrectTitle() {
+        XCTAssertEqual(sut.actionButton.title, viewModel.actionButtonTitle)
+    }
+    
+    // MARK: - Actions
+    
+    func test_copyCodeAction_invokesViewModelCopy() {
+        sut.copyCode()
+        XCTAssertTrue(viewModel.copyCodeCalled, "copyCode action should call viewModel.copyCode")
```
atmamont: It is a very basic way to check if necessary call was made. There is one improvements that comes to my mind:
- check for the call count (Int) instead of factual call (Bool). This is how you will check for duplicate calls for free.

Tests/IntegrationTests/Actions Tests/QRCode/QRCodeViewControllerTests.swift
```diff
@@ -0,0 +1,99 @@
...
+    }
+    
+    func test_saveQRCodeImage_invokesViewModelSave() {
+        sut.saveQRCodeImage()
+        XCTAssertNotNil(viewModel.saveQRCodeCalled, "saveQRCodeImage should call viewModel.saveQRCode")
+        XCTAssertEqual(viewModel.saveQRCodeCalled?.sourceView, sut.view)
```
atmamont: This check is obscure. `saveQRCodeCalled` reads as `Bool` and following dot syntax makes little sense.

Tests/IntegrationTests/Actions Tests/QRCode/QRCodeViewControllerTests.swift
```diff
@@ -0,0 +1,99 @@
...
+        let size = sut.preferredContentSize
+        XCTAssertEqual(size.width, CGFloat.greatestFiniteMagnitude)
+        XCTAssertEqual(size.height, CGFloat.greatestFiniteMagnitude)
+    }
+    
+    // MARK: - Flow Types
+    
+    func test_copyFlowType_addsCopyLabelView() {
+        viewModel.flowType = .copyCode
+        sut = QRCodeViewController(viewModel: viewModel, style: qrCodeViewStyleMock)
```
atmamont: Here is the perfect example how this quickly loses consistency, can be fixed by extracting sut creation to a separate helper
`private func makeSut(flowType: FlowType = .copyCode) -> QRCodeViewController`

Tests/IntegrationTests/Actions Tests/QRCode/QRCodeViewControllerTests.swift
```diff
@@ -0,0 +1,99 @@
...
+        _ = sut.view
+        XCTAssertNotNil(sut.copyCodeLabel)
+    }
+    
+    func test_saveAsImageFlowType_doesNotAddCopyLabelView() {
+        viewModel.flowType = .saveAsImage
+        sut = QRCodeViewController(viewModel: viewModel, style: qrCodeViewStyleMock)
+        _ = sut.view
+        XCTAssertNil(sut.copyCodeLabel)
```
atmamont: Should this also check for another flow UI controls to be not `nil`?

Tests/IntegrationTests/Actions Tests/QRCode/QRCodeViewModelMock.swift
```diff
@@ -0,0 +1,68 @@
...
+            progressView: .init(progressTintColor: .blue, trackTintColor: .gray),
+            expirationLabel: .init(font: .systemFont(ofSize: 16), color: .black),
+            logoCornerRounding: .fixed(8),
+            backgroundColor: .white
+        )
+    }
+    
+    var loadLogoImageCalled = false
+    func loadLogoImage(completion: @escaping (UIImage?) -> Void) {
+        loadLogoImageCalled = true
```
atmamont, line 33: As mentioned before, this could be improved by counting calls, so this line would be 
`loadLogoImageCallCount = loadLogoImageCallCount + 1`

Tests/IntegrationTests/Actions Tests/QRCode/QRCodeViewModelTests.swift
```diff
@@ -0,0 +1,74 @@
...
+    }
+    
+    func test_copyCode_callsOnCopyCodeClosure() {
+        sut.copyCode()
+        XCTAssertTrue(copyCalled)
+    }
+    
+    func test_saveQRCode_callsOnSaveQRCodeClosure() {
+        sut.saveQRCode(image: UIImage(), sourceView: UIView())
+        XCTAssertTrue(saveCalled)
```
atmamont: One needs to scroll and find what the `saveCalled` is, relying on test case state creates opportunities for potential bugs and makes it hard to read.

Tests/IntegrationTests/Actions Tests/QRCode/QRCodeViewModelTests.swift
```diff
@@ -0,0 +1,74 @@
...
+        sut = QRCodeViewModel(
+            action: action,
+            instructionText: "Scan QR",
+            payment: nil,
+            logoUrl: URL(string: "https://example.com/logo.png")!,
+            observedProgress: nil,
+            expiration: AdyenObservable(nil),
+            imageLoader: imageLoaderMock,
+            localizationParameters: nil,
+            onSaveQRCode: { _, _ in self.saveCalled = true },
```
atmamont: This should belong to every test case to make it clear what is the expectation

atmamont (changes requested):

Tests/IntegrationTests/Actions Tests/QRCode/QRCodeViewControllerTests.swift
nauaros: Used `loadIfNeeded` instead

Tests/IntegrationTests/Actions Tests/QRCode/QRCodeViewControllerTests.swift
nauaros: Updated the mocks and the tests.

Tests/IntegrationTests/Actions Tests/QRCode/QRCodeViewControllerTests.swift
nauaros: SUT creation was extracted to a helper function. Take a look

Tests/IntegrationTests/Actions Tests/QRCode/QRCodeViewModelMock.swift
nauaros: Updated the mock

Tests/IntegrationTests/Actions Tests/QRCode/QRCodeViewModelTests.swift
nauaros: This is part now of the SUT creation method

Tests/IntegrationTests/Actions Tests/QRCode/QRCodeViewModelTests.swift
nauaros: Replaced all of this with expectations. Should be better now

Tests/IntegrationTests/Actions Tests/QRCode/QRCodeViewControllerTests.swift
nauaros: All of this was changed

Tests/IntegrationTests/Actions Tests/QRCode/QRCodeViewControllerTests.swift
nauaros: All tests now have GWT sections

atmamont (approved):

erenbesel (approved):

PR #2296, opened 2025-11-04
Refactor AdyenTheme

atmamont:
# Summary

This PR refactors the `AdyenTheme` API from builder-based methods to constructor-based initialization with default values, improving type safety, making everything explicit, aligning approaches between `AdyenTheme` elements, and aligning its public API with the Android SDK.

Android references:
- [CheckoutTheme.kt](https://github.com/Adyen/adyen-android/blob/main/ui/src/main/java/com/adyen/checkout/ui/theme/CheckoutTheme.kt)
- [CheckoutElements.kt](https://github.com/Adyen/adyen-android/blob/main/ui/src/main/java/com/adyen/checkout/ui/theme/CheckoutElements.kt)
- [CheckoutAttributes.kt](https://github.com/Adyen/adyen-android/blob/main/ui/src/main/java/com/adyen/checkout/ui/theme/CheckoutAttributes.kt)
- [CheckoutElements.kt](https://github.com/Adyen/adyen-android/blob/main/ui/src/main/java/com/adyen/checkout/ui/theme/CheckoutElements.kt)

## Changes

**Before (Builder Pattern):**
```swift
let theme = AdyenTheme()
    .button(AdyenButtonStyles(colorScheme: colorScheme))
    .label(AdyenLabelStyle().font(font).color(color))
```

**After (Constructor-Based):**
```swift
let theme = AdyenTheme(
    elements: AdyenElements(
        buttons: AdyenButtonStyles(primary: AdyenButtonStyle(...)),
        labels: AdyenLabelStyles(body: AdyenLabelStyle(...))
    )
)
```

# Demo

[Demo app BLIK component example of `AdyenTheme` usage](https://github.com/Adyen/adyen-ios/pull/2296/files#diff-9abe25c449cfa387837fd5af9ec592922c41c225797f7938167f2345806681c7R59)

## Ticket 
<ticket>
COSDK-713
</ticket>

## Checklist 

- [x] Tested changes locally

### CashAppPayComponentTests.testUIConfiguration test failure

The new toggle UI style is not propagated properly and it always uses `AdyenSwitchStyle` default configuration. 
In the upcoming refactor, the styling responsibility would be moved to the `FormItemView`, this means the whole chain should be refactored:
`FormViewController` -> `FormViewItemManager` -> `FormItemViewBuilder` -> `FormToggleItemView`
This change is big enough to add even more refactors, so a small TODO will be enough.


AdyenUI/AdyenColors.swift
```diff
@@ -56,7 +56,7 @@ public struct AdyenColorScheme: Equatable {
     // MARK: - Initializers
 
     // A static default ColorScheme
```
nauaros: Let's update this doc to remove the reference to ColorScheme

AdyenUI/Helpers/AdyenTheme+Equatable.swift
```diff
@@ -0,0 +1,60 @@
+//
+// Copyright (c) 2025 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+extension AdyenButtonStyle: Equatable {
```
nauaros, line 7: I guess this is done for testing right?

AdyenUI/AdyenColors.swift
atmamont: This is a good example why I am typically against useless comments, as the maintenance cost increases for no reason. Good catch!

AdyenUI/Styles/AdyenSwitchStyle.swift
```diff
@@ -7,7 +7,7 @@
 import UIKit
 
 /// Contains the styling customization options for a switch item in a form.
-package struct AdyenToggleStyle {
+public struct AdyenSwitchStyle {
```
nauaros: Why is this public while the initializer stays at a package level?
If this is gonna be used by merchants, should be initializer be public too?

AdyenUI/UI/Form/Items/Toggle/FormToggleItemView.swift
```diff
@@ -12,7 +12,7 @@ import UIKit
 public final class FormToggleItemView: FormItemView<FormToggleItem> {
 
     // TODO: TO be passed as a dependency by FormViewController.ItemManager
-    package let style: AdyenToggleStyle = .init()
+    package let style: AdyenSwitchStyle = .init()
```
nauaros: Can this style be `private`? Otherwise, who else is accessing it?

AdyenUI/Styles/Buttons/AdyenButtonStyle.swift
```diff
@@ -0,0 +1,74 @@
...
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Adyen
+import UIKit
+
+// MARK: - AdyenButtonStyle
+
+/// Defines the style for a button.
+public struct AdyenButtonStyle {
```
nauaros: Do you think we could refactor this into something that has a reduced public footprint and make it more DRY?

This is just a suggestion:

```swift
import Adyen
import UIKit

public struct AdyenButtonStyle {

    public enum Variant {
        case primary
        case secondary
        case tertiary
        case destructive
    }

    public var backgroundColor: UIColor
    public var textColor: UIColor
    public var disabledBackgroundColor: UIColor
    public var disabledTextColor: UIColor
    public var cornerRadius: CornerRounding?

    /// Creates a button style for the given color palette and variant.
    public init(
        variant: Variant,
        colors: AdyenColors,
        cornerRadius: CornerRounding? = .fixed(AdyenUIConstants.defaultCornerRadius)
    ) {
        switch variant {
        case .primary:
            backgroundColor = colors.primary
            textColor = colors.textOnPrimary

        case .secondary:
            backgroundColor = colors.container
            textColor = colors.text

        case .tertiary:
            backgroundColor = .clear
            textColor = colors.highlight

        case .destructive:
            backgroundColor = colors.destructive
            textColor = colors.textOnDestructive
        }

        disabledBackgroundColor = colors.disabled
        disabledTextColor = colors.textOnDisabled
        self.cornerRadius = cornerRadius
    }
}
```

AdyenUI/Helpers/AdyenTheme+Equatable.swift
atmamont: Yes, absolutely. I tried putting this into both test targets, but I struggled a lot with using `package` imports, as it's not possible to mix SPI and package imports.

AdyenUI/Styles/AdyenSwitchStyle.swift
atmamont: Good point, I now see all styles are not aligned in terms of internal access modifiers, like label has everything public and buttons have everything package. Let me revisit it.

AdyenUI/Styles/Buttons/AdyenButtonStyle.swift
atmamont: Using public enum:
- makes the initializer code harder to read and and support
- additionally exposes another enum could will cause breaking changes depending on the way merchants would use it

My take is that we should focus on a discussion over button public properties, do we want to expose them or just provide the way to use a custom color palette, so we do the styling internally.

AdyenUI/Helpers/AdyenTheme+Equatable.swift
erenbesel: What was the issue? If this is not needed here, we can look into fixing it.

AdyenUI/Styles/Buttons/AdyenButtonStyle.swift
```diff
@@ -0,0 +1,74 @@
...
+    
+    public var backgroundColor: UIColor
```
erenbesel: Should these properties be accessible and settable directly? If not we can make them internal/package

AdyenUI/Styles/Buttons/AdyenButtonStyles.swift
```diff
@@ -0,0 +1,38 @@
...
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Adyen
+import UIKit
+
+/// Defines styles for different button types.
+public struct AdyenButtonStyles {
+
+    public var primary: AdyenButtonStyle
```
erenbesel: Same as here, are they meant to be accessible/settable as direct properties? Generic approach is that they are not meant to be, which is why we have the init and/or the builder methods to modify them

AdyenUI/Styles/AdyenSwitchStyle.swift
erenbesel: This style is not meant to be public at all. We will apply the theme object by setting the relevant parts of the theme to this struct.

AdyenUI/AdyenElements.swift
```diff
@@ -0,0 +1,36 @@
...
+    package var `switch`: AdyenSwitchStyle
+    package var textField: AdyenTextFieldStyle
+
+    /// Initializes the elements with optional overrides.
+    /// Any parameter left as default will use the default value.
+    ///
+    /// - Parameters:
+    ///   - buttons: The button styles. Defaults to `.default`.
+    ///   - labels: The label styles. Defaults to `.default`.
+    ///   - switch: The switch style. Defaults to `.default`.
```
erenbesel, line 21: We are not exposing switch settings. As part of limiting exposed elements, only the main UI parts will be exposed and we will apply the corresponding parts to the relevant element internally.

AdyenUI/AdyenAttributes.swift
```diff
@@ -0,0 +1,17 @@
+//
+// Copyright (c) 2025 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+public struct AdyenAttributes: Equatable {
```
erenbesel: What exactly is the benefit of having this separate struct?
(I guess this is meant to grow and include more properties but still) 
As for its name, maybe AdyenUIAttributes?

AdyenUI/AdyenElements.swift
```diff
@@ -0,0 +1,36 @@
+//
+// Copyright (c) 2025 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import UIKit
+
+public struct AdyenElements {
```
erenbesel: Missing description.
For its name, perhaps AdyenUIElements?

Demo/Common/IntegrationExamples/AdvancedFlow/Components/BLIKComponentAdvancedFlowExample.swift
```diff
@@ -53,34 +54,33 @@ internal final class BLIKComponentAdvancedFlowExample: InitialDataAdvancedFlowPr
...
-            label: AdyenLabelStyle()
-                .font(AdyenFonts.default.body)
-                .color(AdyenColorScheme.default.textOnPrimary),
-            button: AdyenButtonStyles()
-        )
-        // Providing theme with only custom label
-        .theme(label: AdyenLabelStyle()
-            .font(AdyenFonts.default.body)
-            .color(AdyenColorScheme.default.primary)
+            AdyenTheme(
```
erenbesel: Since our configuration object follows a builder pattern look, I think this change is not fitting. As long as we align on the naming, platform based syntax decisions are fine IMO.

AdyenUI/Helpers/AdyenTheme+Equatable.swift
atmamont: This belongs to the test targets (UnitTests and IntegrationUIKitTests) but it's not possible to mix package and SPI imports in the same test suite as far as I understood. Xcode gives a test target compilation error.

AdyenUI/Styles/AdyenSwitchStyle.swift
atmamont: As agreed today, I'll make the public API as shallow as possible and introduce simple and independent AdyenTheme builder methods in the following PR.

AdyenUI/UI/Form/Items/Toggle/FormToggleItemView.swift
atmamont: Yes, I think it should be `private`

AdyenUI/Styles/Buttons/AdyenButtonStyle.swift
atmamont: As discussed, I'll make all `AdyenTheme` components internal for now.

AdyenUI/AdyenAttributes.swift
atmamont: Purely alignment with the Android SDK, there are links provided in the PR description

AdyenUI/AdyenElements.swift
atmamont: On the naming side, do you think being placed into `AdyenUI` module is enough to imply this?

Demo/Common/IntegrationExamples/AdvancedFlow/Components/BLIKComponentAdvancedFlowExample.swift
atmamont: I'm afraid adding all builder methods to the `AdyenTheme` will add a couple of hundred LOC easily. Do you think it could be done in a separate PR?
I still plan to address access level modifiers in the scope of this PR.

atmamont:
@nauaros @erenbesel 
Last changes include:
- shrinking access levels from public to package and from package to internal
- adding a few builder methods to `AdyenTheme` 
- showcase the approach using BLIK as an example

erenbesel (approved):
I like the approach! Not fully sure on the new additions of attributes/elements properties but let's find the next steps forward

nauaros (approved):
It is looking good with the builder methods!

PR #2313, opened 2025-11-13
Refactor `FormTextInputItemView` to take `AdyenTheme` as dependency

atmamont:
# Summary [Required]

Refactors `FormTextItemView` and its subclasses to accept `AdyenTheme` as a constructor parameter instead of reading from `item.style`. This continues the gradual migration from style-based to theme-based styling.

## What's Changed
- `FormTextItemView`, `FormTextInputItemView`, `FormPhoneNumberItemView`, `FormCardNumberItemView`, and `FormCardSecurityCodeItemView` now accept `theme: AdyenTheme` parameter
- `FormItemViewBuilder` now requires theme (defaults to `.default` if nil)
- Old `init(item:)` methods preserved as convenience initializers for backward compatibility (marked with TODO for cleanup)
- 
## Refactor journey overview
- ✅ Introduce `AdyenTheme` system (#2296)
- ✅ Prepare style-to-theme migration infrastructure (#2310)
- ✅ **Current PR:** Refactor `FormTextItemView` hierarchy to accept theme
- Update all `FormItemView` subclasses to accept theme
- Update `FormViewController` to pass theme to all form items
- Remove backward-compatibility convenience initializers
- Deprecate and remove `FormItemStyle` system

## Ticket [Optional]
<ticket>
COSDK-775
</ticket>

## Checklist [Required]

- [x] Tested changes locally

### DIFF
--- a/AdyenUI/UI/Form/FormItemViewBuilder.swift
+++ b/AdyenUI/UI/Form/FormItemViewBuilder.swift
@@ -9,28 +9,28 @@ import Foundation
 /// Builds different types of `FormItemView's`  from the corresponding concrete `FormItem`.
 @_spi(AdyenInternal)
 public struct FormItemViewBuilder {
-    
-    /// The theme to use for building views. Optional during migration.
-    package let theme: AdyenTheme?
-    
+
+    /// The theme to use for building views.
+    package let theme: AdyenTheme
+
     /// Initializes the form item view builder.
-    /// - Parameter theme: The theme to use for building views. If nil, views will use item.style.
+    /// - Parameter theme: The theme to use for building views. Defaults to `.default` if nil.
     package init(theme: AdyenTheme? = nil) {
-        self.theme = theme
+        self.theme = theme ?? .default
     }
-    
+
     /// Builds `FormToggleItemView` from `FormToggleItem`.
     @_spi(AdyenInternal)
     public func build(with item: FormToggleItem) -> FormItemView<FormToggleItem> {
         FormToggleItemView(item: item)
     }
-    
+
     /// Builds `FormSplitItemView` from `FormSplitItem`.
     @_spi(AdyenInternal)
     public func build(with item: FormSplitItem) -> FormItemView<FormSplitItem> {
         FormSplitItemView(item: item)
     }
-    
+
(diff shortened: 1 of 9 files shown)


AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
```diff
@@ -61,8 +63,51 @@ open class FormTextItemView<ItemType: FormTextItem>: FormValidatableValueItemVie
...
             self?.alertLabel.text = newValue
         }
-    
-        apply(style: theme.elements.textField)
+
+        configure()
+    }
+
+    /// Satisfies parent's required initializer. Delegates to main initializer with default theme.
+    public required convenience init(item: ItemType) {
```
erenbesel, line 71: missed this one because no todo :)

AdyenUI/UI/Form/FormItemViewBuilder.swift
```diff
@@ -9,28 +9,28 @@ import Foundation
...
-    package let theme: AdyenTheme?
-    
+
+    /// The theme to use for building views.
+    package let theme: AdyenTheme
+
     /// Initializes the form item view builder.
-    /// - Parameter theme: The theme to use for building views. If nil, views will use item.style.
+    /// - Parameter theme: The theme to use for building views. Defaults to `.default` if nil.
     package init(theme: AdyenTheme? = nil) {
```
erenbesel: I see theme is not optional any more then we can add it as default parameter in the init

erenbesel (approved):
Small comments, can change on next PR too.

AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
atmamont: Can't do it - this one is actually required!

AdyenUI/UI/Form/FormItemViewBuilder.swift
atmamont: Fair, this is indeed one step forward. I tried to make this initializer easily breakable to quickly see what's left but this can be done with a non-optional theme as well!

PR #2317, opened 2025-11-18
`AdyenTheme`: collect default values in one place

atmamont:
# Summary [Required]

## Part 1
After we made a decision to not expose internal UI elements styling options, we do not need to provide a flexible public API with lots of builder methods and default fallbacks anymore.

One of the downsides of the prior approach was styling values spreaded across multiple structs and their inits.
It is much easier to track theme changes and compare them between platforms when they are applied in one single place, this change tries to achieve this level of maintainability.

## Part 2
I noticed we do not pass the `AdyenTheme` to the `CardViewController`, which is required for testing upcoming changes. This change adds proper `theme` propagation.

Tests are updated in the next PR to make review easier.


nauaros (approved):

AdyenUI/AdyenElements.swift
```diff
@@ -12,25 +12,75 @@ package struct AdyenElements {
...
-    ) {
-        self.buttons = buttons
+    /// - Parameter colors: The color scheme to use.
+    internal init(colors: AdyenColors) {
+        // ALL default styling configuration is defined HERE in one place
+
+        // Define labels first so we can reuse them
+        let labels = AdyenLabelStyles(
+            title: AdyenLabelStyle(
+                font: UIFont.systemFont(ofSize: 34.0, weight: .bold),
```
erenbesel: let's decide whether to use constants or not. We have magic numbers and constants together in this init.

erenbesel (approved):

AdyenUI/AdyenElements.swift
atmamont: Good catch, I'll leverage the existing `FontSize` enum for this to remove magic numbers completely, I completely overlooked it before. This used to be a service enum to feed `AdyenFonts`. I'm deliberately not touching the `AdyenFonts` yet, the decision to keep or ditch it will come later.

PR #2318, opened 2025-11-19
Improve unit/integration tests for `AdyenTheme`

atmamont:
# Summary [Required]

Moving fast means either no tests or right tests. So far unit tests added to cover the `AdyenTheme` functionality were focusing primarily on it's public API and builder methods testing.
Since we decided to shrink the `AdyenTheme` API quite a bit and expose colors only, some changes were made.

A lot of unit tests were broken, and it takes time to update them accordingly every time. It doesn't feel like the right thing to spend a lot of time on.

Instead, I decide to change a testing approach a bit - for fast-changing pieces like `AdyenTheme` it makes less sense to capture its internal structure such as `AdyenLabelStyle` and so on and focus more on the outcomes.

This PR deleted all unit tests that capture public API and internal implementation details, and keeps the focus on integration tests, that follow the structure: 

**"Change the public API input such as color palette, and then verify if the UI layer responds accordingly"**

This change should be merged before this https://github.com/Adyen/adyen-ios/pull/2317


erenbesel (approved):

Tests/IntegrationTests/Components Tests/BLIK Component/BLIKComponentTests.swift
```diff
@@ -131,7 +131,7 @@ class BLIKComponentTests: XCTestCase {
             configuration: configuration
         )
 
-        setupRootViewController(sut.viewController)
+        sut.viewController.loadViewIfNeeded()
```
nauaros, line 134: Nice change!

nauaros (approved):

Tests/IntegrationTests/Actions Tests/Redirect/RedirectComponentTests.swift
```diff
@@ -24,7 +24,13 @@ class RedirectComponentTests: XCTestCase {
         }
     }
 
-    func testUIConfiguration() {
+    func testUIConfiguration() throws {
```
nauaros, line 27: What do you think of splitting this test into two tests based on the iOS version?

`func testUIConfiguration_iOS25AndBelow()`
`func testUIConfiguration_iOS26AndAbove()`

nauaros (approved):

Tests/IntegrationTests/Actions Tests/Redirect/RedirectComponentTests.swift
atmamont: @nauaros, what would the iOS26 test look like then?
We first need to support iOS26+ navigation bar styling in the code and then add the coverage for it, right?

PR #2266, opened 2025-10-09
SDKData for payment data object

erenbesel:
# Summary [Required]
Add a new field called `sdkData` to the `PaymentMethodDetails` to improve the SDK's performance.
- The public property itself is a string while internally combines several types together. 
- SDKData is created via composition by relevant modules.
- Got rid of duplicate code in both DropIn and PaymentComponent, making the important data updates in a single place.

## New
<new>
* The `sdkData` property in the `PaymentMethodDetails` protocol. For Advanced flow: it is highly recommended you include it in your `/payments` request, located under `paymentComponentData.paymentMethod.sdkData`
</new>

## Deprecated
<deprecated>
In the `PaymentComponentData` object: 

 - `checkoutAttemptId`
 - `supportNativeRedirect

In the `CardDetails` object:
   - `threeDS2SDKVersion`

Use the new `sdkData` proerty instead.
</deprecated>

## Ticket [Optional]
<ticket>
COSDK-515
</ticket>

## Checklist [Required]

- [x] Tested changes locally
- [x] Added/updated unit tests
- [x] Verified against acceptance criteria

PR #2323, opened 2025-11-24
Implementation of DropInFlowManager

nauaros:
# Summary

This PR introduces the `DropInFlowManager`, to centralize the handling of payment component submissions, failures, cancellations, and action presentations within the Drop-in component, promoting better separation of concerns.

All payment related modules have been refactored to leverage the `DropInFlowManager`, resulting in a reduced business logic implementation within the view models.

**Future work:**
- [ ] We get rid of the `DropInComponent` being passed in the `DropInComponentDelegate`.
  - It is a pattern no longer relevant for us, now that the `AdyenCheckout` manages the `DropInComponent`.
- [ ] Use the `DropInFlowManager` to setup the `CardComponentDelegate`
- [ ] Leverage `DropInFlowManager` to handle partial payment flows.

## Ticket
<ticket>
COSDK-791
</ticket>

## Checklist [Required]

- [ ] Tested changes locally


AdyenDropIn/Modules/DropInFlowManager.swift
```diff
@@ -0,0 +1,137 @@
...
+            return
+        }
+        updatedData.dataByAddingBrowserInfo { [weak self] newData in
+            guard let self else { return }
+            dropInComponentDelegate?.didSubmit(newData, from: component, in: dropInComponent)
+        }
+    }
+
+    internal func fail(with error: Error, from component: PaymentComponent) {
+        guard let dropInComponent else { return }
```
nauaros, line 92: This is temporary. We will get rid of the `dropInComponent` we pass in the `DropInComponentDelegate`. For now it is there to fulfill the API requirements but as the `AdyenCheckout` is aware of the DropInComponent, there is no need to send it over again.

AdyenDropIn/Modules/Common/Utilities/ActionWrapperViewController.swift
```diff
@@ -34,7 +35,14 @@ internal class ActionWrapperViewController: UINavigationController {
...
         fatalError("init(coder:) has not been implemented")
     }
-    
+
+    // MARK: - View life cycle
+
+    override internal func viewDidDisappear(_ animated: Bool) {
+        super.viewDidDisappear(animated)
+        onCancel?()
+    }
```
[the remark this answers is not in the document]
atmamont: This is a valid concern, right? @nauaros 

Update:
I see the `didTapCancel` doesn't call the onCancel() anymore.
Since we use gemini now, it would be easier for the rest of the team to resolve its comments marking your decision - disregarded or addressed. What do you think?

AdyenDropIn/Modules/ComponentContainer/ComponentContainerRouter.swift
```diff
@@ -48,9 +49,18 @@ internal class ComponentContainerRouter: Router, ComponentContainerRouting {
     // MARK: - ComponentContainerRouting
     
     internal func present(component: any PresentableComponent) {
-        rootViewController.present(component.viewController, animated: true)
+        let componentViewController = component.viewController
+        rootViewController.navigationController?.pushViewController(componentViewController, animated: true)
```
atmamont, line 53: This double dot syntax makes me ask - does this complexity deserves its own place to be contained? Like a dedicated presenter as an example. This router knows too much about the `rootViewController` internal structure and makes decisions that do not belong to its responsibilities (decision to push versus presenting modally)

AdyenDropIn/Modules/ComponentContainer/ComponentContainerRouter.swift
```diff
@@ -48,9 +49,18 @@ internal class ComponentContainerRouter: Router, ComponentContainerRouting {
...
     }
-    
+
+    internal func present(actionComponent: any PresentableComponent, onCancel: (() -> Void)?) {
+        let actionViewController = ActionPresentationHelper.viewController(
+            for: actionComponent,
+            onCancel: onCancel
+        )
+        rootViewController.present(actionViewController, animated: true)
```
atmamont, line 61: This presentation example is fine, as it calls a generic `present` method that could be easily refactored in something like `presenter.present(actionViewController, animated: true)`, keeping the same API while easily changing the way the screen is presented - push, modal, etc

AdyenDropIn/Modules/ComponentContainer/ComponentContainerViewModel.swift
```diff
@@ -36,21 +35,20 @@ internal class ComponentContainerViewModel: ComponentContainerViewModelProtocol
...
         self.context = context
         self.configuration = configuration
-        self.dropInComponent = dropInComponent
-        self.dropInComponentDelegate = dropInComponentDelegate
+        self.dropInFlowManager = dropInFlowManager
         self.cardComponentDelegate = cardComponentDelegate
         self.partialPaymentDelegate = partialPaymentDelegate
         self.onCancel = onCancel
 
+        self.dropInFlowManager.delegate = self
```
atmamont: Is this aligned with the Figma diagram you've shared? `dropInFlowManager` delegates to `AdyenCheckout` not to the `ComponentContainerViewModel` on the diagram

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewModel.swift
```diff
@@ -19,28 +19,26 @@ internal class PaymentMethodListViewModel: PaymentMethodListViewModelProtocol {
...
+        dropInFlowManager: DropInFlowManaging
     ) {
         let components = componentManager.sections
         self.paymentMethodListComponent = PaymentMethodListComponent(
             context: context,
             components: components,
             style: configuration.style.listComponent
         )
+        self.dropInFlowManager = dropInFlowManager
+        self.dropInFlowManager.delegate = self
```
atmamont: [Here](https://github.com/Adyen/adyen-ios/pull/2323/files#diff-6f58e8e14d167d24d1e59b3e4d50dd58df3966c18a203acefc8e30a2ff500e23R51) you set the `dropInFlowManager.delegate` to a component container view model. 

Is there a chance one overrides another?

AdyenDropIn/Modules/DropInFlowManager.swift
```diff
@@ -0,0 +1,137 @@
...
+        actionComponent.configuration.twint = configuration.actionComponent.twint
+        return actionComponent
+    }()
+
+    // MARK: - DropInFlowManaging
+
+    internal func submit(_ data: PaymentComponentData, from component: PaymentComponent) {
+        guard let dropInComponent else { return }
+
+        let checkoutAttemptId = component.context.analyticsProvider?.checkoutAttemptId
```
atmamont, line 76: We should not duplicate this code across the codebase. This is a crucial logic for the whole SDK and it should belong to a single reusable module.

I've counted at least three places so far:
- `PaymentComponent.submit`
- `PaymentMethodListViewModel.didSubmit`
- was `PaymentMethodListViewModel`, now it's here

Let's extract this to a separate reusable helper after we merge the SDKData PR? It has some [nice improvements](https://github.com/Adyen/adyen-ios/pull/2266/files#diff-0c691a9a11c6723d3ee715794146e730cf180e5a516725d4451a79870e6579a5R37) around code complexity for this part

atmamont (approved):
Good job Naufal!

Some points:
- the Figma diagram doesn't give the full picture, so some parts of the code I just skimmed without understanding how all pieces are connected in the flow.
- for the future PRs I'd suggest to always drop a comment to the team on every Gemini comment, so we don't assume and save some time during review. Applies to everyone on the team
- this PR is hugely missing any test coverage. While I understand some pieces of the flow could move sideways, having clear and simple tests that capture the business requirements and the flow expectations not only brings more confidence for future changes, but also makes code review much easier

AdyenDropIn/Modules/Common/Utilities/ActionWrapperViewController.swift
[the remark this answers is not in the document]
nauaros: Agree. These comments were addressed in a draft state. Thanks for the suggestion

AdyenDropIn/Modules/DropInFlowManager.swift
nauaros: Until that PR is merged into v6, it has to stay like this. This is the only place in DropIn where such logic exists, once we incorporate the SDKData, then we'll be able to skip it here.

Indeed, `PaymentMethodListViewModel` used to have it. Now that responsibility belongs only to the `DropInFlowManager`.

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewModel.swift
nauaros: No. There is only a single instance of each module (screen) at a time. Once a screen is initialized, the corresponding view model will set itself as the delegate.

AdyenDropIn/Modules/ComponentContainer/ComponentContainerViewModel.swift
nauaros: There are two parts here:
- `DFM` <> `AdyenCheckout`: DFM relies on the AdyenCheckout instance to perform all the payments related logic.
- `DFM` <> `ComponentContainerViewModel `: Notifies the view model about potential action presentation / cancellation

AdyenDropIn/Modules/ComponentContainer/ComponentContainerRouter.swift
nauaros: I think introducing a dedicated Presenter just for a simple push transition feels like unnecessary complexity. Since the router already owns navigation, it seems reasonable for it to handle the presentation style as well. The DropIn flow is already quite complex, so I'm not convinced adding another abstraction would bring enough value in this case.

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewModel.swift
atmamont: Thinking about "embeddable DropIn" as a feature for the long-term, but it such case we would allow only one instance of DropIn, right? It might be accompanied by one or two separate components. 
@Robert-SD is this the right way to describe this?

AdyenDropIn/Modules/Common/Utilities/ActionWrapperViewController.swift
```diff
@@ -4,26 +4,27 @@
...
     // MARK: - Properties
 
     private var onCancel: (() -> Void)?
-    
+
     // MARK: - Initializers
 
     internal init(
-        rootViewController: UIViewController,
+        actionComponent: any PresentableComponent,
```
erenbesel, line 20: Property name is action component but type is `presentableComponent`. We can pass a payment component here too then? What happens in that case?

AdyenDropIn/Modules/Common/Utilities/ActionPresentationHelper.swift
```diff
@@ -0,0 +1,31 @@
...
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Adyen
+import Foundation
+import SafariServices
+
+/// A centralized helper to prepare an action `PresentableComponent` for presentation.
+/// Wraps the action component in `ActionWrapperViewController` only if needed.
+internal enum ActionPresentationHelper {
```
erenbesel, line 13: Here also, the name suggests only for action components but expects a presentable component. If this can be used for any presentable component like card component? we need to change the name or the functionality.

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListRouter.swift
```diff
@@ -15,6 +15,7 @@ internal protocol PaymentMethodListRouterListener: AnyObject {
 internal protocol PaymentMethodListRouting: AnyObject {
     func dismiss(completion: (() -> Void)?)
     func present(_ component: PresentableComponent, onCancel: @escaping () -> Void)
```
erenbesel: Next line is a very similar another present function, are we supposed to remove this old one?

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodRouter.swift
```diff
@@ -15,6 +15,7 @@ internal protocol PreselectedPaymentMethodRouterListener: AnyObject {
 internal protocol PreselectedPaymentMethodRouting: AnyObject {
     func presentPaymentMethodList()
     func present(component: any PresentableComponent, onCancel: @escaping (() -> Void))
```
erenbesel: Same here, similar present functions

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodRouter.swift
```diff
@@ -59,7 +60,15 @@ internal class PreselectedPaymentMethodRouter: Router, PreselectedPaymentMethodR
         self.childRouter = componentContainerRouter
         rootViewController.present(componentContainerRouter.rootViewController, animated: true)
     }
-    
+
+    internal func present(actionComponent: any PresentableComponent, onCancel: (() -> Void)?) {
```
erenbesel, line 64: This function keeps repeating, could it be nice to add this on a generic protocol that each following router conforms to?

AdyenDropIn/Modules/Root/DropInRouter.swift
```diff
@@ -10,10 +10,7 @@ import Foundation
 import SafariServices
 import UIKit
 
-internal protocol DropInRouting: Router, AnyObject {
-    func presentActionComponent(_ component: any PresentableComponent)
-    func handle(action: Action)
-}
+internal protocol DropInRouting: Router, AnyObject { /* Empty implementation */ }
```
erenbesel: Small thing but I don't think we need to add the "empty implementation" text, it's redundant.

AdyenDropIn/Modules/DropInFlowManager.swift
```diff
@@ -0,0 +1,146 @@
...
+#endif
+import Foundation
+
+internal protocol ActionPresenter: AnyObject {
+    func present(actionComponent: PresentableComponent)
+    func didCancel(actionComponent: ActionComponent)
+}
+
+internal protocol DropInFlowManaging {
+//    func submit(_ data: PaymentComponentData, from component: PaymentComponent)
```
erenbesel: commented code

AdyenDropIn/Modules/Common/Utilities/ActionWrapperViewController.swift
nauaros: The thing is; there is no "presentable type" to represent only action components. Everything falls under the same type `PresentableComponent`. What do you suggest we can do there? `ActionComponent` does not offer a way of accessing the view.

AdyenDropIn/Modules/DropInFlowManager.swift
nauaros: ups, thanks!

AdyenDropIn/Modules/Root/DropInRouter.swift
nauaros: I prefer to be explicit about it, but I'll remove it

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodRouter.swift
nauaros: There is a `Router` protocol that applies to all routers. However, not all routers will have this capability. What do suggest? Does it provide any benefit having an abstraction for three routers where this is implemented?
- `PreselectedPaymentMethodRouter`
- `PaymentMethodListRouter`
- `ComponentContainerRouter`

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListRouter.swift
nauaros: No, I was intentional about this. It has dedicated logic to present an action. Reason: Actions do not have a navigation bar with a cancel button. For that reason, they need to be wrapped first, `ActionWrapperViewController`, before presenting it. This allows us to have a navigation bar with a cancel button.
Same reason why the `onCancel` callback is passed.

erenbesel (approved):

PR #2326, opened 2025-11-25
Fix ApplePay stale state in tests

atmamont:
# Summary [Required]

This change fixes tests using shared global UI state. Not confirmed, this might be the reason some UI state-related tests are flaky.

image

### DIFF
--- a/Tests/IntegrationTests/Components Tests/Apple Pay/ApplePayComponentTests.swift
+++ b/Tests/IntegrationTests/Components Tests/Apple Pay/ApplePayComponentTests.swift
@@ -63,8 +63,8 @@ class ApplePayComponentTest: XCTestCase {
             self.mockDelegate = nil // to prevent false triggering
         }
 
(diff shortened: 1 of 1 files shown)


goergisn (approved):

erenbesel (approved):

nauaros (approved):

Tests/IntegrationTests/Components Tests/Apple Pay/ApplePayComponentTests.swift
```diff
@@ -63,8 +63,8 @@ class ApplePayComponentTest: XCTestCase {
             self.mockDelegate = nil // to prevent false triggering
         }
 
-        presentOnRoot(viewController)
-        
+        viewController.loadViewIfNeeded()
```
descorp, line 66: I am surprised it works, because `viewController` is not be attached to anything. 🤔

descorp (approved):

PR #2346, opened 2025-12-11
Match Apple Pay supported network ordering to backend brand order

nauaros:
## Summary

This PR updates the Apple Pay network selection to follow the **brand order provided by the backend** (`/paymentMethods`). Previously, the SDK relied on the order returned by `PKPaymentRequest.availableNetworks()`.

## What’s Changed

- Networks are now filtered and ordered based on the backend brand list.
- Added/updated tests to cover ordering, unknown brands, duplicates, and `txVariantName` cases.

## Release Notes

<changed>
- **Apple Pay:** 
  - Network selection (e.g. cobadged cards) now attempts to follow the order of the brands array provided by the backend.
  - Improved error handling when no networks are available.
</changed>

## Ticket
<ticket>
COSDK-816
</ticket>

## Checklist
- [x] Tested changes locally


Tests/UnitTests/Extension/PKPaymentNetworkExtensionsTests.swift
```diff
@@ -0,0 +1,74 @@
...
+
+        // When
+        let receivedVariantName = sut.txVariantName
+
+        // Then
+        XCTAssertEqual(expectedVariantName, receivedVariantName)
+    }
+
+    // Note: Edge case where we transform `cartesBancaires` into `cartebancaire`
+    func test_txVariantName_givenCartesBancaires_shouldReturnMC() throws {
```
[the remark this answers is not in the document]
nauaros: Good catch, fixed

AdyenComponents/Apple Pay/ApplePayComponent.swift
```diff
@@ -119,7 +119,8 @@ public class ApplePayComponent: NSObject, PresentableComponent, PaymentComponent
     }
 
     private static func canMakePaymentWith(_ networks: [PKPaymentNetwork]) -> Bool {
-        PKPaymentAuthorizationViewController.canMakePayments(usingNetworks: networks)
+        guard !networks.isEmpty else { return false }
```
erenbesel, line 132: is this guard check really needed? I'd assume it would return false if array is empty

nauaros, line 132: Adding missing check here as `PKPaymentAuthorizationViewController.canMakePayments(usingNetworks: networks)` will return `true` even if no networks are present. For an empty networks, it evaluates whether the device can handle ApplePay.

It is the responsibility of the developer to check if no networks are available.
Creating an `ApplePayComponent` with no networks should throw `Error.userCannotMakePayment` if ApplePay onboarding is disabled.

Added tests to `ApplePayComponentTests` to cover these edge cases.

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
```diff
@@ -176,15 +176,24 @@ extension ApplePayComponent {
...
+    ) -> [PKPaymentNetwork] {
+        guard let brands else { return [] }
+
+        let networks = provider.availableNetworks()
+
+        // Build a lookup table from txVariantName → PKPaymentNetwork.
+        // Some networks appear more than once on iOS (e.g., cartebancaire), so we
+        // keep the first occurrence and ignore duplicates.
+        let networkByBrand: [String: PKPaymentNetwork] = networks.reduce(into: [:]) { dict, network in
+            if dict[network.txVariantName] == nil {
```
erenbesel, line 197: If the dict value is not nil, we are not setting it again but is there a chance the new value is different?

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
```diff
@@ -176,15 +176,24 @@ extension ApplePayComponent {
...
-            networks = networks.filter { brandsSet.contains($0.txVariantName) }
+    internal func supportedNetworks(
+        provider: ApplePayNetworksProviding = ApplePayNetworksProvider()
+    ) -> [PKPaymentNetwork] {
+        guard let brands else { return [] }
+
+        let networks = provider.availableNetworks()
+
+        // Build a lookup table from txVariantName → PKPaymentNetwork.
+        // Some networks appear more than once on iOS (e.g., cartebancaire), so we
```
erenbesel, line 194: So the same exact value appear more than once?

AdyenComponents/Apple Pay/ApplePayComponent.swift
nauaros: Explained in this comment: https://github.com/Adyen/adyen-ios/pull/2346#discussion_r2610059221

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
nauaros: The networks are unique except some (like `cartebancaire`). We use the first occurrence. This preserves the logic we had before.

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
nauaros: As far as I know it only happens for `carteBancaires` but Apple has never documented that the list is unique.

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
```diff
@@ -176,15 +176,23 @@ extension ApplePayComponent {
...
+    internal func supportedNetworks(
+        provider: ApplePayNetworksProviding = ApplePayNetworksProvider()
+    ) -> [PKPaymentNetwork] {
+        let networks = provider.availableNetworks()
+        guard let brands else { return networks }
+
+        // Build a lookup table from txVariantName → PKPaymentNetwork.
+        // Some networks appear more than once on iOS (e.g., cartebancaire), so we
+        // keep the first occurrence and ignore duplicates.
+        let networkByBrand: [String: PKPaymentNetwork] = networks.reduce(into: [:]) { dict, network in
```
descorp, line 196: [carteBancaire](https://developer.apple.com/documentation/passkit/pkpaymentnetwork/cartebancaire) and [carteBancaires](https://developer.apple.com/documentation/passkit/pkpaymentnetwork/cartebancaires) are depricated. We already remapping Apple network names into Adyen txvarient names, so comparison should be 1-to-1.
No need for dictionary

AdyenComponents/Apple Pay/ApplePayNetworksProvider.swift
```diff
@@ -0,0 +1,19 @@
...
+//
+
+import Foundation
+import PassKit
+
+internal protocol ApplePayNetworksProviding {
+    func availableNetworks() -> [PKPaymentNetwork]
+}
+
+internal struct ApplePayNetworksProvider: ApplePayNetworksProviding {
```
descorp, line 14: IMHO, this is overkill. Apple providing us with a list of networks. It would never be nil or empty; and if it stop providing a network (`mir` for example) - it is not like Adyen can force it to work 😅

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
```diff
@@ -176,15 +176,24 @@ extension ApplePayComponent {
...
+        let networks = provider.availableNetworks()
+
+        // Build a lookup table from txVariantName → PKPaymentNetwork.
+        // Some networks appear more than once on iOS (e.g., cartebancaire), so we
+        // keep the first occurrence and ignore duplicates.
+        let networkByBrand: [String: PKPaymentNetwork] = networks.reduce(into: [:]) { dict, network in
+            if dict[network.txVariantName] == nil {
+                dict[network.txVariantName] = network
+            }
         }
```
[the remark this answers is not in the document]
atmamont: This looks better indeed

AdyenComponents/Apple Pay/ApplePayNetworksProvider.swift
atmamont: Maybe a separate struct is not needed and `ApplePayConfiguration` could just conform to new `ApplePayNetworksProviding`, but the interface itself is necessary to decouple this logic.

Tests/IntegrationTests/Components Tests/Apple Pay/ApplePayPaymentMethodTests.swift
```diff
@@ -0,0 +1,143 @@
...
+        let expectedBrands: [String] = ["visa", "mc"]
+        let sut = makeSUT(brands: nil)
+        let networksProviderMock = ApplePayNetworksProvidingMock()
+        networksProviderMock.availableNetworksReturnValue = [.visa, .masterCard]
+
+        // When
+        let result = sut.supportedNetworks(provider: networksProviderMock)
+
+        // Then
+        XCTAssertEqual(networksProviderMock.availableNetworksCallsCount, 1)
```
atmamont: This creates additional noise, it's better to sparingly use callCount checks, they typically pollute the test code. The fact it was called has no value, we check real brands received in line 29.
There are typically ferw cases when we use them - when it's not possible for a reason to check the logic in a direct way and we cover it by checking if a proper interface was called. 
As a downside if anything is reshuffled, all such tests would immediately break because they capture implementation details, not the outcome.

atmamont (approved):

AdyenComponents/Apple Pay/ApplePayNetworksProvider.swift
nauaros: As Andrii said, the intention of this ApplePayNetworksProvider is to decouple the logic and being able to unit test the brands order. Otherwise, it is not possible to mock the networks response.

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
nauaros: If you take a closer look, you can see this logc preserves the same logic we had before of choosing the first network with the same `txVariantName`.

Moreover, Apple does not say anywhere the networks are unique. There can be multiple networks with same rawValue. For example:
- `visa` (debit)
- `visa` (credit)

Tests/IntegrationTests/Components Tests/Apple Pay/ApplePayPaymentMethodTests.swift
nauaros: Agree, I had doubts if it provides much value. Thanks!

AdyenComponents/Apple Pay/ApplePayComponent.swift
erenbesel: Good to know, but is this not a behavior change? Current version without the guard check will return a different value than this one, is that what refer to?

erenbesel (approved):
Let's update the description for when the order is applied (cobadged cards) and release notes

PR #2359, opened 2025-12-19
Introduce publisher as a single source of truth for form validation events

atmamont:
# Summary [Required]

Refactored form validation display from scattered state management to a reactive architecture using existing `AdyenObservable`. Established a single source of truth for validation state that could be re-streamed and exposed to merchants, unlocking new requested functionality to enable/disable custom Pay button for standalone components integration type.

**Key change**: Introducing `AdyenUIObservable` that **always** triggers stream updates unlike the `AdyenObservable` that eats up stream events in case new value is equal to the old one, breaking animated updates upon forced validation.

**FormValidatableValueItem.swift**
```swift
@AdyenUIObservable(false) public var shouldShowValidationError: Bool
```

Tests adder are based on [specs](https://github.com/Adyen/adyen-checkout-sdk-meta/pull/1)

 To be added

## Ticket [Optional]
<ticket>
COSDK-886
</ticket>

## Checklist [Required]

- [x] Tested changes locally


Adyen.xcodeproj/project.pbxproj
```diff
@@ -168,7 +168,7 @@
 		0019C4DA2E6EEAA100B0C3A3 /* AbstractPersonalInformationComponent+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = F97C829125BB156600D7F85C /* AbstractPersonalInformationComponent+Extensions.swift */; };
 		0019C4DB2E6EEAA100B0C3A3 /* NameFormItemInjector.swift in Sources */ = {isa = PBXBuildFile; fileRef = F9589D0E2601EF8100E4113F /* NameFormItemInjector.swift */; };
 		0019C4DC2E6EEAA100B0C3A3 /* EmailFormItemInjector.swift in Sources */ = {isa = PBXBuildFile; fileRef = F9589D262601F19E00E4113F /* EmailFormItemInjector.swift */; };
-		0019C4DD2E6EEAF500B0C3A3 /* BuildFile in Sources */ = {isa = PBXBuildFile; };
+		0019C4DD2E6EEAF500B0C3A3 /* (null) in Sources */ = {isa = PBXBuildFile; };
```
nauaros: The null references are happening again. It is possible to get rid of them, take a look [here](https://stackoverflow.com/questions/47338451/getting-a-null-reference-in-the-xcode-pbxproj-file).

AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
```diff
@@ -271,6 +271,11 @@ open class FormTextItemView<ItemType: FormTextItem>: FormValidatableValueItemVie
     /// Subclasses can override this method to stay notified when the text field resigns its first responder status.
     open func textFieldDidEndEditing(_ textField: UITextField) {
         isEditing = false
+        // Validate on focus loss (if field has content)
+        let hasContent = !(textField.text ?? "").isEmpty
+        if hasContent {
```
nauaros, line 276: Maybe something like this reads better:

```swift
// Validate on focus loss (if field has content)
if let text = textField.text, !text.isEmpty {
    updateValidationStatus(forced: true)
}
```

AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
```diff
@@ -279,18 +284,25 @@ open class FormTextItemView<ItemType: FormTextItem>: FormValidatableValueItemVie
     /// Subclasses can override this method to stay notified when textField became the first responder.
     open func textFieldDidBeginEditing(_ textField: UITextField) {
         isEditing = true
+        resetValidationStatus()
         updateBorderColor()
         item.onDidBeginEditing?()
     }
     
     override open func updateValidationStatus(forced: Bool = false) {
         let textFieldNotEmpty = !(textField.text ?? "").isEmpty
```
nauaros: I needed to spend some time in order to understand. Particularly when working with `textFieldNotEmpty`.
What do you think of doing this?

```swift
let isTextFieldEmpty = (textField.text ?? "").isEmpty
let shouldShowValidationUI: Bool

if forced {
    shouldShowValidationUI = true
} else if isTextFieldEmpty {
    shouldShowValidationUI = false
} else {
    shouldShowValidationUI = item.allowsValidationWhileEditing || !isEditing
}
```

And it can be even easier to read if we use a computed property:

```swift
let isTextFieldEmpty = (textField.text ?? "").isEmpty

var shouldShowValidationUI: Bool {
    if forced {
        return true
    }

    if isTextFieldEmpty {
        return false
    }

    return item.allowsValidationWhileEditing || !isEditing
}
```

AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
```diff
@@ -299,7 +311,8 @@ open class FormTextItemView<ItemType: FormTextItem>: FormValidatableValueItemVie
         }
         updateBorderColor()
         
-        super.updateValidationStatus(forced: forceShowValidationStatus)
+        // Pass forced to parent for error message display
+        super.updateValidationStatus(forced: forced)
```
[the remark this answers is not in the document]
nauaros: I guess it is intentional based on your comment

AdyenUI/UI/Form/Items/Value/Validatable/FormValidatableValueItemView.swift
```diff
@@ -13,102 +13,132 @@ open class FormValidatableValueItemView<ValueType, ItemType: FormValidatableValu
...
-        showHint()
-        accessibilityLabelView?.accessibilityLabel = item.title
-    }
 
-    // MARK: - Footer Label (Hint/Error Display)
-    
-    package func showHint() {
+    // MARK: - Footer Display (Private)
+
+    private var footerAnimationKey: String { "footerLabel_visibility" }
```
nauaros: It can be a constant:
```swift
private let footerAnimationKey = "footerLabel_visibility"
```

Or even better to be moved into a `Constants` enum possibly

AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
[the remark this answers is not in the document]
atmamont: @nauaros gemini comment also makes sense as it improves readability, `shouldShowValidationUI` reads much better than `forced` so I'll try to take this route! Anyways the goal is to reduce complexity to the state when we don't pass any booleans at all, so this is an intermediate step

atmamont:
Fixed in commit 33b8d2f. Removed the redundant direct call from `resetValidationStatus()` - the observer now handles all footer updates consistently, eliminating double UI updates and animation conflicts.

AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
atmamont: Good catch Naufal. I always advocate for easy readable code, thanks for calling me out on this!

AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
atmamont: I disagree here, extracting complexity to a variable name is common practice to improve readability. Here you don't need to go deep into understanding how `hasContent` is populated, you make a mental shortcut right away.

atmamont:
**Key change**: Introducing `AdyenUIObservable` that **always** triggers stream updates unlike the `AdyenObservable` that eats up stream events in case new value is equal to the old one.

Adyen/Utilities/Observable/AdyenObservable.swift
```diff
@@ -10,19 +10,19 @@ import Foundation
 /// Note: When the `wrappedValue` is updated, the observable will only publish the new value to subscribers
 ///       if it is not equal to the previous value.
 @propertyWrapper
-public final class AdyenObservable<ValueType: Equatable>: EventPublisher {
+open class AdyenObservable<ValueType: Equatable>: EventPublisher {
```
atmamont, line 14: this will become `package`, switching now triggers a chain reaction
This PR is big enough to postpone this for later

AdyenCard/Form/Installments/FormCardInstallmentsItem.swift
```diff
@@ -92,7 +92,6 @@ internal final class FormCardInstallmentsItem: BaseFormPickerItem<InstallmentEle
         // if there is no installment for the current card type
         // clear picker and hide the component
         guard !additionalPickerElements.isEmpty else {
-            isHidden.wrappedValue = true
```
atmamont, line 95: `isHidden` is set in the initializer, and we don't need to reset it again
this change fixes tests

Adyen/Utilities/Observable/AdyenObservable.swift
erenbesel: Maybe a TODO

AdyenCard/Form/Installments/FormCardInstallmentsItem.swift
robertdalmeida: I'm not sure how this is used. 
i see that didset on the `cardType` also expects a trigger to set the isHidden flag. 
not only from the init flow. 

But if this is expected then maybe the comment above needs to be updated that there will be no hiding of the cmponent.

AdyenUI/Utilities/AdyenUIObservable.swift
```diff
@@ -0,0 +1,29 @@
...
+///
+/// Unlike `AdyenObservable`, this variant fires the observer callback on every assignment,
+/// even if the new value equals the old value. This is useful in UI contexts where setting
+/// a value represents an intent to update the UI, regardless of whether the underlying
+/// value actually changed.
+///
+/// Example: Resetting validation state should always update the UI, even if the validation
+/// state was already in the "no error" state.
+@propertyWrapper
+public final class AdyenUIObservable<ValueType: Equatable>: AdyenObservable<ValueType> {
```
robertdalmeida, line 19: Minor thought: 
Maybe this is a question to how you guys see it.
But I feel it a bit misled for my reading that UIObservable inherits from AdyenObservable. I think it should be the other way. 
Here through inheritance we are removing functionality (equality check) vs using inheritance to add more functionality. I may be wrong with my expectation. But just leaving this note here.

descorp (approved):

Tests/IntegrationTests/Card Tests/Items/CardNumberItem/FormCardNumberValidationTests.swift
```diff
@@ -0,0 +1,315 @@
...
+        // AND - no error should be shown on the number item
+        XCTAssertFalse(
+            containerItem.numberItem.shouldShowValidationError,
+            "No validation error should be shown - brand detection is not validation"
+        )
+    }
+
+    // MARK: - UC6: Card Number - Error Hides Brand Logos
+
+    /// UC6: GIVEN a card number field with supported brand logos below
```
robertdalmeida, line 102: I really like the comments here. Especially having an id to reference a test case.  🚀

robertdalmeida (approved):
The FormItem logic for validation is difficult to follow. But if the validation works behaviorially in the UI. I'm good with that. 
But otherwise lgtm.

AdyenUI/Utilities/AdyenUIObservable.swift
atmamont: Both are temporary residents for our codebase and should be replaced by Combine or another Apple framework depending on availability in the min iOS version we support in v6. (Currently it is iOS 15, we might stretch it to 16 based on 2026 data).
The goal of this change was not reduce refactor risks and leave everything as is. We have substantial amount of fake coverage so I prefer to be defensive.

atmamont: Good point tho, I agree with you. Do you see the way of doing this the other way around without putting existing behavior under the risk?

AdyenCard/Form/Installments/FormCardInstallmentsItem.swift
atmamont: Good point, do you feel that removing the comment will make explanation redundant as code will become self-explanatory?

Adyen/Utilities/Observable/AdyenObservable.swift
nauaros: Not sure if it can be `package`. I think we are subclassing it right?

nauaros (approved):

AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
```diff
@@ -270,27 +269,43 @@ open class FormTextItemView<ItemType: FormTextItem>: FormValidatableValueItemVie
...
-        let textFieldNotEmpty = !(textField.text ?? "").isEmpty
+        let isTextFieldEmpty = (textField.text ?? "").isEmpty
         
-        // if validation check is allowed during editing, ignore editing check
-        let forceShowValidationStatus = (forced || textFieldNotEmpty)
-            && (item.allowsValidationWhileEditing || !isEditing)
+        // Determine if we should show validation UI (accessory, border)
+        // When forced (explicit validation like Pay button), always validate
+        // Otherwise, only validate if field has content and not editing
+        let shouldShowValidationUI: Bool
```
erenbesel, line 299: Adding my previous discussion so we don't forget (we don't need to do it on this PR):
This business logic I believe this should be done on the view model (item) and maybe as a state enum?  
If it's purely UI and doesn't belong on the item, maybe still we can consider the enum approach. 
Would save us from the complexity, going through if else blocks and get rid of the priority of the if else order.

AdyenUI/Utilities/AdyenUIObservable.swift
erenbesel: His way would be to create the new class as the base class, existing one (currently used by the codebase) would be inheriting the new one and adding the equality check to publish its changes, while the "new" one  (base) publishes the changes all the time.

erenbesel (approved):
Nice! Approved but Robert's idea may be worth thinking/updating.

AdyenCard/Form/Installments/FormCardInstallmentsItem.swift
robertdalmeida: I still would like to keep this comment for readability. (not that i understand what it is still :D - but looks like some terms i would could ask around.)
```
// if there is no installment for the current card type
```
Than
````
guard !additionalPickerElements.isEmpty else
````

Otherwise removing the comment is good.

AdyenUI/Utilities/AdyenUIObservable.swift
robertdalmeida: image

Yeah anything is fine. These types don't have much logic in them. So even letting them as-is is fine. I don't think we would want to use them in the future.

atmamont: Thanks for explaining, this way seemed to be more risky. 
Now I control the area of potential damage by manually switching every concrete publisher to become `AdyenUIObservable`, while in your case every publisher would immediately depend on the new class chain. Assessing underlying risk and not being confident in real test coverage for each observable in the codebase makes me stick to my initial decision.

AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
atmamont: We can simplify our thinking asking ourselves a question such as "Would the new system easily fit into the SwiftUI", so whenever we are ready to migrate, we don't need to touch the model/viewModel layers.

atmamont: In other words, I fully agree. We will come to this point step by step.

nauaros (approved):

erenbesel (approved):

robertdalmeida (approved):

PR #2376, opened 2026-01-20
Foundation for standalone stored components

erenbesel:
# Summary [Required]
Updates `AdyenCheckout` and `CheckoutComponentBuilder` to create standalone payment and stored payment components as `CheckoutPaymentComponent`, while also simplifying the creation via `PaymentMethodType` instead of the whole `PaymentMethod` object.

## Ticket [Optional]
<ticket>
COSDK-939
</ticket>

## Checklist [Required]

- [x] Tested changes locally

### DIFF
--- a/AdyenCheckout/AdyenCheckout.swift
+++ b/AdyenCheckout/AdyenCheckout.swift
@@ -112,18 +112,56 @@ public final class AdyenCheckout: AdyenCheckoutProtocol {
         )
     }
     
-    public func createComponent(with paymentMethod: any PaymentMethod) -> CheckoutPaymentComponent? {
+    /// Creates a payment component for the specified payment method type.
+    ///
+    /// Use this method to create a component for regular (non-stored) payment methods.
+    ///
+    /// - Parameter type: The type of payment method to create a component for (e.g., `.scheme` for cards, `.ideal` for iDEAL).
+    /// - Returns: A configured payment component, or `nil` if the payment method is not available
+    ///   in the current payment methods.
+    ///
+    /// ## Example
+    /// ```swift
+    /// if let cardComponent = checkout.createPaymentComponent(for: .scheme) {
+    ///     present(cardComponent.viewController, animated: true)
+    /// }
+    /// ```
+    public func createPaymentComponent(for type: PaymentMethodType) -> CheckoutPaymentComponent? {
+        guard let paymentMethod = paymentMethods?.paymentMethod(ofType: type) else { return nil }
+        
         // TODO: Add new v6 style here
-        CheckoutPaymentComponent(
+        return CheckoutPaymentComponent(
             paymentMethod: paymentMethod,
             configuration: configuration,
             delegate: self
         )
     }
     
-    public func createComponent(with action: Action) -> CheckoutPaymentComponent? {
-        CheckoutPaymentComponent(
-            action: action,
+    /// Creates a payment component for a stored payment method.
+    ///
+    /// Use this method to create a component for previously saved payment methods,
+    /// such as stored cards or saved bank accounts. The identifier uniquely identifies
+    /// the stored payment method from the shopper's saved payment methods.
+    ///
+    /// - Parameter identifier: The unique identifier of the stored payment method.
+    ///   This value comes from `StoredPaymentMethod.identifier` in `paymentMethods.stored`.
+    /// - Returns: A configured payment component, or `nil` if no stored payment method
+    ///   with the given identifier exists.
+    ///
+    /// ## Example
+    /// ```swift
+    ///
+    /// // Create component for selected stored method
+    /// if let storedComponent = checkout.createPaymentComponent(for: selectedMethod.identifier) {
+    ///     present(storedComponent.viewController, animated: true)
+    /// }
+    /// ```
+    public func createPaymentComponent(for identifier: String) -> CheckoutPaymentComponent? {
+        guard let storedPaymentMethod = paymentMethods?.stored.first(where: { $0.identifier == identifier }) else { return nil }
+        
+        // TODO: Add new v6 style here
+        return CheckoutPaymentComponent(
+            storedPaymentMethod: storedPaymentMethod,
             configuration: configuration,
             delegate: self
         )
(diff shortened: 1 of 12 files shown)


Tests/UnitTests/AdyenCheckout/AdyenCheckoutTests.swift
```diff
@@ -214,6 +214,147 @@ final class AdyenCheckoutTests: XCTestCase {
...
+            paymentMethods: paymentMethods,
+            checkoutAttemptId: "attemptId",
+            presentationDelegate: nil
+        )
+        
+        // When
+        let component = sut.createPaymentComponent(for: .ideal)
+        
+        // Then
+        XCTAssertNil(component)
```
atmamont, line 249: Can we also assert if an expected error is returned?

Tests/UnitTests/AdyenCheckout/AdyenCheckoutTests.swift
```diff
@@ -214,6 +214,147 @@ final class AdyenCheckoutTests: XCTestCase {
...
+            paymentMethods: paymentMethods,
+            checkoutAttemptId: "attemptId",
+            presentationDelegate: nil
+        )
+        
+        // When
+        let component = sut.createPaymentComponent(for: "non-existent-identifier")
+        
+        // Then
+        XCTAssertNil(component)
```
atmamont, line 317: And same here if possible

Tests/UnitTests/AdyenCheckout/CheckoutComponentBuilderTests.swift
```diff
@@ -263,7 +264,7 @@ final class CheckoutComponentBuilderTests: XCTestCase {
         // Given
         let paymentMethod = try XCTUnwrap(createACHPaymentMethod())
         let customTheme = AdyenTheme()
-            .colors(AdyenColors(primary: .systemPink))
+            .colors(AdyenColors(primary: .yellow))
```
atmamont, line 267: I wonder what color we end up with in the end :)

atmamont (approved):

Tests/UnitTests/AdyenCheckout/AdyenCheckoutTests.swift
erenbesel: There is no throwing in the function. Do you mean we should add a throw here? I don't think that's needed

Tests/UnitTests/AdyenCheckout/CheckoutComponentBuilderTests.swift
erenbesel: This is the same change we did to pass the compability tests as the colors with `system` in it behave differently on iOS 16 for example

AdyenCheckout/CheckoutComponentBuilder.swift
```diff
@@ -62,21 +62,33 @@ internal enum CheckoutComponentBuilder {
...
+                    context: configuration.context
+                )
+        #endif
+            
+        default:
+            StoredPaymentMethodComponent(
+                paymentMethod: storedPaymentMethod,
+                context: configuration.context
+            )
+        }
```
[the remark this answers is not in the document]
erenbesel: Returns are not needed here

Tests/UnitTests/AdyenCheckout/CheckoutComponentBuilderTests.swift
atmamont: Interesting find! 💡

Tests/UnitTests/AdyenCheckout/AdyenCheckoutTests.swift
atmamont: Depends on the merchants expectations

AdyenCheckout/AdyenCheckout.swift
```diff
@@ -99,57 +117,129 @@ public final class AdyenCheckout: AdyenCheckoutProtocol {
...
+    /// if let cardComponent = checkout.createPaymentComponent(for: .scheme) {
+    ///     present(cardComponent.viewController, animated: true)
+    /// }
+    /// ```
+    func createPaymentComponent(for type: PaymentMethodType) -> CheckoutPaymentComponent? {
+        guard let paymentMethod = paymentMethods?.paymentMethod(ofType: type) else { return nil }
+        
         // TODO: Add new v6 style here
-        CheckoutPaymentComponent(
+        return CheckoutPaymentComponent(
```
nauaros, line 160: `AdyenTheme` would be passed as a param here?

AdyenCheckout/AdyenCheckout.swift
```diff
@@ -99,57 +117,129 @@ public final class AdyenCheckout: AdyenCheckoutProtocol {
...
             configuration: configuration,
             delegate: self
         )
     }
     
-    public func createDropIn() -> DropInComponent? {
+    /// Creates a Drop-in component with all available payment methods.
+    ///
+    /// - Returns: A configured Drop-in component, or `nil` if unavailable.
+    func createDropIn() -> DropInComponent? {
```
nauaros, line 200: Most probably we can return a `UIViewController`, while keeping an internal reference to the dropIn component now that the `handle` function is exposed.

Tests/UnitTests/AdyenCheckout/AdyenCheckoutTests.swift
nauaros: Agree with Eren, returning nil value means no payment component can be created for the given txvariant does not exist

nauaros (approved):

AdyenCheckout/AdyenCheckout.swift
erenbesel: It will actually be inside the configuration so nothing to do here actually. will remove the todo!

AdyenCheckout/AdyenCheckout.swift
erenbesel: Yeah we will get back to this, submit function will need to be addressed.

Tests/UnitTests/AdyenCheckout/AdyenCheckoutTests.swift
atmamont: Imagine yourself an SDK user, would you expect `nil` to be returned or a meaningful error thrown/returned?

erenbesel: I considered both options. Throw seems to add more code here without any extra value. If the return value is nil, it will be because it does not exist. There is no exceptional case. Like a dictionary, if value is there, return it. If not, return nil.

PR #2379, opened 2026-01-22
PreselectedPaymentMethodView Initial UI

robertdalmeida:
# Summary [Required]
To pursue PreSelectedPaymentMethod screen implementation. 
Writing tests is in progress. 
This PR lays the basic UI(Not the polished). COSDK-961.
The below remains to be done in a separate PR to avoid increasing the scope. They are also mentioned as // TODO: Robert: in the code. 

1. Localization 
2. Landscape
3. Use of Dropin Assets
4. Height of the view for iOS 15 and lower.
5. Adjusting the card image size based on the spec. 

image

## Ticket [Optional]
<ticket>
COSDK-961
</ticket>


Adyen/Helpers/ImageLoader/ImageLoader+DependencyKey.swift
```diff
@@ -8,9 +8,8 @@ import Foundation
 
 import UIKit
 
-@_spi(AdyenInternal)
-extension AdyenDependencyValues {
-    public var imageLoader: ImageLoading {
+package extension AdyenDependencyValues {
```
nauaros, line 10: Nice! Thanks for being aware and changing to package level 🔥

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewController.swift
```diff
@@ -8,17 +8,42 @@ import Foundation
...
+        static let sheetCornerRadius: CGFloat = 16
+
+        static let labelsSpacing: CGFloat = 8
+        static let buttonsSpacingWithEachOther: CGFloat = 16
+    }
+
     // MARK: - Properties
 
-    private let viewModel: PreselectedPaymentMethodViewModelProtocol
+    private var viewModel: PreselectedPaymentMethodViewModelProtocol
```
nauaros: Why was it changed into a var? Just curious

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewController.swift
```diff
@@ -59,8 +145,156 @@ internal class PreselectedPaymentMethodViewController: UIViewController {
...
+        return stackView
+    }()
+
+    private lazy var primaryButton: FormButton = {
+        let button = FormButton(buttonStyle: viewModel.theme.elements.buttons.primary)
+        button.translatesAutoresizingMaskIntoConstraints = false
+        button.addTarget(self, action: #selector(primaryButtonTapped), for: .touchUpInside)
+        button.accessibilityIdentifier = ViewIdentifierBuilder.build(scopeInstance: self, postfix: "primaryButton")
+        button.leadingImage =
+            (UIImage(named: "custom-lock") ?? UIImage(systemName: "lock"))?.withRenderingMode(.alwaysTemplate)
```
nauaros: 💡Maybe it's a good idea to move this names into constants:

```swift
private enum Images {
  static let bentoLockIcon = "custom-lock"
  static let systemLockIcon = "lock"
}
```

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewController.swift
robertdalmeida: As Controller needs to set a handler to the viewModel(mutating) - `onLoadingStateChange`
Alternatively this can be made let by making a ViewModel protocol `AnyObject`
Both seem fine for me. 

For now changed it to use let and making the protocol conform to `AnyObject`.

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewController.swift
robertdalmeida: For images(my personal preference), it would be nice to have an extension on UIImage itself i would suppose, then call site would be something like; 
`button.leadingImage = .bentoLock`

For now i created the constant enum.

AdyenDropIn/Modules/PreselectedPaymentMethod/Views/CardImageView.swift
```diff
@@ -0,0 +1,191 @@
...
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Adyen
+import UIKit
+
+internal class CardImageItem {
+
+    /// Configuration for how the image size should be determined.
+    internal enum SizeMode {
```
atmamont, line 13: Do we need both cases?

AdyenDropIn/Modules/PreselectedPaymentMethod/Views/CardImageView.swift
```diff
@@ -0,0 +1,191 @@
...
+    /// The URL of the card image.
+    internal var imageURL: URL?
+
+    internal var identifier: String?
+
+    /// The size mode of the card image.
+    internal var sizeMode: SizeMode
+
+    /// The corner radius of the card image. Defaults to 5pt.
+    internal var cornerRadius: CGFloat
```
atmamont: Doest it belong here? It's a responsibility of a View, not the model, would it be possible to move it to CardImageView.Constants?

AdyenDropIn/Modules/PreselectedPaymentMethod/Views/CardImageView.swift
```diff
@@ -0,0 +1,191 @@
...
+            imageView.bottomAnchor.constraint(equalTo: containerView.bottomAnchor),
+            imageView.leadingAnchor.constraint(equalTo: containerView.leadingAnchor),
+            imageView.trailingAnchor.constraint(equalTo: containerView.trailingAnchor)
+        ])
+    }
+
+    // MARK: - Shadow
+
+    private func applyShadow() {
+        containerView.backgroundColor = .systemBackground
```
atmamont, line 132: Should we place a TODO here to use AdyenTheme later?

AdyenDropIn/Modules/PreselectedPaymentMethod/Views/CardImageView.swift
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    // MARK: - Image Loading
+
+    override internal func didMoveToWindow() {
+        super.didMoveToWindow()
+        updateIcon()
+    }
+
+    private func updateIcon() {
```
atmamont: Icon is a bit misleading here and adds unnecessary cognitive load, can we use `updateCardImage` or similar that fits the context?

AdyenDropIn/Modules/PreselectedPaymentMethod/Views/CardImageView.swift
```diff
@@ -0,0 +1,191 @@
...
+        view.backgroundColor = .clear
+        return view
+    }()
+
+    private lazy var imageView: UIImageView = {
+        let imageView = UIImageView()
+        imageView.translatesAutoresizingMaskIntoConstraints = false
+        imageView.contentMode = .scaleAspectFit
+        imageView.clipsToBounds = true
+        imageView.layer.cornerRadius = item.cornerRadius
```
atmamont: So here it seems we can safely use `Constants.cardImageCornerRadius` and remove `cornerRadius` from the item.

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewController.swift
```diff
@@ -30,27 +60,86 @@ internal class PreselectedPaymentMethodViewController: UIViewController {
...
         setupNavigationItem()
+        configureSheetPresentation()
+        viewModel.viewDidLoad()
     }
 
-    // MARK: - Private
+    // MARK: - setup & configurations
+
+    private func setupView() {
+        view.backgroundColor = .systemBackground
```
atmamont, line 68: Let's add a TODO to update with `AdyenTheme`

atmamont (commented):
Massive one! 🚀 
There are few comments mostly aiming to reduce code complexity and improve readability.
I didn't review the full one, the scope of the changes is massive. I'd prefer having separate PR adding small pieces and then the final one connecting the dots.

AdyenDropIn/Modules/PreselectedPaymentMethod/Views/CardImageView.swift
robertdalmeida: Not right now, but when we get the card art from the backend this may come in handy. 
I tested both the modes with the constraints, so there were no breaking constraints.

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodComponent.swift
```diff
@@ -11,178 +11,66 @@
...
     ) {
         self.title = title
         self.style = style
-        self.listItemStyle = listItemStyle
         self.defaultComponent = component
     }
 
     // MARK: - Cancellable
 
+    // TODO: Robert: figure out who needs didCancel.
```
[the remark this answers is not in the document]
robertdalmeida: WIll address the todos in the next MR.

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewController.swift
```diff
@@ -30,27 +60,86 @@ internal class PreselectedPaymentMethodViewController: UIViewController {
...
+    // Two tasks,
+    // 1. Avoid swipe down to dismiss.
+    // 2. to make the screen dynamic sizable.
+    private func configureSheetPresentation() {
+        // Adding this to avoid swiping down to dimiss the controller.
+        isModalInPresentation = true
+
+        // Adding this to make the controller fit its content height.
+        // TODO: Robert: Need to see if there a solution for iOS15 and lesser. But if not maybe just use full screen for below iOS 15 not much traffic there anyway. I would address lower OS's as a separate task.
+        // iOS 16+: Custom detent to fit content exactly This works really well.
```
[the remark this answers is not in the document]
robertdalmeida: I have a separate ticket for this Todo.

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewModel.swift
```diff
@@ -8,46 +8,118 @@ import Foundation
...
     }
 
     // MARK: - PreselectedPaymentMethodViewModelProtocol
 
-    internal var paymentMethodView: UIViewController {
-        preselectedPaymentMethodComponent.viewController
+    internal var cardImageItem: CardImageItem {
+        let paymentMethod = component.paymentMethod
+        let displayInformation = paymentMethod.displayInformation(using: localizationParameters)
+        // TODO: Robert: This will change as we will not rely on DisplayInformation for V6.
```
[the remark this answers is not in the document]
robertdalmeida: WIll address the todos in the next MR.

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewModel.swift
```diff
@@ -8,46 +8,118 @@ import Foundation
...
     private let component: PaymentComponent
-    private let preselectedPaymentMethodComponent: PreselectedPaymentMethodComponent
-    private var dropInFlowManager: DropInFlowManaging
+    internal let theme: AdyenTheme
+    private let localizationParameters: LocalizationParameters?
+    private let dropInFlowManager: DropInFlowManaging
+
+    internal weak var router: PreselectedPaymentMethodRouting?
+
+    // TODO: Robert: DropInComponent needs to send an event on the component being loaded.
```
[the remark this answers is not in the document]
robertdalmeida: WIll address the todos in the next MR.

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewModel.swift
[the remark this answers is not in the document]
robertdalmeida: WIll address the todos in the next MR.

Tests/IntegrationTests/DropIn Tests/PreselectedPaymentMethod/PreselectedPaymentMethodIntegrationTests.swift
```diff
@@ -0,0 +1,297 @@
...
+
+    @Test("PaymentComponent - UI fields", arguments: PaymentComponentTestData.allCases)
+    func componentIsPreSelected_validateWhatIsDisplayed(testDataPaymentType: PaymentComponentTestData) throws {
+        // When paymentType is loaded
+        let sut = SUT_MockedPaymentMethodRouter(type: testDataPaymentType)
+
+        // Expected in the UI
+
+        // TODO: Robert: This is a strange failure, it never seems to match even if the strings are the same.
+        // try #expect(sut.preSelectedViewController.primaryTitleText == testDataPaymentType.expectedTitle, "Card number")
```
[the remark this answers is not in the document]
robertdalmeida: Yes, this is a strange failure, that i will check in the future PR.

AdyenDropIn/Modules/PreselectedPaymentMethod/Views/CardImageView.swift
robertdalmeida: Sure.

AdyenDropIn/Modules/PreselectedPaymentMethod/Views/CardImageView.swift
robertdalmeida: Yes, yes indeed added a todo for this.

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewController.swift
robertdalmeida: Top

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewModel.swift
```diff
@@ -8,46 +8,118 @@ import Foundation
...
+            return ""
+        }
+        return formatted
+    }
+
+    internal var subtitleText: String {
+        localizedString(.preselectedPaymentMethodSubtitle, localizationParameters, component.paymentMethod.name, formattedAmount)
+    }
+
+    internal lazy var primaryButtonConfiguration: PreselectedPaymentMethodViewController.ButtonConfiguration = {
```
erenbesel: Not sure how to feel about these configuration properties. A few concerns:
- Calling this from viewcontroller as viewModel.buttonConfig.action() seems clunky, why not instead directly call `viewModel.didProceed`?
- The type is nested in the ViewController, but created on the ViewModel. ViewModel should not need to know about the VC (this can be moved into the ViewModel yes but still first point, who accesses what for which reason etc)

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewController.swift
nauaros: Both solution are fine with me!

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewModel.swift
robertdalmeida: Sure, removed that type. 

I used it to group button related functions/values. 
But its not needed as the viewmodel protocol is small anyway.

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewController.swift
robertdalmeida: I used the Images enum. Hope that is fine. Closing.

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewController.swift
```diff
@@ -30,27 +55,87 @@ internal class PreselectedPaymentMethodViewController: UIViewController {
...
+        buttonsStackView.addArrangedSubview(secondaryButton)
+
+        configureConstraints()
+        configureContent()
+    }
+
+    /// Two tasks,
+    /// 1. Avoid swipe down to dismiss.
+    /// 2. to make the screen dynamic sizable.
+    private func configureSheetPresentation() {
```
erenbesel:
```suggestion
    private func configureSheetForPresentation() {
```

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewModel.swift
```diff
@@ -8,46 +8,118 @@ import Foundation
...
+
+    internal var subtitleText: String {
+        localizedString(.preselectedPaymentMethodSubtitle, localizationParameters, component.paymentMethod.name, formattedAmount)
+    }
+
+    internal var primaryButtonTitle: String {
+        localizedString(.submitButtonFormatted, localizationParameters, formattedAmount)
+    }
+
+    func primaryButtonTapped() {
```
erenbesel: A bit semantics but viewModel having a function about button tap feels off. Instead it should have a function to do what needs to be done, called by the owner when needed.
So viewController calling `viewModel.proceedToPayment()` sounds more fitting then telling another layer that should not care about a button that it was tapped

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewModel.swift
```diff
@@ -57,18 +129,16 @@ internal class PreselectedPaymentMethodViewModel: PreselectedPaymentMethodViewMo
...
-    // MARK: - PreselectedPaymentMethodComponentDelegate
+    // MARK: - Button Actions to Pay or Other Payment options
 
-    internal func didRequestAllPaymentMethods() {
+    private func didRequestAllPaymentMethods() {
         router?.presentPaymentMethodList()
     }
 
-    internal func didProceed(with component: any PaymentComponent) {
+    private func didProceed(with component: any PaymentComponent) {
```
erenbesel, line 138: These 2 feel like delegate methods but they are not I think. Maybe we can name just without the "did", wdyt?

erenbesel (commented):
Less important comments with naming, but worth a look and discussion

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewModel.swift
robertdalmeida: I think this is more that the viewController shouldn't know what the button is being used for. Which is why it doesn't have any business logic related stuff. 
The ViewModel has the business logic and knows that it is for payment. So it makes the interpretation.

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewController.swift
robertdalmeida: Whats the difference? 🤔 Here I'm not sure if they could mean anything else.

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewModel.swift
robertdalmeida: I agree. But this was existing code, so left it as it is and prefer tthis pr to have the UI layout alone. 
I'll rename this to proceed() in the follow up todos in this feature???

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewModel.swift
erenbesel: Then the viewModel becomes aware that there is a button (UI element) and it is tapped. It's a reverse of who should know what. This also makes it mandatory that whoever uses this viewModel should have a button rather than have a way to call `proceedToPayment`. We shouldn't need a button to call this event, we just need a way to trigger it from the owner.
Viewcontroller should be the only object knowing it has a button and it was tapped, then telling the viewModel what to do with it. Viewmodel should indeed have the business logic of what to do but viewController should trigger the event.

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewModel.swift
erenbesel: Yeah sure!

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewController.swift
erenbesel: Tiny grammatical nitpick :) it's the sheet that is being configured not the sheet presentation. Like `configurePresentationSheet` also works.

OscarSpruit:
The lock icon seems different from the design (see screenshot). Should we update it?

image

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewModel.swift
robertdalmeida: Sure, then i changed it. Pls have a look.
I was thinking the other way around, as it would give flexibility to use the controller for some other viewModel. 
Eg: when/if we wanted to use the same controller for the CVV screen. But maybe this isn't needed now and can easily be ported when/if at all required.

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewController.swift
robertdalmeida: The api to configure is called - `sheetPresentationController` so i expected it to make sense.
Couldn't care less for naming LLM chose. I changed it to `configurePresentationSheet`.

robertdalmeida:
> The lock icon seems different from the design (see screenshot). Should we update it?
> 
> image

Thanks for checking. 🚀
I have added a todo for this, as this needs an interface from AdyenUI target. Which i would add in a separate branch.

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewModel.swift
```diff
@@ -8,46 +8,118 @@ import Foundation
...
+
+    internal var subtitleText: String {
+        localizedString(.preselectedPaymentMethodSubtitle, localizationParameters, component.paymentMethod.name, formattedAmount)
+    }
+
+    internal var submitButtonTitle: String {
+        localizedString(.submitButtonFormatted, localizationParameters, formattedAmount)
+    }
+
+    func submitPayment() {
```
erenbesel: This and `showAllPaymentMethods()` missing `internal`, linter didn't complain about these?

erenbesel (approved):
Nice! tiny linter comment non blocking

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewModel.swift
robertdalmeida: I think it did complain, but i think there are so many so it didn't get flagged. :D Updated

erenbesel (approved):

PR #2380, opened 2026-01-22
Add SessionResponse and Rename all AdyenSession related types.

erenbesel:
# Summary [Required]
Adds `SessionResponse` public data model as the response of the external `/sessions` call to pass into `Checkout.setup` .

Also renames all AdyenSession related types without the Adyen prefix.

Part of refactoring session to be package and having a clean `Checkout` base.

## Ticket [Optional]
<ticket>
COSDK-469
</ticket>

## Checklist [Required]

- [x] Tested changes locally


Adyen/Core/Models/SessionResponse.swift
```diff
@@ -0,0 +1,31 @@
+//
+// Copyright (c) Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+/// Response data obtained from the `/sessions` call.
+public struct SessionResponse: Decodable, Sendable {
+    
+    package let sessionIdentifier: String
```
Robert-SD: Can we name the parameter just "id" like on [Android](https://github.com/Adyen/adyen-android/blob/8a89464aba8c9a9f6d6b10d65609746c0761550d/core/src/main/java/com/adyen/checkout/core/sessions/SessionResponse.kt#L23). It represents the name from the network model from our [API](https://docs.adyen.com/api-explorer/Checkout/70/post/sessions#responses-201-id).

erenbesel: That would be an unconventional naming, our linter won't even allow it. We would need add a bypass for it.

Robert-SD: I would argue that it makes sense cause then we have full alignment across iOS, Android and our API. The API can be the decision maker cause it is the first in the chain. The SDKs follow and reuse the name.

AdyenCheckout/Checkout.swift
```diff
@@ -66,28 +66,24 @@ public final class Checkout: CheckoutProtocol {
     }()
     
     // MARK: - Public
-
-    // TODO: should we replace sessionId/sessionData params with a struct to future proof session init?
```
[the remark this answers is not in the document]
erenbesel: No need, the change clearly shows the new struct in place right below.

README.md
```diff
@@ -206,9 +206,9 @@ present(dropInComponent.viewController, animated: true)
 
 ```
 
-#### Implementing `AdyenSessionDelegate`
+#### Implementing `SessionDelegate`
 
-`AdyenSession` makes the necessary calls to handle the whole flow and notifies your application through its delegate, `AdyenSessionDelegate`. To handle the results of the Drop-in, the following methods of `AdyenSessionDelegate` should be implemented:
+`AdyenSession` makes the necessary calls to handle the whole flow and notifies your application through its delegate, `SessionDelegate`. To handle the results of the Drop-in, the following methods of `SessionDelegate` should be implemented:
```
robertdalmeida: Just asking is this supposed to be `Session` instead of AdyenSessin here too?

robertdalmeida (approved):
Thanks for keeping the renaming into a separate MR. Helps to review it quickly.

Adyen/Core/Models/SessionResponse.swift
atmamont: As discussed with Eren, we can be flexible here, and update our linter.

Demo/Common/IntegrationExamples/Session/Components/ApplePayComponentExample.swift
```diff
@@ -45,12 +45,12 @@ internal final class ApplePayComponentExample: InitialDataFlowProtocol {
...
     // MARK: - Networking
 
-    internal func loadSession(completion: @escaping (Result<AdyenSession, Error>) -> Void) {
+    internal func loadSession(completion: @escaping (Result<Session, Error>) -> Void) {
         requestSessionInitialInfo { [weak self] response in
             guard let self else { return }
             switch response {
             case let .success(model):
-//                AdyenSession.initialize(
+//                Session.initialize(
```
atmamont, line 53: Commented code, let's add a TODO?

Demo/Common/IntegrationExamples/Session/DropIn/DropInExample.swift
```diff
@@ -50,13 +50,13 @@ internal final class DropInExample: InitialDataFlowProtocol {
...
 
-    private func loadSession(completion: @escaping (Result<AdyenSession, Error>) -> Void) {
+    private func loadSession(completion: @escaping (Result<Session, Error>) -> Void) {
         requestSessionInitialInfo { [weak self] response in
             guard let self else { return }
             
             switch response {
             case let .success(model):
-//                AdyenSession.initialize(
+//                Session.initialize(
```
atmamont, line 59: The same

atmamont (approved):

Demo/Common/IntegrationExamples/Session/Components/ApplePayComponentExample.swift
erenbesel: It's demo code, will be rewritten once component is converted to v6

Demo/Common/IntegrationExamples/Session/DropIn/DropInExample.swift
erenbesel: Demo code

README.md
erenbesel: Oh yes it should be, nice catch

nauaros: I believe so too. The type was renamed to `Session`

nauaros (approved):

README.md
erenbesel: Actually it's the other way around. These should not be updated as none of it will be public on v6. Will be changed as a whole for release.

robertdalmeida (approved):

nauaros (approved):

PR #2384, opened 2026-01-26
Make threeDS2SdkVersion public

Robert-SD:
The Adyen iOS SDK defines the constant with Swift `package` access level, which restricts visibility and prevents access from external modules (like this Flutter plugin).
Importing AdyenCard / Adyen3DS2 does not help.


erenbesel (approved):

AdyenCard/Components/Card/ThreeDS2SdkVersion.swift
```diff
@@ -7,4 +7,4 @@
 import Foundation
 
 /// The 3DS2 SDK version.
-package let threeDS2SdkVersion: String = "2.4.4"
+public let threeDS2SdkVersion: String = "2.4.4"
```
erenbesel, line 10: Now that this constant may only be needed for API only card -> SDK ThreeDS2 component usage, let's move this to the Actions module as well.

robertdalmeida (approved):

AdyenCard/Components/Card/ThreeDS2SdkVersion.swift
Robert-SD: Done

atmamont (approved):

nauaros (approved):

AdyenCard/Components/Card/ThreeDS2SdkVersion.swift
erenbesel: After discussion, let's revert back to where it was and only make it public for now. The integration of the Swift 3ds2 will change these more.

PR #2388, opened 2026-01-30
Validation: introduce validation trigger enum

atmamont:
# Summary 

This PR refactors the form validation architecture to use a reactive pattern with ValidationTrigger and ValidationState enums for a cleaner single-source-of-truth approach.

# What's left
| Issue | Current State | Target |
|-------|---------------|--------|
| `isEditing` duplication | Exists in both Item and View | Keep only View's property |
| `showHint()` / `showError()` | Package-level escape hatches | Remove, use reactive pattern |

Move errorMessage and footer/error label color to Item.

# Demo 

## Ticket 
<ticket>
COSDK-886
</ticket>

## Checklist

- [x] Tested changes locally


AdyenUI/UI/Form/Items/Value/Validatable/FormValidatableValueItem.swift
```diff
@@ -38,4 +70,16 @@ open class FormValidatableValueItem<ValueType: Equatable>: FormValueItem<ValueTy
...
         return nil
     }
+
+    /// Resets validation state to initial (clears any error).
+    package func resetValidation() {
+        validationState = .initial
+    }
+    
+    package func triggerValidation(_ trigger: ValidationTrigger) {
+        if trigger == .focusLost, let stringValue = value as? String, stringValue.isEmpty {
```
erenbesel, line 73: This seems very specific. Is this for leaving the field but it's empty so we shouldn't trigger validation? Seems a bit convoluted (no other suggestion yet but wanted to draw attention)

AdyenUI/UI/Form/Items/Value/Validatable/FormValidatableValueItemView.swift
```diff
@@ -38,22 +38,36 @@ open class FormValidatableValueItemView<ValueType, ItemType: FormValidatableValu
     // MARK: - Validation
 
     private func setupValidationObserver() {
-        observe(item.$shouldShowValidationError) { [weak self] _ in
+        observe(item.$validationState) { [weak self] _ in
```
erenbesel: We are not checking the new `state` value in the closure. Is it for now and we will need to check it later?

AdyenUI/UI/Form/Items/Value/Validatable/FormValidatableValueItemView.swift
```diff
@@ -38,22 +38,36 @@ open class FormValidatableValueItemView<ValueType, ItemType: FormValidatableValu
...
     }
 
     private func updateFooterDisplay(animated: Bool) {
-        if item.shouldShowValidationError {
-            displayError(item.validationFailureMessage, animated: animated)
+        let newText: String?
+        let newColor: UIColor
+        let shouldBeVisible: Bool
+        
+        if let errorMessage = item.validationState.errorMessage, !errorMessage.isEmpty {
```
erenbesel: Ideally I would like all this coming the item. View not caring about the order (error first if not then placeholder else etc) and just getting a text and color to show (putting UI code in item is not ideal but in this case would make the view agnostic to this logic)

erenbesel (commented):
Much better! Some small comments

AdyenUI/UI/Form/Items/Value/Validatable/FormValidatableValueItem.swift
atmamont: Yes, you can read this as the use case "if text input loses the focus and is still empty, validation shouldn't triggered". While it should be obviously marked as invalid in the case of forced validation.
This is the last complex logic piece that is left after refactor.

I can extract this line to a method or named var to provide more context, but it would be a bit long, something like `shouldTriggerValidationOnFocussLossWhenInputIsEmpty`. Let me know if you have other ideas.

AdyenUI/UI/Form/Items/Value/Validatable/FormValidatableValueItemView.swift
atmamont: Passing it now down the methods chain. Created a helper method but the one without state still stays, as a lot of callers do not have state, will address it separately and see if it can be simplified.

AdyenUI/UI/Form/Items/Value/Validatable/FormValidatableValueItemView.swift
atmamont: I wIll take care of this in the next PR, as mentioned in the "What's next" section of PS summary.

AdyenCard/Form/FormCardSecurityCodeItemView.swift
```diff
@@ -50,8 +50,8 @@ internal final class FormCardSecurityCodeItemView: FormTextItemView<FormCardSecu
         return view
     }()
 
-    override internal func updateValidationStatus(forced: Bool = false) {
-        super.updateValidationStatus(forced: forced)
+    override public func updateValidation() {
```
nauaros, line 53: Can this be `package` or would it cascasde many changes?

AdyenUI/UI/Helpers/UIViewHelpers.swift
```diff
@@ -49,17 +49,23 @@ extension AdyenScope where Base: UIView {
         animationKey: String,
         _ hidden: Bool
     ) {
-        // Set isHidden immediately for synchronous state updates (tests)
-        base.isHidden = hidden
-        
+        // Find the nearest UIStackView parent for proper layout animation
+        let parentStackView = findParentStackView(from: base)
```
nauaros, line 53: It's too bad we need to capture the parent stack view to perform the layout. I would've expected stackviews to resize themselves with any changes.

I wonder if there's a simpler way of achieving this.

nauaros (approved):

AdyenCard/Form/FormCardSecurityCodeItemView.swift
atmamont: I'll take a look if I can achieve it without a huge bloated diff.

AdyenUI/UI/Helpers/UIViewHelpers.swift
atmamont: I couldn't find the way to make animation work nicely, and card component is too important to have anything not perfect. This is a trade-off I decided to make to not spend another week on finding a better way.

AdyenUI/UI/Form/Items/Value/Validatable/FormValidatableValueItem.swift
```diff
@@ -7,6 +7,31 @@
...
 
+package enum ValidationTrigger {
+    case focusLost
+    case explicit
+}
+
+package enum ValidationState: Equatable {
+    case initial
+    case valid
+    case invalid(String)
```
robertdalmeida, line 18: suggestion: maybe for readability, in the next PR, we could add comment that the string for invalid is actually the message to display as error. I thought initially this was the invalid string. Or that may be only me - then simply ignore this.

AdyenUI/UI/Form/Items/Value/Validatable/FormValidatableValueItem.swift
```diff
@@ -38,4 +63,16 @@ open class FormValidatableValueItem<ValueType: Equatable>: FormValueItem<ValueTy
...
         return nil
     }
+
+    /// Resets validation state to initial (clears any error).
+    package func resetValidation() {
+        validationState = .initial
+    }
+    
+    package func triggerValidation(_ trigger: ValidationTrigger) {
+        if trigger == .focusLost, let stringValue = value as? String, stringValue.isEmpty {
```
robertdalmeida, line 73: 2 thoughts here, (not blocking).
1. I hope eventually we use a switch case on trigger. So it is easier to read what happens on each trigger.
2. `value` is a generic type and we are casting here to something specific to a String. Which defeats the purpose of operating on a generic type, But if we don't have a need for generic types we shouldn't even use them even if we repeat ourselves twice for the sake of readability.

AdyenUI/UI/Form/Items/Value/Validatable/FormValidatableValueItem.swift
atmamont: Do you think there is a way to make it clear without adding a comment?
Like `case error(String)`?

robertdalmeida:
````
case invalid(ErrorMessage)
typealias ErrorMessage = String.  :D
````
Or
````
    case invalid(errorMessage: String)
````

i just find comments to be cheap. Invalid is already a good case.

atmamont: I like this one, `case invalid(errorMessage: String)`
I am stingy on comments :D

AdyenUI/UI/Form/Items/Value/Validatable/FormValidatableValueItem.swift
atmamont: Good point (2), I missed this one focusing on a specific task. 
Would you think going with the current structure and adding the `func isEmpty() -> Bool` along with existing `func isValid() -> Bool` would be a way to go? (Option A)
The other way is to introduce a new protocol such as `EmptyChekable` (Option B), but then the logic becomes inconsistent, so I'd go with option A.

PR #2398, opened 2026-02-05
Validation: remove helper methods 

atmamont:
# Summary

Since all validation logic now relies on validationState observable, we don't need helper methods anymore, and we don't need to decide whether to update validation state or not. This significantly improves codebase readability and maintainability.

This is the final phase of validation refactor.

## Ticket
<ticket>
COSDK-886
</ticket>

## Checklist

- [x] Tested changes locally


AdyenUI/UI/Form/Items/Value/Validatable/FormValidatableValueItemView.swift
erenbesel, line 61: Was moving this logic starting from this line to the item meant to be part of this task as well?

AdyenUI/UI/Form/Items/Value/Selectable/FormSelectableValueItemView.swift
```diff
@@ -173,35 +173,9 @@ package class FormSelectableValueItemView<ValueType, ItemType: FormSelectableVal
...
-            showError(item.validationState.errorMessage)
-            updateContainerBorderColor(isValid: false)
-            accessibilityLabelView?.accessibilityLabel = [
-                item.title,
-                item.validationState.errorMessage
-            ].compactMap { $0 }.joined(separator: ", ")
-
-            if let validationStatus = item.validationStatus(),
-               let error = validationStatus.validationError {
-                item.onDidShowValidationError?(error)
```
erenbesel, line 189: I see this is getting called from somewhere else now, just making sure it's still called on exact behavior right? (this is analytics trigger for validations)

atmamont: 👍 
It is still called from the `FormValidatableItemView.showValidation()`

AdyenUI/UI/Form/Items/Value/Validatable/FormValidatableValueItemView.swift
atmamont: No, I will do it separately. I have it on a separate branch but forgot about it, thanks for reminding.

AdyenUI/UI/Form/Items/Value/Selectable/FormSelectableValueItemView.swift
atmamont: I will add comprehensive test coverage for validation events in the next PR, as this one can become quite big

AdyenUI/UI/Form/Items/Value/Validatable/FormValidatableValueItemView.swift
atmamont: A few attempt to make it nice ended up in hiding the same complexity elsewhere. Parking this for now, will keep an eye on the future use-cases

p.s. We are not moving `AdyenTheme` to `Item`, as this is not aligned with the initial strategy. We can re-evaluate it later.

nauaros (approved):

robertdalmeida (approved):
🚀

AdyenUI/UI/Form/Items/Value/Selectable/FormSelectableValueItemView.swift
```diff
@@ -173,35 +173,9 @@ package class FormSelectableValueItemView<ValueType, ItemType: FormSelectableVal
...
-        updateContainerBorderColor(isValid: true)
-        accessibilityLabelView?.accessibilityLabel = item.title
-    }
-
-    override internal func resetValidationStatus() {
-        super.resetValidationStatus()
-        updateValidation()
+    override package func onValidationStateChanged(state: ValidationState) {
+        super.onValidationStateChanged(state: state)
+        updateContainerBorderColor(isValid: !state.shouldShowError)
```
robertdalmeida, line 178: Non blocking;
        updateContainerBorderColor(isValid: !state.shouldShowError)
I think these is the same. But maybe there is something I'm missing here. 
        updateContainerBorderColor(isValid: state == .valid)

atmamont: Was in the plain sight and I missed it, great pick! 👏

atmamont: So it wouldn't work as `.initial` state would be considered as "error". Adding combined check doesn't look better, so keeping this as is.

PR #2399, opened 2026-02-06
Add tests for validation events

atmamont:
# Summary 

This change adds basic coverage for validation events telemetry.

## Checklist

- [x] Tested changes locally

### DIFF
--- a/AdyenUI/UI/Form/Items/Value/Validatable/FormValidatableValueItemView.swift
+++ b/AdyenUI/UI/Form/Items/Value/Validatable/FormValidatableValueItemView.swift
@@ -105,7 +105,6 @@ open class FormValidatableValueItemView<ValueType, ItemType: FormValidatableValu
 
     private func triggerValidationErrorCallbackIfNeeded() {
         guard item.validationState.shouldShowError,
-              window != nil,
               let validationStatus = item.validationStatus(),
               let error = validationStatus.validationError
         else { return }
--- a/Tests/IntegrationTests/Card Tests/CardComponentEventTests.swift
+++ b/Tests/IntegrationTests/Card Tests/CardComponentEventTests.swift
@@ -5,21 +5,13 @@
 //
 
 @_spi(AdyenInternal) @testable import Adyen
+import XCTest
 @testable @_spi(AdyenInternal) import AdyenCard
 @testable @_spi(AdyenInternal) import AdyenUI
-import XCTest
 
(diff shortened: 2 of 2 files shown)


AdyenUI/UI/Form/Items/Value/Validatable/FormValidatableValueItemView.swift
```diff
@@ -105,7 +105,6 @@ open class FormValidatableValueItemView<ValueType, ItemType: FormValidatableValu
 
     private func triggerValidationErrorCallbackIfNeeded() {
         guard item.validationState.shouldShowError,
-              window != nil,
```
atmamont, line 108: Since the validation refactoring is over, we don't need to check this anymore - no more duplicate events upon DropIn user actions.

erenbesel (dismissed):
🚀

Tests/IntegrationTests/Card Tests/CardComponentEventTests.swift
```diff
@@ -124,81 +133,315 @@ final class CardComponentEventTests: XCTestCase {
...
+
+        XCTAssertEqual(
+            analyticsProviderMock.infos.count, 1, "Expected exactly one validation error event"
+        )
+
+        let validationEvent = analyticsProviderMock.infos.first
+        XCTAssertEqual(validationEvent?.type, .validationError)
+        XCTAssertEqual(validationEvent?.target, target)
+        XCTAssertNotNil(
+            validationEvent?.validationErrorCode, "Validation error should include error code"
```
robertdalmeida, line 431: Thought; non blocking; (having tests for sending events is already good step).
I see that we are checking for availability of the errorCode and message. 
But wouldn't these codes be different for different cases, and does it provide any value to check the specific code and error message?

robertdalmeida (approved):

Tests/IntegrationTests/Card Tests/CardComponentEventTests.swift
atmamont: Indeed, I will look into locking down specific error codes, this will be much more strict.

atmamont: Missed the branch and pinned specific error codes in the next PR 
https://github.com/Adyen/adyen-ios/pull/2401/changes/e7843bea7207ddbb2618eb25c338d4774438e151

PR #2400, opened 2026-02-06
Introduce type into PaymentComponent

nauaros:
# Summary [Required]
This PR introduces:
- A `type` property in `PaymentComponent` in order to distinguish _regular, stored, initiable payment_ components.
  This allow for decisions based on type instead of casting components.
- Re introduce generic stored payment methods. It was commented out.
- Partially decouple the `StoredCardComponent` from the `CardComponent`. When the new stored component views are built, `StoredCardComponent` and `CardComponent` should be two completely independent components.

Points to approach in the coming future but not critical:
- Redefine `PresentableComponent` in a way that it represents components with an associated UI.
- Remove navBarType in `PresentableComponent`

## Ticket [Optional]
<ticket>
COSDK-990
</ticket>

## Checklist [Required]

- [x] Tested changes locally


Adyen/Core/Core Protocols/PaymentComponent.swift
```diff
@@ -14,12 +14,44 @@ public protocol PaymentMethodAware {
     
 }
 
+/// A component that handles stored payment methods.
+public protocol StoredPaymentComponent: PaymentComponent, PresentableComponent {}
+
+public enum ComponentType {
```
erenbesel: I thought this would be package on the PR, also the name is too generic. This is payment component specific after all so maybe`PaymentComponentType`

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodRouter.swift
```diff
@@ -51,7 +51,10 @@ internal class PreselectedPaymentMethodRouter: Router, PreselectedPaymentMethodR
         rootViewController.present(paymentMethodListRouter.rootViewController, animated: true)
     }
 
-    internal func present(paymentComponent: any PresentableComponent, onCancel: @escaping (() -> Void)) {
+    internal func present(
```
erenbesel, line 55: Is this still needed? Now it should be push for payment and present for stored and this is maybe not needed?

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodRouter.swift
```diff
@@ -61,7 +64,24 @@ internal class PreselectedPaymentMethodRouter: Router, PreselectedPaymentMethodR
...
     }
 
-    internal func present(actionComponent: any PresentableComponent, onCancel: (() -> Void)?) {
+    internal func present(
+        component: PaymentComponent,
+        onCancel: @escaping () -> Void
+    ) {
+        switch component.type {
+        case let .regular(regularComponent):
+            presentModalComponent(regularComponent, onCancel: onCancel)
```
erenbesel, line 74: I guess we can't group the 2 cases in one in this switch

AdyenTwint/TwintComponent.swift
```diff
@@ -64,13 +64,11 @@ public final class TwintComponent: PaymentComponent {
     // MARK: - PaymentInitiable
 
     /// Generate the payment details and invoke PaymentsComponentDelegate method.
-    public func initiatePayment() {
+    public func initiatePayment(delegate: PaymentComponentDelegate) {
```
erenbesel, line 67: Interesting, not sure how to feel about this one

erenbesel (commented):
Nice!

AdyenTwint/TwintComponent.swift
nauaros: It's a way of making sure the delegate is always passed. Do you think it is redundant?

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodRouter.swift
nauaros: Unfortunately not with associated types

Adyen/Core/Core Protocols/PaymentComponent.swift
nauaros: The issue is that I cannot make it package without causing a big change from public to package. As it is part of `PaymentComponent`, there are many dependencies. So in order not to create more noise, it's better to raise a dedicated PR for that.

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodRouter.swift
nauaros: The idea is to have a single `present` function for components. The `router` will abstract how each type of component is presented.

Adyen/Core/Core Protocols/PaymentComponent.swift
```diff
@@ -14,12 +14,44 @@ public protocol PaymentMethodAware {
...
+    case regular(PaymentComponent & PresentableComponent)
+    case stored(StoredPaymentComponent)
+    case initiable(InitiablePaymentComponent)
+    case undefined
```
robertdalmeida, line 24: Take what i say with a grain of salt; I'm just reading without much of brains. So you could ignore this if it doesn't make sense. 

I was misled by what undefined could be, and i checked it was `AlreadyPaidPaymentComponent` which was a `PaymentComponent`. 

I would have the cases like.
presentablePaymentComponent(Pay & Present)
paymentComponent(Pay)
// The others are clear enough.

Adyen/Core/Components/AlreadyPaidPaymentComponent.swift
```diff
@@ -10,6 +10,8 @@ import Foundation
 @_spi(AdyenInternal)
 public final class AlreadyPaidPaymentComponent: PaymentComponent {
 
+    public var type: ComponentType = .undefined
```
atmamont: `.undefined` increases complexity, what prevents from not having this obscure state?

Adyen/Core/Core Protocols/PaymentComponent.swift
atmamont: I am confused by `undefined` as well.

Adyen/Core/Core Protocols/PaymentComponent.swift
```diff
@@ -14,12 +14,44 @@ public protocol PaymentMethodAware {
...
+}
+
+public extension StoredPaymentComponent {
+
+    var type: ComponentType {
+        .stored(self)
+    }
+}
+
+public extension InitiablePaymentComponent {
```
atmamont, line 50: Does this mean it has no dedicated UI and runs headless?

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListRouter.swift
```diff
@@ -56,30 +56,61 @@ internal class PaymentMethodListRouter: Router, PaymentMethodListRouting {
...
     }
+
+    // MARK: - Private
+
+    private func pushComponent(
+        _ component: PresentableComponent,
+        onCancel: @escaping () -> Void
+    ) {
+        let componentContainerViewController = componentContainerViewController(for: component, onCancel: onCancel)
+        viewController.navigationController?.pushViewController(componentContainerViewController, animated: true)
```
atmamont: Why not use `navigationController` directly?

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListRouter.swift
```diff
@@ -56,30 +56,61 @@ internal class PaymentMethodListRouter: Router, PaymentMethodListRouting {
...
+    
+    private func presentModalComponent(
+        _ component: PresentableComponent,
+        onCancel: @escaping () -> Void
+    ) {
+        let componentContainerViewController = componentContainerViewController(for: component, onCancel: onCancel)
+        viewController.present(componentContainerViewController, animated: true)
+    }
+
+    private func componentContainerViewController(
```
atmamont, line 104: Just thinking.. Should this method be moved to the `componentContainerAssembler`?

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodRouter.swift
```diff
@@ -75,6 +95,29 @@ internal class PreselectedPaymentMethodRouter: Router, PreselectedPaymentMethodR
...
+        for component: PresentableComponent,
+        onCancel: @escaping () -> Void
+    ) -> UIViewController {
+        let componentContainerRouter = componentContainerAssembler.resolveComponentContainerRouter(
+            for: component,
+            delegate: self,
+            onCancel: onCancel
+        )
+        childRouter = componentContainerRouter
+        return componentContainerRouter.rootViewController
```
atmamont, line 120: This code looks similar to the block in the `PaymentMethodListRouter`

AdyenDropIn/Utilities/ComponentManager+PaymentComponentBuilder.swift
```diff
@@ -24,49 +24,54 @@ import Foundation
 
 extension ComponentManager: PaymentComponentBuilder {
 
-    // TODO: FIX StoredPaymentMethodComponent
     internal func build(paymentMethod: StoredCardPaymentMethod) -> PaymentComponent? {
-        createCardComponent(with: paymentMethod)
+        let cardComponent = createCardComponent(with: paymentMethod)
```
atmamont, line 28: This part misses on localization parameters injection, is it ok?

Adyen/Core/Core Protocols/PaymentComponent.swift
robertdalmeida: I made a PR to your branch just to check my understanding. Alternative 2.
https://github.com/Adyen/adyen-ios/pull/2409

It is an alternative way of how i visualize it by asking the paymentMethod for what to do for a particular action (presentation or paymentInitiation. But NOT asking what type it is.

AdyenTwint/TwintComponent.swift
atmamont: The delegate property is still weak and optional. I guess this just tightens up the contract?

atmamont (approved):

Adyen/Core/Components/AlreadyPaidPaymentComponent.swift
nauaros: Actually it is this very component you are commenting on (`AlreadyPaidPaymentComponent `).

I want to remove this category, but first I need to see how I can get rid of this `AlreadyPaidPaymentComponent `. It is still a bit early to say what I can do with it. I need to continue with the list changes in order to see its consequences.

nauaros: The `AlreadyPaidPaymentComponent` comes from the gift payments flow. It is a complex topic to grasp just by looking at the surface. More time is needed to come to conclusions.

Adyen/Core/Core Protocols/PaymentComponent.swift
nauaros: Correct! Initiable components are those what do not display any user fillable form. They either finish after the `/payments` request or they have a follow up action. Examples of this are: Redirects (`ideal`) or PIX

Adyen/Core/Core Protocols/PaymentComponent.swift
nauaros: Renamed to `PaymentComponentType`

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListRouter.swift
nauaros: It is just a helper function. If you take a better look, you can see that it does use `componentContainerAssembler` to resolve the dependency, assign the child router and return the root view of this last one.

```swift
    private func componentContainerViewController(
        for component: PresentableComponent,
        onCancel: @escaping () -> Void
    ) -> UIViewController {
        let componentContainerRouter = componentContainerAssembler.resolveComponentContainerRouter(
            for: component,
            delegate: self,
            onCancel: onCancel
        )
        childRouter = componentContainerRouter
        return componentContainerRouter.rootViewController
    }
```

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodRouter.swift
nauaros: Indeed, it is the similar helper function.

Adyen/Core/Core Protocols/PaymentComponent.swift
nauaros: Hey Robert, take a look at the response I gave Andrii down below

Adyen/Core/Core Protocols/PaymentComponent.swift
atmamont: Not critical, but maybe we can find a better name, something like `HeadlessPaymentComponent`

robertdalmeida (approved):

AdyenDropIn/Utilities/ComponentManager+PaymentComponentBuilder.swift
nauaros: It's already set in `createCardComponent`:

```swift
    func createCardComponent(with paymentMethod: AnyCardPaymentMethod) -> CardComponent {
        var cardConfiguration = configuration.card.cardComponentConfiguration
        cardConfiguration.style = configuration.style.formComponent
        cardConfiguration.localizationParameters = configuration.localizationParameters
        cardConfiguration.shopperInformation = configuration.shopperInformation
        return CardComponent(
            paymentMethod: paymentMethod,
            context: context,
            configuration: cardConfiguration
        )
    }
```

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListRouter.swift
nauaros: Good catch! Changed

atmamont (approved):

PR #2401, opened 2026-02-06
Telemetry: align validation events funnel with web

atmamont:
# Summary

This change bridges the gap between iOS and Web. iOS SDK needs to start sending validation error events on both "focus loss" and "forced" validation.

[Reasoning](https://github.com/Adyen/adyen-checkout-sdk-meta/pull/23)

## Checklist 

- [x] Tested changes locally
- [x] Verified against [acceptance criteria](https://github.com/Adyen/adyen-checkout-sdk-meta/blob/c5594ef41fd6c7a91ead8bdd2b1dc62ef65353bc/specs/FORM_VALIDATION_EVENTS_USE_CASES.md)

### DIFF
--- a/AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
+++ b/AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
@@ -272,9 +272,9 @@ open class FormTextItemView<ItemType: FormTextItem>: FormValidatableValueItemVie
     /// Subclasses can override this method to stay notified when the text field resigns its first responder status.
     open func textFieldDidEndEditing(_ textField: UITextField) {
         isEditing = false
+        item.onDidEndEditing?()
         item.triggerValidation(.focusLost)
         updateBorderColor()
-        item.onDidEndEditing?()
     }
     
     /// This method hides validation accessories icons.
--- a/AdyenUI/UI/Form/Items/Value/Validatable/FormValidatableValueItem.swift
+++ b/AdyenUI/UI/Form/Items/Value/Validatable/FormValidatableValueItem.swift
@@ -46,7 +46,6 @@ open class FormValidatableValueItem<ValueType: Equatable>: FormValueItem<ValueTy
     /// Views observe this property to update their UI reactively.
     @AdyenUIObservable(.initial) package var validationState: ValidationState
     
-    
     /// Closure that is triggered when there is a validation error.
     public var onDidShowValidationError: ((ValidationError) -> Void)?
     
@@ -76,6 +75,20 @@ open class FormValidatableValueItem<ValueType: Equatable>: FormValueItem<ValueTy
         if trigger == .focusLost, isEmpty() {
             return
         }
-        validationState = isValid() ? .valid : .invalid(validationFailureMessage ?? "")
+        let newState: ValidationState = isValid() ? .valid : .invalid(validationFailureMessage ?? "")
+        validationState = newState
+
+        notifyOnValidationError(for: newState)
+    }
+
+    // MARK: - Private
+
+    private func notifyOnValidationError(for state: ValidationState) {
+        guard case .invalid = state,
+              let validationStatus = validationStatus(),
+              let error = validationStatus.validationError
+        else { return }
+        
+        onDidShowValidationError?(error)
     }
 }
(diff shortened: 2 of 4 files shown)


Tests/IntegrationTests/Card Tests/CardComponentEventTests.swift
```diff
@@ -124,81 +133,315 @@ final class CardComponentEventTests: XCTestCase {
...
+        )
+    }
+
+    // MARK: - Helper methods
+
+    private func makeSUT(
+        with configuration: CardComponentConfiguration = .init(),
+        analyticsProviderMock: AnalyticsProviderMock
+    ) -> CardComponent {
+        let context = Dummy.context(with: analyticsProviderMock)
```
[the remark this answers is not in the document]
atmamont: It's defined here https://github.com/Adyen/adyen-ios/blob/83fb8ca862d4f41e3f8128342202ae2d34007770/Tests/UnitTests/Mocks/DummyData/DummyHelper.swift

erenbesel (approved):

AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
```diff
@@ -272,9 +272,9 @@ open class FormTextItemView<ItemType: FormTextItem>: FormValidatableValueItemVie
     /// Subclasses can override this method to stay notified when the text field resigns its first responder status.
     open func textFieldDidEndEditing(_ textField: UITextField) {
         isEditing = false
+        item.onDidEndEditing?()
```
robertdalmeida, line 275: observation; nonblocking; 
I see there are 2 calls to item. (onDidEndEditing & triggerValidation), and the order of invoking them has changed,  i cannot deduce the reason why?

If we need to enforce order, we could make the call to trigger validation directly from the `item.OnDidEditing`. 
Or maybe add a comment(which i know you don't like to add) why this order of invoking is necessary is necessary. 

Again, nonblocking.

AdyenUI/UI/Form/Items/Value/Validatable/FormValidatableValueItem.swift
```diff
@@ -76,6 +75,20 @@ open class FormValidatableValueItem<ValueType: Equatable>: FormValueItem<ValueTy
         if trigger == .focusLost, isEmpty() {
             return
         }
-        validationState = isValid() ? .valid : .invalid(validationFailureMessage ?? "")
+        let newState: ValidationState = isValid() ? .valid : .invalid(validationFailureMessage ?? "")
```
robertdalmeida, line 78: thought; nonblocking;

shouldn't we instead of triggerring this call `notifyOnValidationError` only at this point. 
Have an observer on validationState which notifies if necessary(calls `notifyOnValidationError`)? 

This runs the risk that if someone else for whatever reason decides to update the validationState elsewhere then this may not be triggered and we may lose a call to `onDidShowValidationError`.

robertdalmeida (approved):

descorp (approved):

AdyenUI/UI/Form/Items/Text/FormTextItemView.swift
atmamont: The order of events **matters** as we check it in tests, and specification captures the existing web behavior.
Code doesn't need comments, if anyone would ever break it, tests would fail.

And it sounds indeed as a great idea to move validation trigger inside `item.OnDidEditing`, as the decision whether to send telemetry belongs to the business logic layer. I'll take a look.

atmamont: I decided to not go into this change, as:
- `onDidEndEditing` is now used purely as analytics observation tool, and moving validation logic inside will create obscurity for external users, as it would be not clear from its name that validation happens inside.
- this is aligned with explicit validation trigger, will create two different event/data flows if moved inside `onDidEndEditing`
- testing becomes harder as we couple two things in one place
- creates tight coupling and makes future variations hard to achieve (like validate on focus gained)

AdyenUI/UI/Form/Items/Value/Validatable/FormValidatableValueItem.swift
atmamont: Tried this, ended up in having a lot of bolierplate code, so there is no clear win in robustness.
validationState can mutate only within this class, and visibility is decent. We can look into this if this grows.

PR #2402, opened 2026-02-10
Add Image assets in AdyenUI target and provide a way to fetch those images

robertdalmeida:
# Summary 
2 things,
1. Attempting to use `SWIFT_PACKAGE` instead of the existing `CoreBundleExtension` pattern to get the images from the assets in the target. If this works. I think the new way reduces some code and configuration in Package.swift. 
‼️ So please validate if this method wouldn't break any condition.
2. Providing extensions on UIImage directly to access the assets. imho this looks a bit neat from the call site. 

Continuation of PreselectionPayment Component.


AdyenUI/Utilities/BundleSPMExtension.swift
```diff
@@ -0,0 +1,23 @@
+//
+// Copyright (c) 2020 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+import UIKit
```
atmamont: We shouldn't import `UIKit` here, right?
Eventually we plan to switch to assets retrieval from Adyen CDN, it is better to plan protocols that will serve as API between layers

robertdalmeida: > We shouldn't import UIKit here, right?
Yes, thanks, removed. 

> Eventually we plan to switch to assets retrieval from Adyen CDN, it is better to plan protocols that will serve as API between layers

Wasn't aware. But shall we scope that out separately? As i think there needs to be some thought work that needs to happen. This PR is simply a way to access images from AdyenUI target from AdyenDropin.

AdyenUI/Helpers/AdyenUIImages.swift
```diff
@@ -0,0 +1,31 @@
...
+    // MARK: - Private
+
+    /// Loads a named image from the AdyenUI resource bundle with an optional system image fallback.
+    ///
+    /// - Parameters:
+    ///   - name: The name of the image asset in the AdyenUI asset catalog.
+    ///   - systemFallback: An SF Symbol name to use when the asset is not found.
+    /// - Returns: The resolved `UIImage`. If both the asset and the system fallback are
+    ///   unavailable, returns an empty `UIImage`.
+    private static func image(named name: String) -> UIImage? {
```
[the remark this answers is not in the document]
robertdalmeida: changed.

Tests/UnitTests/AdyenUI/Assets/AdyenUIAssetsAccessTests.swift
```diff
@@ -0,0 +1,16 @@
...
+// Copyright (c) 2026 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+@_spi(AdyenInternal) @testable import AdyenUI
+import Testing
+import UIKit
+
+struct PreselectedPaymentMethodIntegrationTests {
```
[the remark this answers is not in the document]
robertdalmeida: changed.

AdyenUI/Utilities/BundleSPMExtension.swift
atmamont: Sure, it's just a context to keep in mind!

atmamont (approved):

nauaros (approved):

PR #2403, opened 2026-02-10
preselected screen with adyen theme

robertdalmeida:
# Summary

For the colors of the preselected screen using the colors from AdyenTheme.
- Added one color from the bento guide for the shadow.


AdyenDropIn/Modules/PreselectedPaymentMethod/Views/CardImageView.swift
```diff
@@ -30,25 +33,24 @@ internal class CardImageItem {
     /// The URL of the card image.
     internal var imageURL: URL?
 
-    internal var identifier: String?
-
     /// The size mode of the card image.
     internal var sizeMode: SizeMode
 
+    internal var theme: AdyenTheme
```
atmamont:
```suggestion
    internal let theme: AdyenTheme
```

AdyenDropIn/Modules/PreselectedPaymentMethod/Views/CardImageView.swift
```diff
@@ -128,9 +130,8 @@ internal final class CardImageView: UIView {
     // MARK: - Shadow
 
     private func applyShadow() {
-        // TODO: Robert: Use AdyenTheme
-        containerView.backgroundColor = .systemBackground
-        containerView.layer.shadowColor = UIColor.label.cgColor
+        containerView.backgroundColor = item.theme.colors.background
+        containerView.layer.shadowColor = item.theme.colors.supportShadow.cgColor
         containerView.layer.shadowOffset = CGSize(width: 0, height: Constants.shadowOffsetHeight)
         containerView.layer.shadowRadius = Constants.shadowRadius
```
atmamont: It might be a good moment to extract these to AdyenUIConstants!

AdyenUI/AdyenColors.swift
```diff
@@ -7,53 +7,56 @@
...
-    static let backgroundQuaternary = UIColor.color(hex: 0x525252)
-    static let backgroundDisabled = UIColor.color(hex: 0xEEEFF1)
-    static let critical = UIColor.color(hex: 0xE22D2D)
-    static let success = UIColor.color(hex: 0x07893C)
-    static let highlight = UIColor.color(hex: 0x0070F5)
-    static let labelPrimary = UIColor.color(hex: 0x00112C)
-    static let labelSecondary = UIColor.color(hex: 0x5C687C)
-    static let labelDisabled = UIColor.color(hex: 0x8D95A3)
-    static let separatorPrimary = UIColor.color(hex: 0xDADDDF)
+    internal static let backgroundPrimary = UIColor.color(hex: 0xFFFFFF)
```
atmamont, line 10: 👍

AdyenUI/AdyenColors.swift
```diff
@@ -7,53 +7,56 @@
...
+    package var highlight: UIColor
+    package var destructive: UIColor
+    package var success: UIColor
+    package var textOnDestructive: UIColor
+    package var disabled: UIColor
+    package var textOnDisabled: UIColor
+    package var separator: UIColor
+    package var text: UIColor
+    package var textSecondary: UIColor
+    package var supportShadow: UIColor
```
atmamont, line 55: Is this aligned with Android? There is no such semantic color in https://github.com/Adyen/adyen-android/blob/main/ui/src/main/java/com/adyen/checkout/ui/theme/CheckoutColors.kt

The consensus right now is to keep public theme API as lean as possible. We can always expose it later if there is a demand for this.

atmamont: @OscarSpruit do you have it hardcoded somewhere in constants?

atmamont (dismissed):

AdyenUI/AdyenColors.swift
robertdalmeida: Android seem to already have it. 
image

I picked up the color from the bento guide. 
image

This doesn't seem to be part of the public api of AdyenTheme as i see. 
AdyenColors struct i assume will not be exposed to integrators. 

They would have a few options available under AdyenTheme, which using which we internally populate AdyenColors and then ui components consume the references in AdyenColors. 
If by adding this here this contributes to the public api load, then it is definitly the wrong place to do it. 

Waiting for some input on how to go about including a shadow color. 

cc: @nauaros

AdyenUI/AdyenColors.swift
robertdalmeida: swiftlint was complaining on this file.

AdyenDropIn/Modules/PreselectedPaymentMethod/Views/CardImageView.swift
robertdalmeida: Would ideally prefer an extension on CALayer to apply these shadows - maybe as an improvement.
But for now moved the constants to AdyenUIConstants.

AdyenUI/AdyenColors.swift
atmamont: My point was indeed about the shadow color being a part of public API, it shouldn't be.
At the same time Android way seems to be extreme hardcoding, as they even need to use Suppress("MagicNumber") annotation.

I'd pick a place that is not exposed to a public API, but is still configurable and scalable, so we can apply this color to any future shadows when needed.

nauaros: I also agree there's no need to expose this publicly. A shadow it's part of the UI elements, no need for merchants to alter it. I would challenge this on Android indeed.

The less public API we have in theming, the better

nauaros (approved):

AdyenUI/AdyenColors.swift
robertdalmeida: The shadow color is package level. 
So resolving.

nauaros (approved):

PR #2405, opened 2026-02-11
Create analytics provider in AdyenContext with checkoutAttemptId

robertdalmeida:
# Summary [Required]
Attempting to add checkoutAttemptId to AdyenContext so that it can internally initialize the analytics classes using it. 

# Concerns: 
Added todo's in the code for them. Please have a look.


Adyen/Analytics/AnalyticsProvider/EventAnalyticsProvider.swift
```diff
@@ -8,40 +8,37 @@ import AdyenNetworking
 import Foundation
 
 internal final class EventAnalyticsProvider: AnyEventAnalyticsProvider {
```
robertdalmeida, line 10: @erenbesel as i go through the code. I increasingly think that no one should concern them of the checkoutAttempId (not even the CheckoutProvider & CheckoutAttemptIdProvider) it could be encapsulated completely in `EventAnalyticsProvider`. 
(ofcourse providing a way to get the id for the /payments call as currently). 

Whoever creates the AnalyticsProvider, just needs to create an instance of it - of course passing the APIContext and it should manage its own lifecyle (fetching checkoutID during, sendingEvents when triggerred). 
if it is able to get an id then analytics would work, if it doesn't then no events are sent.

robertdalmeida: Added a todo.

PR #2408, opened 2026-02-13
Fetch public key during Checkout setup

robertdalmeida:
# Summary [Required]

Task - fetch the public key during the creation Checkout object along with the checkout attempt ID.

Context & reasoning:
PublicKeyProvider is an existing type and it contains some additional functionality like retry & unique request management. 
- Extracted the simple functionality to fetch the public key into a PublicKeyFetcher(for lack of a better name). 
- This will be used by the existing PublicKeyProvider & directly from CheckoutProvider. 

What's next? 
- The components need to use the public key fetched from here.
- The existing PublicKeyProvider should be deleted.


AdyenCheckout/CheckoutProvider.swift
```diff
@@ -66,14 +78,22 @@ internal class CheckoutProvider: CheckoutProviding {
...
         )
-        
-        return Checkout(
+
+        let apiClient = APIClient(apiContext: configuration.context.apiContext)
+
+        async let publicKey = try? await publicKeyProvider.fetchPublicKey(
+            apiClient: apiClient,
+            clientKey: configuration.context.apiContext.clientKey
+        )
```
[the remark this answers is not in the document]
robertdalmeida: The setup for session uses the same apiClient to set itself up. So we are re-using the same api client to fetch the publicKey.
So can't really do this for all setups.

AdyenCheckout/CheckoutProvider.swift
```diff
@@ -42,13 +48,19 @@ internal class CheckoutProvider: CheckoutProviding {
             with: configuration
         )
 
+        // TODO: Robert: Note here we are using try? do we already want to fail here if in the 0.0001% of the change that there is a failure. (I assume yes?)
+        async let publicKey = try? await publicKeyProvider.fetchPublicKey(
+            apiClient: apiClient,
+            clientKey: configuration.context.apiContext.clientKey
+        )
```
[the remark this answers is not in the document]
robertdalmeida: I have already a todo that needs some discussion before we throw at this point.
For now we don't throw at this point.

Adyen/Utilities/PublicKeyProvider/PublicKeyFetcher.swift
```diff
@@ -0,0 +1,67 @@
...
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import AdyenNetworking
+import Foundation
+
+package protocol PublicKeyFetching {
+    /// Retaining a completion handler interface as this is needed for the PublicKeyProvider to ensure that only one request is being sent through the apiClient even if there are multiple callers.
+    /// We can remove this when we remove `PublicKeyProvider`
```
atmamont: Should it be `//`?

Adyen/Utilities/PublicKeyProvider/PublicKeyFetcher.swift
```diff
@@ -0,0 +1,67 @@
...
+    func fetchPublicKey(apiClient: APIClientKeyRequestProtocol, clientKey: String, completion: @escaping (Result<String, Error>) -> Void)
+    func fetchPublicKey(apiClient: APIClientKeyRequestProtocol, clientKey: String) async throws -> String
+}
+
+package protocol APIClientKeyRequestProtocol {
+    func perform(request: ClientKeyRequest, completionHandler: @escaping (Result<ClientKeyResponse, Error>) -> Void)
+}
+
+package final class PublicKeyFetcher: PublicKeyFetching {
+    package init() {}
```
atmamont, line 22: We could move dependencies such as `apiClient` and `clientKey` to initializer and declutter protocol and methods definition, what do you think?

AdyenCheckout/CheckoutProvider.swift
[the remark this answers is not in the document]
atmamont: I'd fail here and expose this error to the merchant, assuming we applied retry policy and still failed.

atmamont (approved):

Adyen/Utilities/PublicKeyProvider/PublicKeyFetcher.swift
robertdalmeida: image

We don't have the apiClient and clientKey during initialization of the CheckoutProvider. 
Which is why it is as a parameter in the protocol.

atmamont: I see, thanks for clarifying! Let's move on then 🚀

atmamont (approved):

nauaros (approved):

PR #2413, opened 2026-02-23
Decouple Payment Method List Business Logic

nauaros:
## Summary
 
This PR refactors the `PaymentMethodListComponent` to follow a cleaner MVVM architecture by decoupling business logic from UI components. **The core change: components are now created on-demand when a payment method is selected, rather than being pre-built upfront.**
 
### On-Demand Component Creation (Core Change)
 
**Before:** All payment components were instantiated at list initialization, holding them in memory regardless of whether the user selected them.
 
**After:** The list now displays payment methods directly. A component is only built when the user taps a payment method:
 
```swift
private func select(paymentMethod: PaymentMethod) {
    guard let component = componentManager.buildComponent(for: paymentMethod) else { return }
    // Present the component...
}
```
 
This reduces memory footprint and speeds up initial list load.
 
### Architecture Refactoring
 
- **Removed `PaymentMethodListComponent`** — Transitioned into a dedicated `PaymentMethodListViewController` with MVVM pattern
- **Introduced `PaymentMethodListState`** — New enum (`ready`, `loaded`, `loading`) to manage view state reactively via Combine
- **Created `PaymentMethodListViewModelProtocol`** — Defines the contract for the ViewModel, enabling testability via mocking
 
### ComponentManager Refactoring
 
- **Introduced `ComponentManaging` protocol** — Abstracts component building for dependency injection
- **Added `sections` property** — Returns `[PaymentMethodsSection]` containing payment methods (not components)
- **Renamed `ComponentSections.swift` → `PaymentMethodsSection.swift`** — Now holds `PaymentMethod` arrays instead of built components
 
### Removed/Simplified
 
- **Deleted `AlreadyPaidPaymentComponent`** — Replaced with direct `OrderPaymentMethod` check in selection handler
- **Removed `CoreListDataSource` from `ListViewController`** — Simplified list management (was only used in iOS 12 or lower)
- **Removed `ComponentLoader` from list** — No longer needed with on-demand component creation

## Benefits
 
1. **Lower memory footprint** — Components created only when needed, not upfront
2. **Faster list load** — No pre-instantiation of all payment components
3. **Better separation of concerns** — UI logic in ViewController, business logic in ViewModel
4. **Improved testability** — Protocol-based dependencies allow easy mocking

## Ticket
<ticket>
COSDK-827
</ticket>

## Checklist

- [x] Tested changes locally


AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewController.swift
```diff
@@ -63,8 +77,62 @@ internal class PaymentMethodListViewController: UIViewController {
...
+//            guard success else { return }
+//            // This is to prevent the merchant calling completion closure multiple times
+//            guard let self else { return }
+//            guard viewModel.componentSections[indexPath.section]
+//                .components[indexPath.item]
+//                .paymentMethod == paymentMethod else { return }
+//            self.deleteComponent(at: indexPath)
+//        }
+//        viewModel.delete(paymentMethod, completion: completion)
+    }
```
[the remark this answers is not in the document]
nauaros: It will be added/changed in follow up PRs

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewModel.swift
```diff
@@ -7,96 +7,152 @@
 import Foundation
 import UIKit
 @_spi(AdyenInternal) import Adyen
+#if canImport(AdyenUI)
+    @_spi(AdyenInternal) import AdyenUI
+#endif
+
+internal enum PaymentMethodListState {
+    case ready
+    case loaded(sections: [ListSection])
```
robertdalmeida, line 16: What is the difference between `ready` and `loaded`? They both seem to indicate the same thing. Ofcourse at this moment i don't know how this is used.

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewController.swift
```diff
@@ -63,8 +77,62 @@ internal class PaymentMethodListViewController: UIViewController {
...
         navigationItem.leftBarButtonItem = cancelButton
     }
-    
+
     @objc private func cancelTapped() {
         viewModel.cancel()
     }
+
+    private func observeState() {
+        viewModel.statePublisher
```
robertdalmeida, line 86: nonblocking; a thought; 
Do you see value in using the Publisher here? (in the future). I see that the purpose of using a Publisher is to observe `state` change alone, which can be achieved with a `didSet` which for me is simpler to read? 
My only argument being that I don't need to load `Combine` skill in my brain 🧠  and LLMs may try to repeat this elsewhere and we may have a Combine heavy project.

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewController.swift
```diff
@@ -63,8 +77,62 @@ internal class PaymentMethodListViewController: UIViewController {
...
+                }
+            }.store(in: &cancellables)
+    }
+
+    private func startLoading(for paymentMethod: PaymentMethod) {
+        let expectedIdentifier = viewModel.listItemIdentifier(for: paymentMethod)
+        let listItem = listViewController.sections
+            .flatMap(\.items)
+            .first { $0.identifier == expectedIdentifier }
+        listItem?.startLoading()
```
robertdalmeida: nonblocking;
This seems to operating on the listItem type and none of the types are actually UIKit types for the Controller to operate on. 
Also ListSection is created by the viewModel and then passed via the state.
So maybe this can be triggered from the viewModel? 
By possibly making the [ListSection] a property of the viewModel, so it can be queried based on any of the current state.

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewModel.swift
```diff
@@ -7,96 +7,152 @@
...
+        state = .loaded(sections: listSections)
     }
-}
 
-extension PaymentMethodListViewModel: PaymentMethodListComponentDelegate {
+    // MARK: - Private
 
-    // MARK: - PaymentMethodListComponentDelegate
+    private func select(paymentMethod: PaymentMethod) {
+        guard let component = componentManager.buildComponent(for: paymentMethod) else { return }
```
robertdalmeida, line 86: nonblocking; 
Should we assert if we are unable to build the component?

AdyenDropIn/Utilities/ComponentManager.swift
```diff
@@ -152,24 +87,120 @@ internal final class ComponentManager {
...
+        guard let payment = context.payment,
+              let remainingAmount = order?.remainingAmount else { return }
+        
+        let updatedPayment = Payment(amount: remainingAmount, countryCode: payment.countryCode)
+        context.update(payment: updatedPayment)
+    }
+    
+    // MARK: - Section Builders
+    
+    var paidSection: PaymentMethodsSection {
```
robertdalmeida: nonblocking;
I see paidSection, storedSection & regularSection are used by a lazy `sections`
Maybe this too can be lazy?
Why? - if it is lazy, then it won't create a new object everytime we invoke this property. 
- It helps if hypothetically we would like to reference a value from any of the section directly. It would be the object that is bound to the UI - so we wouldn't accidently create a new one.

robertdalmeida (approved):

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewModel.swift
nauaros: Changed it! Renamed it to `idle`

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewController.swift
nauaros: This is a good suggestion. I moved all this logic to the view model.

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewModel.swift
nauaros: I'm gonna see what is the best course of action here. Probably we should have buildComponent not returnin any optional value

AdyenDropIn/Utilities/ComponentManager.swift
nauaros: Good catch, changed!

robertdalmeida (approved):

AdyenDropIn/Modules/PaymentMethodList/PaymentMethodListViewModel.swift
robertdalmeida: nonblocking; 
Just to continue this discussion, but doesn't block merging. Maybe it could be something to think about. 

From what i could trace, when the viewController is in `idle` state it still has the sections loaded. Which is why I think this is a bit unclear.

To me, state should indicate the current state of the ViewModel/View. So i think maybe something like below is readable to me. 

 `empty` a state where nothing is loaded. The initial state of the controller
`paymentInProgress(sections: [Section], payment: PaymentComponent)` a state where the view is loading a particular paymentComponent. 
`paymentList(sections: [Section])` - the controller shows a list of paymentOptions

Alternatively we could also extract `sections` outside state and then the two states could simply be. 
`paymentList` - shows a list of payments(sections could be a property in the viewModel and just be empyt). 
`paymentInProgress(payment: PaymentMethod)` - something in progress

PR #2420, opened 2026-02-27
Pass prefetched public key to components

robertdalmeida:
**Problem:** 
So in V6 public key is fetched during the initialization/building. 
1. This needs to be passed down to the components
2. Many components are initiated in flows withtout the Checkout root object. So the public key will not available in that flow. 

**Approach**
To ensure that Drop in flow still works.
Created a temporary type `PublicKeyProgramFlow` which has 2 states, `fetched` & `prefetched`.
`fetched` - from the flow using the v6 code. i.e. `Checkout`
`notFetched` - a flow where the public key is not available during creation eg: Dropin flow. 

**Future**
Once all code is migrated to being constructed using the Checkout root object. Then `notFetched` will not be used anymore in the code and then we can remove `PublicKeyProvider`, `PublicKeyConsumer` and remove the need of this enum `PublicKeyProgramFlow`.

TODO: I haven't updated the tests, if this approach seems acceptable then i'll update the tests accordingly.

### DIFF
--- a/Adyen/Core/Components/ComponentFactory.swift
+++ b/Adyen/Core/Components/ComponentFactory.swift
@@ -33,7 +33,7 @@ package protocol PaymentComponentFactory {
         with paymentMethod: Method,
         context: AdyenContext,
         configuration: Configuration,
-        publicKey: String
+        publicKey: PublicKeyFetchingProgramFlow
     ) -> Component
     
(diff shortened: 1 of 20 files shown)


robertdalmeida:
I'll redo this differently as now we would have the publicKey in AdyenContext.

PR #2422, opened 2026-03-03
AdyenContext created after publicKey & Checkoutid fetching

robertdalmeida:
Changes
1. Create analyticsProvider after `checkoutAttemptId` fetching.
- This means that checkoutAttempt ID will be available during initialization of the Analytics and will be passed to the adapted types internally.

2. Moving the creation of AdyenContext from inside CheckoutConfiguration to after `checkoutAttemptId` & `public key` fetching.

## Ticket [Optional]
<ticket>
COSDK-872
</ticket>


Adyen/Analytics/AnalyticsProvider/EventAnalyticsProvider.swift
```diff
@@ -8,40 +8,35 @@ import AdyenNetworking
...
+        checkoutAttemptId: String,
         batchInterval: TimeInterval = Constants.batchInterval
     ) {
         self.apiClient = apiClient
         self.eventDataSource = eventDataSource
         self.context = context
         self.batchInterval = batchInterval
+        self.checkoutAttemptId = checkoutAttemptId
+        // TODO: Robert: Bug: This can be called in a background thread and then the timer will never be scheduled.
+        startNextTimer()
```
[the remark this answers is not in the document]
robertdalmeida: You elaborated my todo. 🙈
WIll check this in the future not in this MR.

AdyenCheckout/CheckoutConfiguration/CheckoutConfiguration.swift
```diff
@@ -117,7 +129,25 @@ public struct CheckoutConfiguration {
         configurations[.action(actionType)] as? T
     }
 
-    // TODO: Robert: Create the analytical APIClient.
+    // TODO: Robert: Make public to private, This public is not needed. But currently using this to support providing analyticsAPIContext in the Integration Examples.
+    public static func createAnalyticsAPIContext(
```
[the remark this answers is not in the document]
robertdalmeida: Nice one. Thanks. made it package.

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
```diff
@@ -11,15 +11,10 @@ internal final class AnalyticsProvider: AnyAnalyticsProvider {
 
     // MARK: - Properties
 
-    // TODO: Robert: This should become required(non optional) and it need not be passed to the eventAnalyticsProvider, and should be available during init.
-    internal var checkoutAttemptId: String? {
-        didSet {
-            eventAnalyticsProvider?.checkoutAttemptId = checkoutAttemptId
-        }
-    }
+    internal var checkoutAttemptId: String?
```
atmamont, line 14: The comment removed says this should not be optional, is this correct? In the case fetch failed we use a static constant, any chance we can make this non-optional in analytics?

`AnalyticsConstants.swift`:
```swift
public static let fetchCheckoutAttemptIdFailed = "fetch-checkoutAttemptId-failed"
```

AdyenCheckout/CheckoutConfiguration/CheckoutConfiguration.swift
```diff
@@ -84,16 +85,28 @@ public struct CheckoutConfiguration {
...
+            analyticsApiContext: analyticsApiContext,
+            analyticsConfiguration: analyticsConfiguration,
+            configurations: configurations
+        )
     }
     
     internal init(
-        context: AdyenContext,
+        apiContext: APIContext,
+        amount: Amount,
```
erenbesel: remember to make this optional

Adyen/Core/AdyenContext/AdyenContext.swift
```diff
@@ -51,61 +52,46 @@ public final class AdyenContext {
...
             )
             return nil
         }
+        let analyticsApiClient = APIClient(apiContext: analyticsApiContext)
 
         var eventAnalyticsProvider: AnyEventAnalyticsProvider?
-        // TODO: Robert: If flag is true we create everything. If false, we send the RequestCheckoutAttemptID and Enrichment call with analytical data.
-        if analyticsConfiguration.isEnabled {
+
+        if let checkoutAttemptId,
```
erenbesel, line 82: If checkoutAttemptId fetch fails, what's the purpose of creating AnalyticsProvider now? 
If there is no attemptId I would expect no provider to be created any more. If there is an analytics provider object, that means we have checkoutAttemptId. And that would also mean attemptId not being optional in AnalyticsProvider too.

AdyenCheckout/CheckoutConfiguration/CheckoutConfiguration.swift
robertdalmeida: Made it optional. Thanks.

Adyen/Core/AdyenContext/AdyenContext.swift
robertdalmeida: Completely agree and the eventAnalyticsProvider need not be nil too.
I've added a todo for this particular thing at this moment. 

Why todo?: this triggers a lot more changes in tests(AdyenContext & AnalyticsProviderTests) and would better be done in the next MR on this topic.

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
robertdalmeida: This will be made non optional in the next MR where AnalyticsProvider's checkoutAttemptID will be required. 

For the usage of the constant - fetch-checkoutAttemptId-failed. 
This is specifically for the /payments call. For now, it will be like  `anlyticsProvider.checkoutAttmptID ?? 'fetch-checkoutAttemptId-failed". 

We will not send actual events to associate with this failed checkoutAttemptID.

atmamont: It's a good question, I do not know the details on this level. Do we skip such events at all? If yes, how do we track them?

atmamont (approved):

Adyen/Analytics/AnalyticsProvider/AnalyticsProvider.swift
robertdalmeida: All platforms web included have the same behavior of not sending events if checkout attempt id fails. The only tracking by sending the PaymentComponentData. If it contains the failure string `fetch-checkoutAttemptId-failed`.

AdyenCheckout/CheckoutProvider.swift
```diff
@@ -130,13 +102,44 @@ internal class CheckoutProvider: CheckoutProviding {
...
+
+    private func setupAdyenContext(
+        configuration: CheckoutConfiguration,
+        apiClient: APIClient
+    ) async throws -> AdyenContext {
+
+        let analyticsApiClient = configuration.analyticsApiContext.flatMap { APIClient(apiContext: $0) }
+
+        // TODO: Improvement: Suggestion: instead of the checkout provider being aware of something as specific as checkoutAttemptId.
+        // - This could be wrapped in something related to Analytics ex: await AnalyticsProvider.init() and internally fetch what is required for analytics.
```
erenbesel, line 123: Good idea, we can try this with the non optional checkoutAttemptId comment together

erenbesel (approved):

AdyenCheckout/CheckoutProvider.swift
robertdalmeida: Wil check if the changes are not too much then possibly combine or else one step at a time. I'm already facing flak for having big MRs. 🙈

PR #2425, opened 2026-03-04
PreselectedPaymentComponent - Sending DidLoad event and deleting unused file

robertdalmeida:
- Sending didLoad event.
- Deleting PreselectedPaymentComponent

# Testing 
- Verified the event attributes sent with V5 branch
- Added a unit test for this specific event.


erenbesel (approved):

AdyenDropIn/Modules/PreselectedPaymentMethod/PreselectedPaymentMethodViewModel.swift
```diff
@@ -188,6 +194,12 @@ extension PreselectedPaymentMethodViewModel: PaymentComponentDelegate {
             dropInFlowManager.fail(with: error, from: component)
         }
     }
+
+    public func sendDidLoadEvent() {
+        var infoEvent = AnalyticsEventInfo(component: "dropin", type: .rendered)
```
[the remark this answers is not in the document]
atmamont: Agreed, let's avoid using string literals and keep all of them defined in a single souce

Tests/IntegrationTests/DropIn Tests/PreselectedPaymentMethod/PreselectedPaymentMethodIntegrationTests.swift
```diff
@@ -72,6 +72,20 @@ struct PreselectedPaymentMethodIntegrationTests {
         #expect(sut.preselectedPaymentMethodRouter.presentComponentOnCancelCallsCount == 0)
     }
 
+    // MARK: - Sending event on didLoad
+
+    @Test("On preselected payment method load - we send an info event on load")
+    func paymentComponent_onLoad_sendsInfoEvent() throws {
+        // Given
+        let sut = SUT_MockedPaymentMethodRouter(type: .visa)
```
atmamont: SUT cannot be a mock. System under test should be a real data type, and we use mocks for it's dependencies if it makes sense to ease in its intantiation.

In this case you test the mock, not the [real view model behavior ](https://github.com/Adyen/adyen-ios/pull/2425/changes#diff-e61e17101546b228c31fe39039830284cefe180586c0186c04c292bc5848f3efR132)

atmamont (changes requested):

Tests/IntegrationTests/DropIn Tests/PreselectedPaymentMethod/PreselectedPaymentMethodIntegrationTests.swift
robertdalmeida: The sut isn't a mock but a system of viewModel, controller and router, contained in the struct. 
But i get that it isn't very readable as you said.
thus changed it to `makeSUT` passing specific mocked values that we are interested in from the call site.

atmamont (approved):

PR #2432, opened 2026-03-09
Update address picker to display single-line truncated addresses

atmamont:
# Summary 

This change updates the address picker component to match the Figma design.

# Demo 

| Figma | Implementation |
|--------|--------|
| image | image |

## Technical Information

- localization is not included
- the new address selection flow is not included

## Ticket
<ticket>
COSDK-859
</ticket>

## Checklist

- [x] Tested changes locally


atmamont:
@OscarSpruit tagging for visibility

Adyen/Model/PostalAddress.swift
```diff
@@ -192,6 +192,19 @@ extension PostalAddress {
             .replacingOccurrences(of: "\n", with: ", ")
     }
     
+    @_spi(AdyenInternal)
+    public func formattedSingleLine(using localizationParameters: LocalizationParameters?) -> String {
```
erenbesel: Let's make it package, it's a new code no need for spi

Adyen/Model/PostalAddress.swift
```diff
@@ -192,6 +192,19 @@ extension PostalAddress {
...
+        let components = [
+            [street, houseNumberOrName, apartment].compactMap { $0?.adyen.nilIfEmpty }.joined(separator: " "),
+            postalCode,
+            city,
+            stateOrProvince,
+            country.flatMap { countryName(for: $0, using: localizationParameters) }
+        ].compactMap { $0?.adyen.nilIfEmpty }
+        
+        return components.joined(separator: ", ")
```
erenbesel: Are the current tests covering this addition? Not sure if this is testable in our current integration tests but this change seems important to verify in tests

Adyen/Model/PostalAddress.swift
```diff
@@ -192,6 +192,19 @@ extension PostalAddress {
             .replacingOccurrences(of: "\n", with: ", ")
     }
     
+    @_spi(AdyenInternal)
+    public func formattedSingleLine(using localizationParameters: LocalizationParameters?) -> String {
+        let components = [
```
robertdalmeida: nonblocking; thought;
Is there a possibility to use; https://developer.apple.com/documentation/contacts/cnpostaladdressformatter. 

I assume that for US and Europe the way of putting the state before city and country may be different between them.

robertdalmeida (approved):

Adyen/Model/PostalAddress.swift
atmamont: Good point, I'll add new coverage for this part.

Adyen/Model/PostalAddress.swift
atmamont: Addressed f6c19da249b08f1cf303a9b8972a1e7aee82f11f

Adyen/Model/PostalAddress.swift
atmamont: Coverage added for both NL and US (see Robert's comment)

Adyen/Model/PostalAddress.swift
atmamont: Good catch, thank you.

Adyen/Model/PostalAddress.swift
```diff
@@ -192,6 +192,12 @@ extension PostalAddress {
             .replacingOccurrences(of: "\n", with: ", ")
     }
     
+    package func formattedSingleLine(using localizationParameters: LocalizationParameters?) -> String {
+        // Use CNPostalAddressFormatter for locale-aware formatting, then convert to single line
```
nauaros, line 196: Better to mark this as TODO to not lose it

nauaros (approved):

Adyen/Model/PostalAddress.swift
atmamont: This is a comment explaining the approach, not an action item.

PR #2444, opened 2026-03-16
Optionally create analytics provider in adyen context if checkout attemmpt ID is not available

robertdalmeida:
# Summary [Required]

Addressing a left over todo,

1. CheckoutAttemptID is nil then we don't create AnalyticsProvider in AdyenCheckout.
2. Updated the tests (This is where most of the PR changes are simply because we change AdyenCheckout)

## Checklist [Required]

- [x] Tested changes locally


Tests/IntegrationTests/Components Tests/PaymentComponent/PaymentComponentSubjectTests.swift
```diff
@@ -112,8 +118,7 @@ class PaymentComponentSubjectTests: XCTestCase {
     
     func test_submit_creates_sdkData_with_analytics() {
         // Given
-        let expectedCheckoutAttemptId = "d06da733-ec41-4739-a532-5e8deab1262e16547639430681e1b021221a98c4bf13f7366b30fec4b376cc8450067ff98998682dd24fc9bda"
-        analyticsProviderMock.checkoutAttemptId = expectedCheckoutAttemptId
+        let expectedCheckoutAttemptId = AnalyticsProviderMock.testCheckoutAttemptId
```
nauaros, line 121: better like this!

nauaros (approved):

Adyen/Core/AdyenContext/AdyenContext.swift
```diff
@@ -66,31 +66,22 @@ public final class AdyenContext {
         checkoutAttemptId: String?,
         analyticsConfiguration: AnalyticsConfiguration
     ) -> AnyAnalyticsProvider? {
-        guard let analyticsApiContext else {
-            AdyenAssertion.assertionFailure(
-                message: "AnalyticsProvider couldn't be created as AnalyticsAPIContext is not available."
-            )
+        guard let analyticsApiContext,
+              let checkoutAttemptId,
+              analyticsConfiguration.isEnabled else {
```
erenbesel: Adding isEnabled flag check here changes the logic I'm afraid. The isEnabled logic only applies to creation of `EventAnalyticsProvider`.  The regular `AnalyticsProvider` should still be created if there is a checkoutAttemptID and isEnabled is true or false.

So AnalyticsProvider --> checkoutAttemptid, analyticsApiContext
EventAnalyticsProvider --> the above + isEnabled

erenbesel (commented):
Great update! Small comment

Adyen/Core/AdyenContext/AdyenContext.swift
robertdalmeida: A very nice catch, thank you. I did change the logic unintentionally. 
I've reverted that back. So now it respects the `analyticsConfiguration.isEnabled` flag

erenbesel (approved):

PR #2455, opened 2026-03-18
v6: Swapping to async apiClient

erenbesel:
# Summary [Required]
In our efforts to move to async/await, first and major part of the swap to using async apiclient via `AsyncAPIClientProtocol`.
Since the return values of the old `perform` and the new async `perform` don't match, this PR also introduces a convenience `performAsync` method to match the old signature.

Currently RetryAPIClient and UniqueAssetAPIClient does not have async support so they will be done later, not as important.

## Ticket [Optional]
<ticket>
COSDK-1056
</ticket>

## Checklist [Required]

- [x] Tested changes locally


AdyenSession/Session+ActionComponentDelegate.swift
```diff
@@ -56,14 +56,20 @@ extension Session {
...
             sessionData: state.data,
             paymentData: actionComponentData.paymentData,
             details: actionComponentData.details
         )
-        apiClient.perform(request) { @MainActor [weak self] in
-            self?.handle(paymentResponseResult: $0, for: component)
+        Task {
+            do {
+                let response: PaymentsResponse = try await apiClient.performAsync(request)
+                handle(paymentResponse: response, for: component, in: nil)
```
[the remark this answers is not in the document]
erenbesel: Good catch

Adyen/Analytics/AnalyticsProvider/EventAnalyticsProvider.swift
```diff
@@ -62,16 +67,12 @@ internal final class EventAnalyticsProvider: AnyEventAnalyticsProvider {
...
-            guard let self else { return }
-            // clear the sent events on successful send
-            switch result {
-            case .success:
-                self.removeEvents(sentBy: request)
-                self.startNextTimer()
-            case .failure:
-                break
-            }
+        Task { @MainActor in
```
atmamont, line 70: While the apiCLient itself is async, everything else happend on the main thread. Does it have to be this way?
Imagine `removeEvents` and `startNextTimer` are resource-heavy, it may cause UI stuttering.

It also opens up a door for someone in the future chipping in something in between, like `parse100MegabytesJsonResponse` to this block.

AdyenCard/Utilities/BinLookup/BinLookupService.swift
```diff
@@ -52,9 +52,14 @@ internal final class BinLookupService: AnyBinLookupService {
         }
         
         let request = BinLookupRequest(encryptedBin: encryptedBin, supportedBrands: supportedCardTypes, type: binLookupType)
-        apiClient.perform(request) { [weak self] result in
-            _ = result.map { self?.cache[bin] = $0 }
-            completion(result)
+        Task { @MainActor in
+            do {
+                let response: BinLookupResponse = try await apiClient.performAsync(request)
+                cache[bin] = response
```
atmamont, line 62: In theory this can be so massive it will cause UI hiccups.

atmamont (commented):
Good stuff, there are some main thread related comments from my side

AdyenSession/Session+StoredPaymentMethodsDelegate.swift
```diff
@@ -21,12 +21,12 @@ extension Session: SessionStoredPaymentMethodsDelegate {
...
+            do {
+                _ = try await apiClient.performAsync(request)
                 completion(true)
-            case let .failure(error):
-                self?.showAlert(with: error, on: dropInComponent)
+            } catch {
+                showAlert(with: error, on: dropInComponent)
                 completion(false)
             }
         }
```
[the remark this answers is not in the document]
erenbesel: Nice one

descorp (approved):

AdyenCard/Utilities/BinLookup/BinLookupService.swift
erenbesel: Considered this. But at the moment it's a very small operation, so decided not worthy of adding thread safety complexity here.

Adyen/Analytics/AnalyticsProvider/EventAnalyticsProvider.swift
erenbesel: The timer part has be on the main thread. So we would only be thinking of removeEvents which is at the moment not a big job (and thread safe in itself). Similar to the other one, adding complexity for a future potential.

```

Task {
    do {
        _ = try await apiClient.performAsync(request)
        removeEvents(sentBy: request)
        await MainActor.run { startNextTimer() }
    } catch {}
}

```
 Looks worse 

If like you said, someone else wanted to add a heavy operation, they should see current the intent entering the block and make necessary modifications

AdyenCard/Utilities/BinLookup/BinLookupService.swift
atmamont: It is your call, I'd be as defensive as possible with this topic on iOS.

atmamont (approved):

Adyen/Analytics/AnalyticsProvider/EventAnalyticsProvider.swift
```diff
@@ -40,9 +40,14 @@ internal final class EventAnalyticsProvider: AnyEventAnalyticsProvider {
     }
     
     deinit {
-        // attempt to send remaining events on deallocation
         batchTimer?.invalidate()
-        sendEventsIfNeeded()
+        // fire-and-forget remaining events without capturing self
+        if let request = requestWithAllEvents() {
+            let apiClient = apiClient
+            Task { @MainActor in
```
robertdalmeida: nonblocking; but i think mainActor is not needed here? Just spawning a task should be enough.

Adyen/Core/APIClient/AsyncAPIClientProtocol+Extensions.swift
```diff
@@ -0,0 +1,22 @@
...
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import AdyenNetworking
+import Foundation
+
+extension AsyncAPIClientProtocol {
+
+    /// Convenience method to match the signature of the response to the non async version.
+    package func performAsync<R: Request>(_ request: R) async throws -> R.ResponseType {
```
robertdalmeida, line 13: We could potentially use the overloaded `perform` method directly right? 
It does build for me except for one error where the compiler couldn't figure our the type.
```
let response: RedirectDetails = try await apiClient.perform(request)
```

Reason: having async in the method name isn't a standard practice i would guess, the method signature should indicate it is async.

erenbesel: Yeah I couldn't trust xcode here and didn't want to create confusion as well. I'm not a huge fan of async in the name either

AdyenActions/Components/3DS2/Fingerprint Submitter/ThreeDS2FingerprintSubmitter.swift
```diff
@@ -43,21 +43,14 @@ internal final class ThreeDS2FingerprintSubmitter: AnyThreeDS2FingerprintSubmitt
...
-        _ result: Result<Submit3DS2FingerprintResponse, Swift.Error>,
-        completionHandler: (Result<ThreeDSActionHandlerResult, Swift.Error>) -> Void
-    ) {
-        switch result {
-        case let .success(response):
-            completionHandler(.success(response.result))
-        case let .failure(error):
-            sendApiErrorEvent()
-            completionHandler(.failure(error))
+        Task { @MainActor in
```
robertdalmeida, line 46: Do we get an error or a waning if we use main actor? Why can't we simply use Task { and inherit the task context.
also do we need to capture self here?

AdyenCard/Utilities/BinLookup/BinLookupService.swift
```diff
@@ -23,13 +23,13 @@ internal final class BinLookupService: AnyBinLookupService {
...
     
-    private let apiClient: APIClientProtocol
+    private let apiClient: AsyncAPIClientProtocol
 
     private var cache = [String: BinLookupResponse]()
     
     private let binLookupType: BinLookupRequestType
     
-    internal init(publicKey: String, apiClient: APIClientProtocol, binLookupType: BinLookupRequestType) {
+    internal init(publicKey: String, apiClient: AsyncAPIClientProtocol, binLookupType: BinLookupRequestType) {
```
nauaros: non-blocking; Can be nice to do some formatting here so its easier to read and follow the multiline pattern

AdyenCard/Utilities/BinLookup/BinLookupService.swift
nauaros: Perfect way would be to convert this `BinLookupService` into an `actor` so it's thread safe. It may trigger many more changes for the service consumers

AdyenActions/Components/3DS2/Fingerprint Submitter/ThreeDS2FingerprintSubmitter.swift
erenbesel: You mean if we don't use main actor at all here? The class is not main actor so the return won't be on main and will cause a crash as the next step is the challenge which needs UI, right?

AdyenSession/API/SessionAPIClient.swift
```diff
@@ -11,39 +11,42 @@ import Foundation
...
+        return response
+    }
+    
+    internal func perform<R: Request>(_ request: R) async throws -> HTTPResponse<R.ResponseType> where R.ResponseType == DownloadResponse {
+        try await apiClient.perform(request)
+    }
+    
+    private func updateSession(with response: Any) async {
+        if let response = response as? SessionDataAware {
+            await onSessionDataUpdate?(response)
```
robertdalmeida: Is there any reasoning why we want to run this on the main actor for the `onSessionDataUpdate` and `onSessionResultUpdate`? 
I like that it is Sendable, but maybe @ mainactor is not necessary here? we seem to be updating a model object (state) anyways.
I don't see xcode complaining if i remove it actually. 

In principle, MainActor sparingly. 
When we make the swift 6 migration we can think about just using Actor for these cases rather than main actor - If we need to handle isolation.

AdyenSession/Session+PaymentComponentDelegate.swift
```diff
@@ -65,27 +65,18 @@ extension Session {
             sessionData: state.data,
             data: paymentComponentData
         )
-        apiClient.perform(request) { @MainActor [weak self] in
-            self?.handle(paymentResponseResult: $0, for: component, in: dropInComponent)
+        Task {
```
robertdalmeida, line 68: maybe we could capture self weakly in Tasks that we spawn? 
(just for safety- i do assume at times it is not necessary - but tracing that is a pain.

AdyenActions/Components/3DS2/Fingerprint Submitter/ThreeDS2FingerprintSubmitter.swift
robertdalmeida: Yeah, meaning no mainActor for this type at all. 
This type isn't presenting the challenge - it is submitting the fingerprint(nothing requiring main thread) - whoever that guy is who presents the challenge - should be acting in the main thread(MainActor). 

But then again, no harm in this now, but this may add up.

robertdalmeida (approved):

nauaros (approved):

AdyenSession/API/SessionAPIClient.swift
erenbesel: Good point, it was to be safe since the previous implementation always return back to main. But this should be ok to remove.

AdyenActions/Components/3DS2/Fingerprint Submitter/ThreeDS2FingerprintSubmitter.swift
erenbesel: On this one it feels too risky to remove it here and add main actor on the callers

AdyenCard/Utilities/BinLookup/BinLookupService.swift
erenbesel: Done

Adyen/Core/APIClient/AsyncAPIClientProtocol+Extensions.swift
erenbesel: Tried it, we would need to specify the return type every time and won't be able to write `let response = try await apiClient.perform(request)` as xcode cant infer the response type with same signatures with 2 different return types.

Adyen/Analytics/AnalyticsProvider/EventAnalyticsProvider.swift
erenbesel: Good catch

robertdalmeida (approved):

AdyenActions/Components/3DS2/Fingerprint Submitter/ThreeDS2FingerprintSubmitter.swift
nauaros: I think @robertdalmeida means making `ThreeDS2FingerprintSubmitter` main actor so the task inherits the context.

Adyen/Core/APIClient/AsyncAPIClientProtocol+Extensions.swift
robertdalmeida: XCode was able to infer it at all places in this MR for me except for `RedirectDetails` type for which we can explicitly specify type as below. But its totally fine to keep it as it is. Just wanted to check what you thought of the possiblity. 

let response: RedirectDetails = try await apiClient.perform(request)

AdyenActions/Components/3DS2/Fingerprint Submitter/ThreeDS2FingerprintSubmitter.swift
robertdalmeida: Actually no, i would have preferred if this wasn't on the main actor at all. :) 
But all these can be revisitied when we want to optimize for this or even when we do swift 6 stuff.

Adyen/Core/APIClient/AsyncAPIClientProtocol+Extensions.swift
erenbesel: Exactly we would need to specify the response type each time

PR #2458, opened 2026-03-19
Add new StoredCardInputView UI (Screen only) nothing connected  (1/6)

robertdalmeida:
The first screen in the stored card flow. 

- Adding the StoredCardInput ViewController and ViewModel. (This isn't final yet) but tweaks will do it later.
- Moved CardImageView from `AdyenDropIn` to `AdyenUI` (Added some additional tests here) - so these lines are not counted as changed in this MR ;) 

## Some stats
**Actual functional code**: Only 487 lines (42% of total)
**Test**: 684 lines (58% of total)

## In the next MR, 
- Will try to connect this and display and remove the StoredAlertManager. 
- Also todos which i will keep adding on this topic. 

## Checklist [Required]

- [x] Tested changes locally
- [x] Added/updated unit test

COSDK-1140 

How it will look when connected. But not with only this PR. 
Simulator Screen Recording - co-v6-(16+,26 1) - 2026-02-26 at 16 20 36


Tests/IntegrationTests/Card Tests/StoredCardInputViewModelTests.swift
```diff
@@ -0,0 +1,600 @@
...
+        AmountData(amount: Amount(value: 100, currencyCode: "EUR"), expectedFormatted: "€1.00"),
+        AmountData(amount: Amount(value: 14098, currencyCode: "USD"), expectedFormatted: "$140.98"),
+        AmountData(amount: Amount(value: 0, currencyCode: "GBP"), expectedFormatted: "£0.00")
+    ]
+}
+
+// MARK: - StoredCardInputViewControllerTests
+
+@MainActor
+struct StoredCardInputViewControllerTests {
```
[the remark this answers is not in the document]
robertdalmeida: Happenned due to moving to another branch. But it has been fixed.

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewController.swift
```diff
@@ -0,0 +1,236 @@
...
+        buttonsStackView.addArrangedSubview(primaryButton)
+
+        configureConstraints()
+        configureContent()
+        setupNavigationBackButton()
+        configurePresentationSheet()
+    }
+
+    private func configurePresentationSheet() {
+        // Adding this to avoid swiping down to dismiss the controller.
```
atmamont: Let's remove the comment and rename this method to 
```swift
func disablePresentationSheetInteractiveDismissal()
```
?

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewController.swift
```diff
@@ -0,0 +1,236 @@
...
+            scrollView.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor),
+            scrollView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            scrollView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            scrollView.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor),
+
+            contentStackView.topAnchor.constraint(equalTo: scrollView.topAnchor),
+            contentStackView.leadingAnchor.constraint(equalTo: scrollView.leadingAnchor, constant: Constants.contentPadding),
+            contentStackView.trailingAnchor.constraint(equalTo: scrollView.trailingAnchor, constant: -Constants.contentPadding),
+            contentStackView.bottomAnchor.constraint(equalTo: scrollView.bottomAnchor, constant: -Constants.buttonsBottomPadding),
+            contentStackView.widthAnchor.constraint(equalTo: scrollView.widthAnchor, constant: -Constants.contentPadding * 2)
```
atmamont: Do we need to set width contrants explicitly if we have leading and trailing constraints set?

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewController.swift
```diff
@@ -0,0 +1,236 @@
...
+        ])
+    }
+
+    private func setupNavigationBackButton() {
+        let backButton = UIBarButtonItem(
+            image: UIImage(systemName: Constants.leftChevronImage),
```
atmamont: Would it be correct for right-to-left languages?  Let's use `chevron.backward` instead?

Image

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewController.swift
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    private func configureContent() {
+        titleLabel.text = viewModel.titleText
+        subtitleLabel.attributedText = viewModel.subtitleText
+        primaryButton.title = viewModel.submitButtonTitle
+    }
+
+    private func updateLoadingState(_ isLoading: Bool) {
+        primaryButton.isUserInteractionEnabled = !isLoading
```
atmamont: This is usually redundant when used along the `isEnabled`, any particular reason for this? Apart from redundancy aspect, it hides the element from the accessibility framework and this might be harmful.

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewController.swift
```diff
@@ -0,0 +1,236 @@
...
+        primaryButton.isEnabled = !isLoading
+        primaryButton.showsActivityIndicator = isLoading
+    }
+
+    private func setupViewInstructionHandler() {
+        viewModel.onViewInstruction = { [weak self] instruction in
```
atmamont: `onViewState`?

atmamont (commented):
Great stuff, was easy to review (I didn't review tests btw).
As this is a UI change, it would be really helpful to decorate this PR with some images!

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewController.swift
robertdalmeida: They do not indeed. Good catch. Thanks.

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewController.swift
robertdalmeida: Whoah, I honestly never thought about this. But this makes total sense, I'm so curious how apps look in R to L languages. I've changed it to backward. 
Thanks.

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewController.swift
robertdalmeida: Removed it. Thanks.

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewController.swift
robertdalmeida: I would say these are smaller instructions to the view from the viewModel. 
I may need to introduce state for the entire viewController, but after we get the whole picture running.

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewController.swift
robertdalmeida: I'm sorry but for me, 
// Adding this to avoid swiping down to dismiss the controller
gives a more targeted understanding about the user behavior than
// disablePresentationSheetInteractiveDismissal

To give a bit more reasoning, 
on reading the comment, i would immediately know swiping down dismisses the view if this is remved. 

on reading the disablePresentationSheetInteractiveDismissal(), it is difficult for me to come to that specific understanding.

Alternatively, i would be fine to rename it to something like this?? wdyt?
disableSwipeDownToDismissScreen()

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewController.swift
atmamont: Just trying to think forward a bit. Imagine we convert this from a closure-based to an observer based, then this doesn't hold up. It would be safer to use "state", later on it will land smoothly on the SwiftUI architecture of choice.

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewController.swift
atmamont: 👍 
Yes, I would always pick a better method name to improve clarity and reduce ambiguity over patching naming with code comments.

robertdalmeida: disableSwipeDownToDismissScreen it is . thanks.

Adyen/Assets/en-US.lproj/Localizable.strings
```diff
@@ -33,6 +33,7 @@
 "adyen.card.cvcItem.invalid" = "Invalid CVC / CVV format";
 "adyen.card.cvcItem.title" = "Security code";
 "adyen.card.cvcItem.placeholder" = "123";
+"adyen.card.securityCode.description" = "Enter the security code for %@ to complete the payment of %@";
```
nauaros, line 36: This localization will be overriden when the new key is introduced in the localization repo (shared by all platforms). Probably, it is good to take it into account and a TODO wherever it is used.

Adyen/Assets/Generated/LocalizationKey.swift
```diff
@@ -76,6 +76,10 @@ public struct LocalizationKey {
     public static let cardCvcItemInvalid = LocalizationKey(key: "adyen.card.cvcItem.invalid")
     /// Security code
     public static let cardCvcItemTitle = LocalizationKey(key: "adyen.card.cvcItem.title")
+    /// Enter security code
+    public static let cardComponentInputTitle = LocalizationKey(key: "adyen.card.securityCode.title")
```
nauaros: Same here. This keys are autogenerated with a script. We can go through it together

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewController.swift
```diff
@@ -0,0 +1,233 @@
...
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+import UIKit
+@_spi(AdyenInternal) import Adyen
+
+#if canImport(AdyenUI)
+    @_spi(AdyenInternal) import AdyenUI
```
nauaros, line 13: non-blocking and only if possible;
It there is anything `spi `in AdyenUI that can be easily converted into `package`, let's take the opportunity so no new classes use `spi`

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewController.swift
```diff
@@ -0,0 +1,233 @@
...
+
+    private func setupView() {
+        view.backgroundColor = theme.colors.background
+        // TODO: Robert: StoredView: 🐞 The scroll view isn't working in this screen. Needs separate investigation.
+        view.addSubview(scrollView)
+        scrollView.addSubview(contentStackView)
+
+        contentStackView.addArrangedSubview(topContentStackView)
+        contentStackView.addArrangedSubview(securityCodeItemView)
+        contentStackView.addArrangedSubview(buttonsStackView)
```
nauaros: 💡 Suggestion (it's also good like this)

```swift
[
  topContentStackView,
  securityCodeItemView,
  buttonsStackView
].forEach { 
  contentStackView.addArrangedSubview($0) 
}
```

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewController.swift
nauaros: Yeah, a published state from the viewModel would easily fit here.

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewController.swift
```diff
@@ -0,0 +1,233 @@
...
+                securityCodeItemView.resignFirstResponder()
+                securityCodeItemView.showValidation()
+            }
+        }
+    }
+
+    // MARK: - User Actions
+
+    @objc private func primaryButtonTapped() {
+        Task { @MainActor [weak self] in
```
nauaros, line 240: 💡 What about making the view controller as `@MainActor` so the Task inherits the context?

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewModel.swift
```diff
@@ -0,0 +1,221 @@
...
+#if canImport(AdyenUI)
+    @_spi(AdyenInternal) import AdyenUI
+#endif
+#if canImport(AdyenEncryption)
+    import AdyenEncryption
+#endif
+import Foundation
+
+// sourcery: AutoMockable
+internal protocol StoredCardInputViewModelProtocol: AnyObject {
```
nauaros, line 18: And here it would safe to assume this is main actor. What do you think?

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewModel.swift
```diff
@@ -0,0 +1,221 @@
...
+        return attributed
+    }
+
+    private var formattedAmount: String {
+        guard let amount,
+              let formatted = AmountFormatter.formatted(
+                  amount: amount.value,
+                  currencyCode: amount.currencyCode
+              ) else {
+            return ""
```
nauaros, line 132: Safe to assume this until Arjen comes up with an alternative for the amount not being there. I also did it in the Payment Method List.

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewModel.swift
```diff
@@ -0,0 +1,221 @@
...
+                    throw error
+                }
+            }()
+            cardDetailsCompletionHandler?(.success(encryptedCardDetails))
+        } catch {
+            cardDetailsCompletionHandler?(.failure(error))
+        }
+    }
+
+    @MainActor internal func resetSecurityCodeField() {
```
nauaros, line 185: When is this reset func being called?

nauaros: I see now it is being called on dismiss. What is the reason for it?

nauaros (approved):
Nice job Robert! 💪 Just a couple of comments but nothing blocking.

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewController.swift
robertdalmeida: making the whole view controller to adapt based on a `state`. : 
in priniciple i like this pattern for swiftui views & I look forward to this type of implementation as this would make the UI a state machine. But i cannot do this due to the dependency on the `FormCardSecurityCodeItem`

The only other "reactive" way of implementing the screen is the inProgress variable. That i have converted now.

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewController.swift
robertdalmeida: I could remove one with ViewIdentifier, but the others if i change, have cascading effects. So i could remove one @spi(adyenInternal). But thanks for making me think in this direction with every pr.

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewController.swift
robertdalmeida: I have changed it. Thanks. 

image

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewController.swift
robertdalmeida: ViewControllers are implicitly `@MainActor`. They inherit from UIVIewController.

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewModel.swift
robertdalmeida: I've made changes to ensure fine grained sec for methods that explicitly need to be called on the main actor. 

For example, why do i need `title`, `subtitle`, `cardItem` to be called on the main actor from the ViewController. 
But i do expect the dismiss & submit to be called from a main actor.

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewModel.swift
robertdalmeida: I have a todo for this. it exists also for the preselected screen. This will indeed change in a bit.

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewModel.swift
robertdalmeida: When the user navigates away from the screen like on a card challenge. 
The screen on the background has the security code visible. 
This is to clear it after consuming it or navigating away.

Also, if we clear it we can resuse the viewcontroller if the user decides to come back to the security code screen after going back.

Adyen/Assets/en-US.lproj/Localizable.strings
robertdalmeida: Good point, Thanks, i didn't notice that indeed. I've corrected the namig of the variable and added a todo for the localization keys in general for this component.

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewModel.swift
nauaros: Makes sense!

nauaros (approved):

PR #2459, opened 2026-03-20
MainActor on DropInComponent

erenbesel:
# Summary [Required]
- Add MainActor to DropIn and a few remaining UI bound components.
- Add weak self to some Task blocks for safety.

## Ticket [Optional]
<ticket>
COSDK-1050
</ticket>

## Checklist [Required]

- [ ] Tested changes locally

### DIFF
--- a/Adyen/Analytics/AnalyticsProvider/EventAnalyticsProvider.swift
+++ b/Adyen/Analytics/AnalyticsProvider/EventAnalyticsProvider.swift
@@ -67,11 +67,11 @@ internal final class EventAnalyticsProvider: AnyEventAnalyticsProvider {
     internal func sendEventsIfNeeded() {
         guard let request = requestWithAllEvents() else { return }
         
-        Task { @MainActor in
+        Task {
             do {
                 _ = try await apiClient.performAsync(request)
                 removeEvents(sentBy: request)
-                startNextTimer()
+                await MainActor.run { startNextTimer() }
             } catch {}
         }
     }
--- a/Adyen/Analytics/AnalyticsProvider/EventAnalyticsProvider.swift
+++ b/Adyen/Analytics/AnalyticsProvider/EventAnalyticsProvider.swift
@@ -35,8 +35,8 @@ internal final class EventAnalyticsProvider: AnyEventAnalyticsProvider {
         self.context = context
         self.batchInterval = batchInterval
         self.checkoutAttemptId = checkoutAttemptId
-        // TODO: Robert: Bug: This can be called in a background thread and then the timer will never be scheduled.
-        startNextTimer()
+        
+        Task { await startNextTimer() }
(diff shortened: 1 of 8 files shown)


atmamont (approved):

robertdalmeida (approved):

Adyen/Analytics/AnalyticsProvider/EventAnalyticsProvider.swift
```diff
@@ -67,11 +67,11 @@ internal final class EventAnalyticsProvider: AnyEventAnalyticsProvider {
     internal func sendEventsIfNeeded() {
         guard let request = requestWithAllEvents() else { return }
         
-        Task { @MainActor in
+        Task {
             do {
                 _ = try await apiClient.performAsync(request)
                 removeEvents(sentBy: request)
-                startNextTimer()
+                await MainActor.run { startNextTimer() }
```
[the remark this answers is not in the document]
nauaros: This comment is nonsense right? @erenbesel

nauaros (approved):

Adyen/Analytics/AnalyticsProvider/EventAnalyticsProvider.swift
robertdalmeida: Just renoticed this today; something we can think about. Monday morning fresh look;

This versus annotating the startNextTimer() with @ MainActor. 

I would prefer if we annotate startNextTimer() with @ MainActor, instead of explicitly using MainActor.run. 

Why? 
1. if we annotate `startNextTimer` then any caller in the future will be forced to call this method on the mainActor. Currently they are still allowed to make the mistake of calling in a non main thread. 
2. In the init() we directly call startNextTimer(i had kept a todo there about this). But the init is not annotated to be run on the mainActor, so if this type in isolation is called in a non main thread the timer would not be started. 
````
init() {
Task { @MainActor in 
 startNextTimer()
}}

func sendEventsIfNeeded() {
Task {
   // send events. 
   // startNextTimer()
}}

@MainActor func startNextTimer() {}
````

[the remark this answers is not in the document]
atmamont: Yes, best we can do is 👎 
Gemini can then use it to fine-tune their code review agent

erenbesel: That was part of a bigger change to split timer logic to an entire new layer as I mentioned but we can add that small change here indeed

robertdalmeida: If this is eventually changing then no change needed here. Just put my thoughts.

robertdalmeida (approved):

PR #2462, opened 2026-03-25
Threeds2Component theme migration - AdyenTheme to ADYAppearanceConfiguration

robertdalmeida:
# Summary [Required]
- An initial version to migrate AdyenTheme to ADYAppearanceConfiguration. -> The `ADYAppearanceConfigurationBuilder` : Note this is not perfect and finalized, there is still some dicussions i need to have with Android and Arjen about this mapping, and it is based on best effort right now. Will iterate on this, but those changes can be more targeted and specific. 

- Passing AdyenTheme to ThreeDS2Component, even in the default configuration - i think theme is minimally required. @erenbesel  please confirm this part specifically. 

- Took the liberty to add a test app configuration to switch some themes. I can remove this if not required, but thought this would be something good to have for the test app in general. Especially during development.

https://github.com/user-attachments/assets/cd1d6b69-d646-4bca-936b-21cdf06bde9e

https://github.com/user-attachments/assets/8381c588-65b7-4702-a2b4-25fa8a81b85b

https://github.com/user-attachments/assets/eccbedeb-ee0f-443d-b0eb-2545aeca2c82

https://github.com/user-attachments/assets/3bd882bc-74da-425c-b0b2-03bf43f09eb9

## Ticket [Optional]
COSDK-1099

## Checklist [Required]
- [x] Tested changes locally

### DIFF
--- a/AdyenActions/Components/3DS2/Action handlers/3DS2+Delegated Authentication/ThreeDS2PlusDACoreActionHandler.swift
+++ b/AdyenActions/Components/3DS2/Action handlers/3DS2+Delegated Authentication/ThreeDS2PlusDACoreActionHandler.swift
@@ -12,6 +12,7 @@ internal typealias VoidHandler = () -> Void
     @_spi(AdyenInternal) import Adyen
     import Adyen3DS2
     import AdyenAuthentication
+    import AdyenUI
     import Foundation
     import UIKit
     
@@ -44,7 +45,7 @@ internal typealias VoidHandler = () -> Void
         internal convenience init(
             context: AdyenContext,
             service: ThreeDSService,
-            appearanceConfiguration: ADYAppearanceConfiguration,
+            theme: AdyenTheme,
             delegatedAuthenticationConfiguration: ThreeDS2ActionConfiguration.DelegatedAuthentication
         ) {
             self.init(
@@ -55,7 +56,7 @@ internal typealias VoidHandler = () -> Void
                     localizedParameters: delegatedAuthenticationConfiguration.localizationParameters,
                     context: context
                 ),
-                appearanceConfiguration: appearanceConfiguration,
+                theme: theme,
                 style: delegatedAuthenticationConfiguration.style,
                 delegatedAuthenticationConfiguration: delegatedAuthenticationConfiguration
             )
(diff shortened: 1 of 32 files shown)


AdyenUI/AdyenTheme.swift
Robert-SD, line 1: Can we rename the `AdyenTheme` into `CheckoutTheme` similar as on [Android](https://github.com/Adyen/adyen-android/blob/6109c2778bf10944a49f7c560b31e66218845c32/ui/src/main/java/com/adyen/checkout/ui/theme/CheckoutTheme.kt#L15)? We could align the names with the other public api building blocks like `CheckoutConfiguration`, `CheckoutResult` etc.

AdyenActions/Components/3DS2/Action handlers/3DS2+Delegated Authentication/ThreeDS2PlusDACoreActionHandler.swift
```diff
@@ -44,7 +47,7 @@ internal typealias VoidHandler = () -> Void
         internal convenience init(
             context: AdyenContext,
             service: ThreeDSService,
-            appearanceConfiguration: ADYAppearanceConfiguration,
+            theme: AdyenTheme,
```
erenbesel, line 49: Asking to clarify. Since all the inner 3ds2 classes work with ADYAppearanceConfiguration, instead of passing AdyenTheme to all of them, would it make sense to keep ADYAppearanceConfiguration for all of them except one on the top level that will be mapped from AdyenTheme? So we won't have to add AdyenTheme to all of these files?

robertdalmeida: Excellent question. 
On this topic i had forgotten to [remove some unnecessary imports of Adyen3ds2](https://github.com/Adyen/adyen-ios/pull/2462/changes/e252d8be05a8f4cef89a8a442bcfe656f1800f04). Removed them now - thanks. 

Reasoning;
1. ADYAppearanceConfiguration is part of "external" sdk so want to keep its footprint limited to just one layer/folder "legacy SDK". 

2. AdyenTheme on the other hand belongs to this project so passing that around until it reaches the layer of the legacy sdk and then it is built, just when it is required. 

Hope this reasoning make sense.

AdyenUI/AdyenTheme.swift
robertdalmeida: I agree, i like then name CheckoutTheme which is company agnostic. 

@erenbesel & @nauaros if i have your blessings on this? I could do this in [the parallel PR](https://github.com/Adyen/adyen-ios/pull/2465).

erenbesel: Absolutely!

AdyenActions/Components/3DS2/Action handlers/3DS2+Delegated Authentication/ThreeDS2PlusDACoreActionHandler.swift
erenbesel: Great, nice!

erenbesel (approved):
🚀

atmamont:
Great effort
Thanks for adding switchable themes to the Demo app! 💪

atmamont (approved):

PR #2467, opened 2026-03-26
Localization: fix simulator localizations not being picked up in debug mode

atmamont:
# Summary
This PR:
- fixes the current "legacy" localization resolution in debug/simulator runs where missing strings can be surfaced as uppercased keys instead of falling through to the expected fallback layers. The resolver now keeps treating those debug placeholders as missing values, so localization continues through the existing fallback chain instead of rendering raw localization keys in the UI. This allows us to use Simulator debug capabilities to find missing localizations. This is also addressed in an automated way by introducing [localization coverage unit tests](https://github.com/Adyen/adyen-ios/pull/2464)

# Demo

| Debug localizations ON before the fix | Debug localizations ON after the fix |
|-|-|
| image |  image |

## Technical Information
Main localization logic changes in this PR:
- preserve and document the current legacy fallback order: `Bundle.main` -> optional `LocalizationParameters.bundle` -> SDK localized resources -> SDK `en-US` -> raw key
- keep enforced-locale resolution working by reading the requested `.lproj` from each candidate bundle before falling back to SDK resources
- treat Xcode's `Show non-localized strings` uppercase placeholder output as a missing translation so fallback continues correctly
- keep custom table and key-separator lookup behavior unchanged
- add focused localization test coverage for the debug placeholder case

## Testing Instructions
- Enable Xcode's debug option to show non-localized strings for `AdyenUIHost`
- Launch the Demo app and open a flow that renders SDK-owned strings such as the preselected payment method screen
- Verify missing app-level translations no longer render as raw localization keys and instead fall back to the SDK translation chain
- Run `LocalizationTests` and confirm the debug placeholder case passes

## Ticket
<ticket>
COSDK-1097
</ticket>

## Checklist

- [x] Tested changes locally


Adyen/Utilities/Localization.swift
```diff
@@ -29,19 +30,19 @@ private struct LocalizationInput {
 /// - Returns: The localized string for the given key, or the key itself if the localized string could not be found.
 @_spi(AdyenInternal)
 public func localizedString(_ key: LocalizationKey, _ parameters: LocalizationParameters?, _ arguments: CVarArg...) -> String {
```
erenbesel: Maybe a good time to make package, only if it won't require cascading changes

erenbesel (approved):

nauaros (approved):

Adyen/Utilities/Localization.swift
atmamont: Done

PR #2470, opened 2026-03-30
Localization: internalize legacy APIs

atmamont:
# Summary
This PR completes Phase 1 of the iOS custom localization plan by removing the legacy localization API from the merchant-facing v6 surface while keeping the existing localization path available internally. The main goal is to avoid exposing two competing localization contracts in v6 before the new provider-based API is introduced. This narrows the public API now and keeps the follow-up Phase 2 migration clearer and safer.

# Progress tracker
✅ Phase 0 — [Baseline safety net for current localization behavior](https://github.com/Adyen/adyen-ios/pull/2463)
✅ Phase 1 — Internalize legacy localization API
Phase 2 — Introduce the new public contract
Phase 3 — Wire provider resolution into card flow
Phase 4 — Unsupported locale warnings
Phase 5 — Final confidence pass for card alpha scope

## Technical Information 
Grouped changes:
- Internalized `LocalizationParameters`, `Localizable`, and related localization helpers by changing them from public/SPI exposure to `package`
- Updated component, action, and Drop-in configuration types so `localizationParameters` stays internal to the package
- Added public initializer overloads where needed so public configuration APIs remain usable without exposing `LocalizationParameters`

## Ticket
<ticket>
COSDK-1097
</ticket>

## Checklist [Required]
- [x] Tested changes locally

### DIFF
--- a/Adyen/Core/Core Protocols/PaymentMethod.swift
+++ b/Adyen/Core/Core Protocols/PaymentMethod.swift
@@ -41,20 +41,10 @@ public extension PaymentMethod {
 /// A protocol to define any partial payment method such as gift cards, `MealVoucher` etc.
 public protocol PartialPaymentMethod: PaymentMethod {}
 
-package protocol LocalizedPaymentMethod: PaymentMethod {
-    func defaultDisplayInformation(using parameters: LocalizationParameters?) -> DisplayInformation
-}
-
 package extension PaymentMethod {
     
     func displayInformation(using parameters: LocalizationParameters?) -> DisplayInformation {
-        let defaultDisplayInformation: DisplayInformation
-        if let localizedPaymentMethod = self as? any LocalizedPaymentMethod {
-            defaultDisplayInformation = localizedPaymentMethod.defaultDisplayInformation(using: parameters)
-        } else {
-            defaultDisplayInformation = DisplayInformation(title: name, subtitle: nil, logoName: type.rawValue)
-        }
-
+        let defaultDisplayInformation = defaultDisplayInformation(using: parameters)
         if let merchantProvidedDisplayInformation {
             let subtitle = merchantProvidedDisplayInformation.subtitle ?? defaultDisplayInformation.subtitle
             return DisplayInformation(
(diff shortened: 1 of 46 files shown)


Adyen/Core/Core Protocols/PaymentMethod.swift
```diff
@@ -48,11 +41,20 @@ public extension PaymentMethod {
...
+package protocol LocalizedPaymentMethod: PaymentMethod {
+    func defaultDisplayInformation(using parameters: LocalizationParameters?) -> DisplayInformation
+}
+
+package extension PaymentMethod {
     
     func displayInformation(using parameters: LocalizationParameters?) -> DisplayInformation {
-        let defaultDisplayInformation = defaultDisplayInformation(using: parameters)
+        let defaultDisplayInformation: DisplayInformation
+        if let localizedPaymentMethod = self as? any LocalizedPaymentMethod {
```
erenbesel: Can all payment methods be LocalizedPaymentMethod? Maybe it's good to review this and make all payment methods localized and we won't have to have this if else block? Also wouldn't have to conform to this protocol on every payment method.

We can make PaymentMethod conform to LocalizedPaymentMethod as a package extension if we don't want to expose this.

AdyenActions/Utilities/BundleSPMExtension.swift
```diff
@@ -18,7 +18,7 @@ internal extension Bundle {
     static let actions: Bundle = .init(for: RedirectComponent.self)
 
     /// The bundle in which the framework's resources are located.
-    static let actionsInternalResources: Bundle = .module
+    static let actionsInternalResources: Bundle = actions
```
erenbesel: I assume the previous one was the wrong one?

erenbesel (commented):
Nice! small comments

Adyen/Core/Core Protocols/PaymentMethod.swift
atmamont: Interesting! I will take a look if this change can be done in without blowing up the scope

AdyenActions/Utilities/BundleSPMExtension.swift
```diff
@@ -18,7 +18,7 @@ internal extension Bundle {
     static let actions: Bundle = .init(for: RedirectComponent.self)
 
     /// The bundle in which the framework's resources are located.
-    static let actionsInternalResources: Bundle = .module
```
robertdalmeida, line 21: https://developer.apple.com/documentation/xcode/bundling-resources-with-a-swift-package
I think it should still be .module. It seems to be recommended. 
Any particular reason to change it adyenUI?

Adyen/Core/Payment Methods/PayByBankUSPaymentMethod.swift
```diff
@@ -42,3 +42,5 @@ public struct PayByBankUSPaymentMethod: PaymentMethod {
         case name
     }
 }
+
+extension PayByBankUSPaymentMethod: LocalizedPaymentMethod {}
```
robertdalmeida: thought; nonblocking; 

Usually when i see this, i would expect the conforming is free 🆓
But here conforming isn't free but has the `defaultDisplayInformation` method requirement. 

We could either have the conforming directly in the type definition. 
Or 
move the defaultDisplayInformation at the point of conforming (I personally prefer this)

Adyen/Core/Core Protocols/PaymentMethod.swift
robertdalmeida: thought;
To me `if let localizedPaymentMethod = self as? any LocalizedPaymentMethod` is a code smell.
If i look at the code in isolation in the root context, i'm trying to conditionally cast it to something that is specialized.  Ofcourse i don't understand the depth and breath of this feature, so maybe our hands are tied here. 

But maybe the below code works?(it atleast builds for me) as an alternative if it makes sense? ofcourse the naming of the methods are very similar and confusing as-is, but llms can help in that i suppose. 


````
internal protocol LocalizedPaymentMethod: PaymentMethod {
    func defaultDisplayInformation(using parameters: LocalizationParameters?) -> DisplayInformation
}

internal extension PaymentMethod where Self: LocalizedPaymentMethod {
    func displayInformation(parameters: LocalizationParameters?) -> DisplayInformation {
        defaultDisplayInformation(using: parameters)
    }
}

internal extension PaymentMethod {
    func displayInformation(parameters: LocalizationParameters?) -> DisplayInformation {
        DisplayInformation(title: name, subtitle: nil, logoName: type.rawValue)
    }
}

package extension PaymentMethod {

    func displayInformation(using parameters: LocalizationParameters?) -> DisplayInformation {
        let defaultDisplayInformation: DisplayInformation = displayInformation(parameters: parameters)

````

**Review by robertdalmeida — APPROVED** [review]

File: Adyen/Core/Payment Methods/PayByBankUSPaymentMethod.swift (hunk 4 above)
Comment (atmamont) [reply Adyen/Core/Payment Methods/PayByBankUSPaymentMethod.swift]: We talked with Eren about this and a broader API topic yesterday, I will remove this new protocol and simplify things a bit. An simple edge case that is currently not covered in the demo app code and integration tests - merchants overriding payment method title and subtitle in DropIn made us think a bit differently about this approach.

File: AdyenActions/Utilities/BundleSPMExtension.swift (hunk 3 above)
Comment (atmamont) [reply AdyenActions/Utilities/BundleSPMExtension.swift:21]: This change slipped through my self-review. Lesson learned 👍

File: Adyen/Core/Core Protocols/PaymentMethod.swift [hunk 5]
```diff
@@ -66,9 +58,44 @@ public extension PaymentMethod {
         return defaultDisplayInformation
     }
 
-    @_spi(AdyenInternal)
+    // swiftlint:disable:next cyclomatic_complexity
     func defaultDisplayInformation(using parameters: LocalizationParameters?) -> DisplayInformation {
-        DisplayInformation(title: name, subtitle: nil, logoName: type.rawValue)
+        switch self {
```
atmamont, line 63: This is a temporary solution to make a change without big changes in the logic. This is required to change accessibility modifier for `LocalizationParameters` to `package`.

Adyen/Core/Core Protocols/PaymentMethod.swift
atmamont: Reverted structural changes as discussed.
Had to workaround a bit and move `func defaultDisplayInformation` to a default extension (package) because `PaymentMethod` is `public`.

AdyenActions/Utilities/BundleSPMExtension.swift
atmamont: Yes, reverted

robertdalmeida (approved):

erenbesel (approved):

PR #2473, opened 2026-03-31
V6 - Run BrowserInfo on the main thread

robertdalmeida:
# Summary [Required]
1. BrowserInfo run things on WKWebView which needs to be run on the main thread.
2. Refactored the browserInfo consuming api to be async instead of completion to a point that was logically possible.

Context: When trying to connect the StoredView, found this intermittent crash on browserInfo. 

## Checklist [Required]

- [x] Tested changes locally

### DIFF
--- a/Adyen/Core/Core Protocols/PaymentComponent.swift
+++ b/Adyen/Core/Core Protocols/PaymentComponent.swift
@@ -62,12 +62,11 @@ extension PaymentComponent {
     ///   - data: The Payment data to be submitted
     ///   - component: The component from which the payment originates.
     public func submit(data: PaymentComponentData, component: PaymentComponent? = nil) {
-        sendSubmitEvent()
-        
-        let component = component ?? self
-        
-        prepareSubmitData(from: data) { [weak self] updatedData in
+        Task { [weak self] in
             guard let self else { return }
+            sendSubmitEvent()
+            let component = component ?? self
+            let updatedData = await prepareSubmitData(from: data)
             self.delegate?.didSubmit(updatedData, from: component)
         }
     }
@@ -77,18 +76,17 @@ extension PaymentComponent {
     }
     
     /// Adds SDK related info to payment data object and returns the final data in the completion.
-    public func prepareSubmitData(from data: PaymentComponentData, completion: @escaping (PaymentComponentData) -> Void) {
-        
+    public func prepareSubmitData(from data: PaymentComponentData) async -> PaymentComponentData {
+
         let sdkData = SDKData(
             checkoutAttemptId: checkoutAttemptId,
             authenticationProvider: data.paymentMethod as? SDKDataAuthenticationProvider
         )
         
-        let updatedData = data
+        return await data
             .replacing(checkoutAttemptId: checkoutAttemptId)
             .replacing(sdkData: sdkData)
-        
-        updatedData.dataByAddingBrowserInfo(completion: completion)
+            .replacing(browserInfo: BrowserInfo.initialize())
     }
     
     private func sendSubmitEvent() {
--- a/Adyen/Core/Core Protocols/PaymentComponent.swift
+++ b/Adyen/Core/Core Protocols/PaymentComponent.swift
@@ -86,7 +86,7 @@ extension PaymentComponent {
         return await data
             .replacing(checkoutAttemptId: checkoutAttemptId)
             .replacing(sdkData: sdkData)
-            .replacing(browserInfo: BrowserInfo.initialize())
+            .replacing(browserInfo: BrowserInfo())
     }
     
     private func sendSubmitEvent() {
--- a/Adyen/Model/BrowserInfo.swift
+++ b/Adyen/Model/BrowserInfo.swift
@@ -13,23 +13,21 @@ public struct BrowserInfo: Encodable {
     /// The device default user-agent.
     public var userAgent: String?
     
-    /// Initializes a `BrowserInfo` instance asynchronously.
-    ///
-    /// - Parameters:
-    ///   - completion: A call back when the `BrowserInfo` instance is ready or when initialization fails.
-    public static func initialize(completion: @escaping ((_ info: BrowserInfo?) -> Void)) {
+    @MainActor public static func initialize() async -> BrowserInfo? {
         guard cachedUserAgent == nil else {
-            completion(BrowserInfo(userAgent: cachedUserAgent))
-            return
+            return BrowserInfo(userAgent: cachedUserAgent)
         }
-        webView = WKWebView()
-        webView?.evaluateJavaScript("navigator.userAgent") { result, _ in
-            webView = nil
-            guard let result = result as? String else {
-                return completion(nil)
+        return await withCheckedContinuation { continuation in
+            webView = WKWebView()
+            webView?.evaluateJavaScript("navigator.userAgent") { result, _ in
+                webView = nil
+                guard let result = result as? String else {
+                    continuation.resume(returning: nil)
+                    return
+                }
+                BrowserInfo.cachedUserAgent = result
+                continuation.resume(returning: BrowserInfo(userAgent: result))
             }
-            BrowserInfo.cachedUserAgent = result
-            completion(BrowserInfo(userAgent: cachedUserAgent))
         }
     }
     
--- a/Adyen/Model/BrowserInfo.swift
+++ b/Adyen/Model/BrowserInfo.swift
@@ -21,6 +21,9 @@ public struct BrowserInfo: Encodable {
         case userAgent
     }
 
+    /// Initializes a `BrowserInfo` instance asynchronously by retrieving the device's user agent.
+    ///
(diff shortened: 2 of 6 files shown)


Adyen/Core/Core Protocols/PaymentComponent.swift
```diff
@@ -77,18 +76,17 @@ extension PaymentComponent {
...
             authenticationProvider: data.paymentMethod as? SDKDataAuthenticationProvider
         )
         
-        let updatedData = data
+        return await data
             .replacing(checkoutAttemptId: checkoutAttemptId)
             .replacing(sdkData: sdkData)
-        
-        updatedData.dataByAddingBrowserInfo(completion: completion)
+            .replacing(browserInfo: BrowserInfo.initialize())
```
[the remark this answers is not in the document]
robertdalmeida: Sure. Why not

Adyen/Model/BrowserInfo.swift
```diff
@@ -13,23 +13,21 @@ public struct BrowserInfo: Encodable {
...
+                guard let result = result as? String else {
+                    continuation.resume(returning: nil)
+                    return
+                }
+                BrowserInfo.cachedUserAgent = result
+                continuation.resume(returning: BrowserInfo(userAgent: result))
             }
-            BrowserInfo.cachedUserAgent = result
-            completion(BrowserInfo(userAgent: cachedUserAgent))
         }
```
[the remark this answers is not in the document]
robertdalmeida: I didn't want to change the implementation. 
But since even you pointed it out, made the WKWebView a property of the BrowerInfo. 
No more static.

Adyen/Core/Core Protocols/PaymentComponent.swift
```diff
@@ -77,18 +76,17 @@ extension PaymentComponent {
     }
     
     /// Adds SDK related info to payment data object and returns the final data in the completion.
-    public func prepareSubmitData(from data: PaymentComponentData, completion: @escaping (PaymentComponentData) -> Void) {
-        
+    public func prepareSubmitData(from data: PaymentComponentData) async -> PaymentComponentData {
```
nauaros, line 79: Nice!! 👍

Adyen/Model/BrowserInfo.swift
```diff
@@ -9,31 +9,44 @@ import WebKit
 
 /// Provides the device default browser info.
 public struct BrowserInfo: Encodable {
```
nauaros, line 11: Wouldn't it be better to have `BrowserInfo` completely isolated to the main actor?

Then we wouldn't have mixed property isolation, and its execution can be predictable (only on main actor) avoiding confusing concurrency boundaries

Adyen/Model/BrowserInfo.swift
```diff
@@ -9,31 +9,44 @@ import WebKit
...
-    public var userAgent: String?
-    
-    /// Initializes a `BrowserInfo` instance asynchronously.
-    ///
-    /// - Parameters:
-    ///   - completion: A call back when the `BrowserInfo` instance is ready or when initialization fails.
-    public static func initialize(completion: @escaping ((_ info: BrowserInfo?) -> Void)) {
-        guard cachedUserAgent == nil else {
-            completion(BrowserInfo(userAgent: cachedUserAgent))
+    public var userAgent: String
```
nauaros: It could be declared as a constant right? Since it is only written on initialization.

nauaros (approved):

Tests/UnitTests/Core/BrowserInfoTests.swift
```diff
@@ -4,27 +4,27 @@
 // This file is open source and available under the MIT license. See the LICENSE file for more info.
 //
 
-@_spi(AdyenInternal) @testable import Adyen
-import XCTest
+@testable import Adyen
+import Testing
```
nauaros, line 8: Very cool you are adopting SwiftTesting! Thanks!

Adyen/Model/BrowserInfo.swift
robertdalmeida: My reasonings here are;
1. It is the init that needs MainActor isolation due to Webview. 
2. Once initialized this object is only a wrapper for a (now constant `let`) String. Which would then become Sendable. I don't find a reason that access to that constant should be restricted to the mainActor.

Adyen/Model/BrowserInfo.swift
robertdalmeida: Indeed, Thanks.

Adyen/Model/BrowserInfo.swift
```diff
@@ -9,31 +9,44 @@ import WebKit
...
-        webView?.evaluateJavaScript("navigator.userAgent") { result, _ in
-            webView = nil
-            guard let result = result as? String else {
-                return completion(nil)
+
+        self.webView = WKWebView()
+
+        let userAgent: String? = await withCheckedContinuation { [webView] continuation in
+            guard let webView else { return }
+            webView.evaluateJavaScript("navigator.userAgent") { result, _ in
```
erenbesel: I thought when we made this class async we would use the async version of `evaluateJavaScript` method.  Was there an issue with that one?

Adyen/Model/BrowserInfo.swift
```diff
@@ -9,31 +9,44 @@ import WebKit
...
-            completion(BrowserInfo(userAgent: cachedUserAgent))
+    public let userAgent: String
+
+    @MainActor internal var webView: WKWebView?
+
+    private enum CodingKeys: CodingKey {
+        case userAgent
+    }
+
+    @MainActor public init?() async {
```
erenbesel, line 27: Is there no race condition possibilities to set and read the properties of this class? If the class itself is an actor we wouldn't have to worry about these.

Adyen/Model/BrowserInfo.swift
robertdalmeida: Nice one, i hadn't check that possibility at all. 
I replaced it with the async one and it looks much better. Thanks.

Adyen/Model/BrowserInfo.swift
robertdalmeida: The type and its accessible properties is non mutable so it doesn't need to be isolated. It is free to be accessed from any thread.
The only isolation needed is in the init where we interact with WKWebKit which is isolated to MainActor.

nauaros (approved):

Adyen/Model/BrowserInfo.swift
```diff
@@ -9,31 +9,31 @@ import WebKit
...
-    public var userAgent: String?
-    
-    /// Initializes a `BrowserInfo` instance asynchronously.
-    ///
-    /// - Parameters:
-    ///   - completion: A call back when the `BrowserInfo` instance is ready or when initialization fails.
-    public static func initialize(completion: @escaping ((_ info: BrowserInfo?) -> Void)) {
-        guard cachedUserAgent == nil else {
-            completion(BrowserInfo(userAgent: cachedUserAgent))
+    public let userAgent: String
```
atmamont, line 16: It is still public and the docc comment is lost

Tests/IntegrationTests/Components Tests/IssuerList/IssuerListComponentTests.swift
```diff
@@ -59,9 +59,10 @@ class IssuerListComponentTests: XCTestCase {
             let details = paymentData.paymentMethod as! IssuerListDetails
             XCTAssertEqual(details.issuer, expectedIssuer.identifier)
             
-            XCTAssertEqual(analyticsProviderMock.infos.count, 1)
+            XCTAssertEqual(analyticsProviderMock.infos.count, 2)
```
atmamont, line 62: This change is not described in the PR description and I couldn't easily map it to any code change. Can you please explain what is going on here?

robertdalmeida: In short it is a timing issue, due to a suspension point(Task) introduced, now the entire block(selectionHandler) is executed. Previously it assumed that the block wouldn't execute completely. 

In IssuerListComponet, `listItem.selectionHandler` does 3 tasks
1. self.submit(data: PaymentComponentData. 
2. self.sendIssuerSelectedEvent
3. listItem.startLoading

in submit(data) we introduce a suspension Task {} which is why steps 2 and 3 are executed and then the delegate method in submit is called. 

Note: this doesn't change the order of the events sent, due to a suspension the next event is sent before calling the delegate.

Adyen/Model/BrowserInfo.swift
robertdalmeida: Nice, Thanks. 
You mean for the init right? i added that back just now. For userAgent it is still there.

atmamont: Yes. Sorry, I think I tapped wrong line when commenting 🤷

Tests/IntegrationTests/Components Tests/IssuerList/IssuerListComponentTests.swift
atmamont: Thanks!

atmamont (approved):

PR #2479, opened 2026-04-02
StoredCard and Preselected screen amount based changes & tests (2/6)

robertdalmeida:
# Summary [Required]
- Amount can have 3 states, so added tests for them using existing copies(note: the copies may change but that will be a smaller change). The todo for this change will still remain until it is confirmed. 
- Fixing contraints for the width and duplicate constraints for the StoredView. 
- Passing CheckoutTheme to the StoredCardComponent
- Removing unused code in the viewModel. 

COSDK-1140 

## Checklist [Required]

- [x] Tested changes locally


Adyen/Assets/Generated/LocalizationKey.swift
```diff
@@ -78,7 +78,7 @@ public struct LocalizationKey {
     public static let cardCvcItemTitle = LocalizationKey(key: "adyen.card.cvcItem.title")
     /// Enter security code
     public static let cardSecurityCodeTitle = LocalizationKey(key: "adyen.card.securityCode.title")
-    /// Enter the security code for %@ to complete the payment of %@
+    /// Enter the security code for %@
     public static let cardSecurityCodeDescription = LocalizationKey(key: "adyen.card.securityCode.description")
```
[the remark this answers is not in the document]
robertdalmeida: Done.

Adyen/Assets/Generated/LocalizationKey.swift
```diff
@@ -560,7 +560,7 @@ public struct LocalizationKey {
     public static let cardScannerCameraAccessDeniedAlertMessage = LocalizationKey(key: "adyen.card.scanner.camera.access.denied.alert.message")
     /// Open Settings
     public static let cardScannerCameraAccessDeniedAlertSettingsButtonTitle = LocalizationKey(key: "adyen.card.scanner.camera.access.denied.alert.settingsButton.title")
-    /// Use %@ to pay %@
+    /// Use %@
     public static let preselectedPaymentMethodSubtitle = LocalizationKey(key: "adyen.preselectedPaymentMethod.subtitle")
```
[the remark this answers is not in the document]
robertdalmeida: Done.

AdyenCard/Components/Stored Card/StoredCardInputView/StoredCardInputViewController.swift
```diff
@@ -193,8 +189,8 @@ internal class StoredCardInputViewController: UIViewController {
 
             contentStackView.topAnchor.constraint(equalTo: scrollView.topAnchor),
             contentStackView.leadingAnchor.constraint(equalTo: scrollView.leadingAnchor, constant: Constants.contentPadding),
-            contentStackView.trailingAnchor.constraint(equalTo: scrollView.trailingAnchor, constant: -Constants.contentPadding),
-            contentStackView.bottomAnchor.constraint(equalTo: scrollView.bottomAnchor, constant: -Constants.buttonsBottomPadding)
+            contentStackView.bottomAnchor.constraint(equalTo: scrollView.bottomAnchor, constant: -Constants.buttonsBottomPadding),
+            contentStackView.widthAnchor.constraint(equalTo: scrollView.widthAnchor, constant: -2 * Constants.contentPadding)
```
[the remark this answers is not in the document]
robertdalmeida: I need to revisit these constraints again. So i will change them anyway. 
But for clarity, for now, I have removed the trailing anchor.
i have a leading and a width (without a trailing constraint).

AdyenCard/Components/Stored Card/StoredCardComponent.swift
```diff
@@ -7,6 +7,9 @@
 @_spi(AdyenInternal) import Adyen
 import Foundation
 import UIKit
+#if canImport(AdyenUI)
+    @_spi(AdyenInternal) import AdyenUI
```
erenbesel: I would hope we shouldn't need spi import on a new module

erenbesel (approved):

AdyenCard/Components/Stored Card/StoredCardComponent.swift
robertdalmeida: Yeah it doesn't, thanks for pointing that out.

Adyen/Assets/en-US.lproj/Localizable.strings
```diff
@@ -273,5 +274,5 @@
 "adyen.card.scanner.camera.access.denied.alert.title" = "Allow camera access";
 "adyen.card.scanner.camera.access.denied.alert.message" = "Access was previously denied. To scan cards, please grant access from Settings.";
 "adyen.card.scanner.camera.access.denied.alert.settingsButton.title" = "Open Settings";
-"adyen.preselectedPaymentMethod.subtitle" = "Use %@ to pay %@";
+"adyen.preselectedPaymentMethod.subtitle" = "Use %@";
```
atmamont, line 277: Is this a v6 new key? If we have it in v5, we might need to add new key, otherwise we'll have troubles releasing localization changes in v5. It applies to all keys.

Tests/IntegrationTests/Card Tests/StoredCardComponentTests.swift
```diff
@@ -208,7 +209,8 @@ class StoredCardComponentTests: XCTestCase {
         let paymentMethod = storedCardPaymentMethod(brand: .masterCard)
         let sut = StoredCardComponent(
             storedCardPaymentMethod: paymentMethod,
-            context: context
```
atmamont, line 211: This test class can benefit from extracting the sut creation into `makeSUT()` helper method

Tests/IntegrationTests/Card Tests/StoredCardInputViewModelTests.swift
```diff
@@ -66,19 +66,50 @@ struct StoredCardInputViewModelTests {
...
+        // Given
+        let amount = Amount(value: 0, currencyCode: "USD")
+        let sut = makeSUT(name: "VISA", lastFour: "4556", amount: amount)
+
+        // Then
+
+        #expect(sut.titleText == "Enter security code")
+        // subtitleText contains payment method info
+        let subtitle = sut.subtitleText.string
+        #expect(subtitle == "Enter the security code for VISA •••• 4556")
```
atmamont: If subtitle is not affected by the amount passed, it is better to scope it out to make this easier to read, it is being checked repetitively in each test case.

Tests/IntegrationTests/Card Tests/StoredCardInputViewModelTests.swift
```diff
@@ -66,19 +66,50 @@ struct StoredCardInputViewModelTests {
...
+        // submitButtonTitle contains formatted amount
+        #expect(sut.submitButtonTitle == "Confirm preauthorization")
```
atmamont: We can make it more readable by extracting string literals into 
```
let amount = Amount(value: 0, currencyCode: "USD")
let expectedButtonTitle = "Confirm preauthorization"
...
#expect(sut.submitButtonTitle == expectedButtonTitle)
``` 

The idea is to remove all constants from the assertions block so no one actually needs to pay attention, and it becomes just a harness.

atmamont (approved):

Tests/IntegrationTests/Card Tests/StoredCardComponentTests.swift
robertdalmeida: many of them will be removed in the next PR on this topic as i remove the StoredAlertManager. So leaving this for now.

Tests/IntegrationTests/Card Tests/StoredCardInputViewModelTests.swift
robertdalmeida: I like what you are implying, that we only validate the test data and not worry about the manner of checking -  this is the advantage of parameterized tests. 
Din't want to spend time making this more optimized using parameterized tests, which would be a good improvement imho.

For now i did as you said and moved the constants to a variable and kept them in the // Given section.

Tests/IntegrationTests/Card Tests/StoredCardInputViewModelTests.swift
robertdalmeida: I left the title and subtile checks in each of the 3 cases, even though it repeats. Would like to keep it this way as;
- We are in one unit test able to see the 3 different ui texts that impact the screen on the whole. So it can be ead in context with the other texts.

Adyen/Assets/en-US.lproj/Localizable.strings
robertdalmeida: Good point that i need to be aware of. 
In this case this is a new string = new key.

erenbesel (approved):

Tests/IntegrationTests/Card Tests/StoredCardInputViewModelTests.swift
robertdalmeida: For now I've kept it as-is.

Tests/IntegrationTests/Card Tests/StoredCardInputViewModelTests.swift
robertdalmeida: Hope this was fine. But considering these are tests and there will be more follow up MRs on this, feel free to bring this up again. For now want to progress on this.

PR #2484, opened 2026-04-08
[v5] Add new SDK data fields

nauaros:
## Summary

This PR introduces some key changes related with `SDKData`:
- Adds 4 new fields to `SDKData` in order to support hosted checkout components. These are:
  - `channel`
  - `platform`
  - `sdkVersion`
  - `paymentMethodBehavior`
- Every usage of `channel` and `platform` needs to be **lowercase**. Thus, every instance of `iOS` has been replaced with `ios`.
- Introduces `CheckoutPlatformParams`. [Similar to Android](https://github.com/Adyen/adyen-android/blob/main/core/src/main/java/com/adyen/checkout/core/common/internal/helper/CheckoutPlatformParams.kt), there's a centralized place where _cross-platform_ can override platform parameters such as `version` and `channel`.

## Release Notes

<changed>
- The payload returned in the `onSubmit` callback has increased in size, because of changes to the included `sdkData` field.
</changed>

## Ticket
<ticket>
COSDK-1149
</ticket>

## Checklist

- [x] Tested changes locally

### DIFF
--- a/Adyen/Analytics/Models/AnalyticsData.swift
+++ b/Adyen/Analytics/Models/AnalyticsData.swift
@@ -7,34 +7,6 @@
 import Foundation
 import UIKit
 
-/// The context in which the SDK operates
-///
-/// Used to e.g. override the version + platform from within the Flutter SDK
-@_spi(AdyenInternal)
-public struct AnalyticsContext {
-    
-    internal let version: String
-    internal let platform: Platform
-    
-    public init(
-        version: String = adyenSdkVersion,
-        platform: Platform = .iOS
-    ) {
-        self.version = version
-        self.platform = platform
-    }
-}
-
-@_spi(AdyenInternal)
-public extension AnalyticsContext {
-
-    enum Platform: String {
-        case iOS = "iOS"
-        case reactNative = "react-native"
-        case flutter
-    }
-}
-
 internal struct AnalyticsData: Encodable {
 
     // MARK: - Properties
@@ -100,8 +72,8 @@ internal struct AnalyticsData: Encodable {
         self.amount = additionalFields?.amount
         self.sessionId = additionalFields?.sessionId
         
-        self.version = configuration.context.version
-        self.platform = configuration.context.platform.rawValue
+        self.version = checkoutPlatformParams.version
+        self.platform = checkoutPlatformParams.platform.rawValue
         
         self.level = configuration.analyticsLevel
 
--- a/Adyen/Analytics/Models/AnalyticsData.swift
+++ b/Adyen/Analytics/Models/AnalyticsData.swift
@@ -14,7 +14,7 @@ internal struct AnalyticsData: Encodable {
     /// The version of the SDK
     internal let version: String
 
-    internal let channel: String = "iOS"
+    internal let channel: String = "ios"
     
     /// The platform the SDK is running on (e.g. ios, flutter, react-native)
     internal let platform: String
@@ -74,7 +74,7 @@ internal struct AnalyticsData: Encodable {
         
         self.version = checkoutPlatformParams.version
         self.platform = checkoutPlatformParams.platform.rawValue
-        
+
         self.level = configuration.analyticsLevel
 
         switch flavor {
--- a/Adyen/Analytics/Models/AnalyticsData.swift
+++ b/Adyen/Analytics/Models/AnalyticsData.swift
@@ -14,8 +14,8 @@ internal struct AnalyticsData: Encodable {
     /// The version of the SDK
     internal let version: String
 
-    internal let channel: String = "ios"
-    
+    internal let channel: String = CheckoutPlatformParams.shared.channel
+
     /// The platform the SDK is running on (e.g. ios, flutter, react-native)
     internal let platform: String
 
--- a/Adyen/Core/Core Protocols/PaymentComponent.swift
+++ b/Adyen/Core/Core Protocols/PaymentComponent.swift
@@ -19,7 +19,9 @@ public protocol PaymentComponent: Component, PartialPaymentOrderAware, PaymentMe
     
     /// The delegate of the payment component.
     var delegate: PaymentComponentDelegate? { get set }
-    
+
+    @_spi(AdyenInternal)
+    var paymentMethodBehavior: SDKData.PaymentMethodBehavior { get }
 }
 
 @_spi(AdyenInternal)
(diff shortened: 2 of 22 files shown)


Adyen/Analytics/Models/AnalyticsData.swift
```diff
@@ -7,42 +7,14 @@
...
-
 internal struct AnalyticsData: Encodable {
 
     // MARK: - Properties
 
     /// The version of the SDK
     internal let version: String
 
-    internal let channel: String = "iOS"
+    internal let channel: String = "ios"
```
Robert-SD: Will this conflict with our existing data collection? Is case sensitivity important?

Adyen/Helpers/CheckoutPlatformParams.swift
```diff
@@ -0,0 +1,49 @@
...
+    // MARK: - Initializers
+
+    private init() {}
+
+    // MARK: - Mutation (restricted)
+
+    private let lock = NSLock()
+
+    @_spi(AdyenInternal)
+    public func override(version: String, platform: Platform) {
```
Robert-SD: Can we rename `override` into `overrideForCrossPlatform` to use the same name as Android [has](https://github.com/Adyen/adyen-android/blob/d10384c43fc3ca1d2efbdcfd834c84281b3fe3e8/core/src/main/java/com/adyen/checkout/core/common/internal/helper/CheckoutPlatformParams.kt#L25)?

Robert-SD: Shall we also order the parameter accordingly?

Adyen/Helpers/CheckoutPlatformParams.swift
```diff
@@ -0,0 +1,49 @@
...
+
+    // MARK: - Global shared instance
+
+    public static let shared = CheckoutPlatformParams()
+
+    // MARK: - Properties
+
+    public private(set) var version: String = adyenSdkVersion
+    public private(set) var platform: Platform = .ios
+    public let channel: String = "ios"
```
Robert-SD, line 29: The channel is initialized in AnalyticsData and AnalyticsRequest. Shall we use one central public constant?

Adyen/Helpers/CheckoutPlatformParams.swift
nauaros: yeah, I can order it exactly like Android has it

nauaros: Changed!

Adyen/Helpers/CheckoutPlatformParams.swift
nauaros: Good point. I set them to `CheckoutPlatformParams.shared.channel`

Adyen/Analytics/Models/AnalyticsData.swift
nauaros: It doesn't matter what value we send. This is handled on the backend side and mapped into an internal enum that's send to the Big Data platform.

Robert-SD (approved):

Adyen/Core/Core Protocols/PaymentComponent.swift
```diff
@@ -46,9 +46,14 @@ extension PaymentComponent {
     
     /// Adds SDK related info to payment data object and returns the final data in the completion.
     public func prepareSubmitData(from data: PaymentComponentData, completion: @escaping (PaymentComponentData) -> Void) {
-        
+        let isInstantPaymentComponent = self is InstantPaymentComponent
```
erenbesel: We can have this property `paymentMethodBehavior` or something as a protocol property with a default extension and on InstantPaymentComponent return the other value.

erenbesel (commented):
not a fan of the params name but one other comment

Adyen/Helpers/CheckoutPlatformParams.swift
```diff
@@ -0,0 +1,52 @@
+//
+// Copyright (c) 2026 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+@_spi(AdyenInternal)
+public class CheckoutPlatformParams {
```
atmamont: CheckoutPlatformParameters?

Adyen/Helpers/CheckoutPlatformParams.swift
```diff
@@ -0,0 +1,52 @@
...
+
+    // MARK: - Types
+
+    public enum Platform: String {
```
atmamont, line 15: I'd add a comment linking the Android counterpart here

Adyen/Helpers/CheckoutPlatformParams.swift
```diff
@@ -0,0 +1,52 @@
...
+    // MARK: - Mutation (restricted)
+
+    private let lock = NSLock()
+
+    @_spi(AdyenInternal)
+    public func overrideForCrossPlatform(
+        platform: Platform,
+        version: String
+    ) {
+        lock.lock()
```
atmamont, line 50: Can we make `CheckoutPlatformParams` an actor?

Tests/UnitTests/Core/CheckoutPlatformParamsTests.swift
```diff
@@ -0,0 +1,78 @@
...
+
+        // When
+        sut.overrideForCrossPlatform(platform: .flutter, version: "1.0.0")
+
+        // Then
+        XCTAssertEqual(sut.version, "1.0.0")
+        XCTAssertEqual(sut.platform, .flutter)
+
+        // Cleanup - restore original values
+        sut.overrideForCrossPlatform(platform: originalPlatform, version: originalVersion)
```
atmamont: Instead of cleaning this up you can just instantiate a completely new instance (do not use .shared)
It will allow to reduce the amount of code (storing original values, restoring them)

Adyen/Helpers/CheckoutPlatformParams.swift
nauaros: Following the exact naming convention Android already had in place, and this way making it consistent for cross platform. You can find Android version [here](https://github.com/Adyen/adyen-android/blob/main/core/src/main/java/com/adyen/checkout/core/common/internal/helper/CheckoutPlatformParams.kt)

I can change to any other name

atmamont: Makes sense then!

Adyen/Helpers/CheckoutPlatformParams.swift
nauaros: I actually tried modeling it as an actor but it’s not really worth it. It forces async access for simple config reads and that leaks into SDKData consumers and the Flutter bridge layer. Since this is just shared configuration with controlled mutation, a synchronous approach with a lock is much simpler and integrates better with Fluter

Adyen/Core/Core Protocols/PaymentComponent.swift
nauaros: Change to what we discussed. You can find the extension in the `SDKData` file

Adyen/Helpers/CheckoutPlatformParams.swift
nauaros: Done

Tests/UnitTests/Core/CheckoutPlatformParamsTests.swift
nauaros: Since this is a singleton, I added an internal init for testing purposes so new instances can created.

atmamont (approved):

PR #2486, opened 2026-04-09
[v5] UPI Smart Intent - Analytics Events

nauaros:
## Summary

Added analytics tracking for UPI payment flow to improve visibility into user interactions.

### Events Added

| Event Type | Component | Description |
|------------|-----------|-------------|
| **Initial** | `upi` | Sent on `viewDidLoad` to track component initialization |
| **Rendered** | `upi` | Sent on `viewDidLoad` to track component rendering |
| **Displayed** | `upi_intent` | Sent when UPI Intent flow is shown (with `target: listDetected` for installed apps, `target: issuerList` for fallback list) |
| **Displayed** | `upi_collect` | Sent when UPI Collect flow is shown |
| **Selected** | `upi_intent` | Sent when user selects a UPI app (includes `issuer` ID and `target`) |

### Implementation Details

- Displayed events are sent from a single place (`updateInterface()`) triggered by both `viewDidAppear` and tab switches
- `presentedValues` array contains available UPI app identifiers
- `target` differentiates between detected installed apps (`.listDetected`) and fallback issuer list (`.issuerList`)

### Ticket
<ticket>
COSDK-1089
</ticket>

### Checklist

- [x] Tested changes locally


AdyenComponents/UPI/UPIComponent.swift
```diff
@@ -367,6 +371,51 @@ internal extension UPIComponent {
...
+            sendUPIIntentDisplayedEvent()
+        case .upiCollect:
+            sendUPICollectDisplayedEvent()
+        default:
+            break
+        }
+    }
+
+    func sendUPIIntentDisplayedEvent() {
+        var eventInfo = AnalyticsEventInfo(
```
erenbesel: small nitpick: `infoEvent` is the right naming to prevent future confusions. They are all events, and can be infoEvent, logEvent, errorEvent etc.

AdyenComponents/UPI/UPIComponent.swift
```diff
@@ -463,12 +513,32 @@ extension UPIComponent: AdyenObserver {}
...
+
     public func validate() -> Bool {
         formViewController.validate()
     }
 }
+
+// MARK: - TrackableComponent
+
+@_spi(AdyenInternal)
+extension UPIComponent: TrackableComponent {}
```
erenbesel, line 529: 👀

AdyenComponents/UPI/UPIComponent.swift
```diff
@@ -367,6 +371,51 @@ internal extension UPIComponent {
...
+        )
+        eventInfo.target = installedUPIApps.isEmpty ? .issuerList : .listDetected
+        eventInfo.presentedValues = availableUPIApps.map(\.identifier)
+        context.analyticsProvider?.add(info: eventInfo)
+    }
+
+    func sendUPIIntentAppSelectionEvent(for issuer: Issuer?) {
+        guard let issuer else { return }
+        var eventInfo = AnalyticsEventInfo(
+            component: UPIFlowType.upiIntent.value,
```
erenbesel, line 402: I notice component parameter is not following the other components naming which is the txvariant name. Not only that, this will not be the same with this component's own renderEvent which will send the txVariant on the component parameter. Is this ok and accepted? 
Otherwise I would suggest sending another parameter or so and keep this value consistent.

robertdalmeida (approved):

AdyenComponents/UPI/UPIComponent.swift
nauaros: Okkk, I change it then.

AdyenComponents/UPI/UPIComponent.swift
nauaros: Yep, I noticed UPIComponent was not tracked 😬

AdyenComponents/UPI/UPIComponent.swift
nauaros: Yep, this has been decided together with the web team (Scotti) and Vish. They are doing the same doing.
Where do you think we should use the UPI flow?

erenbesel (approved):
I saw the updated description, if the rendered event is using the txvariant, I think that's the important one that needs to follow the convention

PR #2487, opened 2026-04-10
Co-badge UI & Logic update

erenbesel:
# Summary [Required]
Updating the logic while removing the old co-badge form item completely.
- Use the newly added co badge picker logic in DualBrandAccessoryView
- Remove cobadge form item
- `BrandDisplayMode` enum to make the state clean and managable from a single point
- Clean up code in these areas to make it more straightforward

https://github.com/user-attachments/assets/cc2fea5a-98a0-4111-9ba0-9f388fb4c1f9

## Ticket [Optional]
<ticket>
COSDK-1064
</ticket>

## Checklist [Required]

- [ ] Tested changes locally


AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -223,37 +223,47 @@ internal class CardViewController: FormViewController {
...
+            // no dual branding on regex response, get the 1st element
+            brands = Array(response.brands?.prefix(1) ?? [])
         } else {
-            brands = binInfo.brands ?? []
-            showCoBadgedCardsUI(for: brands)
+            brands = response.brands ?? []
         }
-        issuingCountryCode = binInfo.issuingCountryCode
+        
+        cardNumberItem.isLocalBrand = response.isCreatedLocally
```
atmamont: The name `isLocalBrand` creates more questions than gives answers to someone who is not deep into the logic. Is there a way to improve name clarity? Now it reads as something that matters _inside_ the card number item

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -223,37 +223,47 @@ internal class CardViewController: FormViewController {
...
         }
-        if !coBadgedBrands.isEmpty { // Check if there are any co-badged brands
-            items.coBadgedCardItem.updateItems(brands, cardLogos: cardLogos)
+        
+        let isDualBranded = brands.count == 2 && brands.allSatisfy(\.isSupported)
+        let containsAllowedBrands = brands.contains { allowedCoBadgedCardTypes.contains($0.type) }
+        
+        if isDualBranded, containsAllowedBrands {
+            cardNumberItem.brandDisplayMode = .dualSelectable
+            triggerBrandEvent(of: .displayed, brands: brands)
```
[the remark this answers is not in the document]
erenbesel: @gemini-code-assist this is already what is happening here, it's only triggered on dual brand detection and mode is `.dualSelectable`

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -223,37 +223,47 @@ internal class CardViewController: FormViewController {
...
+        brands: [CardBrand]
+    ) {
+        items.triggerInfoEvent(of: type, target: .dualBrandButton, brands: brands)
     }
 
-    internal func showCoBadgedCardsUI(for brands: [CardBrand]) {
-        let coBadgedBrands = brands.filter { brand in
-            allowedCoBadgedCardTypes.contains(brand.type)
+    private func updateBrandDisplayMode(for brands: [CardBrand], isLocal: Bool) {
+        guard !isLocal else {
```
atmamont, line 252: I see there are some booleans defining the source of origin for the actual state. Would it help to store a local `brandDisplayMode` instead, and move all logic inside specific helper methods to have a single pass here without the need of additional checks and early returns? 

Some other booleans: `isDualBranded`, `containsAllowedBrands`

AdyenCard/Form/DualBrandAccessoryView.swift
```diff
@@ -144,40 +150,45 @@ internal class DualBrandAccessoryView: UIView {
...
         applySelectionStyle(to: primaryOptionView, selected: false)
+        applySelectionStyle(to: secondaryOptionView, selected: false)
+        
+        switch displayMode {
+        case .single:
+            NSLayoutConstraint.activate(singleBrandConstraints)
+            setOptionPadding(0)
+            setSecondaryVisible(false)
+            
+        case .dualDisplay:
```
atmamont: WHat's the difference between `.dualDisplay` and `.dualSelectable`? Not gonna lie, this is quite hard to read :)

AdyenCard/Form/FormCardNumberItem.swift
```diff
@@ -165,42 +174,38 @@ internal final class FormCardNumberItem: FormTextItem, AdyenObserver {
...
+            updateSelectedBrand(firstSupportedBrand)
         case (2, nil):
-            // if there are 2 brands and neither is supported,
-            // need to show unsupported text
-            update(initialBrand: nil, defaultSupportedValue: false)
+            updateSelectedBrand(nil, defaultSupportedValue: false)
         default:
-            update(initialBrand: nil)
+            updateSelectedBrand(nil)
         }
```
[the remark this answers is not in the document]
erenbesel: No, this is triggered on a new response, meaning it should override the current one. Preserving previous brand selection is only risky and unsafe

AdyenCard/Form/FormCardNumberItem.swift
```diff
@@ -52,13 +49,25 @@ internal final class FormCardNumberItem: FormTextItem, AdyenObserver {
...
+        case dualDisplay
+        case dualSelectable
+    }
+    
+    /// Returns the brand to include in the payment request.
+    /// Returns `nil` for local (regex) brands and `.dualDisplay` mode.
     internal var currentBrand: CardBrand? {
-        isDualBranded ? selectedDualBrand : initialBrand
+        guard !isLocalBrand else { return nil }
+        return brandDisplayMode == .dualDisplay ? nil : selectedBrand
```
atmamont: This is interesting, what prevents us from making it simpler and just returning the current brand no matter which mode item has set?

atmamont (approved):
Not an easy walk, but let's be pragmatic to chose which parts could be improved and what we can live with.

AdyenCard/Components/Card/CardViewController.swift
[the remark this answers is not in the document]
erenbesel: Bin look up happens only once after a certain threshold and the response is cached for the same bin. The response also cannot be different for the same bin

AdyenCard/Components/Card/CardViewController.swift
erenbesel: I will find a better name indeed

AdyenCard/Form/DualBrandAccessoryView.swift
erenbesel: I thought of adding descriptions to each case :). It's for when dual brands are detected but not selectable (like eftpos in austraila)

AdyenCard/Form/FormCardNumberItem.swift
erenbesel: If it's a dual brand case but not selectable like above, we are not supposed to send the brand. Same with regex response

AdyenCard/Components/Card/CardViewController.swift
erenbesel: The new way is actually quite straight forward and concise with 15 lines of code. Maybe due to the changes on the file it feels confusing?

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -223,37 +223,47 @@ internal class CardViewController: FormViewController {
...
-        // no dual branding if response is from regex (fallback)
-        if binInfo.isCreatedLocally, let firstBrand = binInfo.brands?.first {
-            brands = [firstBrand]
-            items.coBadgedCardItem.isHidden.wrappedValue = true
-            items.coBadgedCardItem.resetItems()
+    internal func handleBinLookupResponse(_ response: BinLookupResponse) {
+        let brands: [CardBrand]
+        if response.isCreatedLocally {
+            // no dual branding on regex response, get the 1st element
+            brands = Array(response.brands?.prefix(1) ?? [])
```
robertdalmeida, line 230: nonblocking; 
```
brands?.prefix(1)
```
isn't .first more readable. I have never used .prefix(1) 
no change needed if you are already used to it.

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -223,37 +223,47 @@ internal class CardViewController: FormViewController {
...
+        of type: AnalyticsEventInfo.InfoType,
+        brands: [CardBrand]
+    ) {
+        items.triggerInfoEvent(of: type, target: .dualBrandButton, brands: brands)
     }
 
-    internal func showCoBadgedCardsUI(for brands: [CardBrand]) {
-        let coBadgedBrands = brands.filter { brand in
-            allowedCoBadgedCardTypes.contains(brand.type)
+    private func updateBrandDisplayMode(for brands: [CardBrand], isLocal: Bool) {
```
robertdalmeida, line 251: thought; nonblocking; suggestion;

this not only updatesBrandDisplayMode but also sends an analytical event. 
maybe at the handleBinResponse after we 
1. updateBrandDisplayMode
2. sendAnalyticalEvent 
So that the method does one particular task.

robertdalmeida (approved):
I like the state - BrandDisplayMode and the view reacting to it.

AdyenCard/Form/DualBrandAccessoryView.swift
erenbesel: Updated to `isBrandDetectedLocally`

AdyenCard/Components/Card/CardViewController.swift
erenbesel: Me neither, but try writing that with `.first` :p

robertdalmeida: haha... i see the problem. Both are criptic enough.

response.bands?.first.map { [$0] } ?? []

AdyenCard/Components/Card/CardViewController.swift
erenbesel: it is not ideal indeed but how would your suggestion look like in the code?

atmamont (approved):

PR #2500, opened 2026-04-20
Apple pay v6 improvements

erenbesel:
# Summary [Required]
Apple Pay flow improvements before migrating it to v6.
Callbacks converted to async.
State simplified with a continuation.

## Ticket [Optional]
<ticket>
COSDK-
</ticket>

## Checklist [Required]

- [x] Tested changes locally


AdyenCheckout/Checkout+DelegateCallbacks.swift
```diff
@@ -9,26 +9,24 @@
...
-// If there is a callback, regardless of session, we call it first.
-// If not, we check session and pass the work to it.
-// Finally if neither, we will fail/assert/show error.
+// MARK: - PaymentComponentDelegate
 
 extension Checkout: PaymentComponentDelegate {
-    
+
     public func didSubmit(_ data: PaymentComponentData, from component: any PaymentComponent) {
+        pendingPaymentComponent = component
```
atmamont, line 21: Just thinking of this state here in the case when multiple payment methods could reuse the same Checkout object, overwriting `pendingPaymentComponent`. It shouldn't be possible, right?

AdyenCheckout/Checkout+DelegateCallbacks.swift
```diff
@@ -41,34 +39,28 @@ extension Checkout: PaymentComponentDelegate {
...
                 do {
                     let response = try await onAdditionalDetails(data)
                     guard !Task.isCancelled else { return }
-                    self?.handle(response)
+                    self?.handle(response, from: self?.pendingPaymentComponent)
                 } catch {
                     // Ignore if this was a cancellation (task superseded or Checkout torn down).
                     guard !(error is CancellationError), !Task.isCancelled else { return }
-                    self?.finish(with: error)
+                    self?.finish(with: error, from: self?.pendingPaymentComponent)
```
atmamont: I notice we have two semantics here:
- handle
- finish

Could it make the flow even easier to understand and read if we add two handle methods:
- handle(response, ...)
- handle(error, ...)

At least both `handle` and `finish` would live in the same code block, making any future debugging much easier

AdyenCheckout/Checkout+DelegateCallbacks.swift
```diff
@@ -81,32 +73,75 @@ extension Checkout: ActionComponentDelegate {
...
+private extension Checkout {
+
+    /// If the response carries an action, dispatch it and keep `pendingPaymentComponent`
+    /// set so the final result (arriving later via `didProvide`) can still finalize
+    /// the component that originally submitted.
+    func handle(_ paymentsResponse: CheckoutPaymentsResponse, from component: (any PaymentComponent)?) {
+        if let action = paymentsResponse.action {
+            actionHandlingComponent.handle(action)
+        } else {
+            // TODO: check for error cases here
```
atmamont, line 116: This could reuse the handle(error, ...) method I mentioned above, consolidating all error handling in one place for both external and internal errors. The whole picture is still a bit vague for me, so take this as a conversation point.

AdyenCheckout/Checkout+DelegateCallbacks.swift
```diff
@@ -81,32 +73,75 @@ extension Checkout: ActionComponentDelegate {
...
+    func finish(with result: CheckoutResult, from component: (any PaymentComponent)?) {
+        finalize(component, success: result.resultCode.isSuccessful)
         configuration.onComplete?(result)
     }
-    
-    private func finish(with error: Error) {
-        // TODO: add any finalizing code if needed
+
+    func finish(with error: Error, from component: (any PaymentComponent)?) {
+        finalize(component, success: false)
```
atmamont: This part is something I was not happy about in the pas when debugging Apple Pay flow. Now we have even more semantics:
- handle
- finish
- finalize

Which makes zero sense when you try understanding the data flow of the component. `finish` and `finalize` are the same thing semantically. Let's try to improve it and remove all ambiguity points?

AdyenCheckout/Checkout+DelegateCallbacks.swift
```diff
@@ -81,32 +73,75 @@ extension Checkout: ActionComponentDelegate {
...
         configuration.onError?(CheckoutError(error: error))
     }
+
+    /// Resumes the originating component if it needs finalization (e.g. Apple Pay sheet)
+    /// and clears the pending reference so a new flow can start clean.
+    func finalize(_ component: (any PaymentComponent)?, success: Bool) {
+        (component as? any FinalizableComponent)?.didFinalize(with: success, completion: nil)
```
atmamont: Just saying - FinalizableComponent only raises more questions than gives answers, such as "oh, what? do we have components that cannot be finalized?"

AdyenComponents/Apple Pay/ApplePayComponent.swift
```diff
@@ -75,84 +73,68 @@ public class ApplePayComponent: NSObject, PresentableComponent, PaymentComponent
...
         self.context = context
-        self.paymentAuthorizationViewController = viewController
         self.applePayPaymentMethod = paymentMethod
         super.init()
 
-        paymentAuthorizationViewController?.delegate = self
+        let controller = PKPaymentAuthorizationViewController(paymentRequest: paymentRequest)
+        guard let controller else {
+            throw UnknownError(
+                errorDescription: "Failed to instantiate PKPaymentAuthorizationViewController because of unknown error"
```
atmamont: > If the user can’t make payments on any of the payment request’s supported networks, initialization fails and this method returns nil.

Can we make it more specific? [Docs](https://developer.apple.com/documentation/passkit/pkpaymentauthorizationviewcontroller/init(paymentrequest:)#Discussion)

AdyenComponents/Apple Pay/ApplePayComponent.swift
```diff
@@ -75,84 +73,68 @@ public class ApplePayComponent: NSObject, PresentableComponent, PaymentComponent
...
-        newItems.append(PKPaymentSummaryItem(label: lastItem.label, amount: decimalAmount))
-        paymentRequest.paymentSummaryItems = newItems
+        return controller
     }
 
-    // MARK: - Private
+    /// Cancels a pending authorization when the user dismisses the Apple Pay sheet
+    /// before the async `didAuthorizePayment` flow has completed.
+    internal func cancelPendingAuthorization() {
+        resumeContinuation(returning: false)
```
atmamont: This it quite obscure, specifically the `returning` part that is boolean. I see it's coming from the Swift itself, but here we call internal method and we can make its signature a bit better.
Something like `resume(success: false)` or even a separate method `cancelContinuation()` that will do the rest inside.

atmamont (commented):
Great improvements on code reductions, Eren!
I'd do an additional round to zoom out and try to focus on improving readability and clarity to make future us lives even easier.

AdyenCheckout/Checkout+DelegateCallbacks.swift
erenbesel: What name would you recommend? We need this logic unfortunately

AdyenCheckout/Checkout+DelegateCallbacks.swift
erenbesel: Did some improvements. We still need handle/finish separately IMO. But this class and the naming here are not Apple Pay flow specific, so your comment about getting confused for Apple Pay on this file is confusing me :D

AdyenComponents/Apple Pay/ApplePayComponentExtensions.swift
```diff
@@ -9,134 +9,81 @@ import Foundation
 import PassKit
 
 @_spi(AdyenInternal)
-extension ApplePayComponent: PKPaymentAuthorizationViewControllerDelegate {
-    
+extension ApplePayComponent: @preconcurrency PKPaymentAuthorizationViewControllerDelegate {
```
robertdalmeida: I'm not sure why `@preconcurrency` is needed here? 
I would think that having nonisolated extension would be expected as we conform to `PKPayment...` which is nonisolated.

AdyenComponents/Apple Pay/ApplePayComponentExtensions.swift
```diff
@@ -9,134 +9,81 @@ import Foundation
...
+
+    // MARK: - Did Finish
+
     public func paymentAuthorizationViewControllerDidFinish(_ controller: PKPaymentAuthorizationViewController) {
-        if configuration.dismissesAutomatically {
-            controller.dismiss(animated: true) { [weak self] in
-                self?.handleViewControllerDidFinish()
+        controller.dismiss(animated: true) { [weak self] in
+            MainActor.assumeIsolated {
```
robertdalmeida, line 18: Can't we use Task { `@MainActor` in 
here instead of MainActor.assumeIsolation?

Why: 
this wouldn't crash, while assume would crash if non main thread. (i think).

Demo/Shared.entitlements
```diff
@@ -6,5 +6,7 @@
 	<array>
 		<string>webcredentials:test-authentication-adyen.netlify.app</string>
 	</array>
+	<key>com.apple.developer.in-app-payments</key>
```
robertdalmeida, line 9: question; what shuld be the value of this array? 
Is there smething that we need to add like adyen domain for the test app?

AdyenDropIn/Modules/PreApplePay/PreApplePayComponent.swift
```diff
@@ -91,7 +90,9 @@ internal final class PreApplePayComponent: PresentableComponent,
     }
 
     internal func didFinalize(with success: Bool, completion: (() -> Void)?) {
-        applePayComponent.didFinalize(with: success, completion: completion)
+        Task { @MainActor in
```
robertdalmeida: PreApplyPayComponent is already on the mainactor do we need to spawn a task here?

AdyenComponents/Apple Pay/ApplePayComponent.swift
```diff
@@ -14,10 +14,19 @@ public class ApplePayComponent: NSObject, PresentableComponent, PaymentComponent
...
+    /// The continuation that bridges the gap between `submit(data:)` (fire-and-forget)
+    /// and the backend result delivered via `didFinalize(with:completion:)`.
+    /// While suspended, the Apple Pay sheet stays on screen waiting for a `PKPaymentAuthorizationResult`.
+    internal var paymentResultContinuation: CheckedContinuation<Bool, Never>?
+
+    /// Tracks whether `handleDidAuthorize` ran and returned a result to Apple.
+    /// Used by `didFinish` to distinguish user cancellation from normal sheet dismissal.
+    /// Set to `true` in every exit path of `handleDidAuthorize` — including early returns
+    /// for invalid tokens or failed `onAuthorize` — to prevent a spurious `didFail(.cancelled)`.
+    internal var authorizationHandled = false
```
robertdalmeida, line 28: Is there a possibility for the same apple component instance to be used for 2 consecutive transactions if the user cancelled the first one? 
I was looking for a place where this variable is reset to authorizationHandled = false, but it seems to be in the init alone.

AdyenDropIn/Modules/PreApplePay/PreApplePayComponent.swift
erenbesel: Must be a commit before main actor adoption

AdyenComponents/Apple Pay/ApplePayComponentExtensions.swift
erenbesel: We could but this needs to be sync, Task causes it to be scheduled on the next loop. This method is an Apple UIKit one, has to run on main so assumeIsolated should be safe unless Apple wants to break the world.

AdyenComponents/Apple Pay/ApplePayComponent.swift
erenbesel: Ideally not. But this flag is not related to that scenario, do we need to update the description?

AdyenDropIn/Modules/PreApplePay/PreApplePayComponent.swift
```diff
@@ -91,7 +90,9 @@ internal final class PreApplePayComponent: PresentableComponent,
...
+            applePayComponent.didFinalize(with: success, completion: nil)
+        }
```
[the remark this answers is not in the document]
erenbesel: This class is being removed, added a todo reminder.

Adyen.xcodeproj/project.pbxproj
```diff
@@ -10383,7 +10375,7 @@
 				CODE_SIGN_IDENTITY = "Apple Development";
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = 15;
-				DEVELOPMENT_TEAM = 8GDUWCHDXJ;
+				DEVELOPMENT_TEAM = B2NYSS5932;
```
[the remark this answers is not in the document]
erenbesel: The previous one could have been wrongly commited perhaps as it was not working. If this causes issues we will update it.

AdyenComponents/Apple Pay/ApplePayComponent.swift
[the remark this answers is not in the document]
erenbesel: It is already MainActor. Superflous comment @gemini-code-assist

AdyenComponents/Apple Pay/ApplePayComponentExtensions.swift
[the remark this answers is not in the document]
erenbesel: We are upping the Swift version to 5.9, see package.swift

AdyenCheckout/Checkout+DelegateCallbacks.swift
erenbesel: Ideally no. When a submit is triggered, there should be a UI loading state by the merchant app that we can't control.

AdyenCheckout/Checkout+DelegateCallbacks.swift
erenbesel: Once we add our errors, we will use the same function but with a different error type.

AdyenComponents/Apple Pay/ApplePayComponent.swift
robertdalmeida: No description update is needed. 
If this is not being used for 2 transactions then it doesn't matter.

AdyenDropIn/Modules/PreApplePay/PreApplePayComponent.swift
robertdalmeida: I see you removed main actor. btw, we don't need a task here at all right?

erenbesel: yes removed

atmamont (approved):

Demo/Configuration.swift
```diff
@@ -360,12 +360,26 @@ internal extension PKPaymentRequest {
         paymentRequest.paymentSummaryItems = [
             PKPaymentSummaryItem(label: ConfigurationConstants.appName, amount: decimalAmount)
         ]
-        paymentRequest.merchantCapabilities = .capability3DS
+        paymentRequest.merchantCapabilities = [.capability3DS, .credit, .debit]
```
descorp, line 363: If I am not mistaken you can omit ` .credit, .debit`, because if neither is passed, all card types will be available.

PR #2504, opened 2026-04-22
Align v6 checkout callbacks (SubmitResult / AdditionalDetailsResult)

Robert-SD:
# Summary

Aligns the v6 advanced-flow checkout callback surface with Android by splitting the shared callback result type into stage-specific enums.

- Renames `CheckoutPaymentsResponse.swift` to `CheckoutResult.swift`, keeping only the types with live call sites (`CheckoutResult`, `CheckoutError`).
- Introduces `SubmitResult` (`finished | action | error | partialPayment`) and `AdditionalDetailsResult` (`finished | error`) so `onSubmit` and `onAdditionalDetails` report results in a shape that matches each stage's valid transitions.
- `SubmitResult.partialPayment` carries a `PartialPayment` payload (`order: PartialPaymentOrder`, `paymentMethods: PaymentMethods?`) to support a future Drop-in advanced (non-session) continuation.
- Hardens `Checkout` advanced-flow dispatch in `Checkout+DelegateCallbacks.swift`: a single `CancellationError` catch plus a single post-catch `Task.isCancelled` guard, uniform across `didSubmit` / `didProvide`. A pending submit/details `Task` is cancelled before a new one is started.
- Updates advanced-flow demo examples (`CardComponentAdvancedFlowExample`, `BLIKComponentAdvancedFlowExample`, `DummyActionComponentExample`) to return the new result types.

## Scope

This PR is scoped to the **advanced flow only**. Earlier Session integration changes were reverted in `6d8a37a5b` to keep the PR focused; Session continues to go through the existing `SessionDelegate` path unchanged.

`SubmitResult.partialPayment` is an intentional no-op in the advanced Components flow today: the merchant owns the continuation (they instantiate a new component for the remaining amount). A TODO is left in `Checkout+DelegateCallbacks.swift` for Drop-in advanced partial-payment support.

## Follow-up work (not in this PR)

- `Checkout.didComplete(from:)` currently has no resultCode to forward; the advanced non-session action terminal path needs a design decision for a merchant-visible code (tracked via code TODO).
- Drop-in advanced (non-session) flow must consume `SubmitResult.partialPayment` and loop the continuation (tracked via code TODO).
- Session path could later be migrated to the same stage-specific result types in a follow-up.

## Testing Instructions

- `xcodebuild -scheme Adyen build` → succeeds.
- New/updated unit tests in `Tests/UnitTests/AdyenCheckout/CheckoutTests.swift`:
  - `test_didSubmit_callsOnSubmit_whenSet`
  - `test_didSubmit_responseWithoutAction_callsOnComplete`
  - `test_didSubmit_errorThrown_callsOnError`
  - `test_didSubmit_returnsErrorBranch_callsOnError`
  - `test_didSubmit_cancellationErrorThrown_doesNotCallOnError`
  - `test_didProvide_callsOnAdditionalDetails_whenSet`
  - `test_didProvide_returnsFinished_callsOnComplete`
  - `test_didProvide_returnsErrorBranch_callsOnError`
  - `test_didProvide_errorThrown_callsOnError`
  - `test_didProvide_cancellationErrorThrown_doesNotCallOnError`

## Checklist

- [x] Tested changes locally


Adyen/Core/Configuration/SubmitResult.swift
```diff
@@ -0,0 +1,59 @@
...
+public enum SubmitResult: Sendable {
+    
+    /// The `/payments` call finished and carries a final `resultCode`.
+    case finished(resultCode: String)
+    
+    /// The `/payments` call returned an action that must be handled by the SDK.
+    case action(Action)
+    
+    /// An error occurred while handling the submit callback.
+    case error(Error)
```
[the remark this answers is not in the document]
Robert-SD: Declining this suggestion.

Constraining the associated value to `any Error & Sendable` forces every error producer in the merchant's `onSubmit` implementation to return a `Sendable`-conforming error. Merchants commonly wrap errors from third-party backends / networking SDKs that do not yet conform to `Sendable`, and pushing that burden onto integrators would be a significant API ergonomics regression for marginal benefit.

The existential `any Error` is treated as implicitly `@unchecked Sendable` when crossing actor boundaries in practice (the callback sites here are `@MainActor`-isolated, so the error is only ever observed on the main actor anyway). If full strict-concurrency enforcement lands project-wide later, the right fix is to gate that through a project-level `SWIFT_STRICT_CONCURRENCY` setting and revisit — not to impose `Sendable` on the public enum.

Adyen/Core/Configuration/CheckoutCallbackResults.swift
```diff
@@ -0,0 +1,59 @@
...
+    /// The `/payments/details` call finished and carries a final `resultCode`.
+    ///
+    /// `resultCode` is always a `String`. When the underlying SDK delegate path does not carry a
+    /// resultCode (e.g. `Checkout.didComplete(from:)` in the advanced, non-session flow), the SDK
+    /// emits an empty string to keep the type uniform with `SubmitResult.finished` and with the
+    /// Android callback surface.
+    case finished(resultCode: String)
+    
+    /// An error occurred while handling the additional-details callback.
+    case error(Error)
```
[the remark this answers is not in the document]
Robert-SD: Declining — same reasoning as on `SubmitResult.error` above. Forcing `any Error & Sendable` here would require merchants' `onAdditionalDetails` implementations to only surface errors that conform to `Sendable`, which is not a constraint most backend / networking error types satisfy out of the box. The callback is `@MainActor`-isolated at both ends, so the error never actually crosses actor boundaries in a way that requires strict `Sendable` enforcement.

AdyenSession/Session+ActionComponentDelegate.swift
```diff
@@ -65,12 +65,64 @@ extension Session {
         )
         Task { [weak self] in
             guard let self else { return }
+            let additionalDetailsResult: AdditionalDetailsResult
             do {
                 let response: PaymentsResponse = try await apiClient.performAsync(request)
-                handle(paymentResponse: response, for: component, in: dropInComponent)
+                additionalDetailsResult = mapToAdditionalDetailsResult(response)
```
Robert-SD: We don't need that migration in sessions, because it will use its own internal logic. Revert this change.

AdyenSession/Session+PaymentComponentDelegate.swift
```diff
@@ -67,49 +67,83 @@ extension Session {
         )
         Task { [weak self] in
             guard let self else { return }
+            let submitResult: SubmitResult
             do {
                 let response: PaymentsResponse = try await apiClient.performAsync(request)
-                handle(paymentResponse: response, for: component, in: dropInComponent)
+                submitResult = mapToSubmitResult(response)
```
Robert-SD: Same here

AdyenSession/Session+ActionComponentDelegate.swift
Robert-SD: Addressed in the revert pushed 2026-04-23: `mapToAdditionalDetailsResult` and the stage-enum migration in `Session+ActionComponentDelegate.swift` have been removed. `didProvide` is restored to its pre-refactor `handle(paymentResponse:...)` direct routing. Session continues to use its own `CheckoutPaymentsResponse`-based internal logic; stage-enum alignment is deferred to a follow-up PR.

AdyenSession/Session+PaymentComponentDelegate.swift
Robert-SD: Addressed in the revert pushed 2026-04-23: `mapToSubmitResult` and the stage-enum migration in `Session+PaymentComponentDelegate.swift` have been removed. `didSubmit` is restored to its pre-refactor `handle(paymentResponse:...)` direct routing, along with `handle(order:resultCode:...)` and `showPaymentFailedAlert` — the refused-partial-payment alert is back. Session-side stage-enum alignment is deferred to a follow-up PR.

AdyenCheckout/Checkout+DelegateCallbacks.swift
```diff
@@ -47,6 +48,29 @@ extension Checkout: PaymentComponentDelegate {
...
+            finish(with: CheckoutResult(resultCode: CheckoutResultCode(rawValue: resultCode)), from: component)
+        case let .error(error):
+            finish(with: error, from: component)
+        case .partialPayment:
+            // Components (advanced flow): the SDK intentionally performs no work here.
+            // The merchant owns the continuation — they decide whether to instantiate a new
+            // payment component for the remaining amount based on the PartialPayment payload
+            // they returned from `onSubmit`.
+            // TODO: add partial-payment support for Drop-in on the advanced (non-session) flow.
+            break
```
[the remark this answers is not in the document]
Robert-SD: This is a TODO for later. Not this PR.

Adyen/ResponseModels/CheckoutResult.swift
```diff
@@ -0,0 +1,34 @@
...
+/// Wrapper type to contain checkout related errors.
+// TODO: if not needed, can remove?
+public struct CheckoutError: Error {
+    
+    private let error: Error
+    
+    public init(error: Error) {
+        self.error = error
+    }
+}
```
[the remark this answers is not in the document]
Robert-SD: We will use it later, I removed the TODO comment.

Adyen/ResponseModels/CheckoutResult.swift
```diff
@@ -0,0 +1,40 @@
...
+    public let sessionResult: String?
+    
+    package init(resultCode: CheckoutResultCode, sessionResult: String? = nil) {
+        self.resultCode = resultCode
+        self.sessionResult = sessionResult
+    }
+}
+
+/// Wrapper type to contain checkout related errors.
+public struct CheckoutError: Error {
```
descorp: Maybe move `CheckoutError` to a standalone file?

Adyen/Core/Configuration/SubmitResult.swift
```diff
@@ -0,0 +1,42 @@
...
+
+/// Payload carried on `SubmitResult.partialPayment`, describing the partial-payment continuation.
+public struct PartialPayment: Sendable {
+    
+    /// The partial-payment order. `remainingAmount` is reachable via `order.remainingAmount`.
+    public let order: PartialPaymentOrder
+    
+    /// Updated payment methods for the continuation flow. Optional because a session reload
+    /// does not always produce a new list.
+    public let paymentMethods: PaymentMethods?
```
descorp: Should this be optional?

Adyen/Core/Configuration/AdditionalDetailsResult.swift
```diff
@@ -0,0 +1,24 @@
...
+///
+/// Mirrors the generic advanced-flow branch view for the details stage: `Finished | Error`.
+public enum AdditionalDetailsResult: Sendable {
+    
+    /// The `/payments/details` call finished and carries a final `resultCode`.
+    ///
+    /// `resultCode` is always a `String`. When the underlying SDK delegate path does not carry a
+    /// resultCode (e.g. `Checkout.didComplete(from:)` in the advanced, non-session flow), the SDK
+    /// emits an empty string to keep the type uniform with `SubmitResult.finished` and with the
+    /// Android callback surface.
```
atmamont: Should we mention generic Adyen Checkout SDK here? We aim to align with Web eventually

Adyen/ResponseModels/CheckoutResult.swift
```diff
@@ -0,0 +1,40 @@
...
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+/// A model that holds information regarding the result status of a payment.
+public struct CheckoutResult {
+    
+    public let resultCode: CheckoutResultCode
```
atmamont, line 12: The `CheckoutResultCode` is public, but we expose `resultCode` as `String` above. Wouldn't it create a confusion from the merchant side?

AdyenCheckout/Checkout+DelegateCallbacks.swift
```diff
@@ -87,6 +110,20 @@ extension Checkout: ActionComponentDelegate {
...
+
+    @MainActor
+    private func handle(additionalDetailsResult: AdditionalDetailsResult, from component: (any PaymentComponent)?) {
+        switch additionalDetailsResult {
+        case let .finished(resultCode):
+            finish(with: CheckoutResult(resultCode: CheckoutResultCode(rawValue: resultCode)), from: component)
+        case let .error(error):
+            finish(with: error, from: component)
+        @unknown default:
+            AdyenAssertion.assertionFailure(
```
atmamont: This had me stopped to think for a moment. Would this kind of future-proofing lead to the flow to be stuck if both case let failed to match?

Adyen/ResponseModels/CheckoutResult.swift
Robert-SD: I will remove this class, it is not needed and we will introduce the `CheckoutError` when we implement the error alignment.

Adyen/ResponseModels/CheckoutResult.swift
Robert-SD: The `resultCode` as a string is what we ask the merchant to provide. The CheckoutResult is used by us to finalize the payment flow. This is part of the `onComplete()` alignment of the ADR.

I moved the class into its own file. I was already present int the deleted `CheckoutPaymentsResponse`.

AdyenCheckout/Checkout+DelegateCallbacks.swift
Robert-SD: We could discuss on how to proceed in this case. We could also follow the finish with error flow.

AdyenCheckout/Checkout+DelegateCallbacks.swift
```diff
@@ -47,6 +47,31 @@ extension Checkout: PaymentComponentDelegate {
...
+            // `errorMessage` in the UI before re-prompting.
+            break
+        case .partialPayment:
+            // Components (advanced flow): the SDK intentionally performs no work here.
+            // The merchant owns the continuation — they decide whether to instantiate a new
+            // payment component for the remaining amount based on the PartialPayment payload
+            // they returned from `onSubmit`.
+            // TODO: add partial-payment support for Drop-in on the advanced (non-session) flow.
+            break
+        @unknown default:
```
robertdalmeida: I think `@unknown default` is not needed for both these enums. 

Why: The merchant cannot add its own cases. The enum is controlled by us. If we add an additional case we can add it here too.

robertdalmeida (approved):

Adyen/ResponseModels/CheckoutResult.swift
Robert-SD: As discussed, we keep the `String` to move the mapping  the control onto our side.

AdyenCheckout/Checkout+DelegateCallbacks.swift
Robert-SD: Agreed. Since these enums are controlled by the SDK, explicit exhaustiveness is clearer here. I removed the `@unknown default` branches so future enum changes must be handled directly in the switch.

AdyenCheckout/Checkout+DelegateCallbacks.swift
Robert-SD: As mentioned above, I removed the `@unknown default:` case.

robertdalmeida (approved):

Adyen/Core/Configuration/AdditionalDetailsResult.swift
```diff
@@ -0,0 +1,19 @@
+//
+// Copyright (c) 2026 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+/// Result returned by the merchant's `onAdditionalDetails` callback in the advanced flow.
```
erenbesel: On public API let's avoid keywords like merchant. We can write "Result returned by `onAdditionalDetails` callback in the advanced flow."

Adyen/Core/Configuration/AdditionalDetailsResult.swift
```diff
@@ -0,0 +1,19 @@
...
+public enum AdditionalDetailsResult: Sendable {
+
+    /// The `/payments/details` call completed and carries a final `resultCode`.
```
erenbesel: This text is too long, verbose IMO. We can shorten it.

Adyen/Core/Configuration/SubmitResult.swift
```diff
@@ -0,0 +1,47 @@
+//
+// Copyright (c) 2026 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+/// Result returned by the merchant's `onSubmit` callback in the advanced flow.
```
erenbesel: Same here

Adyen/Core/Configuration/SubmitResult.swift
```diff
@@ -0,0 +1,47 @@
...
+    ///   re-prompting.
+    case retry(errorMessage: String? = nil)
+    
+    /// A partial-payment continuation is required. The SDK will loop back into `onSubmit`
+    /// with the updated order and payment methods.
+    case partialPayment(PartialPayment)
+}
+
+/// Payload carried on `SubmitResult.partialPayment`, describing the partial-payment continuation.
+public struct PartialPayment: Sendable {
```
erenbesel, line 31: Not sure where to put this struct. Inside SubmitResult or as its own. Let's just add a TODO to check when we do Dropin

erenbesel (commented):
Amazing! Just tiny remarks on text

Tests/UnitTests/AdyenCheckout/CheckoutTests.swift
```diff
@@ -14,6 +14,10 @@ import XCTest
 
 @MainActor
 final class CheckoutTests: XCTestCase {
+    /// Sentinel used when the advanced (non-session) action flow has no server-provided
```
erenbesel: What is Sentinel?

erenbesel (approved):
🚀

AdyenCheckout/Checkout+DelegateCallbacks.swift
```diff
@@ -148,6 +162,6 @@ private extension Checkout {
     func finish(with error: Error, from component: (any PaymentComponent)?) {
         (component as? any FinalizableComponent)?.didFinalize(with: false, completion: nil)
         pendingPaymentComponent = nil
-        configuration.onError?(CheckoutError(error: error))
```
erenbesel, line 151: I just noticed CheckoutError is deleted, why is that?

Tests/UnitTests/AdyenCheckout/CheckoutTests.swift
atmamont: image

robertdalmeida (approved):

erenbesel (approved):

atmamont (approved):

PR #2511, opened 2026-04-23
ApplePay v6: Factory and related changes

erenbesel:
# Summary [Required]
- Adding Apple pay factory and related changes.
 - Move configuration to its own object.
- Making factory default configuration optional to support configurations that can't have a default.

Next: Make the component package, update demo code to use via Checkout.

## Ticket [Optional]
<ticket>
COSDK-931
</ticket>

## Checklist [Required]

- [x] Tested changes locally


AdyenCheckout/Checkout+DelegateCallbacks.swift
```diff
@@ -16,7 +16,7 @@ extension Checkout: PaymentComponentDelegate {
     public func didSubmit(_ data: PaymentComponentData, from component: any PaymentComponent) {
         AdyenAssertion.assert(
             message: "A new payment component submitted while another flow is still pending.",
-            condition: pendingPaymentComponent == nil
+            condition: pendingPaymentComponent != nil
```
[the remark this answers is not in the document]
erenbesel: @gemini-code-assist how so? shouldn't it be the assertion will trigger if the pendingPaymentComponent is not nil, as in, another submit was called and the value was set beforehand?

AdyenComponents/Apple Pay/ApplePayComponentFactory.swift
```diff
@@ -0,0 +1,37 @@
...
+    ) throws -> ApplePayComponent {
+        try ApplePayComponent(
+            paymentMethod: paymentMethod,
+            context: context,
+            configuration: configuration
+        )
+    }
+
+    package func defaultConfiguration() -> ApplePayConfiguration? {
+        nil
```
atmamont: We can think of stopping here instead of making it optional. By doing this we could shift the responsibility to integrators while keeping internal structure tight (introducing an optional makes it loose)

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
```diff
@@ -8,132 +8,135 @@
...
+    package var showsSubmitButton: Bool = true
+
+    package var theme: CheckoutTheme = .default
+
+    package var localizationParameters: LocalizationParameters?
+
+    /// The flag to toggle onboarding.
+    /// If true, allow the shopper to add cards to Apple Pay if non exists yet.
+    /// If false, then Apple Pay is disabled if the shopper doesn't have supported cards on Apple Pay wallet.
+    /// Default is false.
```
atmamont, line 30: What drives this decision to make it false by default?

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
```diff
@@ -8,132 +8,135 @@
...
-            return newConfig
+    internal static func validate(summaryItems: [PKPaymentSummaryItem]) throws {
+        guard !summaryItems.isEmpty else {
+            throw ApplePayComponent.Error.emptySummaryItems
+        }
+        guard let lastItem = summaryItems.last,
+              lastItem.amount.doubleValue >= 0 else {
+            throw ApplePayComponent.Error.negativeGrandTotal
+        }
+        guard !summaryItems.map(\.amount).contains(NSDecimalNumber.notANumber) else {
```
atmamont, line 106: I'd sacrifice fancy syntax and unwrap it to make reading easier

Tests/UnitTests/AdyenCheckout/Component Factory Tests/ApplePayComponentFactoryTests.swift
```diff
@@ -0,0 +1,90 @@
...
+        )
+
+        // Then
+        XCTAssertTrue(component.configuration.allowOnboarding)
+    }
+
+    // MARK: - Type Conformance Tests
+
+    func testFactory_ConformsToPaymentComponentFactory() {
+        XCTAssertNotNil(factory as any PaymentComponentFactory)
```
atmamont, line 85: I wonder if this could be even `nil`, as the compiler doesn't ask for `as?`

Tests/UnitTests/AdyenCheckout/CheckoutComponentBuilderTests.swift
```diff
@@ -402,6 +403,48 @@ final class CheckoutComponentBuilderTests: XCTestCase {
...
+        // When / Then
+        XCTAssertThrowsError(
+            try CheckoutComponentBuilder.build(
+                for: paymentMethod,
+                configuration: checkoutConfiguration,
+                context: context
+            ),
+            "Builder should throw when Apple Pay has no default and no user-supplied config"
+        ) { error in
+            XCTAssertTrue(error is UnknownError, "Expected UnknownError, got \(type(of: error))")
```
atmamont, line 444: We will update this later to a specific error, right?

atmamont (approved):
Left some comments as a conversation starter, looks good!
The only thing that didn't land smoothly for me is making everything optional (the consequences are evident in test updates)

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
```diff
@@ -8,132 +8,135 @@
...
+            throw ApplePayComponent.Error.invalidCountryCode
         }
-
-        internal mutating func paymentRequest(with supportedNetworks: [PKPaymentNetwork]) -> PKPaymentRequest {
-            paymentRequest.supportedNetworks = supportedNetworks
-            return paymentRequest
+        guard CurrencyCodeValidator().isValid(paymentRequest.currencyCode) else {
+            throw ApplePayComponent.Error.invalidCurrencyCode
         }
+        try Self.validate(summaryItems: paymentRequest.paymentSummaryItems)
```
descorp, line 92: We can also add check for `merchantCapabilities` since we no longer defining it.

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
```diff
@@ -8,132 +8,135 @@
...
+            throw ApplePayComponent.Error.invalidSummaryItem
         }
     }
+
+    internal mutating func paymentRequest(with supportedNetworks: [PKPaymentNetwork]) -> PKPaymentRequest {
+        paymentRequest.supportedNetworks = supportedNetworks
+        return paymentRequest
+    }
+
+    package var currentAmount: Amount? {
```
descorp, line 111: Not sure if this is still used. Maybe we can remove (less Decimal2Int conversion we do, less chance to have an precision error)

AdyenComponents/Apple Pay/ApplePayComponent.swift
descorp, line 80: We need to add one more check that verify that `context.amount` contains same value as `paymentRequest.summeryItems.last` (at least throw console warning if it is not matching)

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
erenbesel: What should we validate here? Should it be us doing that?

Tests/UnitTests/AdyenCheckout/CheckoutComponentBuilderTests.swift
erenbesel: yes

Tests/UnitTests/AdyenCheckout/Component Factory Tests/ApplePayComponentFactoryTests.swift
erenbesel: Yeah don't think so. Interesting

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
erenbesel: This was readable to me, what do you refer to by unwwapping?

atmamont: Something like this, on top `map` processes all items no matter what, while `allSatisfy` will exit early

```swift
guard summaryItems.allSatisfy({ $0.amount != .notANumber }) else {

}
```

atmamont (approved):
🚀  Looks much better now, and easier to follow/debug!
There are some unanswered comments though, it would be great to resolve them.

Adyen/Core/Components/ComponentFactory.swift
```diff
@@ -36,7 +36,13 @@ package protocol PaymentComponentFactory {
...
-    func defaultConfiguration() -> Configuration?
+    /// Returns a default configuration for this factory's component.
+    ///
+    /// Most factories can construct a usable default. Some (e.g. Apple Pay) cannot, because
+    /// their configuration requires merchant-specific values; those factories throw to make
+    /// the missing-configuration case explicit at the call site.
+    ///
+    /// - Returns: A configured ``Configuration`` instance.
+    /// - Throws: An error if no usable default exists for this factory.
+    func defaultConfiguration() throws -> Configuration
```
descorp, line 47: IMHO - good API is the one you can use in first attempt. 

Throwing is confusing because merchant can't figure when to use `defaultConfiguration` with payment method X until it throw. 

Also API would look odd:

```swift
let config: XXXConfiguration
do { 
   let defaultConfig = defaultConfiguration()
   config = defaultConfig
} catch { 
   // You can't return here, because this is "expected scenario"
   config = XXXConfiguration(...)
   config.yyy = ...
}
component = XXXComponent(config)
...
```

Keeping `optional` could be more logical:

```swift
let config: XXXConfiguration
if let defaultConfig = defaultConfiguration() {
   config = defaultConfig
} else { 
   // Still weird
   config = XXXConfiguration(...)
}
component = XXXComponent(config)
...
```

---

I would suggest to not make `defaultConfiguration` as part of `ComponentFactory` and keep it "per case". It is more boilerplate code, but API would be more straitforward

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
erenbesel: This will be removed once amount is removed from paymentData

Adyen/Core/Components/ComponentFactory.swift
erenbesel: This is package, not merchant facing code. The throw happens when they want to use a component but forgot to provide a necessary config

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
erenbesel: Will do on next PR

AdyenComponents/Apple Pay/ApplePayConfiguration.swift
erenbesel: WIll discuss on a followup

PR #2521, opened 2026-05-01
Update AnyPaymentMethodDecoder to enable IRIS as InstantPaymentMethod

nauaros:
### Summary
- Added IRIS as a supported payment method, decoded as `InstantPaymentMethod` (same as iDEAL)
- Refactored `AnyPaymentMethodDecoder.decode()` to prioritize type-specific decoders from the dictionary
- Moved `isIssuersList` handling to a fallback for unknown payment methods with issuers

### Why
- Cleaner, data-driven approach via the decoder dictionary
- Removes special-case branching for individual payment methods
- `isIssuersList` now only catches truly unknown payment methods with issuers

### Testing
- Added IRIS test data with issuers to verify it decodes as `InstantPaymentMethod`, not `IssuerListPaymentMethod`

### Demo

This is a demo of IRIS as a Hosted Checkout component.

https://github.com/user-attachments/assets/d9eb6d4f-f45b-4c84-bf11-280fe635fc56

## Ticket
<ticket>
COSDK-1149
</ticket>

## Checklist

- [x] Tested changes locally

### DIFF
--- a/Adyen/Core/Payment Methods/Abstract/AnyPaymentMethodDecoder.swift
+++ b/Adyen/Core/Payment Methods/Abstract/AnyPaymentMethodDecoder.swift
@@ -92,7 +92,8 @@ internal enum AnyPaymentMethodDecoder {
         .twint: TwintPaymentMethodDecoder(),
         .payByBankAISDD: PayByBankUSPaymentMethodDecoder(),
         .payTo: PayToPaymentMethodDecoder(),
-        .iris: IrisPaymentMethodDecoder()
+        .iris: IrisPaymentMethodDecoder(),
+        .ideal: InstantPaymentMethodDecoder()
     ]
     
     private static var defaultDecoder: PaymentMethodDecoder = InstantPaymentMethodDecoder()
@@ -105,18 +106,6 @@ internal enum AnyPaymentMethodDecoder {
             let brand = try? container.decode(String.self, forKey: .brand)
             let isIssuersList = try container.containsValue(.issuers)
             
-            if type == .ideal || type == .iris {
-                return try InstantPaymentMethodDecoder().decode(from: decoder, isStored: isStored)
-            }
-            
-            if isIssuersList {
-                if type == .onlineBankingCZ || type == .onlineBankingSK {
-                    return try OnlineBankingPaymentMethodDecoder().decode(from: decoder, isStored: isStored)
-                }
-                
-                return try IssuerListPaymentMethodDecoder().decode(from: decoder, isStored: isStored)
-            }
-
             // This is a hack to handle stored Bancontact as a separate
             // payment method, even though Bancontact is just another
             // scheme of a card payment method,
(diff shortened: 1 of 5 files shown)


Adyen/Core/Payment Methods/Abstract/PaymentMethodType.swift
```diff
@@ -64,6 +64,7 @@ public enum PaymentMethodType: RawRepresentable, Hashable, Codable {
     case twint
     case payByBankAISDD
     case payTo
+    case iris
```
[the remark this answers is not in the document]
nauaros: good catch, added it there

Adyen/Core/Payment Methods/Abstract/AnyPaymentMethodDecoder.swift
```diff
@@ -128,8 +118,17 @@ internal enum AnyPaymentMethodDecoder {
...
             
-            let paymentDecoder = type.map { decoders[$0, default: defaultDecoder] } ?? defaultDecoder
-            return try paymentDecoder.decode(from: decoder, isStored: isStored)
+            // Use type-specific decoder if available
+            if let type, let paymentDecoder = decoders[type] {
+                return try paymentDecoder.decode(from: decoder, isStored: isStored)
+            }
+            
+            // Fallback: decode unknown payment methods with issuers as IssuerListPaymentMethod
+            if isIssuersList {
```
erenbesel, line 127: Are we still keeping this as it will mean any new Issuer that we want to enable as Hosted will be issuer on older versions.

erenbesel (approved):
Nice improvement! one question about issuer decoding but maybe we want to keep changes minimal for v5?

descorp (approved):

Adyen/Core/Payment Methods/Abstract/AnyPaymentMethodDecoder.swift
nauaros: You raise a good point. The only reason for this check to exist is to parse as issuerListComp any PM that we are currently not explicit about.

Let's discuss as a team if its worth the risk of removing it on v5. Added to Monday's agenda

Adyen/Core/Payment Methods/Abstract/AnyPaymentMethodDecoder.swift
```diff
@@ -104,18 +106,6 @@ internal enum AnyPaymentMethodDecoder {
             let brand = try? container.decode(String.self, forKey: .brand)
             let isIssuersList = try container.containsValue(.issuers)
             
-            if type == .ideal {
```
atmamont, line 107: Every change made here should have coverage that passes on the existing v5 branch and it should pass with this change. Please ensure not a single change here remains uncovered.

atmamont (commented):
Looks good, but to make sure we don't break it, let's ensure we have all cases covered.

Adyen/Core/Payment Methods/Abstract/AnyPaymentMethodDecoder.swift
nauaros: These changes did not trigger any red, thus the logic remains exactly the same but cleaner and simplified.
Following your advice, I added an extra test for iDeal which probably we should have had all this time. Here's the [commit](https://github.com/Adyen/adyen-ios/pull/2521/changes/33d9a02cef832734fbb0d91b53b8e684ec2ff7f7). The DummyPaymentMethods is accurate 1:1 with the backend response.

Tests/UnitTests/Core/PaymentMethodTests.swift
```diff
@@ -336,8 +336,9 @@ class PaymentMethodTests: XCTestCase {
         XCTAssertEqual(cashAppPay.clientId, "testClient")
         XCTAssertEqual(cashAppPay.scopeId, "testScope")
         
+        // iDEAL has issuers but should decode as InstantPaymentMethod, not IssuerListPaymentMethod
```
atmamont, line 339: Let's extract?
```swift
let iDealPaymentMethod = paymentMethods.regular[34]
```
It will make the following assertions read much easier

atmamont (approved):

Tests/UnitTests/Core/PaymentMethodTests.swift
nauaros: Good suggestion. Changed it to:

```swift
        let iDealPaymentMethod = try XCTUnwrap(paymentMethods.regular[34] as? InstantPaymentMethod)
```

erenbesel (approved):

PR #2533, opened 2026-05-07
Session Refactor: Async callbacks

erenbesel:
# Summary [Required]
Remove checkout being session delegate and update session protocol to use async methods.

## Ticket [Optional]
<ticket>
COSDK-907
</ticket>

## Checklist [Required]

- [x] Tested changes locally


AdyenSession/Session.swift
```diff
@@ -28,6 +28,11 @@ public final class Session: SessionProtocol {
     
     internal let context: AdyenContext
     
+    package var currentResult: CheckoutResult? {
+        guard let resultCode = state.resultCode else { return nil }
```
atmamont, line 32: _Thinking:_ Do we allow users technically to somehow create multiple sessions? Thinking of multi-session setup here and whether it can break the code.

AdyenSession/Session.swift
```diff
@@ -127,6 +132,31 @@ public final class Session: SessionProtocol {
         )
     }
     
+    // MARK: - API
+    
+    @MainActor
```
atmamont, line 137: Why does it need to be on main thread?

atmamont (approved):

AdyenSession/SessionProtocol.swift
```diff
@@ -12,9 +12,16 @@ import AdyenNetworking
 
 package protocol SessionProtocol: AnyObject {
     var state: Session.State { get }
+    var currentResult: CheckoutResult? { get }
     var delegate: SessionDelegate? { get set }
     var presentationDelegate: PresentationDelegate? { get set }
     
+    @MainActor
```
robertdalmeida, line 19: nonblocking; 
Is there any reason to have this on the main actor for these methods. they seem to only send an api request.

robertdalmeida (approved):

AdyenSession/Session.swift
erenbesel: No, one session per checkout.

AdyenSession/Session.swift
erenbesel: Good catches from both of you. Don't think these are needed indeed.

AdyenSession/Session.swift
```diff
@@ -127,6 +132,31 @@ public final class Session: SessionProtocol {
...
+            sessionId: state.identifier,
+            sessionData: state.data,
+            data: data
+        )
+        let response: PaymentsResponse = try await apiClient.performAsync(request)
+        return try response.asSubmitResult()
+    }
+    
+    @MainActor
+    package func performAdditionalDetails(_ data: ActionComponentData) async throws -> AdditionalDetailsResult {
```
nauaros, line 149: Same here right?

erenbesel: Yes updating both on the next PR already

AdyenSession/SessionProtocol.swift
erenbesel: Updating on the next PR

PR #2539, opened 2026-05-11
[v6] Add new SDK data fields

nauaros:
## Summary

_Disclaimer:_ This PR introduces the same changes applied in _v5_

This PR introduces some key changes related with `SDKData`:
- Adds 4 new fields to `SDKData` in order to support hosted checkout components. These are:
  - `channel`
  - `platform`
  - `sdkVersion`
  - `paymentMethodBehavior`
- Every usage of `channel` and `platform` needs to be **lowercase**. Thus, every instance of `iOS` has been replaced with `ios`.
- Introduces `CheckoutPlatformParams`. [Similar to Android](https://github.com/Adyen/adyen-android/blob/main/core/src/main/java/com/adyen/checkout/core/common/internal/helper/CheckoutPlatformParams.kt), there's a centralized place where _cross-platform_ can override platform parameters such as `version` and `channel`.

## Ticket
<ticket>
COSDK-1215
</ticket>

## Checklist

- [x] Tested changes locally


Adyen/Helpers/CheckoutPlatformParams.swift
```diff
@@ -0,0 +1,59 @@
...
+// Copyright (c) 2026 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+@_spi(AdyenInternal)
+/// Reference to the Android counterpart: https://github.com/Adyen/adyen-android/blob/main/core/src/main/java/com/adyen/checkout/core/common/internal/helper/CheckoutPlatformParams.kt
+public final class CheckoutPlatformParams {
```
nauaros, line 11: This should remain spi-public as it is accessed by cross platform.

Adyen/Model/SDKData.swift
```diff
@@ -6,51 +6,90 @@
 
 import Foundation
 
-package struct SDKData: Codable {
-    
+@_spi(AdyenInternal)
+public struct SDKData: Codable {
```
nauaros, line 11: Temporarily switching `SDKData` into spi-public as it is required due to its usage in `PaymentComponent`.

Of course, to be changed back to `package` once we convert `PaymentComponent` and all its dependencies.

nauaros: @robertdalmeida I don't know if you already tackled or plan to transform `PaymetComponent`

erenbesel: Let's add a todo to remember otherwise we may think this spi was intentional and keep it.

erenbesel (approved):

Adyen/Model/SDKData.swift
nauaros: Good point, I'll do that

robertdalmeida (approved):

PR #2542, opened 2026-05-12
[v6] Remove automatic issuer-based payment method decoding

nauaros:
## Summary

This PR removes the implicit logic in `AnyPaymentMethodDecoder` that automatically decodes any payment method containing `issuers` as an `IssuerListPaymentMethod`.

A payment method having issuers does not necessarily mean it should be treated as an issuer list payment method. Going forward, payment methods must be explicitly added to the decoding strategy.

If a payment method type is not explicitly handled, it will now fallback to `InstantPaymentMethod`.

## Changes

- Removed issuer-based detection logic from `AnyPaymentMethodDecoder`
- Updated fallback decoding to use `InstantPaymentMethodDecoder`
- Added any missing `IssuerListPaymentMethod` types explicitly to the decoding strategy
- Ensured unknown payment methods are safely decoded as `InstantPaymentMethod`

## Expected Behavior

- Unknown payment methods are decoded as `InstantPaymentMethod`
- Payment methods with issuers are only decoded as `IssuerListPaymentMethod` when explicitly registered in the decoding strategy
- Payment methods with issuers that are not explicitly registered will fallback to `InstantPaymentMethod`

## Ticket
<ticket>
COSDK-1220
</ticket>

## Checklist

- [x] Tested changes locally


erenbesel (approved):

robertdalmeida (approved):

Adyen/Core/Payment Methods/Abstract/AnyPaymentMethodDecoder.swift
```diff
@@ -62,7 +62,6 @@ internal enum AnyPaymentMethodDecoder {
         .achDirectDebit: ACHDirectDebitPaymentMethodDecoder(),
         .applePay: ApplePayPaymentMethodDecoder(),
         .payPal: PayPalPaymentMethodDecoder(),
-        .bizum: InstantPaymentMethodDecoder(),
```
robertdalmeida, line 65: nonblocking; I wan't able to understand why bizum is removed, but i hope it was intentional.

nauaros: As discussed, anything not explicitly added in the decoder will be handled as a `InstantPaymentMethod`

Tests/UnitTests/Core/PaymentMethodTests.swift
```diff
@@ -1224,6 +1231,33 @@ class PaymentMethodTests: XCTestCase {
         wait(for: [expectation], timeout: 10)
     }
     
+    // MARK: - Pay by Bank
+    
+    func test_decodingPayByBankPaymentMethod_withoutIssuers_decodesAsInstant() throws {
```
atmamont, line 1236: Could you please ensure Android has similar test coverage capturing this logic?

atmamont (approved):

atmamont:
I'd be happy if you can run the "check-api-alignment" skill on this PR to see if it's valuable

Tests/UnitTests/Core/PaymentMethodTests.swift
nauaros: Android already has equivalent test coverage for this logic. The tests are in [DefaultPayByBankDelegateTest.kt](https://github.com/Adyen/adyen-android/blob/main/paybybank/src/test/java/com/adyen/checkout/paybybank/internal/ui/DefaultPayByBankDelegateTest.kt#L155)

PR #2544, opened 2026-05-13
package access control adoption - make `PaymentComponent` to `package` (2/4)

robertdalmeida:
# Summary

1. Making `PaymentComponent` package & all the components

- As a consequence, the way `ComponentManager` builds components changes. see: `buildComponent(for paymentMethod: PaymentMethod)`
- Had to remove the default implementation to create the `PaymentComponet` in `PaymentMethod.swift` which asserts, now the expectation will be when we add a payment method to explicitly conform to this protocol or else it will default to build a `InstantPaymentComponent` 

2. AdyenCheckout.PaymentCheckout.createDropIn code has been commented out. 
public func createDropIn() -> (any AnyDropInComponent)

- Dropin is not yet ready to be exposed. And exposing AnyDropInComponent leads to a lot of related types to be public. 

3. ActionComponent has been made public to package. 

What's next?
1. Make a list of the existing Public types and check which to make package. (multiple rounds)

## Ticket [Optional]
<ticket>
COSDK-470
</ticket>

## Checklist [Required]

- [x] Tested changes locally
- Did not Added/updated unit tests


Adyen/Core/Core Protocols/PaymentMethod.swift
```diff
@@ -25,24 +25,11 @@ public protocol PaymentMethod: Codable {
...
-    @_spi(AdyenInternal)
-    func buildComponent(using builder: PaymentComponentBuilder) -> PaymentComponent? {
-        AdyenAssertion.assertionFailure(
-            message: "`@_spi(AdyenInternal) \(#function)` needs to be implemented on `\(String(describing: Self.self))`"
-        )
-        
-        return nil
-    }
+/// A payment method that can build a `PaymentComponent` using a `PaymentComponentBuilder`.
+package protocol PaymentComponentBuildable {
```
erenbesel, line 21: The whole PaymentComponentBuilder is going to be removed, to be replaced by the new CheckooutComponentBuilder/factories. So we don't need a new protocol here. Can we somehow make it without adding these?

robertdalmeida: ~~Good to hear `CheckooutComponentBuilder` will be removed. (i don't like that pattern at all)~~
This isn't actually a "new-new protocol" i.e. adding any new interface, it moves the protocol function interface from the public protocol to package. 

When we remove PaymentComponentBuilder, then i would think this package protocol will need to be extracted into whichever pattern. Which wouldn't become difficult by this change but rather help towards that end imho.

erenbesel: CheckooutComponentBuilder will not be removed, that's the new one. So you don't like the new one with factories? :)

The one that ties payment method strongly to its building (this one) will be removed, so this new protocol would have to be removed as well

robertdalmeida: typo 🙈: `ComponentManager: PaymentComponentBuilder` i meant this one. As i understand this one will be removed. 🤞
CheckooutComponentBuilder is 🚀 :) 

> so this new protocol would have to be removed as well

Yes, we can/should remove this. This protocol is only for splitting and package ACL.

nauaros (approved):

Adyen/Core/Payment Methods/ACHDirectDebitPaymentMethod.swift
```diff
@@ -75,3 +65,17 @@ public struct StoredACHDirectDebitPaymentMethod: StoredPaymentMethod, PaymentMet
         case bankAccountNumber
     }
 }
+
+// MARK: - PaymentComponentBuildable
+
+extension ACHDirectDebitPaymentMethod: PaymentComponentBuildable {
```
erenbesel, line 71: So is there no other way than to add a intermediate protocol that all payment methods must conform only to be removed when we move fully to factories? :D

AdyenCheckout/Flows/PaymentCheckout.swift
```diff
@@ -29,8 +29,10 @@ public class PaymentCheckout: BaseCheckout {
...
     }
 
-    /// Creates a Drop-in component with all available payment methods.
-    public func createDropIn() -> (any AnyDropInComponent)? {
-        core.createDropIn()
-    }
+    // TODO: Dropin: When dropin is supported then we should expose this method.
+    // Currently by exposing this method we make public all related types to AnyDropInComponent. Which are V5 api and not v6.
+//    /// Creates a Drop-in component with all available payment methods.
+//    public func createDropIn() -> (any AnyDropInComponent)? {
```
erenbesel, line 54: There is no need for this method to return a protocol, must be an oversight. We can return DropInComponent directly. Then can wekeep it uncommented?

erenbesel (commented):
Ultra nice changes. Small questions for clarity

AdyenCheckout/Flows/PaymentCheckout.swift
robertdalmeida: By returning either DropInComponent or AnyDropInComponent, i would need to make `DropInComponentDelegate` public too which then would mean even `PaymentComponent`. 
So there is a ripple effect. 

I see 2 options, which reduces public footprint.
1. Comment the code and uncomment when ready. (current)
2. Create a temporary public protocol without any methods. 

Which one do you think is better for this?

Adyen/Core/Payment Methods/ACHDirectDebitPaymentMethod.swift
robertdalmeida: yeah, i don't see any other approach here, other than making the factories ofcourse. 

Having this Buildable intermediate protocol atleast keeps it composable and we can indeed remove this Adyen Internal protocol whenever needed.

AdyenCheckout/Flows/PaymentCheckout.swift
erenbesel: Alright let's go with 1 indeed

erenbesel (approved):

PR #2553, opened 2026-05-20
Session Patching/onBeforeSubmit

erenbesel:
Introduces the `onBeforeSubmit` callback for the session flow, allowing merchants to get access to some of the payment data and modify them if needed while also giving them the opportunity to update the `amount` via patching the session (v72+) and returning a new sessionData.

- onBeforeSubmit public callback
- BeforeSubmitCallbackHandler with composition to handle its result.
- If aborted, reset component UI and don’t call onSubmit

## Ticket [Optional]
<ticket>
COSDK-907
</ticket>

## Checklist [Required]

- [x] Tested changes locally

### DIFF
--- a/Adyen/Model/PaymentComponentData.swift
+++ b/Adyen/Model/PaymentComponentData.swift
@@ -160,7 +160,7 @@ public struct PaymentComponentData {
         )
     }
 
-    package func applying(_ beforeSubmitData: BeforeSubmitData) -> PaymentComponentData {
+    package func replacing(beforeSubmitData: BeforeSubmitData) -> PaymentComponentData {
         var copy = self
         copy.beforeSubmitData = beforeSubmitData
         return copy
--- a/Adyen/Model/PaymentComponentData.swift
+++ b/Adyen/Model/PaymentComponentData.swift
@@ -29,16 +29,17 @@ public struct PaymentComponentData {
     /// The installments object.
     public let installments: Installments?
     
+    /// Data applied via onBeforeSubmit callback. Takes priority for its fields.
+    private var beforeSubmitData: BeforeSubmitData?
+    
     /// Shopper name.
     public var shopperName: ShopperName? {
-        guard let shopperInfo = paymentMethod as? ShopperInformation else { return nil }
-        return shopperInfo.shopperName
+        beforeSubmitData?.shopperName ?? (paymentMethod as? ShopperInformation)?.shopperName
     }
 
     /// The email address.
     public var emailAddress: String? {
-        guard let shopperInfo = paymentMethod as? ShopperInformation else { return nil }
-        return shopperInfo.emailAddress
+        beforeSubmitData?.shopperEmail ?? (paymentMethod as? ShopperInformation)?.emailAddress
     }
 
     /// The telephone number.
@@ -52,14 +53,12 @@ public struct PaymentComponentData {
 
     /// The billing address information.
     public var billingAddress: PostalAddress? {
-        guard let shopperInfo = paymentMethod as? ShopperInformation else { return nil }
-        return shopperInfo.billingAddress
+        beforeSubmitData?.billingAddress ?? (paymentMethod as? ShopperInformation)?.billingAddress
     }
     
     /// The delivery address information.
     public var deliveryAddress: PostalAddress? {
-        guard let shopperInfo = paymentMethod as? ShopperInformation else { return nil }
-        return shopperInfo.deliveryAddress
+        beforeSubmitData?.deliveryAddress ?? (paymentMethod as? ShopperInformation)?.deliveryAddress
     }
     
     /// The social security number.
@@ -159,5 +158,11 @@ public struct PaymentComponentData {
             browserInfo: browserInfo,
             installments: installments
         )
+    }
+
+    package func applying(_ beforeSubmitData: BeforeSubmitData) -> PaymentComponentData {
+        var copy = self
+        copy.beforeSubmitData = beforeSubmitData
+        return copy
     }
 }
--- a/AdyenCheckout/Callbacks/CheckoutCallbackHandlers.swift
+++ b/AdyenCheckout/Callbacks/CheckoutCallbackHandlers.swift
@@ -38,6 +38,51 @@ package final class SessionCallbackHandler: CheckoutCallbackHandling {
 }
 
 @MainActor
+package final class BeforeSubmitCallbackHandler: CheckoutCallbackHandling {
+
+    private let innerHandler: any CheckoutCallbackHandling
+    private let session: SessionProtocol
+    private let callbackStore: SessionCheckoutCallbackStore
+
+    package init(
+        inner: any CheckoutCallbackHandling,
+        session: SessionProtocol,
+        callbackStore: SessionCheckoutCallbackStore
+    ) {
+        self.innerHandler = inner
+        self.session = session
+        self.callbackStore = callbackStore
+    }
+
+    package func handleSubmit(_ data: PaymentComponentData) async throws -> SubmitResult {
+        var submitData = data
+        if let onBeforeSubmit = callbackStore.onBeforeSubmit {
+            let inputData = BeforeSubmitData(
+                billingAddress: data.billingAddress,
+                deliveryAddress: data.deliveryAddress,
+                shopperName: data.shopperName,
+                shopperEmail: data.emailAddress
+            )
+            let result = try await onBeforeSubmit(inputData)
+            switch result {
+            case let .proceed(modifiedData, sessionData):
+                if let sessionData {
+                    try await session.refreshSessionState(with: sessionData)
+                }
+                submitData = data.applying(modifiedData)
+            case .abort:
+                throw CallbackError.beforeSubmitAborted
+            }
+        }
+        return try await innerHandler.handleSubmit(submitData)
+    }
+
+    package func handleAdditionalDetails(_ data: ActionComponentData) async throws -> AdditionalDetailsResult {
+        try await innerHandler.handleAdditionalDetails(data)
+    }
+}
+
+@MainActor
 package final class AdvancedCallbackHandler: CheckoutCallbackHandling {
     private let callbackStore: AdvancedCheckoutCallbackStore
 
--- a/AdyenCheckout/Callbacks/CheckoutCallbackHandlers.swift
+++ b/AdyenCheckout/Callbacks/CheckoutCallbackHandlers.swift
@@ -40,45 +40,51 @@ package final class SessionCallbackHandler: CheckoutCallbackHandling {
 @MainActor
 package final class BeforeSubmitCallbackHandler: CheckoutCallbackHandling {
 
-    private let innerHandler: any CheckoutCallbackHandling
+    private let handler: any CheckoutCallbackHandling
     private let session: SessionProtocol
     private let callbackStore: SessionCheckoutCallbackStore
 
     package init(
-        inner: any CheckoutCallbackHandling,
+        handler: any CheckoutCallbackHandling,
         session: SessionProtocol,
         callbackStore: SessionCheckoutCallbackStore
     ) {
-        self.innerHandler = inner
+        self.handler = handler
         self.session = session
         self.callbackStore = callbackStore
     }
 
     package func handleSubmit(_ data: PaymentComponentData) async throws -> SubmitResult {
-        var submitData = data
-        if let onBeforeSubmit = callbackStore.onBeforeSubmit {
-            let inputData = BeforeSubmitData(
-                billingAddress: data.billingAddress,
-                deliveryAddress: data.deliveryAddress,
-                shopperName: data.shopperName,
-                shopperEmail: data.emailAddress
-            )
-            let result = try await onBeforeSubmit(inputData)
-            switch result {
-            case let .proceed(modifiedData, sessionData):
-                if let sessionData {
-                    try await session.refreshSessionState(with: sessionData)
-                }
-                submitData = data.applying(modifiedData)
-            case .abort:
-                throw CallbackError.beforeSubmitAborted
-            }
-        }
-        return try await innerHandler.handleSubmit(submitData)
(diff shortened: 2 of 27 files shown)


Adyen/Model/PaymentComponentData.swift
```diff
@@ -29,16 +29,17 @@ public struct PaymentComponentData {
...
     public let installments: Installments?
     
+    /// Data applied via onBeforeSubmit callback. Takes priority for its fields.
+    private var beforeSubmitData: BeforeSubmitData?
+    
     /// Shopper name.
     public var shopperName: ShopperName? {
-        guard let shopperInfo = paymentMethod as? ShopperInformation else { return nil }
-        return shopperInfo.shopperName
+        beforeSubmitData?.shopperName ?? (paymentMethod as? ShopperInformation)?.shopperName
```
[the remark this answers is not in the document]
atmamont: Is there a use case when one needs to nullify any of these fields?

Adyen/Model/PaymentComponentData.swift
```diff
@@ -160,4 +159,10 @@ public struct PaymentComponentData {
             installments: installments
         )
     }
+
+    package func applying(_ beforeSubmitData: BeforeSubmitData) -> PaymentComponentData {
```
atmamont: Is there a semantic difference between updating `BrowserData` and `BeforeSubmitData`?
Should we use the same approach and use `replacing()`?

AdyenCheckout/Callbacks/CheckoutCallbackHandlers.swift
```diff
@@ -37,6 +37,51 @@ package final class SessionCallbackHandler: CheckoutCallbackHandling {
     }
 }
 
+@MainActor
+package final class BeforeSubmitCallbackHandler: CheckoutCallbackHandling {
+
+    private let innerHandler: any CheckoutCallbackHandling
```
atmamont: I don't have a better name from the top of my head, but `inner` feels like a crutch. To me a handler would be sufficient along with the `private` access modifier.

AdyenCheckout/Callbacks/CheckoutCallbackHandlers.swift
```diff
@@ -37,6 +37,51 @@ package final class SessionCallbackHandler: CheckoutCallbackHandling {
...
+    private let session: SessionProtocol
+    private let callbackStore: SessionCheckoutCallbackStore
+
+    package init(
+        inner: any CheckoutCallbackHandling,
```
atmamont: Adding to the comment above, `inner` as an init parameter name is quite confusing.

AdyenCheckout/Callbacks/CheckoutCallbackHandlers.swift
```diff
@@ -37,6 +37,51 @@ package final class SessionCallbackHandler: CheckoutCallbackHandling {
...
+    ) {
+        self.innerHandler = inner
+        self.session = session
+        self.callbackStore = callbackStore
+    }
+
+    package func handleSubmit(_ data: PaymentComponentData) async throws -> SubmitResult {
+        var submitData = data
+        if let onBeforeSubmit = callbackStore.onBeforeSubmit {
+            let inputData = BeforeSubmitData(
```
atmamont: Can we extract this part of the handling logic to a private helper method `handleOnBeforeSubmit` to make this less cluttered?

AdyenSession/Session.swift
```diff
@@ -123,6 +123,11 @@ extension Session {
 
 extension Session {
     
+    package func refreshSessionState(with sessionData: String) async throws {
+        state.data = sessionData
+        state = try await refreshedState()
+    }
```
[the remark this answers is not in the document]
atmamont: This sounds as a valid concern

Demo/Common/IntegrationExamples/Session/Components/ApplePayComponentExample.swift
```diff
@@ -27,10 +29,12 @@ internal final class ApplePayComponentExample: InitialDataFlowProtocol {
 
     internal func start() {
         startLoading()
+        latestApplePayAmount = ConfigurationConstants.current.amount
 
         Task {
             do {
-                let sessionResponse = try await requestSessionInitialInfo()
+                let patchableSession = ConfigurationConstants.current.applePaySettings.onBeforeSubmitMode == .patchSession
+                let sessionResponse = try await requestSessionInitialInfo(payable: !patchableSession)
```
atmamont, line 37: I can't follow the logic connection between `onBeforeSubmitMode == .patchSession` and "payable" value. I can think of it making some sense as other two values are "abort" and "patch", but it really takes some time to understand and make assumptions.

Demo/Common/IntegrationExamples/Session/Components/ApplePayComponentExample.swift
```diff
@@ -63,6 +67,58 @@ internal final class ApplePayComponentExample: InitialDataFlowProtocol {
                         )
                         return PKPaymentAuthorizationResult(status: .failure, errors: [postalCodeError])
                     }
+                }.onShippingContactChange { contact, summaryItems in
+                    var items = summaryItems
```
atmamont: Let's extract items mutation to a separate method?

Demo/Common/IntegrationExamples/Session/Components/ApplePayComponentExample.swift
```diff
@@ -63,6 +67,58 @@ internal final class ApplePayComponentExample: InitialDataFlowProtocol {
...
+                            amount: NSDecimalNumber(value: 5.0)
+                        ))
+                        items.append(.init(label: last.label, amount: NSDecimalNumber(value: last.amount.floatValue + 5.0)))
+                    }
+                    self.updateLatestApplePayAmount(using: items)
+                    return PKPaymentRequestShippingContactUpdate(paymentSummaryItems: items)
+                }
+                .onShippingMethodChange { shippingMethod, summaryItems in
+                    var items = summaryItems
+                    if let last = items.last {
```
atmamont: The same suggestion here

Demo/Common/IntegrationExamples/Session/Components/ApplePayComponentExample.swift
```diff
@@ -114,6 +189,33 @@ internal final class ApplePayComponentExample: InitialDataFlowProtocol {
             self.presenter?.presentAlert(withTitle: title, message: message)
         }
     }
+
+    private func patchSession(using sessionResponse: SessionResponse) async throws -> SessionPatchResponse {
```
atmamont: What do you think of marking this part of the code as "your backend code"? Or even extracting it to a separate file?

Demo/Common/Networking/Environment.swift
```diff
@@ -28,6 +28,10 @@ internal struct DemoAPIContext: AnyAPIContext {
 internal enum DemoCheckoutAPIEnvironment: String, AnyAPIEnvironment, CaseIterable {
     
     case test, local
+
+    private enum Constants {
+        static let localCheckoutAPIVersion = 72
```
atmamont, line 33: I'd add a comment here explaining the reason, and also mentioning the expected outcome when trying this on a versions <72 (like an expected backend error maybe, in the case if mobile team updates their integration and backend is not ready yet)

Demo/Common/Networking/Environment.swift
```diff
@@ -39,24 +43,22 @@ internal enum DemoCheckoutAPIEnvironment: String, AnyAPIEnvironment, CaseIterabl
...
-    case beta, test, local
-    
-    internal var baseURL: URL {
         switch self {
-        case .beta:
-            return URL(string: "https://pal-beta.adyen.com/pal/servlet/")!
         case .test:
-            return URL(string: "https://pal-test.adyen.com/pal/servlet/")!
+            // The test Checkout API version is taken from MERCHANT_SERVER_HOST in Demo/Secrets*.xcconfig.
+            return checkoutAPIVersion(from: ConfigurationConstants.serverUrl) ?? Constants.localCheckoutAPIVersion
```
atmamont, line 49: Thinking of the version coming from the URL itself, should we think of making session patching unavailable in the settings screen if detected version <72?

Demo/Common/Networking/SessionRequest.swift
```diff
@@ -57,7 +64,7 @@ internal struct SessionRequest: APIRequest {
         if ConfigurationConstants.current.cardSettings.showStorePaymentMethod {
             AdyenAssertion.assert(
                 message: "API version should be v70 or above to apply card component's store payment method field",
-                condition: ConfigurationConstants.current.apiVersion < 70
+                condition: ConfigurationConstants.demoServerEnvironment.version < 70
```
atmamont, line 67: Following the prior comment - let's use similar check for session patching?

atmamont (commented):
Great work, some comments on:
- code clarity and readability
- you can ditch everything that I commented on the demo app, as it has lower priority, just something to discuss and address in the future

Adyen/Model/PaymentComponentData.swift
erenbesel: Browserdata is cached and then replaced next time, this one is usually a one time assignment. But for consistency might be better to change

AdyenCheckout/Callbacks/CheckoutCallbackHandlers.swift
erenbesel: Was hard to name indeed, can go with just handler then

Demo/Common/IntegrationExamples/Session/Components/ApplePayComponentExample.swift
erenbesel: `payable` is coupled with session patching. The values are just settings added on the demo for us (simulate patching, simulate just update data, simulate abort), not expected real life scenarios but we need to be able to test this. The other option would be to always have patching on Apple pay example and no settings? then we wouldn't be showcasing other scenarios of abort and update data without patching

Demo/Common/Networking/Environment.swift
erenbesel: I think these may be too many checks and checks for our demo app :)

Demo/Common/Networking/SessionRequest.swift
erenbesel: Same as above, not sure why we added this check. If the api version is below something, the calls will fail anyway. Don't see the value on these currently

Demo/Common/Networking/Environment.swift
erenbesel: There is another comment explaining this variable is not used for actual endpoints and this is just local.  apiVersion now comes from the secrets.xcconfig file. I cleaned up unused code but kept this for `local` case, which maybe we can even remove that case?

atmamont: Yeah, we can just assert if the value is missing from the `xcconfig` the same way we do it for the rest of critical values.

Demo/Common/Networking/Environment.swift
atmamont: Agreed :)

Demo/Common/Networking/SessionRequest.swift
atmamont: Sure, my point was about code consistency. I am fine if you chose to delete this redundant check instead of adding a new one

Demo/Common/IntegrationExamples/Session/Components/ApplePayComponentExample.swift
atmamont: Would renaming "payable" to something more explanatory be a quicker solution?

AdyenCheckout/Core/CheckoutCore+Orchestration.swift
```diff
@@ -58,6 +58,11 @@ internal extension CheckoutCore {
                 let submitResult = try await onSubmit()
                 guard !Task.isCancelled else { return }
                 self?.handle(submitResult: submitResult, source: source)
+            } catch CallbackError.beforeSubmitAborted {
+                // catch beforeSubmit abort here and reset the component UI
+                guard !Task.isCancelled else { return }
+                source.paymentComponent.stopLoading()
```
robertdalmeida: nonblocking; i assume that we want the source to stop loading (either dropIn's paymentComponent or paymentComponent). 
So maybe a call to source.stopLoading() and the source decide which one to stop loading.

robertdalmeida (approved):

Demo/Common/IntegrationExamples/Session/Components/ApplePayComponentExample.swift
erenbesel: `payable` was the direct name of the backend property. It's a tough combo :D
should we rename? suggestions?

robertdalmeida (approved):

atmamont (approved):

AdyenSession/Session.swift
```diff
@@ -211,6 +216,18 @@ private extension Session {
...
     
+    func refreshedState() async throws -> State {
+        let initialInfo = SessionResponse(
+            id: state.identifier,
+            sessionData: state.data
+        )
+        return try await Self.makeSetupCall(
+            with: initialInfo,
+            baseAPIClient: apiClient,
+            order: nil
```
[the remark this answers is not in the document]
nauaros: We should add a TODO here to handle partial payments in sessions.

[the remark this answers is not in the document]
erenbesel: Session setup call step has been removed

AdyenSession/Session.swift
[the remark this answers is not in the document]
erenbesel: Done

Adyen/Model/PaymentComponentData.swift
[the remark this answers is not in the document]
erenbesel: Now we only allow non null values to be passed with public modifiers

atmamont (approved):

atmamont (approved):

PR #2562, opened 2026-05-22
Error handling foundation

erenbesel:
# Summary [Required]
Base changes of public error handling API with CheckoutError and its codes.

## Ticket [Optional]
<ticket>
COSDK-589
</ticket>

## Checklist [Required]

- [x] Tested changes locally


AdyenCheckout/Core/CheckoutCore+Orchestration.swift
```diff
@@ -155,11 +155,17 @@ internal extension CheckoutCore {
...
-        resultCallbacks.onError?(CheckoutError(error: error))
+        resultCallbacks.onError?(checkoutError(from: error))
     }
 }
 
 private extension CheckoutCore {
+
+    func checkoutError(from error: Error) -> CheckoutError {
+        if let checkoutError = error as? CheckoutError { return checkoutError }
+        if case ComponentError.cancelled = error { return CheckoutError(code: .cancelled, message: nil, underlyingError: error) }
```
atmamont: Do you foresee any other mapping we would need to do here?
I was thinking of `func map(_ error: Error) -> CheckoutError` or at least having `switch` here to make it easier to add cases.

AdyenCheckout/Flows/PaymentCheckout.swift
```diff
@@ -20,13 +20,32 @@ public class PaymentCheckout: BaseCheckout {
...
+    /// - Throws: ``CheckoutError`` with code ``CheckoutError/Code/unknown`` if the payment method
+    ///   is not available in the current payment methods, is not supported, or cannot be
+    ///   initialised on this device (e.g. Apple Pay hardware check failed).
     public func createPaymentComponent(for type: PaymentMethodType) throws -> CheckoutPaymentComponent {
-        try core.createPaymentComponent(for: type)
+        do {
+            return try core.createPaymentComponent(for: type)
+        } catch {
+            throw (error as? CheckoutError) ?? CheckoutError(code: .unknown, message: error.localizedDescription, underlyingError: error)
+        }
```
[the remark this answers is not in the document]
atmamont: Regardless of what gemini suggests, let's have all error mapping in one place?

AdyenCheckout/Flows/PaymentCheckout.swift
```diff
@@ -20,13 +20,32 @@ public class PaymentCheckout: BaseCheckout {
...
+    /// - Parameter identifier: The unique identifier of the stored payment method.
+    /// - Returns: A configured ``CheckoutPaymentComponent``.
+    /// - Throws: ``CheckoutError`` with code ``CheckoutError/Code/unknown`` if no stored
+    ///   payment method matching `identifier` exists.
     public func createPaymentComponent(for identifier: String) throws -> CheckoutPaymentComponent {
-        try core.createPaymentComponent(for: identifier)
+        do {
+            return try core.createPaymentComponent(for: identifier)
+        } catch {
+            throw (error as? CheckoutError) ?? CheckoutError(code: .unknown, message: error.localizedDescription, underlyingError: error)
```
atmamont: Same here, I'd love to have all of this cases handled once

AdyenCheckout/CheckoutConfiguration/CheckoutConfiguration.swift
```diff
@@ -147,6 +152,20 @@ public struct CheckoutConfiguration {
 
         return analyticsApiContext
     }
+    
+    private static func validateAmount(_ amount: Amount) throws {
+        guard Locale.Currency.isoCurrencies.contains(where: { $0.identifier == amount.currencyCode }) else {
```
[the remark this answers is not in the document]
erenbesel: It's 16.0 minimum

AdyenCheckout/Flows/PaymentCheckout.swift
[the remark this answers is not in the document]
erenbesel: Done, a new mapper in AdyenCheckout module which will convert all internal errors to CheckoutError

AdyenCheckout/Core/CheckoutCore+Orchestration.swift
erenbesel: Done

AdyenCheckout/Flows/PaymentCheckout.swift
erenbesel: Done

AdyenCheckout/CheckoutConfiguration/CheckoutConfiguration.swift
```diff
@@ -56,7 +56,11 @@ public struct CheckoutConfiguration {
     ///   - clientKey: The client key that corresponds to the web service user you will use for initiating the payment.
     ///   - content: Configuration builder to provide the desired configuration instances.
     ///   See https://docs.adyen.com/user-management/client-side-authentication for more information.
-    /// - Throws: `ClientKeyError.invalidClientKey` if the client key is invalid.
+    /// - Throws: `CheckoutError` with one of the following codes if the configuration is invalid:
```
robertdalmeida, line 59: nonblocking; 
Do we want to list all the error codes that would be possible to be thrown from this method? 
or do we want to simple use CheckoutError(lesser doc maintainence). 

The only draw back i see is that this would be a list that we would need to maintain in this public doc comment.

AdyenCheckout/CheckoutConfiguration/CheckoutConfiguration.swift
```diff
@@ -162,6 +167,20 @@ public struct CheckoutConfiguration {
...
+        guard Locale.Currency.isoCurrencies.contains(where: { $0.identifier == amount.currencyCode }) else {
+            throw CheckoutError(code: .invalidCurrencyCode, message: "Invalid currency code")
+        }
+        if let localeIdentifier = amount.localeIdentifier {
+            guard Locale.availableIdentifiers.contains(localeIdentifier) else {
+                throw CheckoutError(code: .invalidLocale, message: "Invalid locale")
+            }
+        }
+        guard amount.value >= 0 else {
+            throw CheckoutError(code: .invalidAmountValue, message: "Invalid amount value")
```
robertdalmeida, line 181: though; If the message is the same for the different codes. 
Maybe then we can allow the initializer to have only the code and CheckoutError internally map to the correct message. 

To make the callsite become `throw CheckoutError(code: .invalidAmountValue)`.

AdyenCheckout/Setup/Checkout.swift
```diff
@@ -20,6 +20,7 @@ public enum Checkout {
     ///   - configuration: The checkout configuration.
     ///   - presentationDelegate: Optional delegate for handling UI presentation.
     /// - Returns: A session checkout instance exposing session-compatible APIs.
+    /// - Throws: ``CheckoutError`` with code ``CheckoutError/Code/sessionSetupFailure`` if the session cannot be established.
```
robertdalmeida: Here too it reads as if this method could return only one error. 
But there are many try statements at multiple points and i would assume that we would be potentially throwing multiple errors.

AdyenCheckout/Core/CheckoutError+Mapping.swift
```diff
@@ -0,0 +1,31 @@
...
+    /// Maps any `Error` to a `CheckoutError`.
+    ///
+    /// `AdyenCheckout` is the boundary module that imports all internal SDK modules,
+    /// making it the single place where every internal error type can be exhaustively mapped.
+    ///
+    /// - If `error` is already a `CheckoutError`, it is returned as-is.
+    /// - Known SDK error types are mapped to their corresponding ``Code``.
+    /// - All other errors use `fallback` as the code (defaults to ``Code/unknown``).
+    /// - Parameter fallback: The code to use when no specific mapping is found.
+    static func map(_ error: Error, fallback: Code = .unknown) -> CheckoutError {
```
robertdalmeida: This sounds like more like an `init` to me. 

`map` usually is on the instance method level as i think.

````
    init(error: Error) {
        if let checkoutError = error as? CheckoutError { self = checkoutError }
        switch error {
        // Adyen
        case ComponentError.cancelled:
            self.init(code: .cancelled, message: nil, underlyingError: error)
        default:
            self.init(code: .unknown, message: error.localizedDescription, underlyingError: error)
        }
    }

````

**Review by robertdalmeida — APPROVED** [review]

File: AdyenCheckout/CheckoutConfiguration/CheckoutConfiguration.swift (hunk 5 above)
Comment (nauaros) [reply AdyenCheckout/CheckoutConfiguration/CheckoutConfiguration.swift:59]: I would keep it simple as it is and then merchant can always access the definition of `CheckoutError` in the source file.

File: Adyen/Core/APIClient/APIContext.swift [hunk 9]
```diff
@@ -37,11 +37,12 @@ public struct APIContext: AnyAPIContext {
     /// - Parameters:
     ///   - environment: The environment to retrieve internal resources from.
     ///   - clientKey: The client key that corresponds to the web service user you will use for initiating the payment.
-    /// - Throws: `ClientKeyError.invalidClientKey` if the client key is invalid.
+    /// - Throws: `CheckoutError` with code ``CheckoutError/Code/invalidClientKey`` if the client key is invalid.
     /// - Note: Always use the provided `Environment` type to ensure a correct environment.
     public init(environment: AnyAPIEnvironment, clientKey: String) throws {
         guard ClientKeyValidator().isValid(clientKey) else {
-            throw ClientKeyError.invalidClientKey
+            let message = "The entered client key is invalid. Valid client key starts with the environment name (e.g. live_XXXXXXXXXX)."
```
atmamont, line 44: Should this be localized? We recently received a support request for Apple Pay errors localization

atmamont (approved):

Adyen/Core/APIClient/APIContext.swift
erenbesel: Oh dear that would be overkill IMO. These are API error codes explaining the error in short.

AdyenCheckout/Core/CheckoutError+Mapping.swift
erenbesel: Hmm, good point. It was actually an init(error: Error) before, LLM changed to this.

AdyenCheckout/CheckoutConfiguration/CheckoutConfiguration.swift
erenbesel: But this would add more logic to Code type when now it is a plain rawrepsentable.

AdyenCheckout/Core/CheckoutError+Mapping.swift
erenbesel: Updated

AdyenCheckout/Setup/CheckoutProvider.swift
```diff
@@ -113,44 +113,51 @@ internal class CheckoutProvider: CheckoutProviding {
...
-
-        return try await AdyenContext(
-            apiContext: configuration.apiContext,
-            amount: configuration.amount,
-            publicKey: publicKey,
-            checkoutAttemptId: checkoutAttemptId,
-            analyticsAPIContext: configuration.analyticsApiContext,
-            analyticsConfiguration: configuration.analyticsConfiguration
-        )
+            // TODO: Improvement: Suggestion: instead of fetching the publicKey, which requires a bit of searching to understand
```
nauaros, line 141: If this is an improvement, does it make sense to have it as a YT ticket under tech debt instead of an actual TODO here?

nauaros (approved):

AdyenCheckout/Setup/CheckoutProvider.swift
erenbesel: we can convert all todos to tickets indeed

AdyenCheckout/CheckoutConfiguration/CheckoutConfiguration.swift
robertdalmeida: nonblocking; Just for my understanding is there any particular reason to list some of the errors here in the doc?

erenbesel: Now it contains all possible errors. Is it better?

robertdalmeida (approved):

PR #2564, opened 2026-05-27
switch to using xcode folder instead of groups

robertdalmeida:
Made all framework targets to use folders. 
Demo app and Snapshot folders are still group type.
- Demo app - because of the configuration files.
- Snapshot testing folder - Which is why in the project file you will see about 40 explicit .swift from the snapshot testing forlder in the `membershipExceptions`. For now I will let this be. 

**Hint to review**:  checkout the branch and go through the folders.

Changes
1. AdyenCard, AdyenActions now use the BundleSPMExtension using the #if def SWIFT_PACKAGE macro.
2. Moved mocks & helper types in the test folder to a common folder as they are used by both integration target and unit test target. 

Xcode project (project.pbxproj):
- Before: 12,741 lines
- After: 5,637 lines (-7,104 lines, ~56% reduction)


atmamont:
Do you see a clear path for both Demo and Snapshot targets conversion?

atmamont (approved):

nauaros (approved):

robertdalmeida:
> Do you see a clear path for both Demo and Snapshot targets conversion?

Not a clear path, but definitly possilbe. But as it is currently it is benificial with minimal effort.

PR #2568, opened 2026-05-29
Scaffold v6 docs and migration guide

atmamont:
# Summary

Introducing basic documentation:
- v5 to v6 migration guide
- card component documentation for both sessions and advanced flow
- theming and localization overrides

## Ticket 
<ticket>
COSDK-1235
</ticket>

## Checklist

- [x] Tested changes locally


docs/v6/card.md
```diff
@@ -0,0 +1,178 @@
...
+    amount: amount,
+    clientKey: clientKey
+) {
+    CardConfiguration()
+    AuthenticationConfiguration()
+        .requestorAppURL(URL(string: "your-app://adyen")!)
+}
+```
+
+## Session flow
```
nauaros: @atmamont what about if inside card you had the two integrations flows in different MDs:
- `card-session-flow.md` (default as we want to encourage merchants to adopt session)
- `card-advanced-flow.md`

docs/v6/README.md
```diff
@@ -0,0 +1,249 @@
...
+viewController.navigationItem.leftBarButtonItem = .init(
+    barButtonSystemItem: .cancel,
+    target: self,
+    action: #selector(cancelPressed)
+)
+
+present(navigationController, animated: true)
+```
+
+## Theme
```
nauaros, line 187: Thinking out loud there, should be have a dedicated page around `CheckoutTheme` with all its configurable properties?

atmamont: Yes! Though we don't expose much to merchants on purpose as we want to own the overall style, only allowing to override semantic colors.
Do you think it should go into the first docs version? If yes, I'd still make a separate PR for that.

docs/v6/card.md
atmamont: Yes, it makes sense to split them

docs/v6/README.md
nauaros: I would say so. It's one of the most drastic changes in v6, I believe having its own section would be nice. And from there, build on top of it if there's any confusion on how to use it.

docs/v6/README.md
```diff
@@ -0,0 +1,249 @@
...
+The public integration surface is centered around four concepts:
+
+- `Checkout.setup(...)` is the entry point for checkout flows.
+- `CheckoutConfiguration` is the shared configuration container.
+- Flow callbacks are configured with closures.
+- Payment components are created from `SessionCheckout` or `AdvancedCheckout`.
+
+## Installation
+
+Import `AdyenCheckout` together with the modules used by your flow:
```
erenbesel: Currently importing AdyenCheckout is enough but once/if we move to optional modules, then each one will be needed separately.

erenbesel (commented):
Tiny disclaimer but looking great!

MIGRATION.md
```diff
@@ -1,7 +1,224 @@
...
+}
+```
+
+#### Summary
+
+- `Checkout.setup(...)` replaces `AdyenSession.initialize(...)` for the new public v6 flows.
+- `CheckoutConfiguration` replaces flow-specific setup objects as the main integration entry point.
+- Closure callbacks replace the public delegate-first flow setup for submission and completion handling.
+- `SessionCheckout` and `AdvancedCheckout` create payment components and Drop-in instances for the active flow.
+- Theme and localization are configured at the checkout level through `theme(_:)` and `localizationProvider(_:)`.
```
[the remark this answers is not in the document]
atmamont: Good point, addressed.

docs/v6/README.md
```diff
@@ -0,0 +1,249 @@
...
+.onAdditionalDetails { data in
+    try await submitAdditionalDetailsToYourServer(data)
+}
+.onComplete { result in
+    print(result.resultCode)
+}
+.onError { error in
+    print(error.localizedDescription)
+}
+
```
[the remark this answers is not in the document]
atmamont: Good point, addressed.

docs/v6/card.md
atmamont: Done

docs/v6/README.md
atmamont: Fair, agreed. It's not that big actually, I've added a separate `theme.md` to this PR

docs/v6/README.md
atmamont: Good point, we can add this whenever we have our vision refined and the approach validated.

MIGRATION.md
```diff
@@ -1,7 +1,225 @@
...
+##### After (v6)
+
+```swift
+let checkout = try await Checkout.setup(
+    with: paymentMethods,
+    configuration: configuration,
+    presentationDelegate: self
+)
+.onSubmit { data in
+    try await submitToYourServer(data)
```
robertdalmeida: thought; Just generally onSubmit, onAdditionally now expects a return type SubmitResult, AdditionalDetailsResult. 
Does it need to be reflected here? (I'm not even sure - as the audience for this migration would potentially jus be some llm)

MIGRATION.md
```diff
@@ -1,7 +1,225 @@
...
+
+```swift
+let configuration = try CheckoutConfiguration(
+    environment: .test,
+    amount: amount,
+    clientKey: clientKey
+) {
+    CardConfiguration()
+    AuthenticationConfiguration()
+        .requestorAppURL(URL(string: "your-app://adyen")!)
```
robertdalmeida: Could you make this sample link as a universal link instead. This is also reccommended for 3ds2. 
https://your-domain

docs/v6/theme.md
```diff
@@ -0,0 +1,67 @@
...
+import AdyenUI
+import UIKit
+```
+
+## Apply a theme
+
+Configure the theme on `CheckoutConfiguration` before calling `Checkout.setup(...)`:
+
+```swift
+let theme = CheckoutTheme.default
```
atmamont: @OscarSpruit does this match the Android customization capabilities? We had discussions to shrink it down to only allow customizing semantic colors, and this is the right moment to make a call.

@erenbesel @nauaros @robertdalmeida ☝️

robertdalmeida (approved):

MIGRATION.md
atmamont: Good catch!

docs/v6/theme.md
OscarSpruit: Last time we discussed indeed to keep it very minimal. Based on that I made changes on Android to reflect that. Here is the structure we have on Android:
```kotlin
data class CheckoutTheme(
    val colors: CheckoutColors = CheckoutColors.light(),
    val textStyles: CheckoutTextStyles = CheckoutTextStyles.default(),
    val attributes: CheckoutAttributes = CheckoutAttributes.default(),
)

data class CheckoutColors(
    val background: CheckoutColor,
    val container: CheckoutColor,
    val containerOutline: CheckoutColor,
    val primary: CheckoutColor,
    val textOnPrimary: CheckoutColor,
    val highlight: CheckoutColor,
    val destructive: CheckoutColor,
    val textOnDestructive: CheckoutColor,
    val disabled: CheckoutColor,
    val textOnDisabled: CheckoutColor,
    val separator: CheckoutColor,
    val text: CheckoutColor,
    val textSecondary: CheckoutColor,
)

data class CheckoutTextStyles(
    // This points to a XML resource to change the general font used
    @FontRes val font: Int?,
)

data class CheckoutAttributes(
    val cornerRadius: Int,
)
```
So customizing a specific UI component or text style/font is no longer possible

robertdalmeida: I do have to do the sync between platforms about theming. COSDK-1107. (if needed then i'll update this section). I'll pick that up again next. 

But that shouldn't stop this guide which should reflect the current state.

atmamont: @robertdalmeida I removed styling from the public docs, and my next PR will make a code change to make init with colors public and hide other inits, so the docs are aligned with the code.

MIGRATION.md
atmamont: Done

MIGRATION.md
atmamont: Renamed to improve clarity and added description of return types.

robertdalmeida (approved):

PR #2571, opened 2026-06-03
Align theming API

atmamont:
# Summary 

This is a follow-up to the [docs PR](https://github.com/Adyen/adyen-ios/pull/2568).
Changes:
- rename `AdyenColors` to `CheckoutColors`
- make `AdyenColors` and `CheckoutTheme(colors: CheckoutColors)` public
- expose `cornerRadius` helper to match Android API

[Android PR](https://github.com/Adyen/adyen-android/pull/2804)

## Ticket
<ticket>
COSDK-1251
</ticket>

## Checklist

- [x] Tested changes locally


nauaros (approved):

erenbesel (approved):

AdyenUI/CheckoutTheme.swift
```diff
@@ -9,49 +9,49 @@ import UIKit
 
 public struct CheckoutTheme {
 
-    package private(set) var colors: AdyenColors
+    public private(set) var colors: CheckoutColors
```
robertdalmeida, line 12: question; 
Do we need `CheckoutColors` get only property to be public?
if public then maybe we need to add some docs to it?

AdyenUI/CheckoutColors.swift
```diff
@@ -36,27 +36,28 @@ internal enum DefaultColorsDark {
...
-package struct AdyenColors: Equatable {
-
-    package var background: UIColor
-    package var container: UIColor
-    package var containerOutline: UIColor
-    package var primary: UIColor
-    package var textOnPrimary: UIColor
-    package var highlight: UIColor
-    package var destructive: UIColor
+public struct CheckoutColors: Equatable {
```
robertdalmeida, line 39: nonblocking; maybe we could also rename the filename to CheckoutColors too.

AdyenUI/CheckoutTheme.swift
```diff
@@ -116,7 +116,7 @@ extension CheckoutTheme {
     ///   - disabledTextColor: The disabled text color. Defaults to current value.
     ///   - cornerRadius: The corner radius. Defaults to current value.
     /// - Returns: A new `CheckoutTheme` instance with the updated primary button style.
-    public func primaryButton(
+    package func primaryButton(
```
robertdalmeida: thought; nonblocking;
I see that this isn't needed anymore but will be introduced only when requested by merchant. 
So do you think we can then delete `bodyLabel`, `primaryButton` & `destructiveButton`. otherwise this would be zombie code until then? 
I vote to delete.

robertdalmeida (approved):

AdyenUI/CheckoutTheme.swift
```diff
@@ -9,49 +9,49 @@ import UIKit
...
-extension CheckoutTheme {
-    
     /// Returns a new theme with the specified colors.
     /// - Parameter colors: The color scheme to apply.
     /// - Returns: A new `CheckoutTheme` instance.
-    package func colors(_ colors: AdyenColors) -> CheckoutTheme {
+    public func colors(_ colors: CheckoutColors) -> CheckoutTheme {
         CheckoutTheme(
             colors: colors,
             elements: elements,
```
OscarSpruit: Shouldn't the elements be recreated with the new colors here?

AdyenUI/CheckoutColors.swift
nauaros: Good catch

AdyenUI/CheckoutTheme.swift
atmamont: `elements: elements` part means we use existing `self.elements` and create new `CheckoutTheme` using them

AdyenUI/CheckoutTheme.swift
atmamont: I had the same thought following the principle "if you don't use it - delete it". 
@erenbesel wdyt?

AdyenUI/CheckoutColors.swift
atmamont: Yes definitely, good catch

AdyenUI/CheckoutTheme.swift
atmamont: Yes, as it's a basic principle "if allow setting it, reading should be allowed"

AdyenUI/CheckoutTheme.swift
erenbesel: sure, chop it up

nauaros: I agree with Robert and you Andrii, let's not maintain code that we don't need right now.

AdyenUI/CheckoutTheme.swift
robertdalmeida: elements does indeed depends on Colors. I too think that elements needs to be recreated here.

atmamont: gotcha, missed the dependency part

AdyenUI/CheckoutTheme.swift
robertdalmeida: Just saying; i could also do this later, but if you can.

i see cornerRadius & font configurable in Android. 
But in iOS cornerRadius can be made configurable by making this method public. 

And the current `AdyenFonts` is not connected with `CheckoutTheme`. I think we want it? 
If yes, AdyenFonts needs to be added as a depency to AdyenElements and consumed. 

Thank you for using your tokens.

AdyenUI/CheckoutColors.swift
atmamont: Renamed

AdyenUI/CheckoutTheme.swift
atmamont: Addressed

AdyenUI/CheckoutTheme.swift
atmamont: Let's keep cornerRadius helper public for now, and tackle fonts topic later.

AdyenUI/CheckoutTheme.swift
atmamont: Done, moved the helpers to test target to keep the coverage, and removed usage examples from the Demo app code

robertdalmeida (approved):

nauaros (approved):

PR #2576, opened 2026-06-04
[v6] (1/5) Non-UI components: Public API

nauaros:
## Summary

This PR introduces the public API needed to support non-UI component integrations (e.g., instant payment methods like PayPal, ApplePay, stored PMs, etc).
 
## New Public API
 
The `CheckoutPaymentComponent` exposes two new public APIs:
 
### `var requiresUserInteraction: Bool`

Indicates whether the payment component requires user interaction before submitting.
- `true`: Component has UI that the shopper must interact with (e.g., card details form)
- `false`: Component can submit immediately without presenting UI (e.g., instant payment methods)
 
### `func submit()`

Programmatically triggers the payment submission. For components with UI, this validates the form first. For instant payment methods, this immediately initiates the payment.
 
### Usage Example
```swift
let component = try checkout.createPaymentComponent(for: paymentMethod)
 
if component.requiresUserInteraction, let componentViewController = component.viewController {
    present(componentViewController, animated: true)
} else {
    component.submit()
}
```

### Protocol Consolidation
- Deleted `SubmittableComponent.swift` protocol
- Removed `InitiablePaymentComponent` protocol  
- Added `func submit()` requirement to `PaymentComponent` protocol
- Simplified `PaymentComponentType` enum: `.initiable(InitiablePaymentComponent)` → `.initiable(PaymentComponent)`

### InstantPaymentComponent
- Replaced `initiatePayment(delegate:)` with `submit()` for consistent API
- Removed delegate parameter from initiation method (delegate is set during component setup)

### Component Updates
All payment components now implement `submit()` directly instead of conforming to `SubmittableComponent`:
- `CardComponent`, `GiftCardComponent`, `StoredCardComponent`
- `BLIKComponent`, `SEPADirectDebitComponent`, `ACHDirectDebitComponent`
- `OnlineBankingComponent`, `UPIComponent`, `PayToComponent`
- `BoletoComponent`, `CashAppPayComponent`, `TwintComponent`
- `AbstractPersonalInformationComponent`, `IssuerListComponent`
- `ApplePayComponent`, `BACSDirectDebitComponent`

### Drop-in Integration
- Updated `PaymentMethodListViewModel` and `PreselectedPaymentMethodViewModel` to call `submit()` instead of `initiatePayment(delegate:)`

### CheckoutPaymentComponent
- Added `requiresUserInteraction` property
- Added `submit()` method that forwards to underlying `paymentComponent.submit()`

## Next PRs:
- [x] Fix broken tests (#2578)
- [ ] Handle edge case scenarios:
  - [ ] ApplePayComponent
  - [ ] IssuerListComponent
  - [ ] BACSDirectDebitComponent
  - [ ] StoredCardComponent

## Ticket
<ticket>
COSDK-1103
</ticket>

## Checklist [Required]

- [x] Tested changes locally
- [ ] Added/updated unit tests **[Future PR]**


Adyen.xcodeproj/project.pbxproj
```diff
@@ -2318,6 +2318,8 @@
 				"",
 				"$swiftlintPath --config \"$SRCROOT/.swiftlint.yml\"",
 				"",
+				"",
```
erenbesel, line 2321: New quotation marks, very important! 😄

AdyenCheckout/Component/CheckoutPaymentComponent.swift
```diff
@@ -64,4 +56,44 @@ public final class CheckoutPaymentComponent {
...
+
+    /// Indicates whether the payment component requires user interaction before submitting.
+    ///
+    /// When `true`, the component has a UI that the shopper must interact with (e.g., filling in card details).
+    /// When `false`, the component can submit immediately without presenting any UI (e.g., instant payment methods).
+    ///
+    /// Use this property to determine whether to present the component's `viewController` or call `submit()` directly.
+    ///
+    /// ```swift
+    /// if component.requiresUserInteraction {
```
erenbesel: Code sample here is missing `if let component.viewController` part.

AdyenCheckout/Component/CheckoutPaymentComponent.swift
```diff
@@ -64,4 +56,44 @@ public final class CheckoutPaymentComponent {
...
+    /// }
+    /// ```
+    public var requiresUserInteraction: Bool {
+        viewController != nil
+    }
+
+    /// Submits the payment request to initiate the payment process.
+    ///
+    /// Call this method to programmatically trigger the payment submission. For components with UI (`requiresUserInteraction == true`),
+    /// this validates the form and submits if valid. For instant payment methods (`requiresUserInteraction == false`),
```
erenbesel: I think we are moving away from the `instant` keyword since it means different things for Web/mobile. Maybe we can for now use direct payment methods?

AdyenComponents/Apple Pay/ApplePayComponent.swift
```diff
@@ -139,6 +139,11 @@ package class ApplePayComponent: NSObject, PresentableComponent, PaymentComponen
         resumeContinuation(success: success)
         completion?()
     }
+
+    package func submit() {
+        // TODO: - Naufal: Check where to perform submit in ApplePay.
```
erenbesel, line 144: Yeah, ApplePay can't be triggered via submit() indeed. Can only be triggered via its own with passcode/biometrics. Which made question whether we should keep the submittable protocol and only not conform in ApplePay. 
or we can keep this function empty for ApplePay. 

Ideally I vote for the calling of submit() directly for components that has a UI and using SDK's pay button to have no effect (it would include ApplePay component automatically)

AdyenComponents/SEPA Direct Debit/SEPADirectDebitComponent.swift
```diff
@@ -48,6 +48,10 @@ package final class SEPADirectDebitComponent: PaymentComponent, PresentableCompo
         self.context = context
         self.configuration = configuration
     }
+
+    package func submit() {
```
erenbesel, line 52: I noticed now every conforming component has this submit function implemented exactly the same.
So instead of adding this function to all components, we can write an extension on the protocol and call `didSelectSubmitButton` (which we add to the protocol definition) and save a lot repeated code. Because didSelectSubmitButton is already implemented everywhere, we don't have to repeat the same for submit() imo

erenbesel (approved):
Nice nice! Some questions, remarks

AdyenCheckout/Component/CheckoutPaymentComponent.swift
```diff
@@ -64,4 +56,44 @@ public final class CheckoutPaymentComponent {
...
+    /// ```swift
+    /// if component.requiresUserInteraction {
+    ///     present(component.viewController!, animated: true)
+    /// } else {
+    ///     component.submit()
+    /// }
+    /// ```
+    public var requiresUserInteraction: Bool {
+        viewController != nil
+    }
```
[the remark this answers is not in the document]
atmamont: Valid point

AdyenCard/Components/Card/CardComponent.swift
```diff
@@ -134,6 +134,10 @@ package class CardComponent: PresentableComponent,
         self.supportedCardTypes = configuration.supportedCardBrands ?? paymentMethod.brands
     }
 
+    package func submit() {
+        didSelectSubmitButton()
```
atmamont, line 138: Is this the right time to migrate to something more clear like `onSumbitButtonTap()`?

AdyenCard/Components/Card/CardComponentExtensions.swift
```diff
@@ -12,19 +12,23 @@ import Adyen
 #if canImport(AdyenAuthentication)
     import AdyenAuthentication
 #endif
+#if canImport(AdyenUI)
+    import AdyenUI
+    @_spi(AdyenInternal) import class AdyenUI.FormViewController
```
atmamont, line 16: Do you have a vision of how this would work after `FormViewController` becomes package?

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
```diff
@@ -5,6 +5,7 @@
 //
 
 import Adyen
+@_spi(AdyenInternal) import protocol Adyen.PaymentComponent
```
atmamont: Same note here and for the rest of @spi imports

AdyenCheckout/Component/CheckoutPaymentComponent.swift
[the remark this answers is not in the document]
atmamont: I am not 100% sure, as it's a protocol property, but if we take a look in the Card component code, then we have this 100% true:

```swift
    package var viewController: UIViewController {
        if let storedCardComponent {
            return storedCardComponent.viewController
        }
        return securedViewController
    }

    private lazy var securedViewController = SecuredViewController(child: cardViewController, style: configuration.style)

```

atmamont (commented):
Let's not merge this one until it's green (we merge tests into here first)?

AdyenCard/Components/Stored Card/StoredCardComponent.swift
```diff
@@ -77,6 +77,10 @@ package final class StoredCardComponent: StoredPaymentComponent, Localizable {
 
         return StoredCardInputViewController(viewModel: viewModel)
     }()
+
+    package func submit() {
+        // TODO: - Naufal: Implement submit logic in  StoredCardComponent
+    }
```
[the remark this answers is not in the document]
nauaros: Its part of the development plan

AdyenComponents/Apple Pay/ApplePayComponent.swift
```diff
@@ -139,6 +139,11 @@ package class ApplePayComponent: NSObject, PresentableComponent, PaymentComponen
...
+        // Web triggers the ApplePay system view.
+    }
```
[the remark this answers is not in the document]
nauaros: It's part of the development plan

AdyenComponents/Issuer List/IssuerListComponent.swift
```diff
@@ -57,7 +57,12 @@ package final class IssuerListComponent: PaymentComponent, PresentableComponent,
         self.configuration = configuration
         self.title = paymentMethod.displayInformation(using: configuration.localizationParameters).title
     }
-    
+
+    package func submit() {
+        // TODO: - Naufal: What should we do in the IssuerListComponent?
+        // IssuerListComponent needs to be changed so it validates shopper issuer selection.
+    }
```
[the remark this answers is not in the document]
nauaros: It's part of the development plan

AdyenCheckout/Component/CheckoutPaymentComponent.swift
nauaros: True, updated it

AdyenCard/Components/GiftCardComponent/GiftCardComponent.swift
nauaros: Good catch here, I spi imported before Robert changes

AdyenCard/Components/Card/CardComponentExtensions.swift
nauaros: @robertdalmeida The reason for not making `FormViewController` package is due to the fact that it is an open class and it's being subclassed in the `CardComponent` right?

AdyenComponents/BACS Direct Debit/BACSDirectDebitComponent.swift
```diff
@@ -102,6 +102,12 @@ package final class BACSDirectDebitComponent: PaymentComponent, PresentableCompo
         inputFormViewController.presenter = inputPresenter
         
     }
+
+    package func submit() {
+        // TODO: - Naufal: How to handle submit in BACS?
+        // - submit() flow does not fit into a two step component.
+        // - Zero traffic on BACS. We either drop support or BACS has to be redesigned into a single step component.
+    }
```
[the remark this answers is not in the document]
nauaros: It's part of the development plan

AdyenComponents/SEPA Direct Debit/SEPADirectDebitComponent.swift
nauaros: Yes, good point. Once the `submit` implementations are done in every component, I take a look to see how we can make it more DRY.

I had the exact same thought as yours

AdyenCard/Components/Card/CardComponent.swift
nauaros: It's a private/internal method. We could name it however we want. No strong opinions on it

AdyenCheckout/Component/CheckoutPaymentComponent.swift
nauaros: I like direct PMs as Vlad suggested. @OscarSpruit also likes it so let's go with that. I bring it up in our v6 session so everyone in the teams knows.

AdyenCard/Components/Card/CardComponentExtensions.swift
robertdalmeida: > @robertdalmeida The reason for not making FormViewController package is due to the fact that it is an open class and it's being subclassed in the CardComponent right?

Yes ineed. open classes cannot be made package :(

erenbesel (approved):

erenbesel (approved):

PR #2580, opened 2026-06-05
Add MOCKING.md and improve TESTING.md

nauaros:
## Summary

- **New [MOCKING.md](cci:7://file:///Users/naufal/Documents/Adyen/adyen-checkout-sdk-meta/sdks/adyen-ios/MOCKING.md:0:0-0:0)**: Documents Sourcery mock generation via `GenerateSourcery` scheme, `// sourcery: AutoMockable` annotation, and generated mock features
- **Updated [TESTING.md](cci:7://file:///Users/naufal/Documents/Adyen/adyen-checkout-sdk-meta/sdks/adyen-ios/TESTING.md:0:0-0:0)**: Simplified to focus on `UnitTests` as the default test scheme; references [MOCKING.md](cci:7://file:///Users/naufal/Documents/Adyen/adyen-checkout-sdk-meta/sdks/adyen-ios/MOCKING.md:0:0-0:0) for mock guidance

## Why

- Clear guidance on default test target (`UnitTests`) for developers and LLM agents
- Reduces overhead by not recommending `IntegrationUIKitTests` by default

### DIFF
--- a/MOCKING.md
+++ b/MOCKING.md
@@ -1,12 +1,8 @@
 # Mocking Guide
 
-This file provides guidance for generating and using mocks in the Adyen iOS SDK.
+The project uses [Sourcery](https://github.com/krzysztofzablocki/Sourcery) to auto-generate mock implementations from protocols.
 
-## Generating Mocks with Sourcery
-
-The project uses [Sourcery](https://github.com/krzysztofzablocki/Sourcery) to auto-generate mock implementations from protocols. Generated mocks are output to `Tests/GeneratedMocks/`.
-
-### Running the Generator
+## Running the Generator
 
 Build the `GenerateSourcery` scheme to regenerate mocks:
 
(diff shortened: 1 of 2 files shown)


TESTING.md
```diff
@@ -322,3 +323,7 @@ private func getContainerView(from sut: FormTextItemView<FormTextInputItem>) ->
 ### Benefits
 
 This helper method pattern makes tests **~70% more concise** while maintaining clarity and proper error reporting. Tests read like specifications rather than imperative code.
+
+## Mocking
+
+For guidance on generating and using mocks with Sourcery, see [MOCKING.md](MOCKING.md).
```
atmamont: This has to be extremely explicit, let's exclude any ambiguity?

Something like
`When testing tasks require creating/updating a mock, refer to [MOCKING.md](MOCKING.md) for details`

nauaros: Good suggestion, I updated it to something similar to your example.

atmamont (approved):

robertdalmeida (approved):

MOCKING.md
```diff
@@ -0,0 +1,60 @@
...
+### Marking Protocols as AutoMockable
+
+Add the `// sourcery: AutoMockable` annotation above any protocol to generate a mock:
+
+```swift
+// sourcery: AutoMockable
+protocol MyProtocol {
+    func doSomething()
+}
+```
```
robertdalmeida: nonblocking; thought; imho, i fail to see how as a dev anything below this line would be of use and may fall into somthing that is obvious & llm created noise. Maybe a link to the config file in the last line may help i guess.

nauaros: The idea is to give context so llm can create mocks with the same format as sourcery when a protocol is not explicitly marked as such. But maybe you are right, and it can pick it up from existing mocks. wdyt @atmamont ?

atmamont: I would agree here with @robertdalmeida 
Instead, let's point out to an example mock, keep instructions to add annotation concise (1 line), and explain how to execute mock generation (we do use a specific target run for this)

nauaros: Simplified with your suggestions guys. Thanks

atmamont (approved):

PR #2584, opened 2026-06-09
Update CheckourError mapping, add missing cases

erenbesel:
# Summary [Required]
- Update the error codes to reflect changes on [PR](https://github.com/Adyen/adyen-checkout-sdk-meta/pull/57)
- Update CheckoutError mapping to be more precise.

## Ticket [Optional]
<ticket>
COSDK-589
</ticket>

## Checklist [Required]

- [ ] Tested changes locally

### DIFF
--- a/Adyen/Core/Errors/CheckoutError.swift
+++ b/Adyen/Core/Errors/CheckoutError.swift
@@ -42,6 +42,8 @@ public extension CheckoutError {
             self.rawValue = rawValue
         }
 
+        /// A generic or unclassified error occurred.
+        public static let generic = Code(rawValue: "Generic")
         /// The provided client key is malformed or invalid.
         public static let invalidClientKey = Code(rawValue: "InvalidClientKey")
         /// The specified locale is not supported or malformed.
@@ -52,9 +54,11 @@ public extension CheckoutError {
         public static let invalidAmountValue = Code(rawValue: "InvalidAmountValue")
         /// Failed to establish a checkout session.
         public static let sessionSetupFailure = Code(rawValue: "SessionSetupFailure")
-        /// The user cancelled the payment.
-        public static let cancelled = Code(rawValue: "Cancelled")
-        /// An unknown or unexpected error occurred.
-        public static let unknown = Code(rawValue: "Unknown")
+        /// A network or HTTP-level error occurred.
+        public static let httpError = Code(rawValue: "HttpError")
+        /// A payment method is unavailable, unsupported, or cannot be used by the current device or user.
+        public static let paymentMethodFailure = Code(rawValue: "PaymentMethodFailure")
+        /// The component configuration contains invalid or missing values.
+        public static let invalidConfiguration = Code(rawValue: "InvalidConfiguration")
     }
 }
--- a/Adyen/Core/Errors/CheckoutError.swift
+++ b/Adyen/Core/Errors/CheckoutError.swift
@@ -54,6 +54,8 @@ public extension CheckoutError {
         public static let invalidAmountValue = Code(rawValue: "InvalidAmountValue")
         /// Failed to establish a checkout session.
         public static let sessionSetupFailure = Code(rawValue: "SessionSetupFailure")
+        /// The user cancelled the payment.
+        public static let cancelled = Code(rawValue: "Cancelled")
         /// A network or HTTP-level error occurred.
         public static let httpError = Code(rawValue: "HttpError")
         /// A payment method is unavailable, unsupported, or cannot be used by the current device or user.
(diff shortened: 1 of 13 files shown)


Adyen/Core/Errors/CheckoutError.swift
```diff
@@ -52,9 +54,11 @@ public extension CheckoutError {
         public static let invalidAmountValue = Code(rawValue: "InvalidAmountValue")
         /// Failed to establish a checkout session.
         public static let sessionSetupFailure = Code(rawValue: "SessionSetupFailure")
-        /// The user cancelled the payment.
-        public static let cancelled = Code(rawValue: "Cancelled")
```
robertdalmeida, line 56: nonblocking; With the removal of cancelled. 
Does this mean that we won't allow the user to know if a payment has been cancelled and we'll default to .generic?

robertdalmeida (approved):

AdyenCheckout/Core/CheckoutError+Mapping.swift
```diff
@@ -16,18 +19,34 @@ package extension CheckoutError {
...
+            case ApplePayComponent.Error.emptySummaryItems,
+                 ApplePayComponent.Error.emptyMerchantIdentifier,
+                 ApplePayComponent.Error.negativeGrandTotal,
+                 ApplePayComponent.Error.invalidSummaryItem,
+                 ApplePayComponent.Error.invalidCountryCode,
+                 ApplePayComponent.Error.invalidCurrencyCode,
+                 ApplePayComponent.Error.missingConfiguration,
+                 ApplePayComponent.Error.invalidPaymentRequest:
+                self = .init(code: .invalidConfiguration, message: error.localizedDescription, underlyingError: error)
+        #endif
```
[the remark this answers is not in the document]
erenbesel: @gemini-code-assist this is plain wrong. Based on what data do you make this comment?

nauaros (approved):

AdyenCheckout/CheckoutConfiguration/CheckoutConfiguration.swift
```diff
@@ -73,22 +74,26 @@ public struct CheckoutConfiguration {
...
         var configDictionary: [CheckoutComponentType: CheckoutComponentConfiguration] = [:]
-        let content = try content()
-        let configArray = (content as? CompositeCheckoutConfiguration)?.configurations ?? []
+        let configurations: CheckoutConfigurable
+        do {
+            configurations = try content()
+        } catch {
+            throw CheckoutError(error: error, fallback: .invalidConfiguration)
+        }
+        let configArray = (configurations as? CompositeCheckoutConfiguration)?.configurations ?? []
```
atmamont, line 83: What do you think of moving type casting into a `do - catch` block? It also could belong to the `.invalidConfiguration` usecase

AdyenComponents/Apple Pay/ApplePayComponentError.swift
```diff
@@ -61,6 +68,10 @@ extension ApplePayComponent {
                 return "The currency code is invalid."
             case .invalidToken:
                 return "The Apple Pay token is invalid. Make sure you are using physical device, not a Simulator."
+            case .missingConfiguration:
+                return "No Apple Pay configuration was provided. Supply an ApplePayConfiguration via the CheckoutConfiguration DSL."
```
atmamont: Sorry for getting back to this topic, but what's the hidden cost of using a NSLocalizedString here? We offer merchants ability to use this right away in their UI if they want to

atmamont (approved):

AdyenCheckout/CheckoutConfiguration/CheckoutConfiguration.swift
erenbesel: If the type cast fails we fallback to empty array as they can leave the configurations block empty. And it shouldn't throw in such cases and use the default configuraitons.

AdyenComponents/Apple Pay/ApplePayComponentError.swift
erenbesel: You mean encase all of them here in NSLocalizedString? Will that be enough?

AdyenCheckout/CheckoutConfiguration/CheckoutConfiguration.swift
atmamont: Thanks for explaining!

AdyenComponents/Apple Pay/ApplePayComponentError.swift
atmamont: Yes, I think so. Even if we skip adding keys to the TPS (Smartling) and keep English as the only language, there still would be a possibility to let merchants to provide their own overrides.

atmamont (approved):

Adyen/Core/Errors/CheckoutError.swift
erenbesel: Yes since other platforms didn't have this case at all. But removing it was also not a supper happy choice. We may add it back after discussion

nauaros (approved):

PR #2589, opened 2026-06-11
[v6] (6/5) Non-UI components: Migrate standalone Instant component integration to v6

nauaros:
## Summary

This PR migrates the Demo app's standalone `InstantComponent` integration examples to v6.

The following integrations have been updated:
- Advanced flow
- Sessions flow

The changes align the examples with the latest v6 APIs and serve as updated references for merchants integrating the SDK.

# Demo

https://github.com/user-attachments/assets/b3af5784-6a52-48da-9248-9222d210ec76

## Ticket
<ticket>
COSDK-1103
</ticket>

## Checklist

- [x] Tested changes locally

### DIFF
--- a/Demo/Common/IntegrationExamples/AdvancedFlow/Components/InstantPaymentComponentAdvancedFlowExample.swift
+++ b/Demo/Common/IntegrationExamples/AdvancedFlow/Components/InstantPaymentComponentAdvancedFlowExample.swift
@@ -6,132 +6,135 @@
 
 import Adyen
 import AdyenActions
+import AdyenCheckout
 import AdyenComponents
 import Foundation
 
+@MainActor
 internal final class InstantPaymentComponentAdvancedFlow: InitialDataAdvancedFlowProtocol {
 
-    // MARK: - Properties
-
-    internal var instantPaymentComponent: InstantPaymentComponent?
-
     internal weak var presenter: PresenterExampleProtocol?
-    
+
+    private var checkout: AdvancedCheckout?
+    private var adyenComponent: CheckoutPaymentComponent?
+
     internal lazy var apiClient = ApiClientHelper.generateApiClient()
-    
+    private lazy var asyncApiClient = ApiClientHelper.generateAsyncApiClient()
+
     /// comes from demo app protocol, unused on new structure
     internal var context: AdyenContext?
-
-    // MARK: - Action Handling
-
-    private lazy var actionComponent: CheckoutActionComponent = {
-        guard let context else {
-            fatalError("Context not initialized")
-        }
-        let handler = CheckoutActionComponent(context: context)
-        handler.delegate = self
-        handler.presentationDelegate = self
-        return handler
-    }()
-
-    // MARK: - Initializers
 
     internal init() {}
 
     internal func start() {
-        presenter?.showLoadingIndicator()
+        startLoading()
+
         Task {
             do {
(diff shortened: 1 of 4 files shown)


Demo/Common/IntegrationExamples/AdvancedFlow/Components/InstantPaymentComponentAdvancedFlowExample.swift
```diff
@@ -6,198 +6,147 @@
...
         }
-        
-        return InstantPaymentComponent(paymentMethod: paymentMethod, context: context, order: nil)
+        .onError { [weak self] error in
+            self?.dismissAndShowAlert(false, error.localizedDescription)
+        }
+
+        self.checkout = checkout
+
+        return try checkout.createPaymentComponent(for: PaymentMethodType.ideal)
```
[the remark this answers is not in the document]
nauaros: This is a deliberate change in order to test the error flow when the PM is not available for creation.

robertdalmeida (approved):

Demo/Common/IntegrationExamples/AdvancedFlow/Components/InstantPaymentComponentAdvancedFlowExample.swift
```diff
@@ -6,198 +6,147 @@
...
-                        self.presentAlert(with: error)
+                let paymentMethods = try await requestPaymentMethods(order: nil)
+                let component = try await instantPaymentComponent(from: paymentMethods)
+                self.adyenComponent = component
+                hideLoading()
+
+                if component.requiresUserInteraction {
+                    // Present the component UI if user interaction is needed
+                    if let viewController = component.viewController {
+                        presenter?.present(viewController: viewController, completion: nil)
```
erenbesel: Should this part be in the non UI demo example class? Maybe just focus on the submit part on these files?

nauaros: I think its a good opportunity to make usage of the new API in `CheckoutPaymentComponent`.

We could even go a step further and have a single integration for all standalone components instead of repeating the same logic, as we are component agnostic

erenbesel: That's a good idea! We can have a separate demo app file for that. But here that line will never be executed right?
We can still leverage the new api with a guard block, wdyt?

erenbesel (approved):
One small comment about unused code in demo but nice!

Demo/Common/IntegrationExamples/AdvancedFlow/Components/InstantPaymentComponentAdvancedFlowExample.swift
nauaros: Yeah, that is good. Updated with a guard block

nauaros: I will create a task to potentially unify our example integrations (advanced and session) so all shared logic is in one place.

erenbesel: Great! (Unifying them is really nice but advanced/session should be still separate imo)

PR #2590, opened 2026-06-11
Align BillingAddressMode configuration for CardComponent

robertdalmeida:


AdyenCard/Components/Card/BillingAddressConfiguration.swift
```diff
@@ -16,10 +16,24 @@ public enum BillingAddressMode {
...
     /// Displays the full billing address form with all address fields.
-    case full
+    ///
+    /// - Parameters:
+    ///   - supportedCountryCodes: List of ISO country codes supported for the billing address.
+    ///     When empty, all countries are available.
+    ///   - hideForCardTypes: Card types for which the address form should be hidden
+    ///     when detected via BIN lookup.
+    case full(
+        supportedCountryCodes: [String] = [],
```
robertdalmeida, line 34: @erenbesel @nauaros 
I find this part strange, when we provide `nil` then all countries are available. 
I would have preferred to have an `enum` something like below. WDYT?

CountryCode {
case all
case countries([String]) 
}

AdyenCard/Components/Card/BillingAddressConfiguration.swift
```diff
@@ -16,10 +16,24 @@ public enum BillingAddressMode {
...
 
     /// Displays the full billing address form with all address fields.
-    case full
+    ///
+    /// - Parameters:
+    ///   - supportedCountryCodes: List of ISO country codes supported for the billing address.
+    ///     When empty, all countries are available.
+    ///   - hideForCardTypes: Card types for which the address form should be hidden
+    ///     when detected via BIN lookup.
+    case full(
```
robertdalmeida, line 33: @nauaros @erenbesel 
I'm not sure what full means here. I personally think that either `.form` or `.manual` better describes the option where the user has to manually fill up the address without the search functioality. 
Is this something that warrants a change? WDYT?

robertdalmeida:
Created an alternate MR here - https://github.com/Adyen/adyen-ios/pull/2608

PR #2591, opened 2026-06-11
Simplify BeforeSubmitData API

erenbesel:
# Summary

Simplifies the `BeforeSubmitData` API introduced with `onBeforeSubmit`. The previous design exposed `public private(set) var` properties alongside dedicated `replacing()` modifier functions. This replaces that with plain `public var` properties so merchants can mutate fields directly via standard Swift value semantics.

The `package init` is unchanged — merchants cannot construct a fresh `BeforeSubmitData` instance, so the `??\ fallback in `PaymentComponentData` remains correct: setting a field to `nil` has no effect and the original value collected by the component is preserved.

## Ticket

<ticket>
COSDK-907
</ticket>

## Checklist
- [x] Tested changes locally

### DIFF
--- a/Adyen/Core/Configuration/BeforeSubmitResult.swift
+++ b/Adyen/Core/Configuration/BeforeSubmitResult.swift
@@ -8,32 +8,32 @@ import Foundation
 
 /// Result returned by `onBeforeSubmit` callback in the session flow.
 public enum BeforeSubmitResult: Sendable {
-    
+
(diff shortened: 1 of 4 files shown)


Robert-SD (approved):

atmamont (approved):

nauaros (approved):

nauaros:
Looks cleaner like this, its good change

PR #2596, opened 2026-06-12
onComplete callback signature update

erenbesel:
# Summary [Required]
Replace the single CheckoutResult type with two flow-specific types: SessionCheckoutResult (resultCode, sessionId, sessionResult, all non-optional) and AdvancedCheckoutResult (resultCode only). 

The onComplete callback moves from BaseCheckout to each subclass with typed closures, the old CheckoutSuccessHandler/CheckoutErrorHandler public typealiases are removed (inlined publicly, package typealiases added internally), and the dead Session.currentResult property is cleaned up.

## Ticket [Optional]
<ticket>
COSDK-1290
</ticket>

## Checklist [Required]

- [x] Tested changes locally


AdyenCheckout/Callbacks/CheckoutResultCallbackStore.swift
```diff
@@ -8,33 +8,47 @@ import Adyen
...
-    package var onComplete: CheckoutSuccessHandler?
+    package var onComplete: SessionCheckoutCompletionHandler?
 
-    package var onFailure: CheckoutErrorHandler?
+    package var onFailure: CheckoutFailureHandler?
+
+    package func handleCompletion(resultCode: CheckoutResultCode, sessionId: String?, sessionResult: String?) {
+        guard let sessionId, let sessionResult else { return }
+        onComplete?(SessionCheckoutResult(resultCode: resultCode, sessionId: sessionId, sessionResult: sessionResult))
+    }
```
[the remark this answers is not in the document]
erenbesel: @gemini-code-assist this will not happen. instead of an NSerror, how about assertion?

AdyenCheckout/Callbacks/CheckoutResultCallbackStore.swift
```diff
@@ -8,33 +8,50 @@ import Adyen
...
 package final class SessionCheckoutCallbackStore: CheckoutResultCallbackStore {
     package var onBeforeSubmit: BeforeSubmitHandler?
 
-    package var onComplete: CheckoutSuccessHandler?
+    package var onComplete: SessionCheckoutCompletionHandler?
 
-    package var onFailure: CheckoutErrorHandler?
+    package var onFailure: CheckoutFailureHandler?
+
+    package func handleCompletion(resultCode: CheckoutResultCode, sessionId: String?, sessionResult: String?) {
```
atmamont, line 24: This is internal, could we find a way to change the signature to avoid nullability for `sessionId` and `sessionResult`?

AdyenCheckout/Callbacks/CheckoutResultCallbackStore.swift
```diff
@@ -8,33 +8,50 @@ import Adyen
...
     package var onAdditionalDetails: AdditionalDetailsHandler?
 
-    package var onComplete: CheckoutSuccessHandler?
+    package var onComplete: AdvancedCheckoutCompletionHandler?
+
+    package var onFailure: CheckoutFailureHandler?
 
-    package var onFailure: CheckoutErrorHandler?
+    package func handleCompletion(resultCode: CheckoutResultCode, sessionId: String?, sessionResult: String?) {
+        onComplete?(AdvancedCheckoutResult(resultCode: resultCode))
```
atmamont, line 44: This method ignores `sessionId` and `sessionResult`, I wonder if this could be rewritten in a less obscure way or maybe a code comment would be beneficial here.

AdyenCheckout/Callbacks/CheckoutResultCallbackStore.swift
nauaros: What is the purpose of the `sessionId` in the result? Is it to know which session instance the result is coming from in case there are different ones?

erenbesel: Yes, in the endpoint the id is also needed. Will add the description back to explain that part.

AdyenCheckout/Callbacks/CheckoutResultCallbackStore.swift
erenbesel: No other way without exposing session into the other layer I'm afraid.

atmamont (approved):

nauaros (approved):

nauaros (approved):

atmamont (approved):

PR #2598, opened 2026-06-16
Binlook alignment pt1: Logic

erenbesel:
# Summary [Required]
Aligns binlookup public API and trigger logic to the spec.

## New types

- BinLookupData and BinLookupBrand - new public, Decodable, Equatable structs in AdyenCard/Utilities/BinLookup/BinLookupData.swift, aligned with Android's merchant-facing API
Callback changes

- CardConfiguration.onBinLookup signature changed from ([CardBrand]) -> Void to (BinLookupData) -> Void
updateBrand in CardComponent now guards on !binInfo.isCreatedLocally - the callback only fires for actual /binLookup API responses (live or cached), never for client-side regex/fallback detection

## CardBrand internals

- Added paymentMethodVariant: String? property (decoded from API, forwarded into BinLookupBrand)

## Ticket [Optional]
<ticket>
COSDK-1243
</ticket>

## Checklist [Required]

- [ ] Tested changes locally

### DIFF
--- a/AdyenCard/Components/Card/CardComponent.swift
+++ b/AdyenCard/Components/Card/CardComponent.swift
@@ -263,7 +263,11 @@ extension CardComponent: CardViewControllerDelegate {
         binInfoProvider.provide(for: pan, supportedTypes: supportedCardTypes) { [weak self] binInfo in
             guard let self else { return }
             self.cardViewController.update(binInfo: binInfo)
-            self.configuration.onBinLookup?(binInfo.brands ?? [])
+            guard !binInfo.isCreatedLocally else { return }
+            let brands = (binInfo.brands ?? []).map {
+                BinLookupBrand(brand: $0.type.rawValue, supported: $0.isSupported, paymentMethodVariant: $0.paymentMethodVariant)
+            }
+            self.configuration.onBinLookup?(BinLookupData(issuingCountryCode: binInfo.issuingCountryCode, brands: brands))
         }
     }
 }
--- a/AdyenCard/Utilities/BinLookup/BinLookupData.swift
+++ b/AdyenCard/Utilities/BinLookup/BinLookupData.swift
@@ -14,16 +14,6 @@ public struct BinLookupData: Decodable, Equatable {
 
     /// The card brands detected from the BIN. Empty when no brands are detected.
     public let brands: [BinLookupBrand]
-
-    public init(issuingCountryCode: String?, brands: [BinLookupBrand]) {
-        self.issuingCountryCode = issuingCountryCode
-        self.brands = brands
-    }
-
-    private enum CodingKeys: String, CodingKey {
-        case issuingCountryCode
-        case brands
-    }
 }
 
 /// Describes a single card brand entry in a BIN lookup result.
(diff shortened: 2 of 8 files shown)


atmamont (dismissed):

AdyenCard/Components/Card/CardComponent.swift
```diff
@@ -263,7 +263,11 @@ extension CardComponent: CardViewControllerDelegate {
         binInfoProvider.provide(for: pan, supportedTypes: supportedCardTypes) { [weak self] binInfo in
             guard let self else { return }
             self.cardViewController.update(binInfo: binInfo)
-            self.configuration.onBinLookup?(binInfo.brands ?? [])
+            guard !binInfo.isCreatedLocally else { return }
```
nauaros, line 266: Is there anything we can do here instead of just returning when binInfo is created locally?

AdyenCard/Utilities/BinLookup/BinLookupData.swift
```diff
@@ -0,0 +1,52 @@
...
+
+    public init(issuingCountryCode: String?, brands: [BinLookupBrand]) {
+        self.issuingCountryCode = issuingCountryCode
+        self.brands = brands
+    }
+
+    private enum CodingKeys: String, CodingKey {
+        case issuingCountryCode
+        case brands
+    }
```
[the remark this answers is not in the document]
nauaros: True, no need to define them

AdyenCard/Utilities/BinLookup/BinLookupData.swift
```diff
@@ -0,0 +1,52 @@
...
+public struct BinLookupBrand: Decodable, Equatable {
+
+    /// The brand identifier, e.g. `"visa"`, `"mc"`.
+    public let brand: String
+
+    /// Indicates whether the brand is supported by the merchant.
+    public let supported: Bool
+
+    /// The payment method variant, if provided by the backend.
+    public let paymentMethodVariant: String?
```
nauaros, line 29: Do you think we could use `PaymentMethodType` as it is Codable?

AdyenCard/Components/Card/CardComponent.swift
erenbesel: No that's the requirement. Only triggering it when not local

AdyenCard/Utilities/BinLookup/BinLookupData.swift
erenbesel: Alignment was to use string and besides, despite its name the values are not PaymentMethodType but CardType. Then again, non-enum is chosen for safety I believe.

Tests/IntegrationTests/Card Tests/CardComponentTests.swift
```diff
@@ -384,7 +384,7 @@ class CardComponentTests: XCTestCase {
     func test_onBinChangeAndOnBinLookup_whenCardNumberEntered_shouldBeCalledWithCorrectValues() {
         let cardTypeProviderMock = BinInfoProviderMock()
         cardTypeProviderMock.onFetch = {
-            $0(BinLookupResponse(brands: [CardBrand(type: .americanExpress)]))
+            $0(BinLookupResponse(brands: [CardBrand(type: .americanExpress)], isCreatedLocally: false))
```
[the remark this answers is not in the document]
erenbesel: @gemini-code-assist the tests are passing. How then?

AdyenCard/Utilities/BinLookup/BinLookupData.swift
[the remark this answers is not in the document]
erenbesel: Removed, also removed init that were not supposed to be added anyway.

nauaros (dismissed):

atmamont (approved):

PR #2599, opened 2026-06-17
Rename CardBrand to DetectedCardBrand

erenbesel:
Renames CardBrand to DetectedCardBrand as the CardBrand name will be used as the public enum (follow-up PR).

Note: Initial plan was to rename it to DetectedCardType to match Android's internal name but all the properties are named brand and brand is used extensively, so we went DetectedCardBrand

## Ticket [Optional]
<ticket>
COSDK-1243
</ticket>

## Checklist [Required]

- [x] Tested changes locally

### DIFF
--- a/AdyenCard/Utilities/CardType/DetectedCardBrand.swift
+++ b/AdyenCard/Utilities/CardType/DetectedCardBrand.swift
@@ -8,7 +8,7 @@ import Adyen
 import Foundation
 
 /// Describes the Card brand.
-package struct CardBrand: Decodable {
+package struct DetectedCardBrand: Decodable {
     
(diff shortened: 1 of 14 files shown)


AdyenCard/Utilities/CardType/DetectedCardBrand.swift
descorp, line 90: This will fix itself in part 2, right ?

erenbesel: Exactly

atmamont (approved):

atmamont (approved):

PR #2622, opened 2026-07-02
docs: Clarify session-controlled card settings in session flow

atmamont:
# Summary
Clarifies the v6 iOS card docs for session flow so merchants do not rely on component-level `showStorePaymentMethod(_:)` or `installmentConfiguration(_:)` when `/sessions` drives the effective values. The docs now point merchants to the session response values `enableStoreDetails`, `installmentOptions`, and `showInstallmentAmount`, matching the implemented behavior.

[Android PR](https://github.com/Adyen/adyen-android/pull/2869)

## Ticket 
<ticket>
COSDK-1108
</ticket>

## Checklist 
- [x] Tested changes locally


docs/v6/card-session-flow.md
```diff
@@ -51,6 +51,15 @@ let checkout = try await Checkout.setup(
 let component = try checkout.createPaymentComponent(for: .scheme)
 ```
 
+## Session-controlled card settings
+
+When you use `/sessions`, some card settings are controlled by `SessionResponse.configuration` instead of component-level `CardConfiguration` builders:
+
+- `showStorePaymentMethod(_:)` follows `enableStoreDetails`.
+- `installmentConfiguration(_:)` follows `installmentOptions`. `showInstallmentAmount` is also read from the session response.
```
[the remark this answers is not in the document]
atmamont: 🤯

erenbesel (dismissed):

docs/v6/README.md
```diff
@@ -84,6 +84,8 @@ let checkout = try await Checkout.setup(
 let component = try checkout.createPaymentComponent(for: .scheme)
 ```
 
+In session flow, the effective `amount` comes from the `/sessions` response. A client-side `amount` on `CheckoutConfiguration` is ignored and only used as a fallback when the session has no amount. `showStorePaymentMethod(_:)` and `installmentConfiguration(_:)` are likewise controlled by the `/sessions` response through `enableStoreDetails`, `installmentOptions`, and `showInstallmentAmount`. Configure these values in your `/sessions` request and do not rely on component-level values to override them.
```
erenbesel: So for `enableStoreDetails` it's a bit complicated. The parameter that governs this value in external session call is `storePaymentMethodMode` which can have 3 values: "enabled", "disabled", and "askForConsent". When the value is "askForConsent", `enableStoreDetails` becomes true in the response. When it's the other 2, it returns as false and we don't show the toggle.

atmamont: Do you think it deserves a deep explanation here in the root readme? The thing is that enum itself is private and doesn't have any explanations.

atmamont: Re-phrased it according to some feedback on Android side, check it out.

erenbesel (approved):

PR #2648, opened 2026-07-16
(3/3) Redefine Component Abstractions:  Introduce `PresentablePaymentComponent`

nauaros:
## Summary

Replace the generic `PresentableComponent` abstraction with the payment-specific `PresentablePaymentComponent`.

`PresentablePaymentComponent` now inherits from `PaymentComponent`, allowing payment UI flows to use one capability protocol without redundant conformances or casts. The new protocol uses package access because `PaymentComponent` is package-scoped.

## Changes

- Rename `PresentableComponent` to `PresentablePaymentComponent`.
- Make `PresentablePaymentComponent` inherit from `PaymentComponent`.
- Make `StoredPaymentComponent` inherit from `PresentablePaymentComponent`.
- Update `PaymentComponentType` to use the new protocol.
- Migrate payment components to the single inherited conformance.
- Update Drop-in and Checkout presentation boundaries.
- Remove redundant payment-component casts and protocol compositions.
- Give `AnyDropInComponent` its own `viewController` requirement.
- Remove stale action-component casts to the old presentation protocol.
- Update generated mocks, tests, demos, and documentation.


AdyenCheckout/Core/CheckoutCore+Orchestration.swift
```diff
@@ -79,7 +79,6 @@ internal extension CheckoutCore {
         _ data: ActionComponentData,
         from component: any ActionComponent
     ) {
-        (component as? any PresentableComponent)?.viewController.view.isUserInteractionEnabled = false
```
erenbesel, line 82: Don't think this should be completely removed

erenbesel (commented):
Nice! one comment

Adyen/Core/Core Protocols/PaymentComponent.swift
```diff
@@ -15,10 +15,10 @@ package protocol PaymentMethodAware {
 }
 
 /// A component that handles stored payment methods.
-package protocol StoredPaymentComponent: PaymentComponent, PresentableComponent {}
+package protocol StoredPaymentComponent: PresentablePaymentComponent {}
 
 package enum PaymentComponentType {
-    case regular(PaymentComponent & PresentableComponent)
+    case regular(PresentablePaymentComponent)
```
robertdalmeida, line 21: nonblocking;thought; 
May not even be related to these specific changes. But just generally to bring to attention.
currently regular & stored are the same (StoredPaymentComponent = PresentablePaymentCoponent). 
Unless StoredPaymentComponent has something specific in its protocol for Stored

Maybe these types need a revisit? to explain what is regular, initable and stored. 

See also the naming recommendation; 
Image

AdyenCheckout/Core/CheckoutCore+Orchestration.swift
robertdalmeida: Interesting. \AuthenticationComponent\ is an action component that presents.

nauaros: Talked offline. This line is not needed

Adyen/Core/Core Protocols/PaymentComponent.swift
nauaros: As discussed, StoredPaymentComponent won't be PresentablePaymentComponent any more. We can't do the changes right now, as the AlertView still needs to be removed.

erenbesel (approved):

robertdalmeida (approved):

PR #2660, opened 2026-07-31
Fix: Cap card security code to its expected length in stored value

neelSharma12:
# Summary [Required]

Fixes card security code (CVC/CVV) input accepting more than its expected length, being validated and submitted even though the extra digits are masked, because the cap was applied to the formatted value but not to the stored value.

# Demo [Optional]
These demo videos are provided for Non-Amex flow where the expected length is max 3. 

| Before | After |
|--------|-------|
|  |  |

<fixed>
- Fixed the card security code (CVC/CVV) field accepting more than the allowed number of digits, which could submit an over-length code while only showing the truncated value
</fixed>

## Testing Instructions [Optional]

1. Card component, non-Amex card (e.g. Visa): the CVC field stops accepting input after 3 digits; the submitted value is 3 digits.
2. Amex card: the CVC field accepts 4 digits.
3. Paste a long numeric string into the CVC field: it is truncated to the expected length in both the display and the submitted value.

## Ticket [Optional]
<ticket>
COSDK- 572
</ticket>

## Checklist [Required]

- [x] Tested changes locally

### DIFF
--- a/AdyenCard/Form/FormCardSecurityCodeItemView.swift
+++ b/AdyenCard/Form/FormCardSecurityCodeItemView.swift
@@ -82,6 +82,8 @@ internal final class FormCardSecurityCodeItemView: FormTextItemView<FormCardSecu
         replacementString string: String
     ) -> Bool {
         guard !string.isEmpty else { return true }
+
+        guard string.count == 1 else { return true }
         
         let currentText = textField.text ?? ""
         guard let textRange = Range(range, in: currentText) else { return true }
--- a/AdyenCard/Form/FormCardSecurityCodeItemView.swift
+++ b/AdyenCard/Form/FormCardSecurityCodeItemView.swift
@@ -83,7 +83,11 @@ internal final class FormCardSecurityCodeItemView: FormTextItemView<FormCardSecu
     ) -> Bool {
         guard !string.isEmpty else { return true }
 
-        guard string.count == 1 else { return true }
+        // Multi-character replacements (e.g. paste) are allowed through here and truncated
+        // to the expected length by the formatter; a single typed keystroke is rejected
+        // once it would exceed the max length.
+        let isSingleCharacterReplacement = string.count == 1
+        guard isSingleCharacterReplacement else { return true }
         
         let currentText = textField.text ?? ""
         guard let textRange = Range(range, in: currentText) else { return true }
--- a/AdyenCard/Form/FormCardSecurityCodeItemView.swift
+++ b/AdyenCard/Form/FormCardSecurityCodeItemView.swift
@@ -75,6 +75,26 @@ internal final class FormCardSecurityCodeItemView: FormTextItemView<FormCardSecu
         super.textFieldDidEndEditing(text)
         cardHintView.isHighlighted = false
     }
+    
+    override internal func textField(
+        _ textField: UITextField,
+        shouldChangeCharactersIn range: NSRange,
+        replacementString string: String
+    ) -> Bool {
+        guard !string.isEmpty else { return true }
+        
+        let currentText = textField.text ?? ""
+        guard let textRange = Range(range, in: currentText) else { return true }
+        
+        let updatedText = currentText.replacingCharacters(in: textRange, with: string)
+        let digitCount = updatedText.filter(\.isNumber).count
+        
+        return digitCount <= expectedLength
+    }
+    
+    private var expectedLength: Int {
+        item.selectedCard == CardBrand.americanExpress ? 4 : 3
+    }
 }
 
 extension FormCardSecurityCodeItemView {
--- a/Tests/IntegrationTests/Card Tests/Items/CardSecurityCodeItem/FormCardSecurityCodeItemViewTests.swift
+++ b/Tests/IntegrationTests/Card Tests/Items/CardSecurityCodeItem/FormCardSecurityCodeItemViewTests.swift
@@ -35,7 +35,7 @@ class FormCardSecurityCodeItemViewTests: XCTestCase {
         try super.tearDownWithError()
     }
 
-    func testTextFieldSanitizationGivenNonAllowedCharactersShouldSanitizeAndFormatInput() {
+    func test_textFieldSanitization_givenNonAllowedCharacters_shouldSanitizeAndFormatInput() {
         // Given
         let expectedCVV = "458"
         let expectedFormattedCVV = formatter.formattedValue(for: expectedCVV)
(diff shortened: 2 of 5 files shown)


atmamont:
Before video is on v5, after video is on v6, shouldn't it be both on v6?

neelSharma12:
> Before video is on v5, after video is on v6, shouldn't it be both on v6?

The "before" video was captured on v5 indeed, but I'll re-record it on v6 so both videos are on the same version for consistency.

AdyenCard/Form/FormCardSecurityCodeItemView.swift
```diff
@@ -75,6 +75,28 @@ internal final class FormCardSecurityCodeItemView: FormTextItemView<FormCardSecu
...
     }
+    
+    override internal func textField(
+        _ textField: UITextField,
+        shouldChangeCharactersIn range: NSRange,
+        replacementString string: String
+    ) -> Bool {
+        guard !string.isEmpty else { return true }
+
+        guard string.count == 1 else { return true }
```
descorp: This prevents shopper from Copy-Past CVV from somewhere, right ?

neelSharma12: Good question, but no, this does not prevent pasting. On the contrary, this code is added to accept a multi-character paste, and if the pasted value exceeds expectedLength, the formatter truncates it to the max length.The length check here only applies to single keystrokes, so typing past the limit is blocked while pasting still works.

AdyenCard/Form/FormCardSecurityCodeItemView.swift
```diff
@@ -75,6 +75,28 @@ internal final class FormCardSecurityCodeItemView: FormTextItemView<FormCardSecu
...
+        guard let textRange = Range(range, in: currentText) else { return true }
+        
+        let updatedText = currentText.replacingCharacters(in: textRange, with: string)
+        let digitCount = updatedText.filter(\.isNumber).count
+        
+        return digitCount <= expectedLength
+    }
+    
+    private var expectedLength: Int {
+        item.selectedCard == CardBrand.americanExpress ? 4 : 3
```
nauaros, line 102: suggestion - **non-blocking**: What do you think of having the CVC length as a property in `CardBrand` so all the info its in a single place?

nauaros (approved):

AdyenCard/Form/FormCardSecurityCodeItemView.swift
neelSharma12: Thanks for the suggestion. I am making a code improvement for this in a separate PR. Changes will be up for review soon :)

AdyenCard/Form/FormCardSecurityCodeItemView.swift
atmamont: I'd suggest extracting this explanation to a named variable such as 
```swift
let singleKeyStroke = string.count == 1
guard singleKeyStroke else { return true }
```

otherwise it's not clear why this check is needed

neelSharma12: Thanks, that makes sense. However, I am thinking of naming this variable to `isSingleCharacterReplacement`. 

```swift
let isSingleCharacterReplacement = string.count == 1
guard isSingleCharacterReplacement else { return true }
```

Also, I am thinking of adding a short explanation here. What do you think?

AdyenCard/Form/FormCardSecurityCodeItemView.swift
```diff
@@ -75,6 +75,28 @@ internal final class FormCardSecurityCodeItemView: FormTextItemView<FormCardSecu
...
+    ) -> Bool {
+        guard !string.isEmpty else { return true }
+
+        guard string.count == 1 else { return true }
+        
+        let currentText = textField.text ?? ""
+        guard let textRange = Range(range, in: currentText) else { return true }
+        
+        let updatedText = currentText.replacingCharacters(in: textRange, with: string)
+        let digitCount = updatedText.filter(\.isNumber).count
```
atmamont, line 96: There are few places in the code where you check counts, and here there is a filter applied. What is the goal of this, and wouldn't it be false when someone copy-pastes the string value including non-digits?

Tests/IntegrationTests/Card Tests/Items/CardSecurityCodeItem/FormCardSecurityCodeItemViewTests.swift
```diff
@@ -65,4 +65,182 @@ class FormCardSecurityCodeItemViewTests: XCTestCase {
...
+            sut.textField,
+            shouldChangeCharactersIn: NSRange(location: 2, length: 0),
+            replacementString: "3"
+        )
+
+        // Then
+        XCTAssertTrue(shouldChange)
+    }
+
+    func testShouldChangeCharactersGivenAmexAndInputWithinFourDigitsShouldAllow() {
```
atmamont: Can you please reformat test cases names using TESTING.MD guidance?

atmamont (commented):
Looks good, small comments on readability

AdyenCard/Form/FormCardSecurityCodeItemView.swift
atmamont: Sure, you know the semantics better here!

AdyenCard/Form/FormCardSecurityCodeItemView.swift
neelSharma12: I have applied the `filter` here to count the digits rather than the raw characters to keep the check about the thing we actually care about. Non-digits  are stripped out by the formatter anyway, so the result is the same. Counting digits just makes the intent clearer. 

On the copy-paste, this case never actually hits this line. A paste is more than one character, so it's already let through a few lines above (`guard isSingleCharacterReplacement`) and the formatter trims it down to the right length, e.g. `"12ab34" `becomes `"123".`This check only kicks in for a single typed character, so it won't reject a paste.

Tests/IntegrationTests/Card Tests/Items/CardSecurityCodeItem/FormCardSecurityCodeItemViewTests.swift
neelSharma12: Done!

atmamont (approved):

AdyenCard/Form/FormCardSecurityCodeItemView.swift
atmamont: Clear, makes sense!

PR #2663, opened 2026-08-05
Extract CVC/CVV length rule into a single source of truth

neelSharma12:
# Summary [Required]

## Key changes

- Introduces a single source of truth for the card security code length: `CardBrand?.expectedSecurityCodeLength`, which returns **4 for American Express** and **3 for every other or undetected brand**.
- The `CardSecurityCodeFormatter`, `CardSecurityCodeValidator`, and `FormCardSecurityCodeItemView` now derive the length from this one property instead of each re-implementing the `4 : 3` rule.
- Cleans up misleading documentation in `CardSecurityCodeFormatter `. 

## Ticket [Optional]
<ticket>
COSDK-572
</ticket>

## Checklist [Required]

- [x] Tested changes locally

### DIFF
--- a/AdyenCard/Utilities/CardBrand/CardBrand+SecurityCode.swift
+++ b/AdyenCard/Utilities/CardBrand/CardBrand+SecurityCode.swift
@@ -6,10 +6,17 @@
 
 import Adyen
 
+extension CardBrand {
+    /// Expected security code (CVC/CVV) length: 4 for Amex, 3 for any other brand.
+    internal var expectedSecurityCodeLength: Int {
+        self == .americanExpress ? Constants.amexLength : Constants.generalLength
+    }
+}
+
 extension CardBrand? {
     /// Expected security code (CVC/CVV) length: 4 for Amex, 3 for any other or undetected brand.
     internal var expectedSecurityCodeLength: Int {
-        self == .americanExpress ? Constants.amexLength : Constants.generalLength
+        self?.expectedSecurityCodeLength ?? Constants.generalLength
     }
 }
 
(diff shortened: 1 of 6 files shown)


AdyenCard/Utilities/CardBrand/CardBrand+SecurityCode.swift
```diff
@@ -0,0 +1,26 @@
...
+import Adyen
+
+extension CardBrand {
+    /// Expected security code (CVC/CVV) length: 4 for Amex, 3 for any other brand.
+    internal var expectedSecurityCodeLength: Int {
+        self == .americanExpress ? Constants.amexLength : Constants.generalLength
+    }
+}
+
+extension CardBrand? {
```
erenbesel, line 21: Do we really need the same extension on the optional version? Why is the extension on the normal CardBrand not enough?

neelSharma12: We could keep only the concrete extension, but then all four call sites would need `?? Constants.generalLength`. That would repeat the `'no brand → 3 digits'` fallback in all four places, which is exactly the duplication this change removes.

AdyenCard/Utilities/CardBrand/CardBrand+SecurityCode.swift
```diff
@@ -0,0 +1,26 @@
...
+    /// Expected security code (CVC/CVV) length: 4 for Amex, 3 for any other or undetected brand.
+    internal var expectedSecurityCodeLength: Int {
+        self?.expectedSecurityCodeLength ?? Constants.generalLength
+    }
+}
+
+private enum Constants {
```
nauaros: Do you think it makes more sense to move this enum within `CardBrand` instead of it being on its own here in the file?

neelSharma12: Done. Since we have both the concrete `CardBrand` and and optional `CardBrand?` extension, the nested enum is `fileprivate` and referenced as `CardBrand.Constants` from the optional one.

nauaros (approved):

PR #2665, opened 2026-08-05
Update Nightly Firebase workflow

nauaros:
# Summary

Update the following parameters in the Nightly Firebase workflow:
- Build version name
- Update Firebase release notes

### DIFF
--- a/.github/workflows/v6-firebase-nightly.yml
+++ b/.github/workflows/v6-firebase-nightly.yml
@@ -1,4 +1,4 @@
-name: v6_firebase_nightly
+name: 🌙🔥 Nightly Firebase
 
 on:
   # 22:00 UTC Sun-Thu is 00:00 Amsterdam Mon-Fri during CEST (UTC+2). Midnight local falls on the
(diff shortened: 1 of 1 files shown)


erenbesel (approved):

.github/workflows/v6-firebase-nightly.yml
```diff
@@ -23,7 +23,7 @@ jobs:
     # build to Firebase App Distribution.
     uses: ./.github/workflows/firebase-app-distribution.yml
     with:
-      build_version_name: 6.0.0-nightly
+      build_version_name: nightly
       build_number: '1'
-      release_notes: Latest development state of v6
+      release_notes: Latest SDK build
```
atmamont, line 28: Should we maybe mention the branch name here?

nauaros: I can add `Latest SDK build (develop)` until we change it for `main`. The idea is to keep this job even after v6 is finished, so no mention to v6 explicitly

atmamont (approved):

.github/workflows/v6-firebase-nightly.yml
atmamont: My concern was about having both v5 and v6 builds in Firebase along each other.

atmamont: The best way possible is to always have a Demo app build version matching the SDK version

nauaros: yeah. Gonna make sure we have one for alpha-1 and v5 (explicitly mentioned on the build version name)

PR #2678, opened 2026-08-14
Remove superseded nightly builds from Firebase App Distribution

nauaros:
## Summary

Firebase App Distribution always creates a new release instead of replacing an existing one, so in some scenarios duplicate nightly builds pile up for testers.

This adds a `remove_superseded_releases` job to the nightly workflow that runs after a successful distribution and deletes older releases matching the fixed `nightly` version name via the App Distribution REST API. If the distribution fails, this job doesn't run, so the previous nightly build stays available. Mirrors the approach used in `adyen-android`'s `release_nightly_app.yml`.

Also renamed `v6-firebase-nightly.yml` → `firebase-nightly.yml`.

## Checklist [Required]

- [x] Tested changes locally


atmamont (approved):

erenbesel (approved):

neelSharma12 (approved):

.github/workflows/firebase-nightly.yml
```diff
@@ -0,0 +1,97 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    outputs:
+      release_notes: ${{ steps.resolve.outputs.release_notes }}
+
+    steps:
+      - name: Resolve nightly configuration
+        id: resolve
+        run: echo "release_notes=Latest SDK build - (${GITHUB_SHA::7})" >> "$GITHUB_OUTPUT"
```
jreij: Minor: the "nightly" string is hardcoded in the `distribute` job and the `remove_superseded_releases` job. If you ever change the name in one place and forget to in the other, one of the steps will not work as expected. The workflow on Android [propagates the version name](https://github.com/Adyen/adyen-android/blob/159c2c86a5a4d3a64ea8cedbf8593748a5eda3cd/.github/workflows/release_nightly_app.yml#L31) in the `nightly_config` job to avoid this. If you're only using `nightly_config` to propagate the release notes (which is only used in one job), you might as well remove it because right now it's redundant.

nauaros: Good point. Changed so `nightly` is only set in one point.

jreij (approved):

jreij (approved):

PR #2681, opened 2026-08-17
New Installments UI on card component part1

neelSharma12:
# Summary [Required]
This PR wraps the card installments field in a "Payment plan" section header and renames its title to "Installments", keeping the section's visibility in sync with the installments item. This mirrors the existing billing-address section pattern.

Simulator Screenshot - iPhone 17 - 2026-08-17 at 11 07 00

**End goal**: Bring the card installments UI in line with the new design. This part is UI-only and keeps the existing wheel picker

**Note**: the "Payment plan" / "Installments" strings are placeholders (`// TODO: Localize`) pending finalization and Android alignment.

## Ticket [Optional]
<ticket>
COSDK-1431
</ticket>

## Checklist [Required]

- [x] Tested changes locally


AdyenCard/Form/Installments/FormCardInstallmentsItem.swift
```diff
@@ -81,7 +81,8 @@ internal final class FormCardInstallmentsItem: BaseFormPickerItem<InstallmentEle
             style: style
         )
         isHidden.wrappedValue = true
-        title = localizedString(.cardInstallmentsNumberOfInstallments, localizationParameters)
+        // TODO: Localize the "Installments" title once the string is finalized.
+        title = localizedString(LocalizationKey(key: "Installments"), localizationParameters)
         updatePickerContent()
```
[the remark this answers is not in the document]
neelSharma12: Good catch - `updatePickerContent() `never reset `isHidden` to `true` in the empty-options `guard`, so the field and new "Payment plan" header stayed visible after switching to a brand with no options or `nil`. Added `isHidden.wrappedValue = true` in the `guard`, and updated the two nil/no-option `assertions` to expect true.

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -367,6 +370,24 @@ extension CardViewController {
...
+        observe(installmentsFormItem.isHidden) { [weak self] isHidden in
+            self?.installmentsSectionItem?.isVisible = !isHidden
+        }
+    }
+    
+    private var installmentsItem: FormItem? {
+        guard let installmentsItem = items.installmentsItem else { return nil }
+        // TODO: Localize the "Payment plan" section header.
+        return installmentsItem.withSectionHeader(title: localizedString(
+            LocalizationKey(key: "Payment plan"), localizationParameters
```
nauaros, line 387: Better to create a key for this. Since Android already has this feature, you can use their key.

nauaros (approved):

AdyenCard/Components/Card/CardViewController.swift
atmamont: @neelSharma12 according to the ways of working, we put keys for all new strings into the epic. You can keep TODO for the Smartling part, but please use real key.

atmamont: Same applies to another "Installments" string.

atmamont: Android already defined these strings as "public" that can be overriden by merchants. In order to do so in the new v6 API we provide them an opportunity to provide their custom strings provider, see https://github.com/Adyen/adyen-ios/blob/develop/Adyen/Utilities/CheckoutLocalizationProvider.swift#L16, `CheckoutLocalizationKey` and its extensions.
Android had added these public keys for installment strings override so you only need to match them (better to do it in a separate PR): https://github.com/Adyen/adyen-android/blob/main/core/src/main/java/com/adyen/checkout/core/common/localization/CheckoutLocalizationKey.kt#L45

AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -346,8 +347,10 @@ extension CardViewController {
             append(items.socialSecurityNumberItem)
         }
         
-        if let installmentsItem = items.installmentsItem {
+        if let installmentsItem {
+            installmentsSectionItem = installmentsItem
             append(installmentsItem)
+            bindInstallmentsSectionVisibility()
```
atmamont, line 353: This part has no coverage, let's add an integration test that:

1. Sets up the state so installments should render. 
2. Verifies the "Payment plan" header is rendered.
3. Switches the state from one with installments to another without.
4. Verifies the entire section container becomes hidden.
5. Switches back and verifies it becomes visible.

atmamont (commented):
Looks good! Some comments on localization alignment with Android and coverage

AdyenCard/Components/Card/CardViewController.swift
neelSharma12: Added.

erenbesel (approved):
Approved and agreed with localization key comments

AdyenCard/Components/Card/CardViewController.swift
neelSharma12: As discussed, the Android team hasn't implemented the new design (with the different strings) yet. Once the new keys are added, I'll update these strings.

atmamont: Let's not merge this PR with plan English as a key.

atmamont (changes requested):
Requesting changes to make sure we merge localizations once and never get back to them.

PR #2684, opened 2026-08-20
Stored payment method removal screen integration with dropin

erenbesel:
# Summary [Required]
Adds complete stored payment-method management to Drop-in: Manage entry and routing, synchronized list removal, supported-method filtering, expiry/warning presentation, confirmation sheet, empty/error states, and Figma-aligned UI with focused test coverage.

https://github.com/user-attachments/assets/8020b26a-d56c-4e5a-adb2-22e285d33d05

## Ticket [Optional]
<ticket>
COSDK-1430
</ticket>

## Checklist [Required]

- [ ] Tested changes locally

### DIFF
--- a/Adyen/Assets/Generated/LocalizationKey.swift
+++ b/Adyen/Assets/Generated/LocalizationKey.swift
@@ -574,8 +574,6 @@ public struct LocalizationKey {
     public static let storedPaymentMethodManagementCardsTitle = LocalizationKey(key: "adyen.storedPaymentMethodManagement.cardsTitle")
     /// Other
     public static let storedPaymentMethodManagementOtherTitle = LocalizationKey(key: "adyen.storedPaymentMethodManagement.otherTitle")
-    /// Expired
-    public static let storedPaymentMethodManagementExpired = LocalizationKey(key: "adyen.storedPaymentMethodManagement.expired")
     /// Remove %@
     public static let storedPaymentMethodManagementRemoveConfirmationAction = LocalizationKey(key: "adyen.storedPaymentMethodManagement.removeConfirmationAction")
     /// Nothing here!
--- a/Adyen/Assets/Generated/LocalizationKey.swift
+++ b/Adyen/Assets/Generated/LocalizationKey.swift
@@ -568,7 +568,7 @@ public struct LocalizationKey {
     public static let preselectedPaymentMethodOtherOptions = LocalizationKey(key: "adyen.preselectedPaymentMethod.otherOptions")
     /// Manage
     public static let storedPaymentMethodManagementTitle = LocalizationKey(key: "adyen.storedPaymentMethodManagement.title")
-    /// Remove your favorite and saved payment methods
+    /// Remove your saved payment methods
     public static let storedPaymentMethodManagementDescription = LocalizationKey(key: "adyen.storedPaymentMethodManagement.description")
     /// Cards
     public static let storedPaymentMethodManagementCardsTitle = LocalizationKey(key: "adyen.storedPaymentMethodManagement.cardsTitle")
--- a/Adyen/Assets/Generated/LocalizationKey.swift
+++ b/Adyen/Assets/Generated/LocalizationKey.swift
@@ -584,7 +584,7 @@ public struct LocalizationKey {
     public static let storedPaymentMethodManagementEmptyMessage = LocalizationKey(key: "adyen.storedPaymentMethodManagement.emptyMessage")
     /// Payment options
     public static let storedPaymentMethodManagementPaymentOptions = LocalizationKey(key: "adyen.storedPaymentMethodManagement.paymentOptions")
-    /// Error removing payment method. Try again later.
+    /// We were unable to remove your payment method, try again later.
     public static let storedPaymentMethodManagementRemovalErrorMessage = LocalizationKey(key: "adyen.storedPaymentMethodManagement.removalErrorMessage")
     /// Expired
     public static let storedPaymentMethodExpired = LocalizationKey(key: "adyen.storedPaymentMethod.expired")
--- a/AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementViewModel.swift
+++ b/AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementViewModel.swift
@@ -13,12 +13,9 @@ import Foundation
 internal final class StoredPaymentMethodManagementViewModel: ObservableObject {
 
     private enum RemovalFlow {
-        case idle(error: StoredPaymentMethodRemovalError?)
-        case confirming(
-            item: StoredPaymentMethodManagementItem,
-            error: StoredPaymentMethodRemovalError?
-        )
-        case removing(item: StoredPaymentMethodManagementItem)
+        case idle
+        case confirming(StoredPaymentMethodManagementItem)
+        case removing(StoredPaymentMethodManagementItem)
     }
 
     // MARK: - Properties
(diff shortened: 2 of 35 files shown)


AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementViewModel.swift
```diff
@@ -100,10 +97,6 @@ internal final class StoredPaymentMethodManagementViewModel: ObservableObject {
         itemPendingRemoval = nil
     }
 
-    internal func dismissRemovalError() {
-        removalError = nil
-    }
-
     internal func confirmRemoval(of item: StoredPaymentMethodManagementItem) async {
         do {
             try await capability.remove(item.paymentMethod)
```
atmamont, line 139: I couldn't find how the remove button gets disabled state while the removal operation is performed. If we don't disable the button, there is a chance of double-taps.

Maybe something like this is needed with a quick guard check as the first line in this method, then passing it to the `RemovalConfirmationView` in the initializer so it renders proper button state (enabled/disabled)

```swift
@Published internal private(set) var isRemoving = false
```

AdyenDropIn/Modules/StoredPaymentMethodManagement/Views/StoredPaymentMethodManagementListView.swift
```diff
@@ -26,24 +27,29 @@ internal struct StoredPaymentMethodManagementListView: View {
...
-                    StoredPaymentMethodManagementSectionView(
-                        title: viewModel.sectionTitle(for: section.kind),
-                        section: section,
-                        removeButtonTitle: viewModel.removeButtonTitle,
-                        theme: theme,
-                        onRemove: viewModel.requestRemoval
-                    )
-                }
+        VStack(alignment: .leading, spacing: Constants.verticalSpacing) {
+            if viewModel.removalError != nil {
```
atmamont, line 31: If I understand correctly, this renders the error as the first list item. What happens if error happens when user scrolled past the top part of the list and the amount of stored payment methods take more than one visible screen?

AdyenUI/CheckoutColors.swift
```diff
@@ -53,6 +55,7 @@ public struct CheckoutColors: Equatable {
     public var textSecondary: UIColor
 
     package var success: UIColor
+    package var warning: UIColor
```
atmamont, line 58: Do we want to expose it to merchants as well? What's the Android way?

AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementViewModel.swift
erenbesel: The loading indicator itself is another task itself but good point, we could at least block it for now.

AdyenUI/CheckoutColors.swift
erenbesel: This and other points like localization will discussed together.

AdyenDropIn/Modules/StoredPaymentMethodManagement/Views/StoredPaymentMethodManagementListView.swift
erenbesel: There needs to be more than 5 stored pms (which is extremely unlikely) on the smallest phone and even that may still be ok. Other option is to put the header area before scrollview and only the list scrolls. But that looks worse to me only scrolling a smaller area trying to find a payment to remove on a smaller phone. wdyt?

atmamont: We discussed this with the team and, indeed, we can make it better later.

AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementViewModel.swift
erenbesel: Added default loading indicator to the screen for now

AdyenDropIn/Modules/StoredPaymentMethodManagement/Views/StoredPaymentMethodManagementView.swift
```diff
@@ -23,39 +28,62 @@ internal struct StoredPaymentMethodManagementView: View {
...
+                        .presentationBackground(.black.opacity(0.1))
+                } else {
+                    removalConfirmationView(for: item)
                 }
             }
-
-            Button(viewModel.cancelTitle, role: .cancel) {
-                viewModel.dismissRemovalConfirmation()
-            }
         }
```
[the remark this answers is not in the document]
erenbesel: There is no visual glitch, but to preserve animation, this was still a better option. done

Adyen/Assets/Generated/LocalizationKey.swift
```diff
@@ -574,8 +574,6 @@ public struct LocalizationKey {
     public static let storedPaymentMethodManagementCardsTitle = LocalizationKey(key: "adyen.storedPaymentMethodManagement.cardsTitle")
     /// Other
     public static let storedPaymentMethodManagementOtherTitle = LocalizationKey(key: "adyen.storedPaymentMethodManagement.otherTitle")
-    /// Expired
-    public static let storedPaymentMethodManagementExpired = LocalizationKey(key: "adyen.storedPaymentMethodManagement.expired")
```
atmamont, line 578: Is this a duplicate? I see we still show "Expired" label under the payment method that is expired according to designs

AdyenDropIn/Modules/StoredPaymentMethodManagement/Views/StoredPaymentMethodManagementView.swift
```diff
@@ -26,30 +26,46 @@ internal struct StoredPaymentMethodManagementView: View {
...
-        .sheet(isPresented: isRemovalConfirmationPresented) {
-            if let item = viewModel.itemPendingRemoval {
-                if #available(iOS 16.4, *) {
-                    removalConfirmationView(for: item)
-                        .presentationBackground(.black.opacity(0.1))
-                } else {
-                    removalConfirmationView(for: item)
-                }
+        .sheet(item: removalConfirmationItem) { item in
+            if #available(iOS 16.4, *) {
```
atmamont, line 47: This looks like maintenance hell to me.
Would we reuse the same confirmation view anywhere else so we can create our own SwiftUI component and hide this branching inside?

AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementViewModel.swift
```diff
@@ -12,6 +12,15 @@ import Foundation
 @MainActor
 internal final class StoredPaymentMethodManagementViewModel: ObservableObject {
 
+    private enum RemovalFlow {
```
atmamont, line 15: This is a bit hard to understand. Both idle and confirming cases have an optional error. I could imagine confirming is a state that can resolve in a success or an error (idle)

atmamont (approved):

AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementViewModel.swift
erenbesel: Error removed from this state

AdyenDropIn/Modules/StoredPaymentMethodManagement/Views/StoredPaymentMethodManagementView.swift
erenbesel: No idea if we would reuse it yet, but thanks to Apple we have this tiny method only after 16.4 😮‍💨

Adyen/Assets/Generated/LocalizationKey.swift
erenbesel: Duplicate indeed

AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementViewModel.swift
```diff
@@ -123,37 +112,39 @@ internal final class StoredPaymentMethodManagementViewModel: ObservableObject {
...
     }
 
     internal func confirmRemoval() async {
-        guard case let .confirming(item, _) = removalFlow else {
+        guard case let .confirming(item) = removalFlow else {
             return
         }
 
-        removalFlow = .removing(item: item)
+        removalFlow = .removing(item)
```
atmamont, line 135: _Just a thought_ : now the state management takes a shape and projecting this onto all future SwiftUI-friendly view models, I am thinking of a state-action-reduce reusable pattern so we know what exactly to look at and what to expect in every future module.
This is a bigger discussion and def doesn't belong to this scope.

atmamont (approved):

atmamont (approved):

PR #2689, opened 2026-08-28
Add opt-in header to picker search screen via FormPickerConfiguration

neelSharma12:
# Summary [Required]

**What:** Adds an opt-in title and optional subtitle header to the generic picker search screen through `FormPickerConfiguration`. The configuration is threaded through `FormPickerItem` and `FormPickerItemView`, while existing picker flows retain their navigation-title behavior by default.

**Why:** Establishes reusable generic-picker header infrastructure for upcoming picker migrations, including installments, Address Lookup, Issuer List, and Phone Extension, without changing current callers that do not opt in.

## Technical Information [Optional]

- Adds `FormPickerConfiguration.Header` with required title and optional subtitle content.
- Adds `FormPickerHeaderView` using checkout title/body styles and secondary subtitle color.
- Preserves the active checkout theme through the item-to-picker presentation flow.
- Omits headers with an empty title, including subtitle-only configurations, preserving the navigation-title fallback.
- Keeps the header content-sized while allowing the results list to absorb remaining vertical space.
- Leaves Address Lookup, Issuer List, and Phone Extension behavior unchanged until they explicitly migrate.

## Ticket [Optional]
<ticket>
COSDK-1431
</ticket>

## Checklist [Required]
- [x] Tested changes locally
- [x] Aligned public API changes with other platforms (not applicable; no public API change)

### DIFF
--- a/AdyenUI/UI/Form/Items/Value/Picker/FormPickerHeaderView.swift
+++ b/AdyenUI/UI/Form/Items/Value/Picker/FormPickerHeaderView.swift
@@ -52,7 +52,7 @@ internal final class FormPickerHeaderView: UIView {
     // MARK: - Initializers
 
     internal init?(header: FormPickerConfiguration.Header, theme: CheckoutTheme) {
-        guard !header.title.isEmpty || header.subtitle?.isEmpty == false else {
+        guard !header.title.isEmpty else {
             return nil
         }
 
--- a/AdyenUI/UI/Form/Items/Value/Picker/FormPickerHeaderView.swift
+++ b/AdyenUI/UI/Form/Items/Value/Picker/FormPickerHeaderView.swift
@@ -54,7 +54,11 @@ internal final class FormPickerHeaderView: UIView {
 
     // MARK: - Initializers
 
-    internal init(header: FormPickerConfiguration.Header, theme: CheckoutTheme) {
+    internal init?(header: FormPickerConfiguration.Header, theme: CheckoutTheme) {
+        guard !header.title.isEmpty || header.subtitle?.isEmpty == false else {
+            return nil
+        }
+
         self.header = header
         self.theme = theme
         super.init(frame: .zero)
--- a/AdyenUI/UI/Form/Items/Value/Picker/FormPickerHeaderView.swift
+++ b/AdyenUI/UI/Form/Items/Value/Picker/FormPickerHeaderView.swift
@@ -1,0 +1,92 @@
+//
+// Copyright (c) 2026 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Adyen
+import UIKit
+
+/// A header shown at the top of a picker search screen, displaying a title and an optional subtitle.
+///
+/// Mirrors the title/subtitle styling of the payment method list header
+/// (`title` uses `labels.title`, `subtitle` uses `labels.body` + `colors.textSecondary`).
+internal final class FormPickerHeaderView: UIView {
+
+    private enum Layout {
+        static let stackSpacing: CGFloat = 4
+    }
+
+    // MARK: - Subviews
+
+    internal lazy var titleLabel: UILabel = {
+        let label = UILabel()
+        label.text = header.title
+        label.numberOfLines = 0
+        label.adjustsFontForContentSizeCategory = true
+        label.translatesAutoresizingMaskIntoConstraints = false
+        return label
+    }()
+
+    internal lazy var subtitleLabel: UILabel = {
+        let label = UILabel()
+        label.text = header.subtitle
+        label.isHidden = header.subtitle?.isEmpty ?? true
+        label.numberOfLines = 0
+        label.adjustsFontForContentSizeCategory = true
+        label.translatesAutoresizingMaskIntoConstraints = false
+        return label
+    }()
+
+    private lazy var stackView: UIStackView = {
+        let stackView = UIStackView(arrangedSubviews: [titleLabel, subtitleLabel])
+        stackView.axis = .vertical
+        stackView.alignment = .leading
+        stackView.spacing = Layout.stackSpacing
+        stackView.translatesAutoresizingMaskIntoConstraints = false
+        return stackView
+    }()
+
+    // MARK: - Properties
+
+    private let header: FormPickerConfiguration.Header
+    private let theme: CheckoutTheme
+
+    // MARK: - Initializers
+
+    internal init(header: FormPickerConfiguration.Header, theme: CheckoutTheme) {
+        self.header = header
+        self.theme = theme
+        super.init(frame: .zero)
+        setupView()
+    }
+
+    @available(*, unavailable)
+    internal required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    // MARK: - Private
+
+    private func setupView() {
+        layoutMargins = .zero
+
+        addSubview(stackView)
+
+        NSLayoutConstraint.activate([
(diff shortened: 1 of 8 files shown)


AdyenUI/UI/Form/Items/Value/Picker/FormPickerHeaderView.swift
```diff
@@ -0,0 +1,93 @@
...
+
+    private let header: FormPickerConfiguration.Header
+    private let theme: CheckoutTheme
+
+    // MARK: - Initializers
+
+    internal init?(header: FormPickerConfiguration.Header, theme: CheckoutTheme) {
+        guard !header.title.isEmpty || header.subtitle?.isEmpty == false else {
+            return nil
+        }
```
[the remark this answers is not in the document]
neelSharma12: Good catch. The header title is required, so a subtitle-only header should not be created. I updated the `failable` initializer to require a non-empty title and added coverage confirming that an empty title with a populated subtitle falls back to the navigation title.

AdyenUI/UI/Form/Items/Value/Picker/FormPickerHeaderView.swift
```diff
@@ -0,0 +1,93 @@
...
+        layoutMargins = .zero
+
+        addSubview(stackView)
+
+        NSLayoutConstraint.activate([
+            stackView.topAnchor.constraint(equalTo: layoutMarginsGuide.topAnchor),
+            stackView.leadingAnchor.constraint(equalTo: layoutMarginsGuide.leadingAnchor),
+            stackView.trailingAnchor.constraint(equalTo: layoutMarginsGuide.trailingAnchor),
+            stackView.bottomAnchor.constraint(equalTo: layoutMarginsGuide.bottomAnchor)
+        ])
```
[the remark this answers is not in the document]
neelSharma12: Thanks for the suggestion. I’m keeping the stack constrained to `layoutMarginsGuide` intentionally. This matches `PaymentMethodListHeaderView` and preserves the header’s layout-margin ownership, while setting the margins to `zero` avoids the current double inset. Directly constraining to the view anchors would make the margin configuration ineffective and make future explicit spacing require constraint changes.

AdyenUI/UI/View Controllers/SearchViewController/SearchViewController.swift
```diff
@@ -120,11 +130,30 @@ package class SearchViewController: UIViewController, AdyenObserver {
...
+        if let headerView {
+            // Keep the header at its content height so the results list, not the header, absorbs extra vertical space.
+            let headerHeightHug = headerView.heightAnchor.constraint(equalToConstant: 0)
+            headerHeightHug.priority = .defaultLow
+            NSLayoutConstraint.activate([
+                headerView.topAnchor.constraint(equalTo: view.layoutMarginsGuide.topAnchor),
+                headerView.leadingAnchor.constraint(equalTo: view.layoutMarginsGuide.leadingAnchor),
+                headerView.trailingAnchor.constraint(equalTo: view.layoutMarginsGuide.trailingAnchor),
+                headerHeightHug
+            ])
```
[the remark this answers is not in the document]
neelSharma12: Thanks for the suggestion. I’m keeping the low-priority height constraint because the current header labels use `UIKit’s` default vertical hugging priority of 250, not 251, and the results list container has no intrinsic content height that guarantees it absorbs the remaining space. Since `SearchViewController` accepts any `UIView` header, this soft constraint provides a consistent tie-breaker for all callers. At `.defaultLow,` it yields to the header content’s compression resistance while preventing the header from stretching.

AdyenUI/UI/View Controllers/SearchViewController/SearchViewController.swift
```diff
@@ -120,11 +130,30 @@ package class SearchViewController: UIViewController, AdyenObserver {
...
+            searchBarTopAnchor = headerView.bottomAnchor
+            searchBarTopSpacing = 8
```
nauaros: Shall we move this constants to a `Layout` enum?

AdyenUI/UI/View Controllers/SearchViewController/SearchViewController.swift
```diff
@@ -120,11 +130,30 @@ package class SearchViewController: UIViewController, AdyenObserver {
...
+            searchBarTopSpacing = 0
+        }
         
         NSLayoutConstraint.activate([
             searchBar.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: 8),
             searchBar.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -8),
-            searchBar.topAnchor.constraint(equalTo: view.layoutMarginsGuide.topAnchor, constant: 0),
+            searchBar.topAnchor.constraint(equalTo: searchBarTopAnchor, constant: searchBarTopSpacing),
             
             resultsListViewController.view.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: 0),
```
nauaros: suggestion: The default value is 0 already, so the constant part can be removed

nauaros (approved):
Some minor comments. Good job

AdyenUI/UI/View Controllers/SearchViewController/SearchViewController.swift
neelSharma12: Good suggestion. I moved the layout-related constants into a private Layout enum.

AdyenUI/UI/View Controllers/SearchViewController/SearchViewController.swift
neelSharma12: Good catch. I removed the redundant constant: 0 arguments from the anchor-to-anchor constraints in this method.

AdyenUI/UI/Form/Items/Value/Picker/FormPickerHeaderView.swift
```diff
@@ -0,0 +1,93 @@
...
+
+    @available(*, unavailable)
+    internal required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    // MARK: - Private
+
+    private func setupView() {
+        layoutMargins = .zero
```
atmamont, line 73: Would this work with iPhone Duo layout (navigation bar transitioning to the side)? We have to respect platform guidance, as far as I remember we need to take a look into `directionalLayoutMargins`. It's not critical but nice to investigate as we know of upcoming changes.

AdyenUI/UI/Form/Items/Value/Picker/FormPickerHeaderView.swift
```diff
@@ -0,0 +1,93 @@
...
+
+        addSubview(stackView)
+
+        NSLayoutConstraint.activate([
+            stackView.topAnchor.constraint(equalTo: layoutMarginsGuide.topAnchor),
```
atmamont, line 78: My previous comment actually belongs here where we anchor constraints

atmamont (approved):

AdyenUI/UI/Form/Items/Value/Picker/FormPickerHeaderView.swift
atmamont: However, this is just a thought, let's wait for Xcode 27.1 beta to go out and then assess how we approach constraints.

neelSharma12: Agreed. Let’s keep the current constraints unchanged in this PR and revisit `directionalLayoutMargins` once Xcode 27.1 beta is available, so we can validate the iPhone Duo side-navigation behavior against the actual platform guidance.

PR #2687, opened 2026-08-27
Stored payment method removal updates

erenbesel:
# Summary [Required]
Includes the postponed loading state animation per row. Removes the ProgressView that was added as a placeholder.
- Change localization text for header subtitle
- Add icon before the error text
- Revert the new color addition for expired text and use destructive instead
- Nest private views in list view to reduce long names
- Disable back button in the hosting controller while removal is in progress
- Remove the viewmodel state now that state is simple and replace with an optional item

## Ticket [Optional]
<ticket>
COSDK-1430
</ticket>

## Checklist [Required]

- [x] Tested changes locally
- [x] Added/updated unit tests
- [x] Verified against acceptance criteria
- [ ] Aligned public API changes with other platforms (if applicable)

PR #2690, opened 2026-08-28
Showcase: Apply State-Action-Reducer architecture to stored payment methods screen

atmamont:
## Summary

- Models screen coordination as one private `State` value.
- Models each row with an `ItemRemovalState` enum: `.removing`, `.failed`, or absent for idle.
- Routes UI intents and async results through a pure `State + Action -> State?` reducer.
- Preserves the concurrent per-row removal behavior introduced by #2687.
- Binds confirmation to the item rendered by the sheet and qualifies completions by identifier.
- Derives the global error banner, row loading state, and UIKit navigation lock from the same source of truth.

This PR is part of stack #2691 and is stacked directly on #2687. Review the six-file diff against `chore/stored-pm-removal-updates` rather than `develop`.

## Why

#2687 allows several stored payment methods to be removed concurrently. That introduces two levels of state:

- screen state: at most one item awaiting confirmation;
- item state: each row is independently idle, removing, or failed.

The reducer keeps the transition rules in one place. It rejects duplicate, mismatched, stale, and invalid actions without publishing an unchanged state.

This is intentionally a screen-local pilot. It adds no shared framework or dependency. The goal is to evaluate whether this pattern is readable enough to guide future SwiftUI screens before adopting it more broadly.

## State model

The root `State` contains:

- `itemPendingConfirmation`;
- `[StoredPaymentMethodId: ItemRemovalState]`.

`ItemRemovalState` is mutually exclusive:

- `.removing`;
- `.failed`;
- absence from the dictionary means idle.

State helper methods encapsulate dictionary access so the reducer reads in domain terms: `isIdle`, `isRemoving`, `isPendingConfirmation`, `setRemovalState`, and `clearRemovalState`.

The existing global error banner is a view-model projection: it is visible when any item has failed. Retrying one item clears only that item's failure, so errors belonging to other rows remain represented.

`sections` remains outside the reducer because it is screen content rather than transient removal coordination. It changes only after the reducer accepts a successful completion.

## Concurrency guarantees

- Different rows can remove concurrently.
- The same identifier cannot start twice.
- Only the item currently shown by the sheet can be confirmed.
- A row cannot be removing and failed simultaneously.
- Completing one identifier does not alter another row's state.
- Retrying one failed row does not clear failures belonging to other rows.
- Section mutation and router notification happen only after an accepted success.
- Navigation remains locked until no removals are active.

## Review focus

- Does the reducer make valid transitions easier to understand than distributed property mutations?
- Does the screen-state struct plus per-item enum represent the domain clearly?
- Are the helper APIs readable and appropriately scoped?
- Is this a useful pilot for future SwiftUI state management without introducing a framework prematurely?

## Ticket

COSDK-1430

## Test plan

- [x] Single successful removal updates sections and notifies the router.
- [x] Duplicate actions for an in-flight row start one capability call.
- [x] Two rows can remain in flight concurrently.
- [x] Failure of one row does not affect another row's loading state.
- [x] A later success preserves an earlier concurrent error.
- [x] Multiple row failures remain independent; retry clears only the selected row.
- [x] Navigation locks during removal and restores its original state afterward.
- [x] Focused view-model and assembler suites pass: 14 tests.
- [x] SwiftFormat, SwiftLint, and `git diff --check` pass.

### DIFF
--- a/AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementHostingController.swift
+++ b/AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementHostingController.swift
@@ -75,11 +75,9 @@ internal final class StoredPaymentMethodManagementHostingController: UIHostingCo
     }
 
     private func observeRemovalState() {
-        viewModel.$identifiersBeingRemoved
-            .map { !$0.isEmpty }
-            .removeDuplicates()
-            .sink { [weak self] isActive in
-                self?.setNavigationLock(active: isActive)
+        viewModel.isRemovingPublisher
+            .sink { [weak self] isRemoving in
+                self?.setNavigationLock(active: isRemoving)
             }
             .store(in: &cancellables)
     }
--- a/AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementHostingController.swift
+++ b/AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementHostingController.swift
@@ -53,17 +53,20 @@ internal final class StoredPaymentMethodManagementHostingController: UIHostingCo
             return
         }
 
-        originalNavigationState = NavigationState(
-            isUserInteractionEnabled: navigationController.navigationBar.isUserInteractionEnabled,
-            tintColor: navigationController.navigationBar.tintColor,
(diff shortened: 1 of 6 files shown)


atmamont:
Had few discussions, came to agreement this could be the way to go. Ideally, unidirectional architecture could be the best way to deal with simple SwiftUI screens. For this particular (and single SwiftUI) feature the amount of overhead is too much, we might revisit it if this approaches is not failproof.

PR #2693, opened 2026-09-02
Stored payment management screen analytics

erenbesel:
# Summary [Required]
Added analytics to the new screen. Passed the analytics provider on assembler level.

- Render event:
  - component = storedPaymentMethodManagement
  - type = rendered
- Event for confirming removal
  - component = storedPaymentMethodManagement
  - type = clicked
  - target = stored_payment_method_remove_button

## Ticket [Optional]
<ticket>
COSDK-1508
</ticket>

## Checklist [Required]

- [ ] Tested changes locally

### DIFF
--- a/Adyen/Analytics/Models/AnalyticsEventInfo.swift
+++ b/Adyen/Analytics/Models/AnalyticsEventInfo.swift
@@ -40,6 +40,7 @@ package struct AnalyticsEventInfo: AnalyticsEvent {
         case rendered = "Rendered"
         case input = "Input"
         case displayed = "Displayed"
+        case remove = "Remove"
     }
     
     public init(component: String, type: InfoType) {
--- a/Adyen/Analytics/Models/AnalyticsEventInfo.swift
+++ b/Adyen/Analytics/Models/AnalyticsEventInfo.swift
@@ -40,7 +40,7 @@ package struct AnalyticsEventInfo: AnalyticsEvent {
         case rendered = "Rendered"
         case input = "Input"
         case displayed = "Displayed"
-        case remove = "Remove"
+        case clicked = "Clicked"
     }
     
     public init(component: String, type: InfoType) {
--- a/AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementAssembler.swift
+++ b/AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementAssembler.swift
@@ -29,17 +29,20 @@ internal struct StoredPaymentMethodManagementAssembler: StoredPaymentMethodManag
     private let localizationParameters: LocalizationParameters?
     private let logoURLProvider: LogoURLProvider
     private let theme: CheckoutTheme
+    private let analyticsProvider: AnyAnalyticsProvider?
 
     // MARK: - Initializers
 
     internal init(
         localizationParameters: LocalizationParameters?,
         logoURLProvider: LogoURLProvider,
-        theme: CheckoutTheme
+        theme: CheckoutTheme,
+        analyticsProvider: AnyAnalyticsProvider?
     ) {
         self.localizationParameters = localizationParameters
         self.logoURLProvider = logoURLProvider
         self.theme = theme
+        self.analyticsProvider = analyticsProvider
     }
 
     // MARK: - StoredPaymentMethodManagementAssemblerProtocol
@@ -57,7 +60,8 @@ internal struct StoredPaymentMethodManagementAssembler: StoredPaymentMethodManag
             paymentMethods: paymentMethods,
             capability: capability,
             mapper: mapper,
-            localizationParameters: localizationParameters
+            localizationParameters: localizationParameters,
+            analyticsProvider: analyticsProvider
         )
         let hostingController = StoredPaymentMethodManagementHostingController(viewModel: viewModel, theme: theme)
         let router = StoredPaymentMethodManagementRouter(
--- a/AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementViewModel.swift
+++ b/AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementViewModel.swift
@@ -128,7 +128,7 @@ internal final class StoredPaymentMethodManagementViewModel: ObservableObject {
             return
         }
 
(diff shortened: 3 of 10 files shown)


Adyen/Analytics/Models/AnalyticsEventInfo.swift
```diff
@@ -40,6 +40,7 @@ package struct AnalyticsEventInfo: AnalyticsEvent {
         case rendered = "Rendered"
         case input = "Input"
         case displayed = "Displayed"
+        case remove = "Remove"
```
atmamont: Should it be "Removed" to match the rest?

AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementAssembler.swift
```diff
@@ -29,17 +29,20 @@ internal struct StoredPaymentMethodManagementAssembler: StoredPaymentMethodManag
     private let localizationParameters: LocalizationParameters?
     private let logoURLProvider: LogoURLProvider
     private let theme: CheckoutTheme
+    private let analyticsProvider: AnyAnalyticsProvider?
```
atmamont, line 32: _Just asking:_ Can we think of making analytics non-optional in v6?

Adyen/Analytics/Models/AnalyticsEventInfo.swift
erenbesel: Since this is triggered at the start of confirmation, we don't know yet if it's removed or not thus we decided to call it like this.

AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementViewModel.swift
```diff
@@ -117,6 +128,7 @@ internal final class StoredPaymentMethodManagementViewModel: ObservableObject {
             return
         }
 
+        sendAnalyticsEvent(type: .remove)
```
atmamont: I now see the issue - we send this event when the removal is confirmed by a user but the request itself still can fail, so "Removed" is incorrect.

AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementAssembler.swift
erenbesel: If only. That would mean not starting checkout if attempt id fetching fails.

AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementViewModel.swift
erenbesel: See my answer above, it confirms the action meaning the start of the API call, not removal itself

atmamont (approved):

AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementViewModel.swift
atmamont: Clear!

robertdalmeida: nonblocking; From what i can trace the only information we recieve from this event is that storePaymentManage screen tapped on remove. 

Would there be any futuristic value in also knowing which payment method is being removed?

robertdalmeida (approved):

Adyen/Analytics/Models/AnalyticsEventInfo.swift
robertdalmeida: suggestion; thought; 
type: Clicked
target: "remove.{payment-method}"

Image

AdyenDropIn/Modules/StoredPaymentMethodManagement/StoredPaymentMethodManagementViewModel.swift
erenbesel: We discussed it but wasn't about sure sending the unique id of the payment method.

Adyen/Analytics/Models/AnalyticsEventInfo.swift
robertdalmeida: https://github.com/Adyen/adyen-ios/pull/2693#discussion_r3914001665

erenbesel: Done, great suggestion!

AdyenDropIn/Modules/StoredPaymentMethodManagement/Views/StoredPaymentMethodManagementView.swift
```diff
@@ -140,21 +140,29 @@ private extension StoredPaymentMethodManagementView {
...
+                    Text(cancelTitle)
+                        .font(Font(theme.elements.labels.bodyEmphasized.font))
+                        .foregroundStyle(Color(uiColor: theme.colors.highlight))
+                        .frame(maxWidth: .infinity, minHeight: Constants.buttonHeight)
+                        .background(Color(uiColor: theme.colors.background))
+                        .clipShape(RoundedRectangle(cornerRadius: Constants.buttonCornerRadius))
+                        .contentShape(Rectangle())
+                }
+                .buttonStyle(.plain)
+                .accessibilityIdentifier(StoredPaymentMethodManagementAccessibilityIdentifier.cancelRemoval)
```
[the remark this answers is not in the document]
erenbesel: done

atmamont (approved):

Adyen/Analytics/Models/AnalyticsEventInfo.swift
robertdalmeida: just a thought; 
Not a problem per se -
I'm not sure what use will `remove` be without knowing what "paymentMethod" is being removed. 
All we will know is that remove is tapped on this screen. 

We wouldn't be able to determine behaviors like;
This paymnent Method is being removed mostly in comparision to others. 
Or 
The user in one session seems to be deleting this stored payment method but initiating it directly(non stored) immediately after.

robertdalmeida (approved):

PR #2696, opened 2026-09-09
Remove component manager builders & move to Checkout

erenbesel:
- Make Checkout the exclusive owner of Drop-in payment component construction.
- Remove the legacy Drop-in component builders.
- Align Drop-in around one ComponentManager.
- Drop-in demos now use the CheckoutConfiguration (Checkout creation next step)

- Keep partial-payment reload behavior out of scope.

## Ticket [Optional]
<ticket>
COSDK-1525
</ticket>

## Checklist [Required]

- [ ] Tested changes locally


AdyenDropIn/Utilities/ComponentManager.swift
```diff
@@ -217,19 +170,56 @@ private extension ComponentManager {
...
+    }
+
+    func containsSupportedPaymentMethod(_ paymentMethod: PaymentMethod) -> Bool {
+        if let storedPaymentMethod = paymentMethod as? any StoredPaymentMethod {
+            return supportedStoredPaymentMethods.contains { $0 == storedPaymentMethod }
+        }
+
+        return (supportedPaidPaymentMethods + supportedRegularPaymentMethods).contains {
+            $0 == paymentMethod
+        }
```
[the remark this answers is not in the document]
erenbesel: Done

AdyenDropIn/Utilities/ComponentManager.swift
```diff
@@ -217,19 +170,55 @@ private extension ComponentManager {
...
     // MARK: - Payment Method Validation
-    
+
     func isAllowed(_ paymentMethod: PaymentMethod) -> Bool {
         let requiresPhotoLibrary = isVoucherPaymentMethod(paymentMethod) || isQRCodePaymentMethod(paymentMethod)
         guard requiresPhotoLibrary else { return true }
-        
-        return Bundle.main.object(forInfoDictionaryKey: "NSPhotoLibraryAddUsageDescription") != nil
+
+        return hasPhotoLibraryUsageDescription
```
atmamont, line 221: Do we ever ask shopper consent? From this part it looks we would hide such payment method.

AdyenDropIn/Utilities/ComponentManager.swift
```diff
@@ -217,19 +170,55 @@ private extension ComponentManager {
...
         guard let remainingAmount = order?.remainingAmount else { return }
         context.amount = remainingAmount
     }
-    
+
     // MARK: - Payment Method Validation
-    
+
     func isAllowed(_ paymentMethod: PaymentMethod) -> Bool {
         let requiresPhotoLibrary = isVoucherPaymentMethod(paymentMethod) || isQRCodePaymentMethod(paymentMethod)
```
atmamont, line 218: `ComponentManager` should not be aware of payment method specifics, otherwise adding new payment method that required photo library access would force logic update here. Could the availability check be owned by the component builder/factory, or handled by the planned payment-method availability mechanism?

AdyenDropIn/Utilities/ComponentManager.swift
```diff
@@ -217,19 +170,55 @@ private extension ComponentManager {
         paymentMethods.stored
             .filter { $0.supportedShopperInteractions.contains(.shopperPresent) }
     }
-    
+
+    // TODO: To be improved with payment method availability feature.
+    func canBuildComponent(for paymentMethod: PaymentMethod) -> Bool {
+        assembleComponent(for: paymentMethod) != nil
```
atmamont, line 176: Good too see comment above, almost got a heart attack! :)

AdyenDropIn/Utilities/ComponentManager.swift
```diff
@@ -217,19 +170,55 @@ private extension ComponentManager {
...
+    }
+
+    func containsSupportedPaymentMethod(_ paymentMethod: PaymentMethod) -> Bool {
+        if let storedPaymentMethod = paymentMethod as? any StoredPaymentMethod {
+            return supportedStoredPaymentMethods.contains { $0 == storedPaymentMethod }
+        }
+        
+        return supportedRegularPaymentMethods.contains { $0 == paymentMethod }
```
nauaros, line 184: What do you think of building a collection with ALL supported payment methods so its easier to read:

```
func containsSupportedPaymentMethod(_ paymentMethod: PaymentMethod) -> Bool {
    if let storedPaymentMethod = paymentMethod as? any StoredPaymentMethod {
        return supportedStoredPaymentMethods.contains(storedPaymentMethod)
    }

    let supportedPaymentMethods =
        supportedRegularPaymentMethods + supportedPaidPaymentMethods

    return supportedPaymentMethods.contains(paymentMethod)
}
```

AdyenDropIn/Utilities/ComponentManager.swift
nauaros, line 221: I think its better to have the `Info.plist` photo library check here as it was before instead of a property at the top of the `ComponentManager`. Makes it straightforward to understand

nauaros: We don't. We rely on the merchant to add this key to the app's info plist and ask for consent. That's we check before displaying the PM

nauaros (approved):
I dropped two comments for better understanding. Approving to not block

AdyenDropIn/Utilities/ComponentManager.swift
erenbesel: Haha, see gemini's comment on this for how it was before :)

AdyenDropIn/Utilities/ComponentManager.swift
erenbesel: You mean revert it back? It's made a property to be testable, previously it was hardcoded

AdyenDropIn/Utilities/ComponentManager.swift
erenbesel: Good catch and it's part of the availability task. This is just keeping existing behavior without adding more refactoring but only make it testable by making it a property. We will address it separately on the payment method availability task

AdyenDropIn/Utilities/ComponentManager.swift
erenbesel: Not sure if we should, never discussed. Here we are only keeping existing behavior.

atmamont: In general there is a chance an average shopping app never needs photo library access and we could be the only consumers. Current behavior cuts down all payment methods that require photo library access. I'm happy to leave as is and have it tracked somewhere in YT

atmamont: We agreed this check should not belong to the component manager at all.

atmamont (approved):

PR #2697, opened 2026-09-10
Amount in dropin: [1/5] - submit button title logic moved to a namespaced type instead of a Global function & added test coverage

robertdalmeida:
# Summary [Required]
- Moved the logic for amount aware strings to a separate type(`AmountAwarePaymentStringsPolicy`) and added tests to add as much coverage as possible.
- There would be no behavior change in this PR (as those will be made once it is refined). This is to ensure the changes are in one place.

Next:
Will build on this PR to include the changes for the amount.

## Ticket [Optional]
<ticket>
CO-236
</ticket>

## Checklist [Required]

- [x] Tested changes locally

### DIFF
--- a/Adyen/Utilities/AmountAwarePaymentStringsPolicy.swift
+++ b/Adyen/Utilities/AmountAwarePaymentStringsPolicy.swift
@@ -17,14 +17,14 @@ package enum AmountAwarePaymentStringsPolicy {
     package static func payButtonTitle(
         with amount: Amount?,
         style: PaymentStyle,
-        _ localizationParameters: LocalizationParameters?
+        localizationParameters: LocalizationParameters?
     ) -> String {
         guard var amount else {
             return localizedString(.submitButton, localizationParameters)
         }
 
         if amount.value == 0 {
-            return zeroPaymentButtonTitle(style: style, localizationParameters)
+            return zeroPaymentButtonTitle(style: style, localizationParameters: localizationParameters)
         }
 
         amount.localeIdentifier = amount.localeIdentifier ?? localizationParameters?.locale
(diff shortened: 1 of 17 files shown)


Adyen/Utilities/AmountAwarePaymentStringsPolicy.swift
```diff
@@ -0,0 +1,45 @@
+//
+// Copyright (c) 2026 Adyen N.V.
+//
+// This file is open source and available under the MIT license. See the LICENSE file for more info.
+//
+
+import Foundation
+
+package enum PaymentStyle {
```
erenbesel, line 9: I don't really understand these 2 cases and how they affect amount. Perhaps we can rename this enum as well here to make it easier? 

(we can do this on the last PR of the stack as well)

erenbesel (approved):

Tests/UnitTests/Utilities/AmountAwarePaymentStringsPolicyTests.swift
```diff
@@ -0,0 +1,230 @@
...
+        // When
+        let title = sut.payButtonTitle(with: amount, style: scenario.style, localizationParameters: localizationParameters)
+
+        // Then
+        #expect(title == scenario.expectedTitle)
+    }
+
+    private func makeSUT() -> AmountAwarePaymentStringsPolicy.Type {
+        AmountAwarePaymentStringsPolicy.self
+    }
```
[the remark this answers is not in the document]
robertdalmeida: As the team is more used to always have a makeSUT() and operate on the sut, the metatype this returns is an acceptable approach.

Adyen/Utilities/AmountAwarePaymentStringsPolicy.swift
robertdalmeida: It is indeed not used. I'm removing [this in the last PR](https://github.com/Adyen/adyen-ios/pull/2713/changes#top), as in this PR avoiding removing anything.

nauaros (approved):

PR #2700, opened 2026-09-10
Amount in dropin: [2/5] - Amount removed from CheckoutConfiguration as the SDK recieves it from Session.setup()

robertdalmeida:
# Summary [Required]
- Amount doesn't need to be passed in the CheckoutConfiguration, as we are considering session only integration.
- Amount is recieved from Session.setup()
- The SDK will still consider the case for amount not available to support non session flow without amount (if needed)

## Ticket [Optional]
<ticket>
COSDK-1511
</ticket>

## Checklist [Required]

- [x] Tested changes locally
- [x] Aligned public API changes with other platforms (if applicable) - TODO


AdyenCheckout/Setup/CheckoutProvider.swift
```diff
@@ -149,7 +151,7 @@ internal class CheckoutProvider: CheckoutProviding {
 
             return try await AdyenContext(
                 apiContext: configuration.apiContext,
-                amount: configuration.amount,
+                amount: nil,
```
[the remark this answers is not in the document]
robertdalmeida: Any integration without session will not have the amount. v6 is sessions first only and others are an afterthought.

nauaros (approved):

AdyenCheckout/Setup/CheckoutProvider.swift
```diff
@@ -42,6 +42,8 @@ internal class CheckoutProvider: CheckoutProviding {
             adyenContext: adyenContext,
             apiClient: apiClient
         )
+        // Amount is now received from the Session.setup() response.
```
atmamont: This comment makes no sense when advanced flow is removed completely. It's up to you when to remove it, but for an agent it doesn't have enough hooks to catch it when we start COSDK-1513.
Seeing your reply to gemini in the next comment it looks this comment should be removed right away.

atmamont (commented):
Please update v6 readme/migration guides accordingly.

erenbesel (approved):

AdyenCheckout/Setup/CheckoutProvider.swift
robertdalmeida: Sure. Done.

robertdalmeida:
> Please update v6 readme/migration guides accordingly.

Good one. Didn't think of those guides at all.

atmamont (approved):

neelSharma12 (approved):

PR #2703, opened 2026-09-11
fix: Always complete session checkout without session result

descorp:
# Summary

Fixed session checkout completion so `onComplete` is invoked even when `sessionResult` is unavailable. Missing session results now default to an empty string, while unexpected missing values retain assertion diagnostics.

## Release Notes

<fixed>
- Fixed session checkout completion callbacks not being invoked when a session result was unavailable.
</fixed>

## Technical Information

`SessionCheckoutCallbackStore` now always creates `SessionCheckoutResult`. Missing `sessionResult` is expected for `.cancelled` and `.error`; other missing values still trigger an assertion without suppressing the merchant callback.

## Ticket

<ticket>
COSDK-1532
</ticket>

## Checklist

- [x] Tested changes locally

### DIFF
--- a/AdyenCheckout/Callbacks/CheckoutResultCallbackStore.swift
+++ b/AdyenCheckout/Callbacks/CheckoutResultCallbackStore.swift
@@ -22,11 +22,12 @@ package final class SessionCheckoutCallbackStore: CheckoutResultCallbackStore {
     package var onFailure: CheckoutFailureHandler?
 
     package func handleCompletion(resultCode: CheckoutResultCode, sessionId: String?, sessionResult: String?) {
-        guard let sessionId, let sessionResult else {
-            AdyenAssertion.assertionFailure(message: "Session completion called without sessionId or sessionResult.")
-            return
+        if sessionId == nil {
+            AdyenAssertion.assertionFailure(message: "Session completion called without a sessionId.")
+        } else if sessionResult == nil, resultCode != .cancelled, resultCode != .error {
+            AdyenAssertion.assertionFailure(message: "Session completion called without a sessionResult.")
         }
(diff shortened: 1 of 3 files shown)


AdyenCheckout/Callbacks/CheckoutResultCallbackStore.swift
```diff
@@ -22,11 +22,12 @@ package final class SessionCheckoutCallbackStore: CheckoutResultCallbackStore {
...
-        guard let sessionId, let sessionResult else {
-            AdyenAssertion.assertionFailure(message: "Session completion called without sessionId or sessionResult.")
-            return
+        if sessionId == nil {
+            AdyenAssertion.assertionFailure(message: "Session completion called without a sessionId.")
+        } else if sessionResult == nil, resultCode != .cancelled, resultCode != .error {
+            AdyenAssertion.assertionFailure(message: "Session completion called without a sessionResult.")
         }
-        onComplete?(SessionCheckoutResult(resultCode: resultCode, sessionId: sessionId, sessionResult: sessionResult))
+        onComplete?(SessionCheckoutResult(resultCode: resultCode, sessionId: sessionId ?? "", sessionResult: sessionResult ?? ""))
```
erenbesel, line 30: if sessionResult can be null, that is good to catch but not happy with passing empty string for sessionId as they can't make a status call without it. 

So maybe guard sessionId always, then check for sessionResult like this?

descorp: Both are required for status call. Only thing - ID should never be nil. I didn't look too deep to fix it tho.
`sessionResult` can be nil in case of shopper closing payment manually.

PR #2708, opened 2026-09-14
Amount in dropin: [4/5] - Behavior of card/ACH/Cashapp component on zero amount

robertdalmeida:
# Summary [Required]
- When zero amount; Remove the checkbox to store payment credentials from the UI.  A zero amount is always used to create a token. Sending PaymneComponentData.storePaymentMethod: true to our server.

- `StorePaymentMethodPolicy` to keep a common place to decide to show the store checkbocx in the UI. 
- The decision whether to send the storePaymentMethod is a bit different for CashApp due to the CashAppConfiguration.storePaymentMethod parameter so each of components decide on their own. 

## Ticket [Optional]
<ticket>
COSDK-1509
</ticket>

## Checklist [Required]

- [x] Tested changes locally

### DIFF
--- a/AdyenCard/Components/Card/CardViewController.swift
+++ b/AdyenCard/Components/Card/CardViewController.swift
@@ -34,6 +34,7 @@ internal class CardViewController: FormViewController {
     private let cardLogos: [FormCardLogosItem.CardBrandLogo]
     private let allowedCoBadgedCardBrands: [CardBrand] = [.carteBancaire, .bcmc, .dankort]
     private let cardScannerAnalyticsHandler: CardScannerAnalyticsHandler
+
     private lazy var cardScannerController: CardScannerControlling = {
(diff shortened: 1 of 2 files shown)


AdyenCard/Components/Card/CardViewController.swift
```diff
@@ -210,9 +211,17 @@ internal class CardViewController: FormViewController {
     }
     
     internal var storePayment: Bool? {
-        configuration.showStorePaymentMethod ? items.storeDetailsItem.value : nil
+        if amount?.value == 0 {
```
erenbesel, line 214: Unfortunately there are 2 other places this needs to be done. ACH component and Cashapp component.
(So we may wonder if we should move this check a single source)
But even more unfortunately, the one cashapp first needs a fix that was done in v5 to be implemented on v6. Then apply this logic too.

robertdalmeida: Nice call out. Thanks for picking up on this. I will make those changes here.

nauaros (approved):

PR #2712, opened 2026-09-14
Add configurable search visibility to generic picker

neelSharma12:
# Summary [Required]

**What:** Adds configurable search visibility to the generic picker through `FormPickerConfiguration`. Search is still enabled and focused by default, while picker flows can be opt out without leaving an empty search-bar layout gap. 

**Why:** Enables short-list picker flows, starting with installments, to reuse the generic picker without unnecessary search UI while preserving existing behavior for current picker consumers.

## Ticket [Optional]
<ticket>
COSDK-1431
</ticket>

## Checklist [Required]

- [x] Tested changes locally


AdyenUI/UI/Form/Items/Value/Picker/FormPickerSearchViewController.swift
```diff
@@ -41,7 +41,8 @@ package final class FormPickerSearchViewController<Option: FormPickable>: UINavi
             localizationParameters: localizationParameters,
             style: style,
             searchBarPlaceholder: nil,
-            shouldFocusSearchBarOnAppearance: true
+            shouldShowSearchBar: configuration.isSearchEnabled,
```
atmamont, line 44: These two originate from the same `configuration.isSearchEnabled`, do we need to support both parameters?

AdyenUI/UI/View Controllers/SearchViewController/SearchViewController.swift
```diff
@@ -122,7 +126,8 @@ package class SearchViewController: UIViewController, AdyenObserver {
         
         delegate?.viewWillAppear(viewController: self)
         
-        if viewModel.shouldFocusSearchBarOnAppearance {
+        if viewModel.shouldShowSearchBar,
```
atmamont, line 129: Let's check use cases whether we need to show search and NOT focus on it. Maybe we can squash these two into one?

AdyenUI/UI/Form/Items/Value/Picker/FormPickerSearchViewController.swift
nauaros: It looks like the `shouldFocusSearchBarOnAppearance` is disabled by default, unless the search is enabled.

The question, why do we condition the focus search instead of always having it focused when the search is displayed?

AdyenUI/UI/View Controllers/SearchViewController/SearchViewController.swift
neelSharma12: We have a use case for Issuer list where we need to show the search bar without focusing it. We can squash these by introducing different modes for Search (e.g., ` hidden, visible and visibleAndFocused)`. What do you think of this?

atmamont: This gets more complexity than needed, I didn't take the issuer list into consideration (I didn't even know we can search through issuers list). Let's scope this out for now.

nauaros (approved):

PR #2720, opened 2026-09-17
Add `requiresUserInteraction` to `PaymentComponent`

nauaros:
# Summary [Required]
Adds a `requiresUserInteraction` flag to `PaymentComponent`, exposed publicly via `CheckoutPaymentComponent`. This lets integrators reliably know whether a component has any UI to present, or can be submitted directly — aligned with Android's approach to this same problem. 

`type` is kept as-is and continues to drive the UX flow Drop-in needs to trigger per component; `requiresUserInteraction` is a separate, explicit signal for whether that flow actually requires the shopper to interact with anything. Every component now declares its own value explicitly (defaulting to `true`, except `GenericPaymentComponent` which defaults to `false`), and `CheckoutPaymentComponent` simply forwards the component's value instead of inferring it from `type`. 

`CashAppPayComponent` computes it dynamically from its `showsStorePaymentMethodField`/`showsSubmitButton` configuration, since it's the only component whose UI can become fully optional after configuration.

## Technical Information [Optional]
- `type` alone isn't enough to determine UI presence once components can be configured to hide their optional fields/buttons (e.g. `showsSubmitButton`), so `requiresUserInteraction` is a separate, explicit property per component.
- Most components always render at least one mandatory field, so they're hardcoded to `true` regardless of configuration.
- `CashAppPayComponent` is the exception: its store-payment toggle and submit button are both optional, so it's `false` only when both are hidden.

## Checklist [Required]
- [x] Tested changes locally
- [ ] Verified against acceptance criteria (aligned with Android)


Adyen/Core/Core Protocols/PaymentComponent.swift
```diff
@@ -33,6 +33,9 @@ package protocol PaymentComponent: Component, PartialPaymentOrderAware, PaymentM
 
     var type: PaymentComponentType { get }
 
+    /// Indicates whether the component requires user interaction before submitting.
+    var requiresUserInteraction: Bool { get }
```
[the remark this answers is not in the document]
nauaros: We prefer to be explicit in every single component. Also, having a default extension can make developers ignore this property in the future. Its better to enforce it when components are defined

AdyenCashAppPay/CashAppPayComponent.swift
```diff
@@ -50,6 +50,10 @@ package final class CashAppPayComponent: PaymentComponent,
 
     package let type: PaymentComponentType = .regular
 
+    package var requiresUserInteraction: Bool {
+        configuration.showsStorePaymentMethodField || configuration.showsSubmitButton
```
erenbesel: I think the second part is not supposed to be there. Only requires interaction if toggle exists, if not, return no

nauaros: Hmm understood otherwise. @jreij can you confirm if this logic is correct?

jreij: Hmm, on Android I didn't make `showsSubmitButton` affect `requiresUserInteraction` in any component, I don't think cash app pay is different just because it has a custom button.

Think about it simply: merchants should be able to call `if(!requiresUserInteraction()) submit()` on all components to skip a redundant button click while being sure there is no shopper input required.

- If cash app pay is showing a switch and a button -> shoppers must see and react to the switch so requiresUserInteraction = true.
- If cash app pay is only showing a button -> shoppers only need to click a button so requiresUserInteraction = false.
- If cash app pay is neither showing a switch nor a button -> shoppers only need to click the merchant's custom button so requiresUserInteraction = false.

So our button being visible doesn't really affect the outcome, only the switch does. Let me know if this is clearer (hopefully I didn't confuse you more).

Adyen/Core/Core Protocols/PaymentComponent.swift
jreij, line 37: I added these code docs on the merchant facing function in Android, feel free to use something similar:

> Indicates whether the payment method requires user interaction before submitting.
> 
> When this returns `false`, [submit] can be called directly to skip a user action (e.g. a button click).
> Rendering the payment method UI is optional.
> 
> When this returns `true`, the payment method UI should be displayed so the user can provide
> the required input before calling [submit].
> 
> @return `true` if user interaction is needed, `false` otherwise.

AdyenCashAppPay/CashAppPayComponent.swift
nauaros: Gotcha, it is clear now. Gonna remove the `configuration.showsSubmitButton`. Then it is consistent with the rest of the components.

I believe `CashApp` it's the only component where there is no form at all, and a tokenization toggle that can be hidden.

nauaros: Updated

Adyen/Core/Core Protocols/PaymentComponent.swift
nauaros: Nice, thanks for sharing. Updated ours in `CheckoutPaymentComponent`, that's where we expose publicly `requiresUserInteraction`. [Here it is](https://github.com/Adyen/adyen-ios/pull/2720/changes#diff-1ec53f46a5269b7fc1a1569dbda325253e92a1879de5b22944e6540841a7c0a0R31)

atmamont (approved):

PR #2722, opened 2026-09-17
v5: Define currency minor units via CurrencyMinorDigitsProvider

nauaros:
## Summary

Replaces the hardcoded currency minor-digits overrides in `AmountFormatter` with a new `CurrencyMinorDigitsProvider` that uses Adyen’s official currency list. This provides a single source of truth for the number of minor units per ISO 4217 currency and preserves all existing override behavior while fixing more currencies.

Part of the cross-SDK currency alignment effort: [adyen-checkout-sdk-meta#102](https://github.com/Adyen/adyen-checkout-sdk-meta/pull/102) (`alignment/currencies/`, canonical list + audit doc).

## Demo

N/A — this is a backend formatting change with no UI changes.

## Release Notes

<changed>
- Updated amount formatting to use a centralized currency minor-digits provider, improving accuracy for currencies that were previously handled by iOS/ICU defaults.
</changed>

## Technical Information

- Added `CurrencyMinorDigitsProvider`, backed by https://docs.adyen.com/development-resources/currency-codes.
- `AmountFormatter` now delegates all currency minor-digit lookups to this provider.
- Currency codes in Adyen’s official list use the explicitly maintained Adyen minor-digit value.
- Currency codes not in Adyen’s list fall back to Apple’s `NumberFormatter`, which uses the system ICU/CLDR currency data.
- All historically overridden currencies (`ISK`, `CLP`, `COP`, `MRU`, `RSD`, `GHS`, `HUF` → 2; `CVE`, `IDR` → 0) keep the same values.
- The public API of `AmountFormatter` is unchanged: `formatted`, `minorUnitAmount`, and `decimalAmount` retain the same signatures and return types.
- `CurrencyMinorDigitsProviderTests` and an `AmountFormatter` synchronization test cover the provider list end-to-end.

### Why this is not a breaking change

1. **The public API is unchanged.** Existing integrations require no code changes.
2. **Existing overrides are preserved.** Every currency explicitly handled before returns the same number of minor digits.
3. **The fallback preserves platform behavior.** Currency codes not maintained in Adyen’s list use Apple’s `NumberFormatter`, matching the previous system-based behavior.
4. **Behavior becomes more deterministic for supported currencies.** Currencies in Adyen’s list use maintained values instead of relying on platform-dependent iOS/ICU defaults.

### Benefits

- Establishes one source of truth for Adyen-supported currency minor digits.
- Correctly handles currencies including three-decimal currencies such as `JOD`, `KWD`, and `BHD`, and zero-decimal currencies such as `JPY`, `VND`, and `XOF`.
- Retains Apple `NumberFormatter` behavior for currency codes outside Adyen’s maintained list.
- Prevents supported-currency formatting behavior from changing between iOS/ICU versions.
- Makes future currency-data updates localized and easier to review.
- Adds comprehensive regression coverage to keep `AmountFormatter` and the provider synchronized.

## Testing Instructions

1. Run `CurrencyMinorDigitsProviderTests` to verify known currencies, case-insensitive lookup, and the Apple `NumberFormatter` fallback for currencies outside Adyen’s list.
2. Run `AmountFormatterTests` to verify existing overrides and conversion behavior.
3. Confirm `testMinorDigitsMatchCurrencyMinorDigitsProviderForAllCurrencies` validates every currency in the provider data.

## Ticket

<ticket>
COSDK-1060
</ticket>

## Checklist

- [x] Tested changes locally


Tests/UnitTests/Core/Providers/CurrencyMinorDigitsProviderTests.swift
```diff
@@ -0,0 +1,72 @@
...
+
+    override func setUp() {
+        super.setUp()
+        sut = CurrencyMinorDigitsProvider()
+    }
+
+    override func tearDown() {
+        sut = nil
+        super.tearDown()
+    }
```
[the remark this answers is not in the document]
nauaros: Applied

Adyen/Core/Providers/CurrencyMinorDigitsProvider/CurrencyMinorDigitsProvider.swift
```diff
@@ -0,0 +1,42 @@
...
+
+    public func minorDigits(for currencyCode: String) -> Int {
+        let normalizedCurrencyCode = currencyCode.uppercased()
+        return Self.minorDigitsByCurrencyCode[normalizedCurrencyCode] ?? Self.systemMinorDigits(for: normalizedCurrencyCode)
+    }
+
+    /// Falls back to the system's ICU/CLDR currency data for ISO 4217 codes that are valid
+    /// but not on Adyen's official currency list (e.g. `CLF`, `UYW`).
+    private static func systemMinorDigits(for currencyCode: String) -> Int {
+        let formatter = NumberFormatter()
```
atmamont, line 37: Since this is a heavy op, we could think of using an internal default singleton instance to avoid formatter re-creation and keep the static func for convenience.

Adyen/Formatters/AmountFormatter.swift
```diff
@@ -78,27 +78,12 @@ public final class AmountFormatter {
         if let localeIdentifier {
             formatter.locale = Locale(identifier: localeIdentifier)
         }
-        if let maxDigits = maximumFractionDigits(for: currencyCode, localeIdentifier: localeIdentifier) {
-            formatter.maximumFractionDigits = maxDigits
-        }
+        formatter.maximumFractionDigits = minorDigits(for: currencyCode)
```
atmamont, line 81: Fascinating how semantically we went from "max" to "min"

atmamont (approved):

Adyen/Formatters/AmountFormatter.swift
nauaros: With this naming we align with the terminology used across Adyen. Too bad the formatter uses max tho

descorp (approved):

Adyen/Core/Providers/CurrencyMinorDigitsProvider/CurrencyMinorDigitsProvider.swift
nauaros: Good point, I did look into this. The fallback is a cold path tho: it only runs for currency codes that aren't on Adyen's official list, so in practice most integrations never hit it. Reusing a formatter also means mutating `currencyCode` on shared state, which needs synchronization (setting up a lock to make it thread safe). We could have something like this, wdyt? For what it is, I didn't want to over-engineer it

```swift
private final class SystemCurrencyFormatter {
 
    private let lock = NSLock()
 
    private let formatter: NumberFormatter = {
        let formatter = NumberFormatter()
        formatter.numberStyle = .currency
        return formatter
    }()
 
    func minorDigits(for currencyCode: String) -> Int {
        lock.lock()
        defer { lock.unlock() }
 
        formatter.currencyCode = currencyCode
        return formatter.maximumFractionDigits
    }
}
```

nauaros: Discussed this with the team (in the standup). We decided it's all right to keep it simple and recreate as it's a cold path that rarely happens.

